Issue #396, opened 2022-08-05
Produce valid certificate chain for several epochs on Devnet


jpraynaud:
### Issue
* Some epochs are not signed (with empty signers list in the `Pending Certificate`)
* The certificate chain is broken in flaky fashion after several epochs with an `AVKMismatch Error` when the stake distribution is updated
* The Signers get `Lottery for this epoch was lost` error from time to time when the stake distribution is updated

### Tasks

- [x] "Some epochs are not signed" is fixed with #317
- [x] Return an error when Aggregator detects an `Epoch gap` between the last certificate and the current `Epoch` (See #418)
- [x] Fine tune test end to end to reduce flakiness (see #419)
- [x] Add Certificate Chain integration tests (see #424)

jpraynaud:
I close this issue as the remaining fixes/development on the `Certificate Chain` are embedded in #364

PR #419, opened 2022-08-10
Attempt fix CI flakiness

jpraynaud:
This PR attempts to fix the CI flakiness that occurs from time to time and lead to an `AVKMismatch Error` when the Client restores   an archive after a `Stake Distribution` update.

Relates to #396


jpraynaud:
I close this PR as improvements are under development in #438 :slightly_smiling_face:

PR #460, opened 2022-08-30
Greg/444/sql store

ghubertpalo:
Create a SQLite store adapter.


mithril-common/src/store/adapter/mod.rs
```diff
@@ -5,6 +5,7 @@
 
 mod jsonfile_store_adapter;
 mod memory_adapter;
+mod sqlite_adapter;
```
Alenar, line 8: Add a `pub use sqlite_adapter::SQLiteAdapter`, this will also makes some of the clippy warnings disappear :)

mithril-common/src/store/adapter/sqlite_adapter.rs
```diff
@@ -0,0 +1,500 @@
...
+    K: Serialize,
+    V: DeserializeOwned,
+{
+    /// Create a new SQLiteAdapter instance.
+    pub fn new(table_name: &str, file: Option<PathBuf>) -> Result<Self> {
+        let file = file
+            .map(|v| v.to_string_lossy().to_string())
+            .unwrap_or_else(|| ":memory:".to_string());
+        let connection =
+            Connection::open(file).map_err(|e| AdapterError::InitializationError(e.into()))?;
```
Alenar, line 32: use a match here so you won't have to convert the PathBuf to a string (avoiding potential errors if the path contains some characters that don't translate well to a rust string) :
```rust
let connection = match file {
  Some(file) => Connection::open(file),
  None => Connection::open(":memory"),
}.map_err(|e| AdapterError::InitializationError(e.into()))?;
```

mithril-common/src/store/adapter/sqlite_adapter.rs
```diff
@@ -0,0 +1,500 @@
...
+
+    const TABLE_NAME: &str = "key_value_store";
+
+    fn get_file_path(test_name: &str) -> PathBuf {
+        let dirpath = std::env::temp_dir().join("mithril_test");
+
+        if !dirpath.exists() {
+            create_dir_all(&dirpath).expect(&format!(
+                "Expecting to be able to create the test directory '{}'.",
+                dirpath.to_string_lossy()
```
Alenar, line 309: When logging path use their `.display()` function instead of string conversion.

mithril-common/src/store/adapter/sqlite_adapter.rs
```diff
@@ -0,0 +1,500 @@
...
+            1,
+            row[1]
+                .as_integer()
+                .expect("expecting field 1 to be an integer")
+        );
+        assert_eq!(
+            "\"one\"".to_string(),
+            row[2]
+                .as_string()
+                .expect("expecting field 2 to be a string")
```
Alenar, line 359: Maybe you can join those two assert_eq into one using tuples ?

mithril-common/src/store/adapter/sqlite_adapter.rs
```diff
@@ -0,0 +1,500 @@
...
+            .unwrap();
+        let iterator = adapter.get_iter().await.unwrap();
+
+        for (index, element) in iterator.enumerate() {
+            match index {
+                0 => assert_eq!("one", element),
+                1 => assert_eq!("two", element),
+                2 => assert_eq!("three", element),
+                i => panic!("unexpected result index {} with data = '{:?}'.", i, element),
+            }
```
Alenar, line 418: Why don't you simply collect the iterator and compare it's result with a unique `assert_eq` ?

ghubertpalo: I found it more convenient that way to add/remove test cases.

mithril-common/src/store/adapter/sqlite_adapter.rs
ghubertpalo: I found it more explicit and easier to spot which one of the two assertion fails.

mithril-common/src/store/adapter/sqlite_adapter.rs
Alenar: I found it more readable and as convenient to add new cases:
```rust
assert_eq!(
    vec![
        (0, "one".to_string()),
        (1, "two".to_string()),
        (2, "three".to_string()),
    ],
    iterator.enumerate().collect::<Vec<(usize, String)>>(),
);
```
Btw: the test is wrong, the iterator should returns elements in the reverse order (from the latest to the oldest).

Alenar: An alternative using the iterator would be:
```rust
let mut iterator = adapter.get_iter().await.unwrap().enumerate();

assert_eq!(Some((2_usize, "three".to_string())), iterator.next());
assert_eq!(Some((1_usize, "two".to_string())), iterator.next());
assert_eq!(Some((0_usize, "one".to_string())), iterator.next());
assert_eq!(None, iterator.next());
```

Alenar (approved):

PR #466, opened 2022-08-31
Update run-mithril-devnet.md

abailly-iohk:
Seems like there was a supernumerary space character in the path


jpraynaud (approved):
LGTM :+1:

PR #468, opened 2022-08-31
use StoreError in stores

ghubertpalo:
Each stores tended to use their own specific error type, all of them
having only one value: AdapterError. This is now done within one type in
mithril-common::StoreError.


jpraynaud (approved):
LGTM

PR #469, opened 2022-09-01
Remove common test utils from release build

Alenar:
This will prevent us from using common test utilities in production code (since now we don't have to use the fake data anymore).
Hopefully this should also reduce the aggregator compilation time.


Alenar:
This doesn't work as expected: cargo will always include the `test-util` features. The only way to do that correctly would be extracting test-only code in a dedicated crate but ... some of it in common is used by common itself, avoiding circular dependency will be difficult.

Closing :(

PR #472, opened 2022-09-01
Avoid panics in 'StmInitializer'

jpraynaud:
This PR attempts to fix some panics that occur when a `StmInitializer` creates a `StmSigner` if the initializer is not registered.

In that case, the `new_signer` will return an error instead.

### DIFF
--- a/mithril-core/src/error.rs
+++ b/mithril-core/src/error.rs
@@ -115,6 +115,14 @@ pub enum RegisterError {
     SerializationError,
 }
 
+/// Errors which can be outputted by an initializer.
+#[derive(Debug, Clone, thiserror::Error, PartialEq, Eq)]
+pub enum StmInitializerError {
+    /// NotRegistered error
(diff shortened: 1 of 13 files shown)


Alenar (approved):

mithril-core/src/error.rs
```diff
@@ -115,6 +115,14 @@ pub enum RegisterError {
...
 }
 
+/// Errors which can be outputted by an initializer.
+#[derive(Debug, Clone, thiserror::Error, PartialEq, Eq)]
+pub enum StmInitializerError {
+    /// NotRegistered error
+    #[error("Initializer not registered. Cannot participate as a signer.")]
+    NotRegistered,
+}
+
```
iquerejeta, line 125: We are still thinking how to best handle the errors in mithril-core, but we certainly have as a goal to minimise the number of errors. I think it would be better to include this type of error in the `RegisterError` type.

iquerejeta (approved):
LGTM 👍

PR #470, opened 2022-09-01
Move GCP Aggregator to 'preview' network

jpraynaud:
This PR upgrades the Aggregator hosted on GCP cloud so that it uses the `preview` Cardano network instead of the legacy `testnet`:

- [x] Upgrade Cardano node to `1.35.3`
- [x] Parametrize the `docker-compose.yaml` file so that it can be used with different network
- [x] Update the `poolIds` of the Signer nodes (which `poolIds` are selected among the SPOs on the `preview` network) and reduce their number to `2`
- [x] Use `preview` network with `NETWORK=preview` in the `aggregator.tf` file
- [x] Update documentation

After merge, manually duplicate stores data so that we don't need to wait 2 epochs before signing

Relates to #457

PR #474, opened 2022-09-01
Update Blake dependency

iquerejeta:
This PR simply updates the Blake2 dependency. We use the new, more flexible design, of version 0.10.4 to use an output of 32 bytes instead of 64 bytes, making merkle paths half the size.


Alenar (approved):
LGTM, this PR will invalidate data on our testnet environment, but we will replace it with one on preview anyway.
Wait for a GO from @jpraynaud to merge it shortly before #470.

PR #482, opened 2022-09-06
Greg/476/sqlite

ghubertpalo:
This PR fixes a bug in the `SQLiteAdapter`. It does not now update the value when the key already exists.
This PR also makes the aggregator to use SQLite as adapter for all its stores.


ghubertpalo:
Duplicate of #477 
Commit [fd5807a](https://github.com/input-output-hk/mithril/pull/477/commits/fd5807af9fc4be4f855038cfd054b16f5675c7ae) has been cherry-picked into this PR's branch.

PR #484, opened 2022-09-06
Single signature without merkle path

curiecrypt:
Single Signatures without Merkle Path
===

`mithril-core` now provides single signatures that do not contain Merkle tree path. The membership of the signers who issued single signatures is checked by the batch proof that is created in aggregation phase. This update offers a remarkable improvement considering the size of single signatures, aggregate signature, and computation time.
The related changes are as follows:

### `merkle_tree.rs`
- Struct `BatchPath` is added. This struct holds the path of hashes for a batch of indices that contains the hashes and the corresponding merkle tree indices of given batch.
- A new struct `MerkleTreeCommitmentBatchCompat` is added. This is the merkle tree commitment that contains the number of leaves in the merkle tree in addition to the root. 
- `MerkleTree::get_batched_path()` creates the batch path for given list of indices.
- `MerkleTreeCommitmentBatchCompat::check()` checks whether given given list of merkle tree leaves verifies against the given `BatchPath`.

### `stm.rs`
- `Path` removed from `StmSig` and merkle tree index of the signer is included in the single signature.
- `StmSigner::sign()` does not compute the merkle tree path of the signer.
- `StmSig::verify()` does not check the merkle tree path.
- `BatchPath` is included in `StmAggrSig` struct.
- `StmClerk::aggregate()` computes the `batch_proof` by using the signers' merkle tree indexes contained in `StmSig`.
- Since the merkle tree commitment has changed to include the number of leaves in the tree, `StmAggrVerificationKey` now includes `MerkleTreeCommitmentBatchCompat`.

## Performance

We have run the benchmarks on an Apple M1 Pro machine with 16 GB of RAM, on macOS 12.6.

```shell
+----------------------+
| Size of benchmarks   |
+----------------------+
| Results obtained by using the parameters suggested in paper.
+----------------------+
+----------------------+
| Aggregate signatures |
+----------------------+
+----------------------+
| Hash: Blake2b 512    |
+----------------------+
k = 445 | m = 2728 | nr parties = 3000; 118760 bytes (old version = 356632 bytes)
+----------------------+
| Hash: Blake2b 256    |
+----------------------+
k = 445 | m = 2728 | nr parties = 3000; 99384 bytes (old version = 222536 bytes)
+----------------------+
+----------------------+
| Aggregate signatures |
+----------------------+
| Hash: Blake2b 512    |
+----------------------+
k = 554 | m = 3597 | nr parties = 3000; 133936 bytes (old version = 419808 bytes)
+----------------------+
| Hash: Blake2b 256    |
+----------------------+
k = 554 | m = 3597 | nr parties = 3000; 113728 bytes (old version = 261488 bytes)

```

**Note that single signatures in batch compat version does not depend on any variable and size of an individual signature is `176` bytes.**

```shell
STM/Blake2b/Key registration/k: 25, m: 150, nr_parties: 300
                        time:   [409.70 ms 426.81 ms 446.30 ms]
STM/Blake2b/Play all lotteries/k: 25, m: 150, nr_parties: 300
                        time:   [696.58 µs 697.62 µs 698.75 µs]
STM/Blake2b/Aggregation/k: 25, m: 150, nr_parties: 300
                        time:   [18.765 ms 18.775 ms 18.785 ms]
STM/Blake2b/Verification/k: 25, m: 150, nr_parties: 300
                        time:   [2.1577 ms 2.1715 ms 2.1915 ms]

STM/Blake2b/Key registration/k: 250, m: 1523, nr_parties: 2000
                        time:   [2.5807 s 2.5880 s 2.5961 s]
STM/Blake2b/Play all lotteries/k: 250, m: 1523, nr_parties: 2000
                        time:   [5.9318 ms 5.9447 ms 5.9582 ms]
STM/Blake2b/Aggregation/k: 250, m: 1523, nr_parties: 2000
                        time:   [190.81 ms 191.15 ms 191.54 ms]
STM/Blake2b/Verification/k: 250, m: 1523, nr_parties: 2000
                        time:   [13.944 ms 14.010 ms 14.077 ms]


```


mithril-core/src/merkle_tree.rs
```diff
@@ -339,52 +628,84 @@ mod tests {
...
+    }
+
+    #[test]
+    fn test_bytes_tree_commitment((t, values) in arb_tree(5)) {
+        let encoded = bincode::serialize(&t.to_commitment()).unwrap();
+        let decoded: MerkleTreeCommitment::<Blake2b<U32>> = bincode::deserialize(&encoded).unwrap();
+        let tree_commitment = MerkleTree::<Blake2b<U32>>::create(&values).to_commitment();
+        assert_eq!(tree_commitment.root, decoded.root);
+    }
+
```
iquerejeta, line 675: We should also test Ser/Deser of the tree batch commitment. They're different right? And going over the tests, I saw that we don't have tests for failures. We should add one of those for the path verification.

mithril-core/src/stm_batch_compat.rs
```diff
@@ -0,0 +1,932 @@
...
+use std::convert::{From, TryFrom, TryInto};
+use std::hash::{Hash, Hasher};
+use std::marker::PhantomData;
+
+/// Wrapper for StmClerk
+pub type StmClerkBatchCompact<D> = StmClerk<D>;
+
+/// Wrapper for StmInitializer
+pub type StmInitializerBatchCompat = StmInitializer;
+
```
iquerejeta, line 22: Why do we need these wrappers/alias?

mithril-core/src/stm_batch_compat.rs
```diff
@@ -0,0 +1,932 @@
...
+/// This ensures that a `MerkleTree` root is not computed before all participants have registered.
+#[derive(Debug, Clone)]
+pub struct StmSignerBatchCompat<D: Digest> {
+    mt_index: u64,
+    stake: Stake,
+    params: StmParameters,
+    sk: SigningKey,
+    vk: StmVerificationKey,
+    closed_reg: ClosedKeyReg<D>,
+}
```
iquerejeta, line 34: This is the exact same structure as `StmSigner`. Why do we need a new structure?

mithril-core/src/stm_batch_compat.rs
```diff
@@ -0,0 +1,932 @@
...
+
+        Ok(StmSignerBatchCompat {
+            mt_index: my_index.unwrap(),
+            stake: self.stake,
+            params: self.params,
+            sk: self.sk,
+            vk: self.pk.vk,
+            closed_reg,
+        })
+    }
```
iquerejeta, line 108: This seems like the same function as in stm.rs. Why duplicate?

mithril-core/src/stm_batch_compat.rs
```diff
@@ -0,0 +1,932 @@
...
+}
+
+impl<D: Clone + Digest + FixedOutput> StmSignerBatchCompat<D> {
+    /// This function produces a signature following the description of Section 2.4.
+    /// Once the signature is produced, this function checks whether any index in `[0,..,self.params.m]`
+    /// wins the lottery by evaluating the dense mapping.
+    /// It records all the winning indexes in `Self.indexes`.
+    /// If it wins at least one lottery, it produces a list of indexes of merkle path for its corresponding `(VerificationKey, Stake)`.
```
iquerejeta, line 116: I believe this is no longer the case

mithril-core/src/stm_batch_compat.rs
```diff
@@ -0,0 +1,932 @@
...
+}
+
+impl<D: Digest + Clone + FixedOutput> StmClerkBatchCompact<D> {
+    /// Create a Clerk from a signer.
+    pub fn from_signer_batch_compat(signer: &StmSignerBatchCompat<D>) -> Self {
+        Self {
+            params: signer.params,
+            closed_reg: signer.closed_reg.clone(),
+        }
+    }
```
iquerejeta, line 163: Can we use the original?

mithril-core/src/stm_batch_compat.rs
```diff
@@ -0,0 +1,932 @@
...
+            }
+        }
+
+        Err(AggregationError::NotEnoughSignatures(count, self.params.k))
+    }
+
+    /// Compute the `StmAggrVerificationKey` related to the used registration.
+    pub fn compute_avk_batch_compat(&self) -> StmAggrVerificationKey<D> {
+        StmAggrVerificationKey::from(&self.closed_reg)
+    }
```
iquerejeta, line 276: Can we use the original?

mithril-core/src/stm_batch_compat.rs
```diff
@@ -0,0 +1,932 @@
...
+}
+
+impl<D: Clone + Digest + FixedOutput> StmSigBatchCompat<D> {
+    /// Verify an stm signature by checking that the lottery was won, the merkle path is correct,
```
iquerejeta, line 280: We don't check this anymore.

mithril-core/src/error.rs
```diff
@@ -73,6 +73,10 @@ pub enum StmSignatureError<D: Digest> {
     /// This error occurs when the the serialization of the raw bytes failed
     #[error("Invalid bytes")]
     SerializationError,
+
+    /// Invalid merkle batch path
+    #[error("Batch path does not verify against root")]
+    BatchPathInvalid,
```
iquerejeta, line 79: Do we need this message if we have the error `PathInvalid` above?

mithril-core/src/error.rs
```diff
@@ -89,14 +93,18 @@ pub enum AggregationError {
...
-pub enum MerkleTreeError<D: Digest> {
+pub enum MerkleTreeError<D: Digest + FixedOutput> {
     /// Serialization error
     #[error("Serialization of a merkle tree failed")]
     SerializationError,
 
     /// Invalid merkle path
     #[error("Path does not verify against root")]
-    PathInvalid(Path<D>),
+    PathInvalid(BatchPath<D>),
```
iquerejeta, line 103: This should not be the Batch Path, but the normal path.

mithril-core/src/error.rs
```diff
@@ -89,14 +93,18 @@ pub enum AggregationError {
...
+
+    /// Invalid merkle batch path
+    #[error("Batch path does not verify against root")]
+    BatchPathInvalid,
```
iquerejeta, line 107: Here we can include the `BatchPath<D>` as an input.

mithril-core/src/stm.rs
```diff
@@ -200,30 +204,36 @@ pub struct StmSig<D: Clone + Digest> {
     pub stake: Stake,
     /// The index(es) for which the signature is valid
     pub indexes: Vec<Index>,
-    /// The path through the MerkleTree for this party.
-    pub path: Path<D>,
+    /// Merkle tree index of the signer.
+    pub signer_index: Option<Index>,
```
iquerejeta, line 208: Why option?

mithril-core/src/stm.rs
```diff
@@ -200,30 +204,36 @@ pub struct StmSig<D: Clone + Digest> {
...
+    hasher: PhantomData<D>,
```
iquerejeta, line 209: Feels like the Signer no longer needs to have notion of the hash used. This should only be used by the aggregator. However, this means that errors need to be different for aggregate verification and for individual verification. We need to think this through.

mithril-core/src/stm.rs
```diff
@@ -378,16 +393,14 @@ impl<D: Clone + Digest> StmSigner<D> {
...
-                .merkle_tree
-                .get_path(self.mt_index.try_into().ok()?);
             Some(StmSig {
                 sigma,
                 pk: self.vk,
                 stake: self.stake,
                 indexes,
-                path,
+                // path: Path::default(),
+                signer_index: Option::from(self.mt_index),
```
iquerejeta, line 402: Correct notation is 
```suggestion
                signer_index: Some(self.mt_index),
```

but I still don't see why we need an option.

mithril-core/src/stm.rs
```diff
@@ -425,28 +438,37 @@ impl<D: Digest + Clone> StmClerk<D> {
...
         sigs: &[StmSig<D>],
         msg: &[u8],
     ) -> Result<StmAggrSig<D>, AggregationError> {
-        let unique_sigs = self.dedup_sigs_for_indices(msg, sigs)?;
+        let mut unique_sigs = self.dedup_sigs_for_indices(msg, sigs)?;
+        unique_sigs.sort_unstable();
+
+        let mt_index_list = unique_sigs
+            .iter()
+            .map(|sig| sig.signer_index.unwrap() as usize)
```
iquerejeta, line 455: It is not good practice to use unwrap to open `Option`s.

mithril-core/src/stm.rs
```diff
@@ -200,30 +204,36 @@ pub struct StmSig<D: Clone + Digest> {
...
-    serialize = "Path<D>: Serialize",
-    deserialize = "Path<D>: Deserialize<'de>"
+    serialize = "BatchPath<D>: Serialize",
+    deserialize = "BatchPath<D>: Deserialize<'de>"
 ))]
-pub struct StmAggrSig<D: Clone + Digest> {
+pub struct StmAggrSig<D: Clone + Digest + FixedOutput> {
     pub(crate) signatures: Vec<StmSig<D>>,
+    /// The list of unique merkle tree nodes that covers path for all signatures.
+    pub batch_proof: Option<BatchPath<D>>,
```
iquerejeta, line 236: Do we need this to be an option? Do we accept an `StmAggrSig` without a batch_proof?

mithril-core/src/stm.rs
```diff
@@ -569,67 +589,105 @@ impl<D: Clone + Digest> StmSig<D> {
...
         }
-        let pk = StmVerificationKey::from_bytes(&bytes[16 + nr_indexes * 8..])?;
-        let sigma = Signature::from_bytes(&bytes[112 + nr_indexes * 8..])?;
-        let path = Path::from_bytes(&bytes[160 + nr_indexes * 8..])?;
+
+        let offset = 16 + nr_indexes * 8;
+        let pk = StmVerificationKey::from_bytes(&bytes[offset..offset + 96])?;
+        let sigma = Signature::from_bytes(&bytes[offset + 96..offset + 144])?;
+
+        u64_bytes.copy_from_slice(&bytes[offset + 144..]);
```
iquerejeta, line 637: For consistency, you could use 
```suggestion
        u64_bytes.copy_from_slice(&bytes[offset + 144..offset + 152]);
```

mithril-core/src/stm.rs
```diff
@@ -569,67 +589,105 @@ impl<D: Clone + Digest> StmSig<D> {
...
+            hasher: Default::default(),
         })
     }
+
+    /// Compare two `StmSig` by their signers' merkle tree indexes.
+    pub fn cmp_stm_sig(&self, other: &Self) -> Ordering {
+        let result: Ordering = self.signer_index.cmp(&other.signer_index);
+        if result != Ordering::Equal {
+            return result;
+        }
```
iquerejeta, line 656: why do you have this condition if you are anyways going to return the result?

mithril-core/src/stm.rs
```diff
@@ -709,6 +771,7 @@ impl<D: Clone + Digest> StmAggrSig<D> {
     /// * Number of signatures (as u64)
     /// * Size of a signature
     /// * Signatures
+    /// * Batch prof
```
iquerejeta, line 774:
```suggestion
    /// * Batch proof
```

iquerejeta (commented):
Looks good 👍  Just one major comment, and is that of changing the merkle_tree module. We don't want to fully remove the `Path` functionality there. While we want to design STM exclusively with BatchProofs, it is worth for the merkle_tree module, to keep the functionality of individual paths. Those should come back (even if not used). We can see what to do with clippy complaints, if there are any.

iquerejeta (commented):
I didn't have time to go over the merkle_tree.rs file. But we can do that tomorrow together.

mithril-core/src/merkle_tree.rs
```diff
@@ -123,7 +145,72 @@ impl<D: Digest + Clone> Path<D> {
...
+            output.extend_from_slice(value.as_slice())
+        }
+
+        for &index in &self.indices {
+            output.extend_from_slice(&u64::try_from(index).unwrap().to_be_bytes());
+        }
+        output
+    }
+
+    /// Try to convert a byte string into a `BatchPath`.
```
iquerejeta, line 175: Let's add this todo for future reference
```suggestion
    /// Try to convert a byte string into a `BatchPath`.
    // todo: We should not panic if the size of the slice is invalid (I believe `bytes[offset + i * 8..offset + (i + 1) * 8]` will panic if bytes is not large enough. 
```

mithril-core/src/merkle_tree.rs
```diff
@@ -160,7 +247,122 @@ impl<D: Clone + Digest> MerkleTreeCommitment<D> {
...
+
+        msgp
+    }
+
+    /// Check a proof of a batched opening. The indices must be ordered.
+    ///
+    /// # Error
+    /// Returns an error if the proof is invalid.
+    // todo: Update doc.
+    // todo: Simplify the algorithm.
```
iquerejeta, line 270: Let's add for future us
```suggestion
    // todo: Simplify the algorithm.
    // todo: Maybe we want more granular errors, rather than only `BatchPathInvalid`
```

mithril-core/src/merkle_tree.rs
```diff
@@ -377,12 +667,48 @@ mod tests {
...
-            let tree_commitment = MerkleTree::<Blake2b<U32>>::create(&values).to_commitment();
+        fn test_create_batch_proof((t, values) in arb_tree(30)) {
+            let mut mt_index_list :Vec<usize> = Vec::new();
+            for (i, _v) in values.iter().enumerate() {
+                let ind = Some(i as usize);
+                mt_index_list.push(ind.unwrap());
+            }
+            mt_index_list.sort_unstable();
+            let batch_proof = Some(t.get_batched_path(mt_index_list));
+            assert!(t.to_commitment_batch_compat().check(&values, &batch_proof.unwrap()).is_ok());
```
iquerejeta, line 678: Why are you doing `Some()` and then `unwrap()`?
```suggestion
            let batch_proof = t.get_batched_path(mt_index_list);
            assert!(t.to_commitment_batch_compat().check(&values, &batch_proof).is_ok());
```

mithril-core/src/merkle_tree.rs
```diff
@@ -377,12 +667,48 @@ mod tests {
...
         #[test]
-        fn test_bytes_tree_commitment((t, values) in arb_tree(5)) {
-            let encoded = bincode::serialize(&t.to_commitment()).unwrap();
-            let decoded: MerkleTreeCommitment::<Blake2b<U32>> = bincode::deserialize(&encoded).unwrap();
-            let tree_commitment = MerkleTree::<Blake2b<U32>>::create(&values).to_commitment();
+        fn test_create_batch_proof((t, values) in arb_tree(30)) {
+            let mut mt_index_list :Vec<usize> = Vec::new();
+            for (i, _v) in values.iter().enumerate() {
+                let ind = Some(i as usize);
+                mt_index_list.push(ind.unwrap());
```
iquerejeta, line 674: Again, what is the purpose of doing `Some()` and the `unwrap()`?
```suggestion
                mt_index_list.push(i);
```

mithril-core/src/merkle_tree.rs
```diff
@@ -377,12 +667,48 @@ mod tests {
...
+            let batch_proof = Some(t.get_batched_path(mt_index_list));
+            assert!(t.to_commitment_batch_compat().check(&values, &batch_proof.unwrap()).is_ok());
+        }
+
+        #[test]
+        fn test_bytes_batch_path((t, values) in arb_tree(30)) {
+            let mut mt_index_list :Vec<usize> = Vec::new();
+            for (i, _v) in values.iter().enumerate() {
+                let ind = Some(i as usize);
+                mt_index_list.push(ind.unwrap());
```
iquerejeta, line 686:
```suggestion
                mt_index_list.push(i);
```

mithril-core/src/merkle_tree.rs
```diff
@@ -377,12 +667,48 @@ mod tests {
...
+            }
+            mt_index_list.sort_unstable();
+
+            let batch_proof = Some(t.get_batched_path(mt_index_list));
+            let bp = batch_proof.unwrap();
```
iquerejeta, line 691:
```suggestion
            let bp = t.get_batched_path(mt_index_list);
```

mithril-core/src/merkle_tree.rs
```diff
@@ -418,5 +744,30 @@ mod tests {
...
+        ) {
+            let t = MerkleTree::<Blake2b<U32>>::create(&values[1..]);
+            let mut indices = Vec::with_capacity(values.len() / 2);
+            let mut batch_values = Vec::with_capacity(values.len() / 2);
+
+            for _j in 0..values.len() / 2 {
+                let ind = i % (values.len() - 1);
+                indices.push(ind);
+                batch_values.push(values[ind]);
+            }
```
iquerejeta, line 762:
```suggestion
            let t = MerkleTree::<Blake2b<U32>>::create(&values[1..]);
            let indices = vec![i % (values.len() - 1); values.len() / 2];
            let batch_values = vec![values[i % (values.len() - 1)]; values.len() / 2];
```

jpraynaud (approved):
LGTM :+1:

PR #486, opened 2022-09-07
Align signer & aggregator state machines logs

Alenar:
With this PR logs from Signers & Aggregator state machines will use the same semantic.

I've a question: should we pass to info level all logs from the `cycle` methods of both state machines ? I think it would make sense to show what's happening without flooding the logs too much, else on info only level most logs would only be `new cycle [...] current_state: [...]`.

### DIFF
--- a/mithril-aggregator/src/runtime/state_machine.rs
+++ b/mithril-aggregator/src/runtime/state_machine.rs
@@ -94,7 +94,7 @@ impl AggregatorRuntime {
 
         loop {
             if let Err(e) = self.cycle().await {
-                error!("state machine: an error occured: "; "error" => ?e);
+                error!("state machine: an error occurred: "; "error" => ?e);
             }
 
             info!("Sleeping for {} ms", self.state_sleep.as_millis());
@@ -116,7 +116,7 @@ impl AggregatorRuntime {
                         .compare_to_older(state.current_beacon.as_ref().unwrap())?
                         .is_new_beacon()
                 {
-                    debug!(
+                    info!(
                         "→ new Beacon settings found, trying to transition to READY";
                         "new_beacon" => ?chain_beacon
                     );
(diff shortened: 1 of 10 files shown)


mithril-aggregator/src/runtime/state_machine.rs
```diff
@@ -84,14 +79,22 @@ impl AggregatorRuntime {
...
     /// Launches an infinite loop ticking the state machine.
     pub async fn run(&mut self) {
-        info!("Starting runtime");
-        debug!("current state: {}", self.state);
+        info!("state machine: launching");
 
         loop {
             if let Err(e) = self.cycle().await {
-                error!("{:?}", e)
+                error!("state machine: an error occured: "; "error" => ?e);
```
jpraynaud, line 97:
```suggestion
                error!("state machine: an error occurred: "; "error" => ?e);
```

mithril-aggregator/src/runtime/state_machine.rs
```diff
@@ -150,12 +156,13 @@ impl AggregatorRuntime {
                     .await?
                 {
                     // READY > READY
-                    info!("a certificate already exist for this beacon"; "beacon" => ?state.current_beacon);
+                    info!(" ⋅ a certificate already exist for this beacon, waiting…"; "beacon" => ?state.current_beacon);
```
jpraynaud, line 159:
```suggestion
                    info!(" ⋅ a certificate already exists for this beacon, waiting…"; "beacon" => ?state.current_beacon);
```

mithril-aggregator/src/runtime/state_machine.rs
```diff
@@ -169,22 +176,19 @@ impl AggregatorRuntime {
...
                 } else if self.runner.is_multisig_created().await? {
-                    trace!("new multi-signature found");
+                    debug!("→ a multi-signature have been created, build a snapshot & a certificate and transitioning back to IDLE");
                     let new_state = self
                         .transition_from_signing_to_idle_multisignature(state)
                         .await?;
                     self.state = AggregatorState::Idle(new_state);
                 } else {
-                    trace!("nothing to do in SIGNING state")
+                    debug!(" ⋅ not enough signature yet to aggregate a mult-signature, waiting…");
```
jpraynaud, line 191:
```suggestion
                    debug!(" ⋅ not enough signature yet to aggregate a multi-signature, waiting…");
```

jpraynaud (approved):
LGTM :+1:

ghubertpalo (approved):

PR #490, opened 2022-09-08
Upgrade dependencies & enhance makefiles

Alenar:
- Upgrade to latest version dependencies of all cargo projects
- Update the workspace `Cargo.lock`
- CI:
  - :warning: if clippy raise a warning the pipeline will now fail
  - add `--all-targets` to the clippy check so it also check the tests
  - avoid multiple clippy warning for the same problem on shared dependencies by using the `--no-deps` parameter
- Makefiles: Enhance the `make check` target:
  - add `cargo fmt --check` to check code formating
  - add `--all-targets` to the clippy check so it also check the tests

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -95,7 +95,7 @@ jobs:
         with:
           name: clippy-${{ matrix.project }}
           token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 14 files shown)


jpraynaud (approved):
LGTM :+1:

PR #493, opened 2022-09-08
Enhance explorer

Alenar:
## Content

Two changes to the explorer:
- A better message in the modal for Genesis certificate.
- Add a button linking to the source json to all shown data.

## Comments

- :notebook: since the `/pending-certificate` route returns a 204 when there's none and that opening a tab to a 204 resource lead to a "about:blank" url, this new button is only shown for pending certificate when one is available.
- :notebook: I disabled the lint warning when using directly a `img` attribute instead of the nextJs `Image` component since it doesn't handle correctly svg icons _(using 'fill' make them gigantic and they are not centered when forcing heigh/width)_ and one is used for the new button.


jpraynaud (approved):
LGTM :+1:

PR #494, opened 2022-09-09
Implement filtered Signers in Certificate

### LINKED ISSUE #495: Record 'contributing' Signers only in Certificate
### Issue
The `signers` field of the `Certificate` is recording all the **eligible** signers of the epoch which is confusing.
Instead, the `signers` field should include only the Signers that have **contributed** to creating the multi-signature (i.e. those who have sent valid individual signature(s))

### Tasks
- [x] Filter the signers and keep only the ones that have **contributed** when creating a certificate
- [x] Reduce runtime interval of Mithril Signers hosted on GCP to `240s` instead of `60s`. This will avoid having always the same signers contributing (until we update the `Protocol Parameters`)

jpraynaud:
## Content
This PR includes:
* Update the `Certificate` creation so that only Signers with valid `Single Signatures` are included in its `signers` field 
* Reduce the runtime interval of the GCP hosted Signers from `60s` to `240s`

## Pre-submit checklist
- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation

### Issue(s)
Closes #495

### DIFF
--- a/docs/blog/2022-09-07-genesis-certificate-feature.md
+++ b/docs/blog/2022-09-07-genesis-certificate-feature.md
@@ -2,7 +2,7 @@
 title: Genesis Certificate support added
 authors:
 - name: Mithril Team
-tags: [genesis, breaking-change]
+tags: [genesis, certificate, breaking-change]
 ---
 
 **Update**: The PR has been merged and the feature is being deployed on the GCP Mithril Aggregator.
--- a/docs/blog/2022-09-12-certificate-signers-list.md
+++ b/docs/blog/2022-09-12-certificate-signers-list.md
@@ -1,0 +1,22 @@
+---
+title: Signers list computation in Certificates
+authors:
+- name: Mithril Team
+tags: [certificate]
+---
+
+### The way the Signers list is computed inside a Certificate on the Mithril Aggregator is changing
+
+**PR**: `Implement filtered Signers in Certificate` [#494](https://github.com/input-output-hk/mithril/pull/494)
+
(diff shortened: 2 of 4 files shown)


Alenar (approved):

mithril-common/src/entities/signer.rs
```diff
@@ -32,6 +34,26 @@ impl Signer {
     }
 }
 
+impl PartialOrd for Signer {
```
ghubertpalo, line 37: I do not get the business meaning of implementing a `Ord` on Signer. If it is for display purpose, it would be better the handle this on the display side, what do you think?

mithril-common/src/entities/signer.rs
```diff
@@ -68,6 +90,34 @@ impl SignerWithStake {
     }
 }
 
+impl PartialOrd for SignerWithStake {
```
ghubertpalo, line 93: Same as above.

ghubertpalo (changes requested):

mithril-common/src/entities/signer.rs
jpraynaud: This ensures that the `Certificate` hash is computed deterministically, which is not the case if the order is not guaranteed

mithril-common/src/entities/signer.rs
jpraynaud: This ensures that the `Certificate` hash is computed deterministically, which is not the case if the order is not guaranteed

ghubertpalo (approved):

PR #506, opened 2022-09-15
Update PR template

jpraynaud:
## Content
This PR modifies the PR template and removes the first part that is not used.

## Pre-submit checklist

- Branch
- PR
  - [ ] Self-reviewed the diff
  - [ ] Reviewer requested

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,11 +1,3 @@
----
-name: Pull Request Standard
-about: You want to include modifications to the code base
-title: ''
-labels: 'Pull Request'
-assignees: ''
----
-
 ## Content
 <!-- Explain the reason for this change, if a feature is added, a bug is fixed, ... -->
 This PR includes...


.github/pull_request_template.md
```diff
@@ -1,11 +1,3 @@
----
-name: Pull Request Standard
-about: You want to include modifications to the code base
-title: ''
```
ghubertpalo, line 4: Maybe just keep `title`?

jpraynaud: This is a bit redundant with the already existing title of the PR :slightly_smiling_face: 

Also it looks like this part that of the template which is between the `---` delimiters is used only in the case of multiple PR templates in order to show a select menu (as in the `.github/ISSUE_TEMPLATE` dir and the [new issue page](https://github.com/input-output-hk/mithril/issues/new/choose))

ghubertpalo (approved):

PR #531, opened 2022-09-29
Batch verify

### LINKED ISSUE #530: Explore batch verification of mithril certificate chain
## Why
Certificate chain verification could be made more efficient

## What
When verifying a chain of `n` certificates, we currently verify each individually. This means that we compute the final check for each `n` certificate, resulting in `2n` pairing evaluations. We can improve this by batching this check, resulting in `n + 1` pairing evaluations. 

## How
Suppose that you received `n` STM aggregate signatures, `S_1,…, S_n`, where each signature S_i is the aggregate of `k` multi-signatures. For each `n` STM signature, let `PK_i` be it's aggregate public key.  For each STM signature, we need to check that the following equation holds, `e(S_i, G_2) = e(M_i, PK_i)`, where `G_2` is the generator of **G2**, and `M_i` is the message signed by signature `i`. 

The trivial way to verify the batch would be to compute `n` equality checks, resulting in `2n` pairing evaluations. However, we can do better by aggregating all checks into a single one as follows:

`e(\sum_{i=1}^n S_i, G_2) = \prod_{i=1}^n e(M_i, PK_i)`

reducing the number of pairing evaluations to `n + 1`, in exchange of `n` additions in **G1** (a much cheaper operation). 

**NOTE:** This batch verification (without multiplying each `S_i` and `M_i` by a random scalar) is only possible because in mithril, `S_i` and `PK_i` are already multiscalar multiplications of the underlying signatures and public keys respectively with (pseudo)random scalars. If this was not the case, we would need to multiply each `S_i` and `PK_i` with a random scalar.

iquerejeta:
## Content
This PR implements batch-verification for STM certificates, and benchmarks the results. In the final verify of the aggregated multisignatures, we check a pairing equality, meaning that for a chain of `n` STM certificates, we compute `2n` pairings. By using a batching technique, we can reduce this number to `n + 1` pairings. 

## Comments
Benchmarks show that if we only look at the final check of STM signatures (verifying that the aggregate signature is correct with respect to the aggregate public key), the gains are considerable. A x2 improvement for batches of 100 signatures: 

```
MultiSig/Batch Verification/Batch size: 1
                        time:   [586.49 µs 588.12 µs 590.48 µs]

MultiSig/Batch Verification/Batch size: 100
                        time:   [5.4062 ms 5.4660 ms 5.5350 ms]
```

However, STM signature verification, at least in the current state, requires also to compute the aggregate of the signatures and public keys. This requires two multiscalar multiplications of at least `k` (for `k` the security parameter) elements. The overhead caused by the pairing evaluation is minimised, and the gains of batch-verification are less significant: 

```
STM/Blake2b/Batch Verification/k: 25, m: 150, nr_parties: 300/batch size: 1
                        time:   [2.2606 ms 2.2674 ms 2.2747 ms]

STM/Blake2b/Batch Verification/k: 25, m: 150, nr_parties: 300/batch size: 10
                        time:   [18.056 ms 18.089 ms 18.127 ms]

STM/Blake2b/Batch Verification/k: 25, m: 150, nr_parties: 300/batch size: 20
                        time:   [34.939 ms 34.981 ms 35.029 ms]

STM/Blake2b/Batch Verification/k: 25, m: 150, nr_parties: 300/batch size: 100
                        time:   [175.11 ms 175.22 ms 175.34 ms]

STM/Blake2b/Batch Verification/k: 250, m: 1523, nr_parties: 2000/batch size: 1
                        time:   [3.9876 ms 4.0245 ms 4.0574 ms]

STM/Blake2b/Batch Verification/k: 250, m: 1523, nr_parties: 2000/batch size: 10
                        time:   [35.719 ms 35.956 ms 36.202 ms]

STM/Blake2b/Batch Verification/k: 250, m: 1523, nr_parties: 2000/batch size: 20
                        time:   [76.466 ms 78.880 ms 81.210 ms]

STM/Blake2b/Batch Verification/k: 250, m: 1523, nr_parties: 2000/batch size: 100
                        time:   [318.40 ms 319.29 ms 320.23 ms]

```
We can see that for parameters with `k = 25`, we have that the improvements are 23%, while with `k = 250`, the improvements are of 20%. However, and observation can be made that, for `k=25`, batching 10 STM signatures gives a 19% speedup, while going up to batches of size 100 only gives a ~3% improvement (wrt batches of size 10). For `k = 250`, the pattern is less significant, with ~10% and ~10% respectively. 

## Issue(s)
Closes #530


mithril-core/src/merkle_tree.rs
```diff
@@ -0,0 +1,422 @@
+//! Creation and verification of Merkle Trees
```
curiecrypt, line 1: As we talked about it this file is a leftover from rebasing.  We need to remove the entire `mithril-core` folder.

mithril-stm/benches/stm.rs
```diff
@@ -70,15 +67,97 @@ where
...
 
     group.bench_function(BenchmarkId::new("Aggregation", &param_string), |b| {
         b.iter(|| clerk.aggregate(&sigs, &msg))
     });
+}
 
-    group.bench_function(BenchmarkId::new("Verification", &param_string), |b| {
-        b.iter(|| msig.verify(&msg, &clerk.compute_avk(), &params).is_ok())
-    });
+fn batch_benches<H>(
```
curiecrypt, line 76: How do we get the size of benches for the batch version? I couldn't see how we get the output given in the PR description.

mithril-stm/src/multi_sig.rs
```diff
@@ -229,12 +231,12 @@ impl VerificationKeyPoP {
...
         use blst::{
-            blst_fp12, blst_fp12_finalverify, blst_p1_affine_generator, blst_p1_to_affine,
-            blst_p2_affine, blst_p2_affine_generator, BLST_ERROR,
+            blst_fp12, blst_fp12_finalverify, blst_p1_affine_generator, blst_p2_affine_generator,
+            BLST_ERROR,
         };
         let result = unsafe {
             let g1_p = *blst_p1_affine_generator();
-            let mvk_p = std::mem::transmute::<&BlstPk, &blst_p2_affine>(&self.vk.0);
+            let mvk_p = std::mem::transmute::<&BlstVk, &blst_p2_affine>(&self.vk.0);
```
curiecrypt, line 239: We need to have *transmute-free* `multi-sig`, after transmute PR is done, we should change all occurrences in this PR as well.

mithril-stm/src/multi_sig.rs
```diff
@@ -400,46 +402,115 @@ impl Signature {
...
-                &scalars,
-                128,
-            ),
+        let transmuted_vks: Vec<blst_p2> = vks
+            .iter()
+            .map(|vk| unsafe {
+                let mut projective_p2 = blst_p2::default();
+                blst_p2_from_affine(
+                    &mut projective_p2,
+                    &std::mem::transmute::<BlstVk, blst_p2_affine>(vk.0),
```
curiecrypt, line 442: Change transmute.

mithril-stm/src/multi_sig.rs
```diff
@@ -400,46 +402,115 @@ impl Signature {
...
+            })
+            .collect();
+
+        let transmuted_sigs: Vec<blst_p1> = signatures
+            .iter()
+            .map(|&sig| unsafe {
+                let mut projective_p1 = blst_p1::default();
+                blst_p1_from_affine(
+                    &mut projective_p1,
+                    &std::mem::transmute::<BlstSig, blst_p1_affine>(sig),
```
curiecrypt, line 454: Change transmute.

mithril-stm/src/multi_sig.rs
```diff
@@ -400,46 +402,115 @@ impl Signature {
...
+            })
+            .collect();
+
+        let grouped_vks = p2_affines::from(transmuted_vks.as_slice());
+        let grouped_sigs = p1_affines::from(transmuted_sigs.as_slice());
+
+        let aggr_vk: BlstVk = unsafe {
+            let mut affine_p2 = blst_p2_affine::default();
+            blst_p2_to_affine(&mut affine_p2, &grouped_vks.mult(scalars.as_slice(), 128));
+            std::mem::transmute::<blst_p2_affine, BlstVk>(affine_p2)
```
curiecrypt, line 466: Change transmute.

mithril-stm/src/multi_sig.rs
```diff
@@ -400,46 +402,115 @@ impl Signature {
...
+        };
+        let aggr_sig: BlstSig = unsafe {
+            let mut affine_p1 = blst_p1_affine::default();
+            blst_p1_to_affine(&mut affine_p1, &grouped_sigs.mult(scalars.as_slice(), 128));
+            std::mem::transmute::<blst_p1_affine, BlstSig>(affine_p1)
```
curiecrypt, line 471: Change transmute.

curiecrypt (changes requested):
I think this PR can be merged after the minor comments I have made are resolved. Also, we can update the stm `READM.md` to cover batch version descriptions and results.

mithril-stm/benches/stm.rs
iquerejeta: What do you mean the size of benches?

mithril-stm/src/multi_sig.rs
iquerejeta: Yes. My suggestion is that we merge this, and then we work on removing transumute calls from another PR.

mithril-stm/benches/stm.rs
iquerejeta: If you run `cargo bench` you get the output presented in the PR description for STM signatures. If you also want to run the benches for the multi-sig (a module which is not exposed) you need to use the feature benchmark-internals, `cargo bench --features benchmark-internals`. Does this solve the concern?

curiecrypt (approved):

mithril-stm/src/error.rs
```diff
@@ -25,6 +25,10 @@ pub enum MultiSignatureError {
     /// Incorrect proof of possession
     #[error("Key with invalid PoP")]
     KeyInvalid(Box<VerificationKeyPoP>),
+
+    /// At least one signature in the batch is invalid
+    #[error("One signature in the batch is invalid")]
+    BatchInvalid,
```
jpraynaud, line 31: It looks like this error is never used.
Also would it be possible to add the index of the signature that failed?

jpraynaud (approved):
LGTM :+1: 

You should also probably update the version of the crate to `0.2.1`

mithril-stm/src/error.rs
iquerejeta: I've included the error, but unfortunately we cannot add the index of the signature that failed, as we don't know. We simply know that the batch is invalid. If we get this error, we need to individually verify each signature.

PR #559, opened 2022-10-21
remove json store adapter

ghubertpalo:
## Content
This PR removes the JSON store adapter. This adapter is not used anymore and appears to be buggy.


jpraynaud (approved):
LGTM :+1: 
(when no more docs warnings)

PR #609, opened 2022-11-23
Fix KES period verification

### LINKED ISSUE #548: Signer registration fails with key certification mode
## Issue
The Signer registration fails when using the experimental key certification mode. The Aggregator returns a `KESInvalidSignature` error.

After investigation, it appears that the problem is linked with the computation of the `KES Period` when the `KES Signature` is verified.

jpraynaud:
## Content
This PR includes a restriction of the KES period range verification to `[max(0,kes_period-1), min(64,kes_period+1)]`.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Issue(s)
Closes #548

### DIFF
--- a/mithril-common/src/crypto_helper/cardano/key_certification.rs
+++ b/mithril-common/src/crypto_helper/cardano/key_certification.rs
@@ -48,8 +48,8 @@ pub enum ProtocolRegistrationErrorWrapper {
     OpCertInvalid,
 
     /// Error raised when a KES Signature verification fails
-    #[error("KES signature verification error")]
-    KesSignatureInvalid,
+    #[error("KES signature verification error: CurrentKesPeriod={0}, StartKesPeriod={1}")]
+    KesSignatureInvalid(usize, u64),
 
     /// Error raised when a KES Signature is needed but not provided
     #[error("missing KES signature")]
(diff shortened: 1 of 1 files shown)


mithril-common/src/crypto_helper/cardano/key_certification.rs
```diff
@@ -221,18 +221,17 @@ impl KeyRegWrapper {
...
                 .validate()
                 .map_err(|_| ProtocolRegistrationErrorWrapper::OpCertInvalid)?;
-            // TODO: List of eligible indices to be defined by CurrentKesPeriod and StartKesPeriod
             let mut pool_id = None;
             let sig = kes_sig.ok_or(ProtocolRegistrationErrorWrapper::KesSignatureMissing)?;
-            for kes_period_try in 0..64 {
+            let kes_period =
+                kes_period.ok_or(ProtocolRegistrationErrorWrapper::KesPeriodMissing)?;
+            let kes_period_try_min = std::cmp::max(0, kes_period as i32 - 1) as usize;
+            let kes_period_try_max = std::cmp::min(64, kes_period as i32 + 1) as usize;
```
iquerejeta, line 229: Do you need to use i32 here? The original type of kes should work, as we will not result with a negative num.

mithril-common/src/crypto_helper/cardano/key_certification.rs
```diff
@@ -48,8 +48,8 @@ pub enum ProtocolRegistrationErrorWrapper {
     OpCertInvalid,
 
     /// Error raised when a KES Signature verification fails
-    #[error("KES signature verification error")]
-    KesSignatureInvalid,
+    #[error("KES signature verification error: CurrentKesPeriod={0}, StartKesPeriod={0}")]
```
iquerejeta, line 51: Shouldn't this be: 
```suggestion
    #[error("KES signature verification error: CurrentKesPeriod={0}, StartKesPeriod={1}")]
```

iquerejeta (approved):
See if my minor comments make sense, otherwise looks good to merge 👍

iquerejeta (approved):
LGTM 👍

PR #620, opened 2022-11-28
Simplify signature

### LINKED ISSUE #619: Remove VerificationKey and Stake from individual signature
## Why
After the change of removing the merkle path from the individual signature, the protocol relies on the aggregator to have the full Merkle tree to aggregate correctly. Therefore, we don't need the individual signature to contain the public key and stake.

## What
Currently, the signature is defined as:

```
pub struct StmSig {
    pub sigma: Signature,
    pub pk: StmVerificationKey,
    pub stake: Stake,
    pub indexes: Vec<Index>,
    pub signer_index: Index,
}
```

and should be changed to:

```
pub struct StmSig {
    pub sigma: Signature,
    pub indexes: Vec<Index>,
    pub signer_index: Index,
}
```

iquerejeta:
## Content
This PR removes the verification key and the stake from the signature. These values are no longer required, as individual signatures no longer contain the merkle paths (it is now responsibility of the aggregator to include these values in the certificate). This required some changes in the multi_signer and single_signer, as these folders were verifying single signatures. 

Now, in order to verify aggregate signatures, they include not only the individual signatures but also the corresponding `RegParty`, to be able to verify the batched membership proof (batched merkle path).

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Issue(s)
This PR closes #619 .

### DIFF
--- a/mithril-aggregator/src/multi_signer.rs
+++ b/mithril-aggregator/src/multi_signer.rs
@@ -36,6 +36,10 @@ pub enum ProtocolError {
     #[error("signer already registered")]
     ExistingSigner(),
 
+    /// Signer was not registered.
+    #[error("signer did not register")]
+    UnregisteredParty(),
+
     /// Signer registration failed.
     #[error("signer registration failed")]
     FailedSignerRegistration(#[from] ProtocolRegistrationError),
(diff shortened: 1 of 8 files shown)


mithril-aggregator/src/multi_signer.rs
```diff
@@ -703,34 +703,54 @@ impl MultiSigner for MultiSignerImpl {
...
-            .current_beacon
-            .as_ref()
-            .ok_or_else(ProtocolError::UnavailableBeacon)?;
+            .map_err(ProtocolError::Codec)?;
+
+        let avk = clerk.compute_avk();
+
+        // If there is no reg_party, then we simply received a signature from a non-registered
+        // party, and we can ignore the request.
+        if let Some((vk, stake)) = clerk.get_reg_party(&signature.signer_index) {
```
jpraynaud, line 720:
```suggestion
        let (vk, stake) = clerk.get_reg_party(&signature.signer_index).ok_or_else(|| ProtocolError::UnregisteredParty)?
```
This should avoid to create the `if let` and would return an error instead of `Ok(())`

mithril-stm/src/stm.rs
```diff
@@ -535,6 +555,15 @@ impl<D: Digest + Clone + FixedOutput> StmClerk<D> {
     pub fn compute_avk(&self) -> StmAggrVerificationKey<D> {
         StmAggrVerificationKey::from(&self.closed_reg)
     }
+
+    /// Get the (VK, stake) of a party given it's index.
```
jpraynaud, line 559:
```suggestion
    /// Get the (VK, stake) of a party given its index.
```

mithril-stm/src/stm.rs
```diff
@@ -535,6 +555,15 @@ impl<D: Digest + Clone + FixedOutput> StmClerk<D> {
...
+    pub fn get_reg_party(&self, party_index: &Index) -> Option<(StmVerificationKey, Stake)> {
+        if *party_index as usize >= self.closed_reg.reg_parties.len() {
+            return None;
+        }
+
+        Some(self.closed_reg.reg_parties[*party_index as usize].into())
```
jpraynaud, line 565:
```suggestion
       self.closed_reg.reg_parties.get(*party_index as usize).map(|r| r.into())
```
Maybe this would be more readable? WDYT?

jpraynaud (approved):
LGTM :+1: 

I think this is a breaking change as previous multi-signature will not be compatible with the new format. This means that we will probably have to re-genesis the certificate chains of the Mithril networks. Also all the SPOs will have to recompile their node. It is probably a good time to   update the `PROTOCOL_VERSION` (and maybe enforce versions compatibility)? Let's wait before we merge it, WDYT?

curiecrypt (approved):
I think the changes look good, the only change I should request is to update the size bench results in the readme. Since we exclude pk and stake from StmSig, we have no longer the size as 176 bytes for a single sig with 1 winning ticket.

We should update the related line as follows: 
```
Note that the size of an individual signature with one valid index is **72 bytes** and increases linearly in the length of valid indices (where an index is 8 bytes).
```

iquerejeta:
Good catch. But I count 64 bytes + linear increase. 48 from sigma + 8 from the signer index + 8 from _at least_ one index. Do you get the same?
cc: @curiecrypt

curiecrypt:
Yes, I got the same. I wrote 72 bytes because if we create a signature it must include at least one winning index. So, 72 bytes = sigma + signer index + winning index @iquerejeta

curiecrypt:
The extra 8 bytes coming from the number of indexes are included in serialized version of StmSig. @iquerejeta

iquerejeta:
Right! good catch.

curiecrypt (approved):

PR #613, opened 2022-11-24
Doc & Build CI: Switch to pull_request trigger

Alenar:
## Content

This PR switch the event trigger used for the `ci` and `docs` workflows to `pull_request` (from `push`).

Using this trigger add the ability to run those workflows on PR from forks (if we approve them, see: [Approving workflow runs from public forks](https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks)).

## Pre-submit checklist

- Branch
  - [x] Commit sequence broadly makes sense
  - [x] Key commits have useful messages
- PR
  - [x] Self-reviewed the diff
  - [x] Useful pull request description
  - [x] Reviewer requested

## Issue(s)
Fix #597

Issue #597, opened 2022-11-16
CI does not trigger for PR from forks


jpraynaud:
## Issue
The CI workflow does not trigger for PR produced from a fork of the repository, thus not producing artifacts for the commit when it is merged
See https://github.com/input-output-hk/mithril/pull/591 & https://github.com/input-output-hk/mithril/actions/runs/3477670358

## Todo:

- [ ] 1. Switch `pull_request` event trigger for the `ci` and `docs` workflows (see #613)
- [ ] 2. Make a PR from a remote repository to see if this issue is fixed
- [ ] 3. Make a test release to see if the current process still works with packages from workflows using the `pull_request` event trigger.

PR #629, opened 2022-11-30
Deploy using push trigger

Alenar:
## Content

This PR replace the "pull request closed" trigger with a push trigger on the `main` or `hotfix*` branches. It also disable running the `deploy-testing` job when the triggering pull request comes from a fork (since [they don't have all the secrets needed to run it](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflows-in-forked-repositories)).

This means that only merge coming from this repository branches will trigger a deployment and that only those merge can be tagged for a version (else the `pre-release` workflow will fail since no artifacts would have been built).
This is ok since we need anyway to do a commit to update the version and/or the documentation each time we do a new distribution release.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Comments
Hopefully having a push trigger on the main branch will produce proper cache that will be correctly reused by the PR workflow :pray:.

This PR also remove the `dry-run` arg to the crate deployment since a valid `CRATES_IO_API_TOKEN` have been added to this repository secret.

## Issue(s)

Relates to #597 and #613

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,12 +1,18 @@
 name: CI
 
 on:
+  push:
+    branches:
+      - 'main'
+      - 'hotfix**'
   pull_request:
-    types: [opened, reopened, synchronize, closed]
+    types: [opened, reopened, synchronize]
     paths-ignore: # ignore docs only changes since they use a dedicated workflows: docs.yml
       - 'docs/**'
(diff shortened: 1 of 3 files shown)


jpraynaud (approved):
LGTM :+1:

PR #641, opened 2022-12-07
check API version

### LINKED ISSUE #633: Enforcement of API Protocol versions in Client/Signer/Aggregator
## Issue
In order to help signers and client operators know if they are running a stale version of the API protocol and facilitate the overall upgrade process of the nodes, we need to put in place a verification mechanism. This mechanism must be optional, if no version is provider by the HTTP client, the verification is skipped.

We need to have signers and client stop and display an error message requesting for an update of the node in case of incompatible `API_VERSION` detected between signer & aggregator or client & aggregator. 

For instance, the Aggregator could return a specific HTTP error code and this would trigger an exit from the signer/client. 

Protocol versions are compatible if they share the same minor.

## To do

- [ ] If Aggregator version !== Signer/Client version: display an error "You need to upgrade your node to version X.Y"
- [ ] If the version is not provided, the version is ignored.
- [ ] Add units tests or maybe integration tests
- [ ] Maybe add a dev blog post (if it brings added value to SPOs)

Relates to #565

ghubertpalo:
## Content
This PR adds an API version check when requesting the Aggregator. If the API version does not match, the HTTP call is rejected.

## Pre-submit checklist

- Branch
  - [X] Tests are provided (if possible)
  - [X] Commit sequence broadly makes sense
  - [X] Key commits have useful messages
- PR
  - [X] No clippy warnings in the CI
  - [X] Self-reviewed the diff
  - [X] Useful pull request description
  - [X] Reviewer requested
- Documentation

## Comments
None

## Issue(s)

Closes #633

### DIFF
--- a/mithril-aggregator/src/http_server/routes/router.rs
+++ b/mithril-aggregator/src/http_server/routes/router.rs
@@ -24,13 +24,14 @@ pub fn routes(
         "mithril-api-version",
         HeaderValue::from_static(MITHRIL_API_VERSION),
     );
-
+    let header_must_be = warp::header::exact("API_VERSION", MITHRIL_API_VERSION);
     warp::any().and(warp::path(SERVER_BASE_PATH)).and(
         certificate_routes::routes(dependency_manager.clone())
             .or(snapshot_routes::routes(dependency_manager.clone()))
             .or(signer_routes::routes(dependency_manager.clone()))
             .or(signatures_routes::routes(dependency_manager.clone()))
             .or(epoch_routes::routes(dependency_manager))
+            .and(header_must_be)
             .with(cors)
             .with(warp::reply::with::headers(headers)),
     )
--- a/mithril-aggregator/src/http_server/routes/router.rs
+++ b/mithril-aggregator/src/http_server/routes/router.rs
@@ -66,3 +66,40 @@ pub async fn handle_custom(reject: Rejection) -> Result<impl Reply, Rejection> {
         Err(reject)
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[tokio::test]
+    async fn test_no_version() {
+        let filters = header_must_be();
+        warp::test::request()
+            .path("/aggregator/whatever")
+            .filter(&filters)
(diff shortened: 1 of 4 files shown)


mithril-aggregator/src/http_server/routes/router.rs
```diff
@@ -24,14 +31,75 @@ pub fn routes(
...
+                .or(signer_routes::routes(dependency_manager.clone()))
+                .or(signatures_routes::routes(dependency_manager.clone()))
+                .or(epoch_routes::routes(dependency_manager))
+                .with(cors),
+        )
+        .recover(handle_custom)
+        .with(warp::reply::with::headers(headers))
+}
+
+/// API Version verification
```
ghubertpalo, line 49: change the comment according to what the function does

mithril-aggregator/src/http_server/routes/router.rs
```diff
@@ -24,14 +31,75 @@ pub fn routes(
...
+            match maybe_header {
+                None => Ok(()),
+                Some(version) if version == MITHRIL_API_VERSION => Ok(()),
+                Some(_version) => Err(warp::reject::custom(VersionMismatchError)),
+            }
+        })
+        .untuple_one()
+}
+
+pub async fn handle_custom(reject: Rejection) -> Result<impl Reply, Rejection> {
```
ghubertpalo, line 62: handle custom rejection codes

ghubertpalo, line 62: not need to be public

ghubertpalo:
Catch the HTTP error in the signer and the client to display an according error message.

jpraynaud:
We probably need to update the `OpenAPI` specification to add the new error code.

Also I'm not sure that we should use [`412`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/412) HTTP code. Maybe [`426`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/426) is more appropriate. WDYT?

jpraynaud (approved):
LGTM :+1:

PR #674, opened 2023-01-03
Update kes

iquerejeta:
## Content
This PR adapts the code to the new version of the KES library. Signatures and Public keys are compatible with older versions. However, secret keys are not. This should not be a backwards compatibility problem, as new nodes will still be compatible with older versions of the protocol and viceversa. However, secret keys serialised with the current library cannot be deserialised with an old version (or by the haskell node). The opposite, however, does work. New nodes will be able to deserialise previous file formats (or ones generated by the haskell node). 

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

### DIFF
--- a/mithril-common/src/crypto_helper/cardano/key_certification.rs
+++ b/mithril-common/src/crypto_helper/cardano/key_certification.rs
@@ -29,7 +29,7 @@ use thiserror::Error;
 type D = Blake2b<U32>;
 
 /// The KES period that is used to check if the KES keys is expired
-pub type KESPeriod = usize;
+pub type KESPeriod = u32;
 
 /// New registration error
 #[derive(Error, Debug, PartialEq, Eq)]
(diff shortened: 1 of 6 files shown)


mithril-common/src/crypto_helper/cardano/key_certification.rs
```diff
@@ -119,14 +123,23 @@ impl StmInitializerWrapper {
...
+                }
 
-            Some(kes_sk.sign(&stm_initializer.verification_key().to_bytes()))
+                Some(kes_sk.sign(&stm_initializer.verification_key().to_bytes()))
+            } else {
+                return Err(ProtocolInitializerErrorWrapper::KesMismatch(
+                    kes_sk_period,
+                    provided_period,
+                ));
+            }
```
jpraynaud, line 142: Maybe you could first check that the periods don't mismatch for readability:

```
if kes_sk_period > provided_period {
  return Err(ProtocolInitializerErrorWrapper::KesMismatch(
      kes_sk_period,
      provided_period,
  ));
}

// We need to perform the evolutions
for period in kes_sk_period..provided_period {
    kes_sk
        .update()
        .map_err(|_| ProtocolInitializerErrorWrapper::KesUpdate(period))?;
}

Some(kes_sk.sign(&stm_initializer.verification_key().to_bytes()))
```

jpraynaud (approved):
LGTM :+1:

PR #812, opened 2023-03-16
Handle API version with Era Switch

### LINKED ISSUE #727: Handle API Version with Era switch
## Issue
The current API version is handled with a const. We need to be able to make a hot switch when era change (if necessary)

## Questions
- Do we need to reboot the HTTP server or a part of it when the era switch requires a change in the routing?
- How to adapt tests with multiple API specs files? (test multiple files without modifying test if not necessary with `APISpec`)
- How to adapt doc generation with multiple API specs files?
- How to handle API version update on the client (as it does not have access to the Cardano chain)?

## Parent issue
#707

jpraynaud:
## Content
This PR includes the handling of updated API version when an era switch occurs

## Comments
In order to handle a different request object after an era switch, the routers must be modified as follows:
- Import [`bytes`](https://crates.io/crates/bytes) crate in `Cargo.toml` of the aggregator
- Update the route that is concerned by replacing the filter `warp::body::json()` by `warp::body::bytes()`
- Replace the parameter that was parsed by the json filter in the route handler (e.g. `register_signer_message: RegisterSignerMessage,` with `bytes: bytes::Bytes`
- In the handler parse the bytes of the request body to the correct type with `serde_json::from_str(&String::from_utf8(bytes.to_vec()).unwrap()).unwrap()`

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Issue(s)
Closes #727


mithril-client/src/commands/download.rs
```diff
@@ -36,8 +41,18 @@ impl DownloadCommand {
             .map_err(|e| format!("configuration deserialize error: {e}"))?;
         debug!("{:?}", config);
         let runtime = Runtime::new(config.network.clone());
-        let aggregator_handler =
-            AggregatorHTTPClient::new(config.network.clone(), config.aggregator_endpoint);
+        // TODO: This does not allow the client to handle an upgraded API version of a new supported era after the switch
```
ghubertpalo, line 44: This piece of code is repeated in each client's following files, maybe it is possible to refactor this in a function.

mithril-client/src/commands/download.rs
```diff
@@ -36,8 +41,18 @@ impl DownloadCommand {
...
+        // In order to do so, we should retrieve the list of supported versions from the API Version provider and test them sequentially until one hopefully succeeds
+        let era_checker = Arc::new(EraChecker::new(
```
ghubertpalo, line 46: Maybe it would be useful to get rid of the `EraChecker` dependency which is not needed in this case to determine the API version to use.

mithril-aggregator/src/http_server/routes/snapshot_routes.rs
```diff
@@ -338,13 +344,15 @@ mod tests {
...
-            .expect("OpenAPI error");
+        for spec_file in APISpec::get_all_spec_files() {
+            APISpec::from_file(&spec_file)
+                .method(method)
+                .path(path)
+                .validate_request(&Null)
+                .unwrap()
+                .validate_response(&response)
+                .unwrap_or_else(|_| panic!("OpenAPI error in {}", spec_file));
+        }
```
ghubertpalo, line 355: maybe this is factorizable?

ghubertpalo (changes requested):
Few things to pair on IMHO.

mithril-client/src/commands/download.rs
jpraynaud: Done :heavy_check_mark:

mithril-aggregator/src/http_server/routes/snapshot_routes.rs
jpraynaud: Done :heavy_check_mark:

ghubertpalo (approved):

PR #939, opened 2023-05-25
fnv/core verifier

### LINKED ISSUE #839: Full node verifier
## Why
Mithril verifier needs to have guarantees about the validity of keys and stakes of each party since the `AVK` within a certificate should be validated. So, the signers' identities should be provided to a Mithril client. However, if the verification is done by a full node, the verifier can perform the validity check locally, since it knows the identities already.

## What
The aggregate signature and the indices of the signers are enough to validate a certificate for a full node verifier. A verification algorithm for a full-node verifier should be implemented.

----

Note that full-node verification is a different use case, it is not related to the fast-bootstrapping.

curiecrypt:
## Content
A full node verifier feature that does not require checking signers' identities externally.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation

## Comments

The idea is to provide a core verifier (previously full-node verifier) that has
- A list of eligible parties,
- Total stake in the system,

and be able to verify the signatures that are generated without the registration information, i.e., `avk`.
In this case, the signer who generates this signature has all the fields of `StmSigner` except from `closed_reg`. 
So, `closed_reg` field is updated as `Option`. If the signer wants to participate in Mithril certificate generation, then it must have the `closed_reg` and this signer is generated with the function `StmInitializer::new_signer()`. Otherwise, the signer can only generate signatures to be validated by a core verifier. We call this signer *core_signer*, and it is created with the function `StmInitializer::new_core_signer()`. Note that in this version the core signer is built upon the `StmInitializer,` but it does not need to be built upon the `StmInitializer` depending on the functionality it provides.

`StmSigner` now provides two signing functions, one for a core signature and one for a signature to participate in certificate generation. 
A core signature is generated by `StmSigner::core_sign()` function without closed registration and can be verified by the core verifier. This function simply creates a signature for a given message. `StmSigner::sign()` function calls `core_sign()` with `msgp = (Commitment || msg)`, and it produces a signature for the Mithril certificate.

Similarly, `StmSig` is updated to provide
- `core_verify`: Verifies `sigma` and checks the indices. This function verifies the `StmSig`s to be 
  verified by full node verifier.
- `verify`: Runs `core_verify` with `msgp`. This function is responsible for verifying `StmSig`s for **mithril** functionality. 

A new struct `StmSigRegParty` including `StmSig` and `RegParty` is created. Consequently, the field `signatures` of `StmAggrSig` is now of the form `Vec<StmSigRegParty>` instead of `Vec<(StmSig, RegParty)>`. This change is done to avoid duplicate codes since the core verifier and `StmAggrSig` have many common operations.

A struct `CoreVerifier` containing `eligible_parties` and `total_stake` is created. Core verifier implements the following:
- `setup`: Collects unique signers in `eligible_parties`, calculates the `total_stake`, and returns the `CoreVerifier`.
- `dedup_sigs_for_indices`: This function is adopted from `StmClerk::dedup_sigs_for_indices`, and the new function can be used by both `StmClerk` and `CoreVerifier`. 
- `preliminary_verify`: Verify all signatures whether they all won the lottery, if the indices are unique, and if the quorum is achieved.
- `collect_sigs_vks`: Collect and return the signatures and verification keys for given lists of `StmSigRegParty`.
- `verify`: Maps the signatures with the eligible parties by `map_sig_party`. Runs `dedup_sigs_for_indices` to get unique signatures. After running `preliminary_verify`, it collects sigs and vks with `collect_sigs_vks` and verifies the aggregate.

The functions of `CoreVerifier` is designed to be used by `StmAggrSig` since they have similar operations. 

`StmAggrSig::preliminary_verify` verifies all checks from signatures except for the signature verification itself. 
Indices and quorum are checked by `CoreVerifier::preliminary_verify` with `msgp`. 
It collects leaves from signatures and checks the batch proof. After the batch proof is checked, it collects and returns the signatures and verification keys to be used by aggregate verification.

#### Minor changes:
Error enums are updated for new functionality. `to_bytes` and `from_bytes` functions are implemented for `StmSigRegParty`, and they are used in `StmAggrSig`. Unused functions `StmSigner::get_closed_reg()` and `StmSigner::compute_avk()` are removed.

### ToDo
- [x] Implement module test for `CoreVerifier`.
- [x] Remove `StmClerk::dedup_sigs_for_indices()` and its related tests.
- [x] Implement proptest for `CoreVerifier::dedup_sigs_for_indices()`.
- [x] Implement full protocol test for `CoreVerifier`.
- [x] Implement benches for `CoreVerifier`.
- [x] Check the compatibility of the rest of the project with new features.
- [x] Update Mithril change log.
- [x] Update Mithril-stm README.
- [x] Update PR description.

## Issue(s)

Closes #839


mithril-stm/src/stm.rs
```diff
@@ -879,6 +973,69 @@ impl<D: Clone + Digest + FixedOutput + Send + Sync> StmAggrSig<D> {
...
+            .map(|sig| sig.sigma)
+            .collect::<Vec<Signature>>();
+        let vks = signed_parties
+            .iter()
+            .map(|reg_party| reg_party.0)
+            .collect::<Vec<VerificationKey>>();
+
+        Signature::verify_aggregate(msg.to_vec().as_slice(), &vks, &signatures)?;
+
+        Ok(())
```
iquerejeta, line 1015: This looks very similar to `preliminary_verify` and `verify` from `StmAggrSig`. We should try to put them together, and minimise code duplication. Any reason why this would not be possible?

iquerejeta (commented):
Good overall. Left some comments. However, I'm not convinced in the naming. `verify_avk` could be interpreted as verifying the aggregate key. Maybe instead of `verify` and `verify_avk`, we could call `core_verify` and `verify`. And both functions should have good documentation explaining their differences.

mithril-stm/src/stm.rs
```diff
@@ -225,6 +234,15 @@ pub struct StmAggrSig<D: Clone + Digest + FixedOutput> {
     pub batch_proof: BatchPath<D>,
 }
 
+/// Full node verifier
+pub struct FullNodeVerifier<D: Digest> {
+    /// Ordered list of registered parties.
+    pub eligible_parties: HashMap<RegParty, Index>,
+    /// Total stake of registered parties.
+    pub total_stake: Stake,
+    hasher: PhantomData<D>,
```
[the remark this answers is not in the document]
curiecrypt: We need to return `StmAggregateSignatureError<D>` in the implementation of `FullNodeverifier`.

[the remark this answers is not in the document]
iquerejeta: The hasher for the `StmAggregateSignatureError` is for the path error. However, we don't need that for the full node verifier. Maybe we can separate the errors into two enums, for example `StmFnAggregateSignatureError` and `StmAggregateSignatureError`, and make the latter depend on the former, e.g.: 
```rust
pub enum StmAggregateSignatureError<D: Digest + FixedOutput> {
    /// Signature error
    #[error("Error during verification: {0}")]
    VerificationError(StmFnAggregateSignatureError),
    /// and the rest of errors that are related to the enum
    /// [...]
}
```

mithril-stm/src/stm.rs
```diff
@@ -278,22 +295,12 @@ impl StmInitializer {
...
-    /// * the current total stake (according to the registration service)
-    /// # Error
-    /// This function fails if the initializer is not registered.
-    pub fn new_signer<D: Digest + Clone>(
-        self,
+    /// Function that checks whether the initializer is registered.
+    /// Returns the merkle tree index of the party if registered.
+    pub fn check_initializer<D: Digest + Clone>(
+        &self,
         closed_reg: ClosedKeyReg<D>,
```
iquerejeta, line 302: You are consuming the `closed_reg` for checking whether the initializer is registered. It does not look right.

mithril-stm/src/stm.rs
```diff
@@ -304,13 +311,40 @@ impl StmInitializer {
         if my_index.is_none() {
             return Err(RegisterError::UnregisteredInitializer);
         }
+        Ok(my_index)
+    }
 
-        Ok(StmSigner {
-            mt_index: my_index.unwrap(),
+    /// Returns the StmSigner of the initializer for given index.
+    pub fn new_signer(self, index: Index) -> StmSigner {
```
iquerejeta, line 318: The only way to provide an index is via a closed registration. However, we don't want to make a requirement to initialise an StmSigner to have a closed registration, as that might not happen in the full node case. 

Actually, we don't need to create an StmSigner from an StmInitializer. The reason why we have that for the `StmSignerAvk` is because we want to enforce the creation of the latter only if a closed registration exists (which is run by the initializer). However, we don't need that for the StmSigner, and therefore, we could directly have a `new()` function associated to `StmSigner` that takes as input the same values as the `setup` function above, plus the list of all eligible parties (which it already knows as it is running a full node).

mithril-stm/src/stm.rs
```diff
@@ -304,13 +311,40 @@ impl StmInitializer {
...
+    /// * this `StmSigner`'s parameter valuation
+    /// * the `avk` as built from the current registered parties (according to the registration service)
+    /// * the current total stake (according to the registration service)
+    /// # Error
+    /// This function fails if the initializer is not registered.
+    pub fn new_signer_avk<D: Digest + Clone>(
+        self,
+        closed_reg: ClosedKeyReg<D>,
+    ) -> Result<StmSignerAvk<D>, RegisterError> {
+        let my_index = self.check_initializer(closed_reg.clone())?;
```
iquerejeta, line 344: This is exactly what we want to avoid with the closed_reg, cloning it. And that's why the check_initializer should not consume the registration.

mithril-stm/src/stm.rs
```diff
@@ -304,13 +311,40 @@ impl StmInitializer {
...
+    /// then the resulting `StmSigner` may not be able to produce valid signatures.
+    ///
+    /// Returns an `StmSigner` specialized to
+    /// * this `StmSigner`'s ID and current stake
+    /// * this `StmSigner`'s parameter valuation
+    /// * the `avk` as built from the current registered parties (according to the registration service)
+    /// * the current total stake (according to the registration service)
+    /// # Error
+    /// This function fails if the initializer is not registered.
+    pub fn new_signer_avk<D: Digest + Clone>(
```
iquerejeta, line 340: It probably makes sense to keep the `new_signer_avk` function as it was previously defined, and add a new one in the `StmSigner` struct for the case where we don't need the registration.

mithril-stm/src/stm.rs
```diff
@@ -350,38 +384,64 @@ impl StmInitializer {
...
+                signer_index: self.mt_index,
+            })
+        } else {
+            None
+        }
+    }
 
-        // Check which lotteries are won
+    /// Checks whether the indices won the lottery.
+    /// Collects and returns the winning indices.
```
iquerejeta, line 408:
```suggestion
    /// Collects and returns the winning indices.
```

mithril-stm/src/stm.rs
```diff
@@ -350,38 +384,64 @@ impl StmInitializer {
...
+    /// It records all the winning indexes in `Self.indexes`.
+    /// If it wins at least one lottery, it stores the signer's merkle tree index. The proof of membership
+    /// will be handled by the aggregator.
+    pub fn sign(&self, msg: &[u8]) -> Option<StmSig> {
+        let msgp = self
+            .closed_reg
+            .merkle_tree
+            .to_commitment_batch_compat()
+            .concat_with_msg(msg);
+        let signature = StmSigner::sign(&self.stm_signer, &msgp, self.closed_reg.total_stake)?;
```
iquerejeta, line 438: function sign takes `&self` as input. Use `self.stm_signer.sign(...)` instead.

mithril-stm/src/stm.rs
```diff
@@ -350,38 +384,64 @@ impl StmInitializer {
...
+
+        if !signature.indexes.is_empty() {
```
iquerejeta, line 440: The check if `indexes.is_empty` is already performed inside the `sign` function. With the `?` you are implicitly saying that if there are no indices, return `None`, and if there are, continue.

mithril-stm/src/stm.rs
```diff
@@ -705,35 +779,35 @@ impl<D: Clone + Digest + FixedOutput + Send + Sync> StmAggrSig<D> {
...
-
-        for (sig, reg_party) in &self.signatures {
-            let msgp = avk.mt_commitment.concat_with_msg(msg);
-            sig.check_indices(parameters, &reg_party.1, &msgp, avk)?;
-
-            for &index in &sig.indexes {
-                unique_indices.insert(index);
-                nr_indices += 1;
-            }
+    ) -> Result<(Vec<Signature>, Vec<VerificationKey>), StmAggregateSignatureError<D>> {
```
iquerejeta, line 782: I see why this new function signature is useful. However, this change should be reflected in the documentation of the function.

mithril-stm/src/stm.rs
```diff
@@ -705,35 +779,35 @@ impl<D: Clone + Digest + FixedOutput + Send + Sync> StmAggrSig<D> {
...
+            msgp.as_slice(),
+            leaves.clone(),
+        )?;
 
         let proof = &self.batch_proof;
+
         avk.mt_commitment.check(&leaves, &proof.clone())?;
 
-        Ok(())
+        let (sigs, vks) = FullNodeVerifier::<D>::collect_ver_data(signatures.as_slice(), &leaves);
```
iquerejeta, line 809:
```suggestion
        let (sigs, vks) = FullNodeVerifier::<D>::collect_ver_data(&signatures, &leaves);
```

mithril-stm/src/stm.rs
```diff
@@ -705,35 +779,35 @@ impl<D: Clone + Digest + FixedOutput + Send + Sync> StmAggrSig<D> {
...
 
-        if (nr_indices as u64) < parameters.k {
-            return Err(StmAggregateSignatureError::NoQuorum);
-        }
+        FullNodeVerifier::pre_verify(
+            &avk.total_stake,
+            signatures.as_slice(),
+            parameters,
+            msgp.as_slice(),
+            leaves.clone(),
```
iquerejeta, line 802:
```suggestion
            &avk.total_stake,
            &signatures,
            parameters,
            &msgp,
            leaves.clone(),
```
we probably don't want to clone leaves. See comment below.

mithril-stm/src/stm.rs
```diff
@@ -879,6 +942,104 @@ impl<D: Clone + Digest + FixedOutput + Send + Sync> StmAggrSig<D> {
...
 }
 
+impl<D: Digest + FixedOutput> FullNodeVerifier<D> {
+    /// Verify all signatures whether they all won the lottery, if the indices are unique and the quorum is achieved.
+    pub fn pre_verify(
+        total_stake: &Stake,
+        signatures: &[StmSig],
+        parameters: &StmParameters,
+        msg: &[u8],
+        signed_parties: Vec<RegParty>,
```
iquerejeta, line 952:
```suggestion
        signed_parties: &[RegParty],
```

mithril-stm/src/stm.rs
```diff
@@ -879,6 +942,104 @@ impl<D: Clone + Digest + FixedOutput + Send + Sync> StmAggrSig<D> {
...
+        let mut nr_indices = 0;
+        let mut unique_indices = HashSet::new();
+
+        for (i, sig) in signatures.iter().enumerate() {
+            sig.check_indices(parameters, &signed_parties[i].1, msg, total_stake)?;
+            for &index in &sig.indexes {
+                unique_indices.insert(index);
+                nr_indices += 1;
+            }
+        }
```
iquerejeta, line 963: Now we are verifying signature wrt to keys from two different data structures (two independent slices). This is error prone, and a user could provide a list that is not ordered. Previously, the aggregate signature contained a vector that tupled the signature with it's corresponding key. Why are we no longer using that?

iquerejeta (commented):
Made another pass. Still haven't looked at the tests. I would also make the suggestion of not using SignerAvk, as it does not seem clear. I would probably distinguish the two signers (and any other two structures) with the `core` keyword. SignerCore for the one that does not need the merkle tree, and Signer for the one that needs it. 

Will go over the tests once the comments are addressed. Good work!

iquerejeta (commented):
Another round of reviews. Still missing looking into tests.I'll continue when I'm back.

mithril-stm/src/stm.rs
```diff
@@ -853,32 +838,215 @@ impl<D: Clone + Digest + FixedOutput + Send + Sync> StmAggrSig<D> {
...
+            return Err(CoreVerifierError::IndexNotUnique);
+        }
+        if (nr_indices as u64) < parameters.k {
+            return Err(CoreVerifierError::NoQuorum(nr_indices as u64, parameters.k));
+        }
+
+        Ok(())
+    }
+
+    fn map_sig_party(parties: &[RegParty], signatures: &[StmSig]) -> Vec<StmSigRegParty> {
```
iquerejeta, line 912: Let's add documentation even for private functions. Will make it easier for our future selfs. Also, you can extend a bit the todo. I forgot again what we are supposed to check 😓

mithril-stm/src/stm.rs
```diff
@@ -1323,4 +1498,85 @@ mod tests {
...
+    }
+
+    fn setup_core_parties(
+        params: StmParameters,
+        stake: Vec<Stake>,
+    ) -> (Vec<StmSigner<D>>, Vec<(VerificationKey, Stake)>) {
+        let mut trng = TestRng::deterministic_rng(ChaCha);
+        let mut rng = ChaCha20Rng::from_seed(trng.gen());
+
+        #[allow(clippy::needless_collect)]
```
iquerejeta, line 1520: So we need this? Clippy does not complain in my machine if we don't use it, and if the collect is really not necessary, we could remove it.

iquerejeta (changes requested):
Some requested changes. Also, would be good to include documentation even on private functions. While this is not exposed in the public API, we can still use the `--document-private-items` flag to generate the documentation for private items. Will make it easier for our future selfs. 

Good work! 👏

mithril-stm/benches/size_benches.rs
```diff
@@ -60,6 +63,55 @@ where
...
+        .collect();
+
+    let core_verifier = CoreVerifier::setup(&public_signers);
+
+    let mut signatures: Vec<StmSig> = Vec::new();
+    for s in signers {
+        if let Some(sig) = s.core_sign(&msg, core_verifier.total_stake) {
+            signatures.push(sig);
+        }
+    }
```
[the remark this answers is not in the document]
curiecrypt: I called dedup function to get unique signatures, which also checks whether we have at least `k` signatures.  For the second set of parameters, it panics with `NotEnoughSignatures`. So, I used `match` to handle that. However, I'm not sure if it's suitable for benches.

mithril-stm/tests/test_stm_core.rs
```diff
@@ -0,0 +1,73 @@
+use blake2::Blake2b;
```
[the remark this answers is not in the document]
curiecrypt: I also renamed `test_stm_protocol.rs` to `stm_protocol.rs`.

mithril-stm/tests/test_stm_core.rs
```diff
@@ -0,0 +1,73 @@
...
+        .map(|s| s.new_core_signer(&public_signers))
+        .collect();
+
+    let core_verifier = CoreVerifier::setup(&public_signers);
+
+    //////////////////////////
+    ///// operation phase ////
+    //////////////////////////
+
+    let mut signatures: Vec<StmSig> = Vec::new();
```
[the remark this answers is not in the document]
curiecrypt: I used `with_capacity`. However, not sure if it causes to use more memory than we need.

mithril-stm/src/stm.rs
```diff
@@ -211,6 +216,15 @@ pub struct StmAggrVerificationKey<D: Clone + Digest + FixedOutput> {
...
 
+/// Signature with its registered party.
+#[derive(Debug, Clone, Serialize, Deserialize, Hash, Eq, PartialEq, Ord, PartialOrd)]
+pub struct StmSigRegParty {
+    /// Stm signature
+    pub sig: StmSig,
+    /// Registered party
+    pub reg_party: RegParty,
+}
+
```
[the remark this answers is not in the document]
curiecrypt: Still not sure which one would make more sense. Can we discuss it in our sync? @iquerejeta

mithril-stm/src/stm.rs
```diff
@@ -1014,10 +1198,26 @@ mod tests {
                 sigs.push(sig);
             }
 
-            let dedup_result = clerk.dedup_sigs_for_indices(&msg, &sigs);
+            // let sig_reg_list = CoreVerifier::map_sig_party(&clerk.closed_reg.reg_parties, &sigs);
```
iquerejeta, line 1201: Leftover?

iquerejeta (approved):
Minor comments, but other than that, LGTM 👍 good work!

iquerejeta (approved):
With the tests passing, this implies that the serialisation is backwards compatible. It looks good to me 👍

PR #970, opened 2023-06-08
Fix clean epoch Certifier service

jpraynaud:
## Content
This PR includes a fix to the clean of open messages: they must be cleaned only for an epoch strictly lower than the one given by the chain observer.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Issue(s)
Relates to #953

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2030,7 +2030,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-aggregator"
-version = "0.3.26"
+version = "0.3.27"
 dependencies = [
  "async-trait",
  "chrono",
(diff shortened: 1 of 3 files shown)


mithril-aggregator/src/certifier_service.rs
```diff
@@ -170,7 +170,7 @@ impl MithrilCertifierService {
 impl CertifierService for MithrilCertifierService {
     async fn inform_epoch(&self, epoch: Epoch) -> StdResult<()> {
         debug!("CertifierService::inform_epoch(epoch: {epoch:?})");
-        let nb = self.open_message_repository.clean_epoch(epoch).await?;
+        let nb = self.open_message_repository.clean_epoch(epoch - 1).await?;
```
ghubertpalo, line 173: maybe use a strict lower than instead of passing `epoch-1` ?

jpraynaud: Done :heavy_check_mark:

ghubertpalo (approved):

PR #994, opened 2023-06-16
Uniformise datetime usage

### LINKED ISSUE #946: Dates format is not standardized
## Issue
Also, at least two dates formats exist in the application:

1. With timezone, ie: `2023-06-13T16:35:25.104893012Z`
2. Without timezone, ie: `2023-06-13 16:35:28`

One of the best example is the Certificate: The `initiated_at` don't specify the timeonz while the `sealed_at` dates specify it:
```
"initiated_at": "2023-05-26T00:02:23",
"sealed_at": "2023-05-26T00:03:23.998753492Z",
```
They both should use specify a timezone and even better use the same timezone.

This is a problem since it means that clients querying our Aggregator or our SQLite databases have to check themself which field use which format.

To standardize all of this we need to:

1. Choose and use a standard format when we serialize dates: The [RFC 3339](https://datatracker.ietf.org/doc/html/rfc3339) seems perfect for that.
2. Stop using strings to store dates in our datatypes and use a dedicated type: the DateTime<Utc> from the [chrono crate](https://crates.io/crates/chrono) is also perfect, in addition it have dedicated methods to convert from or to **rfc 3339** format and it have a feature that allow to use thoses types for serde (de)serialization (using the rfc 3339 format too).

## Note
In order to do this we will have to change datatypes that are used for some to compute some of our hashes, meaning that those hashes will changes, so:

⚠ **We will need to re-genesis the certificate chains once this issue is merged**

Alenar:
## Content

This PR uniformise all usages of time as `chrono::DateTime<Utc>`, previously we were using DateTime formatted as strings or `chrono::NaiveDateTime`.
The messages are also impacted since a serde feature is available in `chrono`.
Also every time we needs to serialize a date as a string it will be done using the [rfc3339](https://datatracker.ietf.org/doc/html/rfc3339) format (the serde feature of chrono also use this format so everything will still be consistent).

Updating the `CertificateMetadata` fields to use `DateTime<Utc>` means that their hash will changes: this is a breaking change as this impact the certificate chain, we will need to regenesis our aggregator.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation

## Comments

Db Migrations are provided to update previous sqlite database to the new date format and remove `default current_timestamp` clauses from tables fields because its format is not rfc3339 compatible (meaning that old date creation is now done 'software' side).

An ADR is provided alongside this PR to say that we commit to use the rfc3339 format for our dates.

## Issue(s)

Closes #946


Alenar:
The aggregator monitoring database is impacted but is not in the scope of the database migrator.
You can run the following script to update it:
```sql
create table new_event (
    event_id integer primary key asc autoincrement,
    created_at text not null,
    source text not null,
    action text not null,
    content text nul null
);

insert into new_event
    select event_id, strftime('%Y-%m-%dT%H:%M:%fZ', created_at) as created_at, source, action, content
    from event order by rowid asc;

drop table event;
alter table new_event rename to event;
```

mithril-aggregator/src/database/provider/open_message.rs
```diff
@@ -58,7 +50,7 @@ impl OpenMessageRecord {
             signed_entity_type,
             protocol_message: ProtocolMessage::new(),
             is_certified: false,
-            created_at: chrono::Local::now().naive_local(),
+            created_at: chrono::Local::now().with_timezone(&Utc),
```
Alenar, line 53: `Utc::now()`

jpraynaud (approved):
LGTM :+1:

mithril-test-lab/mithril-devnet/devnet-mkfiles.sh
```diff
@@ -606,7 +606,7 @@ done
     cat >> activate.sh <<EOF
 echo ">> Wait for Cardano pools to be activated"
 POOLS_ACTIVATION_WAIT_ROUND_DELAY=2
-POOLS_ACTIVATION_WAIT_ROUNDS_MAX=\$(echo "scale=0; 5 * $EPOCH_LENGTH * $SLOT_LENGTH / \$POOLS_ACTIVATION_WAIT_ROUND_DELAY" | bc)
+POOLS_ACTIVATION_WAIT_ROUNDS_MAX=\$(echo "scale=0; 20 * $EPOCH_LENGTH * $SLOT_LENGTH / \$POOLS_ACTIVATION_WAIT_ROUND_DELAY" | bc)
```
ghubertpalo, line 609: what does that change stand for?

ghubertpalo (approved):

mithril-test-lab/mithril-devnet/devnet-mkfiles.sh
Alenar: This control the max number of attempt that's the script do when starting a node, why use a formula ? Because the network startup time change based on the network parameters, a fast network would start faster than a slow network, so it's difficult to use the same value for everyone.
The problem here was that on my machine this computation lead to a low value of 9, most of the time it was not enough for the network to start (it takes 10 or 12 attempts).
By quadrupling the base factor this should give enough room for everyone for the network to start at the cost of waiting longer in case of a failure.

Issue #834, opened 2023-03-27
Debian package does not install cleanly on older ubuntu versions


abailly-iohk:
## Context & versions

Not sure which version of ubuntu but I would guess ubuntu 20.04 or something.

## Steps to reproduce

1. Download `.deb` package
2. Run `sudo dpkg -i mithril-client_0.2.14+b61945b_amd64.deb `

## Actual behavior

It fails with:

```
$ sudo dpkg -i mithril-client_0.2.14+b61945b_amd64.deb 
Selecting previously unselected package mithril-client.
(Reading database ... 31294 files and directories currently installed.)
Preparing to unpack mithril-client_0.2.14+b61945b_amd64.deb ...
Unpacking mithril-client (0.2.14+b61945b) ...
dpkg: dependency problems prevent configuration of mithril-client:
 mithril-client depends on libc6 (>= 2.34); however:
  Version of libc6:amd64 on system is 2.31-13+deb11u5.
 mithril-client depends on libssl3 (>= 3.0.0~~alpha1); however:
  Package libssl3 is not installed.

dpkg: error processing package mithril-client (--install):
 dependency problems - leaving unconfigured
Errors were encountered while processing:
 mithril-client
```

## Expected behavior

IT should install cleanly or at least warn user about distribution version being too old/Provide some actionable feedback?

abailly-iohk:
Ping @pgrange

pgrange:
Thank you for the issue @abailly-iohk. Problem appears on Debian 11:

```
cat /etc/issue
Debian GNU/Linux 11 \n \l
```

Alenar:
We merged #1043 that changes the ubuntu version used to build the packaged binaries from `ubuntu 22.04` to `ubuntu 20.04`.
That mean that we are now compatible with system bundling libc 2.31 or more (Debian 11+), instead of 2.35 minimum previously (Debian 12).

We also updated the doc to specify the minimum required libc version (ie: [run signer node guide](https://mithril.network/doc/next/manual/getting-started/run-signer-node#download-pre-built-binary)) and it's also specified in the github release notes as well (ie: [last unstable](https://github.com/input-output-hk/mithril/releases/tag/unstable)).

_Note: those changes are only available for unstable releases until the next (pre)release._

Does this solve this ticket for you @abailly-iohk, @pgrange ?

pgrange:
Yes, problem solved @Alenar 

Thank you.

Issue #837, opened 2023-03-30
SQLite compatibility in aggregator


jpraynaud:
## Issue
We have noticed that the Github Action runners don't have the minimum expected SQLite version:
- We need version `3.40+` for running the queries for the monitoring on the aggregator
- We need version  `3.37+` for everything else in the aggregator
- We have also noticed that the packages that we use to install the version `3.40.1` in the `mithril-infra` has a new location suffixed with `3.40.1-2` when it used to be `3.40.1-1`

## To do 
We need to:
- find a solution that is compatible with the CI (on Linux, Mac OS and Windows) that provides fast enough install time on all the CI runners
- probably implement the same install on the infrastructure
- create a tool sub-command for exporting monitoring data?
- create a tool sub-command for compacting the database?

## Dependencies
Should be done once the tasks `Separate data disk from VM system disk` of #767 is completed

Alenar:
With PR #1024 sqlite is now bundled with our binaries, meaning that we don't need it to be installed on the system anymore and that we can use the latest sqlite version (right now `3.42.0`).

This address most of the issues of this ticket.

Todo:

- [ ] Survey that the bundled sqlite works as intended (ie: do a install on a system/docker without sqlite and check that our nodes works)
- [ ] Update doc websites to remove the install of sqlite as a dependency to install our nodes
- [ ] Update dockerfiles to remove sqlite installation.

PR #1043, opened 2023-07-10
CI: Build linux binaries using ubuntu 20-04 and add vulnerabilities check

### LINKED ISSUE #874: Build static binaries in CI
## Issue
We want to build binaries that are compatible with most Linux systems and avoid conflicts with glibc version

### Solution 1: Ubuntu 18.04 or 20.04
Building on an older system version means using an older version of the glibc library, which means avoiding the versions conflict we usually face

### Solution 2: Nix build
The `nix build` commands have been merged in PR #811. These commands build static binaries whereas the `cargo build` commands build dynamic binaries. We have encountered difficulties with the dynamic binaries produced by the CI due to compatibility versions on some of the libraries dynamically linked as described in issue #834.

We need to test in a Poc if the nix commands can be used in replacement on the build jobs of the workflows (if it works and how long it takes vs current workflows)

### Solution 3: MUSL build
We could use MUSL to build the binaries, but it will bring other problems due to cross compilation (openssl, sqlite compilations fialing). We decide to not investigate that option at the moment.

### Solution 1

- [x] Setup build phase on Ubuntu 18.04 or 20.04 (means more complex setup of the env, see example with cargo nextest)

### Solution 2
- [ ] ~~Build binaries with `nix` instead of `cargo`~~
- [ ] ~~Verify build times are acceptable from CI perspective~~
- [ ] ~~Verify that there is no issue regarding evolution (e.g. new cargo versions, ...)~~
- [ ] ~~Verify static builds and that it fixes issue #834~~

### LINKED ISSUE #1037: Check vulnerabilities in CI
## Issue
When security vulnerabilities are identified by a `dependabot alert` in the repository, we usually create a PR that update dependencies, but we don't before it is merged if it fixes the vulnerability.

## To do

- [ ] Implement a new job in the CI that runs the vulnerability detection of the branch (see this [GitHub action](https://github.com/actions/dependency-review-action))

Alenar:
## Content
This PR changes the ubuntu version used to build the linux binaries in order to depends on a earlier version of glibc (2.31 instead of 2.35) to be compatible with more systems.
The minimum required version is now stated in the documentation. 

It also add the [dependency-review](https://github.com/marketplace/actions/dependency-review) action to our check step. 

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation

## Comments
This PR also update the documentation regarding the fact that sqlite is not required anymore to build & run the binaries (see #837). 

## Issue(s)
Closes #874, Closes #1037, Relates to #834 and #837

### DIFF
--- a/docs/root/compiled-binaries.md
+++ b/docs/root/compiled-binaries.md
@@ -6,8 +6,13 @@ hide_table_of_contents: true
 
 ## Download compiled binary
 
-Each Release / Pre-Release distribution comes with pre compiled binaries ready to use or wrapped in a debian package.
+Each Release / Pre-Release distribution comes with pre compiled binaries ready to use or wrapped in a debian package
+for linux¹.
 
-You can download them from the Release / Pre-Release distribution page that depends on the Mithril Network you are targeting.
+You can download them from the Release / Pre-Release distribution page that depends on the Mithril Network you
+are targeting.
 
 These links are available in the **Build From** column of the above **Mithril Networks** table.
+
+¹ The Linux binaries target `glibc`, and have a minimum requirement of `glibc 2.31` (compatible with `Ubuntu 20.04`
+or `Debian Bullseye`). 
--- a/docs/root/compiled-binaries.md
+++ b/docs/root/compiled-binaries.md
@@ -4,8 +4,6 @@ hide_title: true
 hide_table_of_contents: true
 ---
 
-## Download compiled binary
-
 Each Release / Pre-Release distribution comes with pre compiled binaries ready to use or wrapped in a debian package
 for linux¹.
 
--- a/mithril-aggregator/Dockerfile
+++ b/mithril-aggregator/Dockerfile
@@ -1,17 +1,7 @@
 ###############################
 # STEP 1: build rust executable
 ###############################
-FROM ubuntu:20.04 AS rustbuilder
-
-ENV DEBIAN_FRONTEND noninteractive
-
-# Upgrade and install build base
-RUN apt-get update && apt-get install -y libssl-dev curl wget build-essential pkg-config make m4
-
-# Install Rust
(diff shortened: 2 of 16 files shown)


docs/root/compiled-binaries.md
```diff
@@ -4,10 +4,13 @@ hide_title: true
 hide_table_of_contents: true
 ---
 
-## Download compiled binary
```
Alenar, line 7: Sadly this title did not show in the sidebars (and it's level wasn't always appropriate) so I had to remove it from this common file to put it each time this file is used.

mithril-aggregator/Dockerfile
```diff
@@ -37,7 +29,7 @@ RUN /app/target/release/mithril-aggregator --version
 ###############################
 # STEP 2: build a small image
 ###############################
-FROM ubuntu:22.04
+FROM debian:11-slim
```
ghubertpalo, line 32: why isn't the `ubuntu-20.04` image used (the same as the CI)?

ghubertpalo (approved):

Alenar:
> LGTM +1
> 
> Maybe we can create a test Docker image before merging to see if everything works as expected?

Already done locally using debian 11 as the image base and the binaries built by this branch CI :).

mithril-aggregator/Dockerfile
Alenar: Two reasons:

- I want to use debian in the dockerfiles as "compatibility test" to check that our ubuntu 20.04 built binaries are compatible with a later version of debian.
- this allow for the "dev" dockerfiles (the one without the `.ci` suffix) to use the [rust builder images](https://hub.docker.com/_/rust/) that are debian based, allowing a simplification of the builder step.

PR #1076, opened 2023-07-21
Remove legacy store adapters & fix prune of stake pool & epoch settings

### LINKED ISSUE #1053: Remove legacy store adapters from aggregator
## Issue
Remove the implementation of all the store adapters in the aggregator

Alenar:
## Content
This PR remove the `StoreAdapter` implementation that were remaining in the aggregator database providers. Instead of using them in the dependencies it's the trait or even the provider itself that are used now.
This simplify the usage of those database providers.

Also, this PR fix the prune of the Stake Pool & Epoch Settings, their queries  were prepared but not executed since their cursors where not consumed. Tests have been added in order to check those cases.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Issue(s)
Closes #1053


mithril-aggregator/src/database/provider/stake_pool.rs
```diff
@@ -309,24 +317,21 @@ mod tests {
...
-            ("pool1", 2, 1230),
-            ("pool2", 2, 1090),
-            ("pool3", 2, 1300),
-            ("pool1", 3, 1250),
-            ("pool2", 3, 1370),
-            ("pool3", 3, 1300),
-        ];
-        for (pool_id, epoch, stake) in stake_distribution {
+
+        // Note: decreasing stakes for pool3 so we can test that the order by change
```
jpraynaud, line 327: Small typo here?
```
// Note: decreasing stakes for pool3 so we can test that the order has changed
```

mithril-aggregator/src/services/stake_distribution.rs
```diff
@@ -169,10 +169,7 @@ impl StakeDistributionService for MithrilStakeDistributionService {
             .get_stakes(epoch)
             .await
             .map_err(|e| StakePoolDistributionServiceError::technical_subsystem(e.into()))?
-            .ok_or_else(|| StakePoolDistributionServiceError::Technical {
-                message: "The stake distribution should be at least an empty list.".to_string(),
-                error: None,
-            })?;
+            .unwrap_or_default();
```
jpraynaud, line 172: Is this the behavior we expect if we return default instead of an error here? :thinking:

mithril-test-lab/mithril-end-to-end/src/mithril/aggregator.rs
```diff
@@ -41,6 +41,7 @@ impl Aggregator {
                 "CARDANO_NODE_SOCKET_PATH",
                 bft_node.socket_path.to_str().unwrap(),
             ),
+            ("STORE_RETENTION_LIMIT", "5"),
```
jpraynaud, line 44: This is probably an option of the end to end test? I am not sure we want to prune when this is run in the CI.
Or maybe we could test with and without pruning on the CI, WDYT?

mithril-aggregator/src/dependency_injection/builder.rs
```diff
@@ -273,7 +273,10 @@ impl DependenciesBuilder {
     }
 
     async fn build_stake_store(&mut self) -> Result<Arc<StakePoolStore>> {
-        let stake_pool_store = Arc::new(StakePoolStore::new(self.get_sqlite_connection().await?));
+        let stake_pool_store = Arc::new(StakePoolStore::new(
+            self.get_sqlite_connection().await?,
+            self.configuration.store_retention_limit.map(|l| l as u64),
```
jpraynaud, line 278: Do we want to make sure this value is not in a danger zone? Let's say it's safe with `>2`. WDYT?

jpraynaud (approved):
LGTM :+1: 

Except for the pruning of the stores in the end to end test by default, and few typos/questions

mithril-test-lab/mithril-end-to-end/src/mithril/aggregator.rs
Alenar: Before I changed updated the prune to use the configuration retention limit instead of a constant it was the e2e tests that gave away that something was wrong _(the problem was that the protocol parameters were pruned but not the table that have a foreign key with it)_.
So I see value to always use enable pruning in e2e tests.
In this case I would be in favor of increasing the retention limit in order to keep most of the data while ensuring that it works without issue on 2 or 3 epochs.

mithril-aggregator/src/dependency_injection/builder.rs
Alenar: Done with the `safe_epoch_retention_limit` method on the aggregator `Configuration` type :).

mithril-test-lab/mithril-end-to-end/src/mithril/aggregator.rs
Alenar: The value was increased to 10 in order to start pruning at the final third of the e2e tests (they usually go up to epoch 13~14).
Since most problems with the e2e happens in the first two thirds this will not hinders debugging.

ghubertpalo (approved):

Issue #668, opened 2022-12-22
Refactor (de)serialization of crypto entities


jpraynaud:
## Issue
We need to handle the (de)serialization operations more implicitly allowing:
* Avoid handling manually serialize/deserialize operations
* Store non encoded objects in the entities
* Support multiple serialization format (with auto detection)

### Step 1: PoC on `ProtocolVerificationKey` type with JsonHex
- [x] Replace `HexEncodedProtocolVerificationKey` by `ProtocolVerificationKey`
- [x] Handle smoothly (de)serialize when embedded in other types (e.g. `CertificateMetada` that has a list of `SignerWithStake`)

### Step 2: Apply to all other types
- [x] Implement a generic `ProtocolKey<T>` wrapper type of STM types
- [x] Avoid hard coding JsonHex encoded cumbersome values in test by centralizing these values in `fake_data` module (except for golden tests)
- [x] Update all the crypto entities that use a `HexEncoded` prefix in type. Here is the list:
  - [x] `pub type HexEncodedSingleSignature = HexEncodedKey;`
  - [x] `pub type HexEncodedMultiSignature = HexEncodedKey;`
  - [x] `pub type HexEncodedAgregateVerificationKey = HexEncodedKey;`
  - [x] `pub type HexEncodedVerificationKey = HexEncodedKey;`
  - [x] `pub type HexEncodedVerificationKeySignature = HexEncodedKey;`
  - [x] `pub type HexEncodedOpCert = HexEncodedKey;`
  - [x] `pub type HexEncodedGenesisSecretKey = HexEncodedKey;`
  - [x] `pub type HexEncodedGenesisVerificationKey = HexEncodedKey;`
  - [x] `pub type HexEncodedGenesisSignature = HexEncodedKey;`
  - [x] `pub type HexEncodedDigest = HexEncodedKey;`
  - [x] `pub type HexEncodedEraMarkersSecretKey = HexEncodedKey;`
  - [x] `pub type HexEncodedEraMarkersVerificationKey = HexEncodedKey;`
  - [x] `pub type HexEncodedEraMarkersSignature = HexEncodedKey;`

### Update methodology
1. Move the type alias from `mithril-common::crypto_helper::types::alias` to the `mithril-common::crypto_helper::types::wrappers` module
1. Change the type definition to a `ProtocolKey<T>` where T is the concrete stm types that was behind the hex encoded key
1. Try to compile the workspace to see where the compilation fails
[...]

PR #1108, opened 2023-07-28
Common protocol key type for serialization of Stm types

Alenar:
## Content
This PR defines a new type in the common: `ProtocolKey`. This allow to replace previously hex encoded stm types to use the wrapped concrete types without changes to the (de)serialization of the parent types.

The implementation of the two fist types that we worked on, `ProtocolSignerVerificationKey` & `ProtocolSingleSignature`, is replaced with a simple type alias.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation

## Limitation
We can't implement to`ProtocolKey` From/To `String` (and `&str`) alongside the already defined From/To `T` where T is (de)serilizable by serde. This is because String and &str are (de)serilizable by serde so the compiler can't decide which implementation to use if we define both.

To circumvent that we implemented one 'genericly' (the conversion to from strings) and the other conversion is implemented manually for each types, the later operation is made trivial by defining a dedicated macro.

## Issue(s)
Relates to #668


jpraynaud (approved):
LGTM :+1:

PR #1124, opened 2023-08-01
Protocol key for Multi and Genesis signature

Alenar:
## Content
This PR replace most use of hex encoded keys for multi & genesis signatures with a `ProtocolKey` on the concrete crypto type.

In order to do so the `Certificate` entity had to be adapted since it had two mutually excluding string fields (the exclusion was done manually without compiler help), one for the genesis the other for multi-signatures.
This was done by adding a new type, `CertificateSignature` an enum with a variant for both cases.

Since the genesis signature doesn't use the "json hex" encoding for serialization, contrary to all the previous types converted to a `ProtocolKey`, the `ProtocolKey` add to be modified to allow to change the inner encoding that it use.
This was done with a new trait, `ProtocolKeyCodec`, that define how a `ProtocolKey` is encoded or decoded to strings or to using serde.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Comments
The hash changes were done before making the genesis & multi signatures into a protocol key (meaning that they did not changes afterward, ensuring backward compatibility), the changes were necessary because of the change to the Certificate structure.

## Issue(s)
Relates to #668


jpraynaud:
I have a question regarding the Certificate hash computation that changed: 
do we need to recompute hashes when merging this PR?

mithril-aggregator/src/tools/genesis.rs
```diff
@@ -145,7 +145,7 @@ impl GenesisTools {
         to_sign_payload_file.read_to_end(&mut to_sign_payload_buffer)?;
 
         let genesis_signature = genesis_signer.sign(&to_sign_payload_buffer);
-        let signed_payload = genesis_signature.to_bytes();
+        let signed_payload = genesis_signature.key().to_bytes();
```
jpraynaud: Idea: we could maybe implement `Deref` trait for the type so that we don't need to update this code?

jpraynaud (approved):
LGTM :+1:

PR #1195, opened 2023-08-31
Use zstandard compression for snapshot archives

### LINKED ISSUE #876: Use `zstandard` compression for snapshot archives
## Issue
We could use a different compression algorithm for the snapshot archives that we produce. A good candidate seems to be [`zstandard`](https://github.com/facebook/zstd): we could get the size of the archives divided by `2`

### Implement Prototype
- [x] Test the `zstandard` compression for full immutable snapshots
- [x] Assess the optimization factor for the size of the archives as well as the impact on CPU when compressing/decompressing
- [x] Verify that there is no compatibility issue with platforms that we target (Linux, Mac OS, Windows)
- [x] Check if `zstandard` provides easy updating of files in an existing archives
- [x] Capability of the the compression engine to stream the decompression phase

### Implementation MVP
- [x] Add the `zstandard` compression for full immutable snapshots (keep `tar gzip` compression as well)
- [x] Add a compression format option on the aggregator serve command
- [x] Add a `compression format` field on the snapshot artifact message (defaults to `tar.gz`)
- [x] Migration of snapshot artifact to add the new `compression format` field
- [x] Use this new field in the client to negotiate the decompression of the archive

Alenar:
## Content

This PR add the Zstandard compression algorithm as an option for archive creation.

In order to do so a field named `compression_algorithm` is added to the `Snapshot` entity and message. For the message this field is optional and will default to the previous algorithm, `Gunzip`, if missing.

Furthermore Zstandard algorithm can be configured in the aggregator configuration: we can set the compression level (default to `9`) and the number of worker (default to `4`, using `0` will disable multi-threading).

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation

## Comments
Note: the `snapshotter::verify_archive` function add to be updated to allow concurrent execution. Before it always used the same folder to work, now the folder name is appended with the archive name.

## Issue(s)
Closes #876


jpraynaud (approved):
LGTM :+1: 

Just need to make the few adjustments we discussed!

Issue #798, opened 2023-03-13
Errors refactoring


jpraynaud:
## Issue
We need to refactor errors and to standardize them with an ADR (and implement this ADR):
- Standard Result/Error type for the project 
- Separation technical vs business errors
- Separation critical vs non-critical errors (e.g. `HydrationError`)
- Unwrap and panic usage policy (See #1102)
- Templating of errors with useful crates (`thiserror`, ...)

**Note**: A Rust playground is available here with error contextualization with `anyhow` https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=90f962ab001d2ea0321fc5da0d4ec0f1

## To do

- [x] Implement all error types to `anyhow::Result` instead of `StdResult` or equivalent
- [x] Implement `anyhow` context, see below
- [x] Get rid of `error.to_string().contains()` and create appropriate Error with `thiserror`:
  - [x] `mithril-common/src/protocol/multi_signer.rs`
  - [x] `mithril-common/src/protocol/signer_builder.rs`
- [x] Get rid of already wrapping errors (and let `anyhow` do it for us, identify the `from` in the code and use context instead)
- [x] Redact ADR along the way with all decisions and choices (and questions)

### Implement Anyhow context

- [x] mithril-aggregator
  - [x] artifact_builder
  - [x] commands
  - [x] configuration
  - [X] database
    - [x] provider
  - [x] dependency_injection
  - [x] entities
  - [x] event_store
  - [x] http_server
    - [x] routes
      - [x] artifact_routes
  - [x] message_adapters
  - [x] multi_signers
  - [x] runtime
  - [x] services
  - [x] signer_registerer
  - [x] snapshot_uploaders
  - [x] snapshotter 
  - [x] store
  - [x] tools
  - [x] integration tests
- [x] mithril-client
  - [x] aggregator_client
  - [x] commands
    - [x] mithril_stake_distribution
    - [x] snapshot
  - [x] dependencies
  - [x] entities
  - [x] message_adapters
  - [x] services
  - [x] utils
- [x] mithril-common
  - [x] api_version
  - [x] beacon_provider
  - [x] certificate_chain
  - [x] chain_observer *
  - [x] crypto_helper
    - [x] cardano
      - [x] codec
      - [x] cold_key
      - [x] key_certification
      - [x] opcert
    - [X] codec
    - [x] conversions
    - [x] era
    - [x] genesis
    - [x] tests_setup
    - [x] types
      - [x] alias
[...]

PR #1255, opened 2023-09-25
Use anyhow in signer `single-signer`

jpraynaud:
## Content
This PR includes the update of the signer single-signer to use `anyhow`.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Issue(s)
Relates to #798

### DIFF
--- a/mithril-signer/src/single_signer.rs
+++ b/mithril-signer/src/single_signer.rs
@@ -1,4 +1,4 @@
-use anyhow::Context;
+use anyhow::{anyhow, Context};
 use hex::ToHex;
 use slog_scope::{info, trace, warn};
 use std::path::PathBuf;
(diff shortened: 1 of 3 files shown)


mithril-signer/src/single_signer.rs
```diff
@@ -145,11 +146,15 @@ impl SingleSigner for MithrilSingleSigner {
             signers_with_stake,
             &protocol_initializer.get_protocol_parameters().into(),
         )
+        .with_context(|| "Mithril Single Signer can not compute aggregate verification key")
         .map_err(SingleSignerError::AggregateVerificationKeyComputationFailed)?;
```
dlachaumepalo, line 150:
```suggestion
        .map_err(SingleSignerError::AggregateVerificationKeyComputationFailed(anyhow!("Mithril Single Signer can not compute aggregate verification key")))?;
```

`AggregateVerificationKeyComputationFailed` is defined as `AggregateVerificationKeyComputationFailed(StdError)`. Isn't it easier to add context like this?

dlachaumepalo (approved):
Just a question, otherwise LGTM.

mithril-signer/src/single_signer.rs
jpraynaud: The `map_err` takes a closure as argument: this is the shortcut for `.map_err(|e|  SingleSignerError::AggregateVerificationKeyComputationFailed(e)`. 
Your proposition is creating a different error actually which is not the expected behavior here :slightly_smiling_face:

PR #1303, opened 2023-10-18
clarify reference to block producer IP

adrem1:
## Content

This PR includes a minor change to the SPO documentation to setup Mithril. It intends to clarify something that _I_ found a tad confusing, namely the reference to the block producer IP.  

This is reflected in the changes (minor) made to the `squid.conf` file.

## Pre-submit checklist

- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation

## Comments

No worries if this is deemed irrelevant, I might be the only person to be confused by the terminology "internal IP".  

## Issue(s)

This PR does not relate to any specific issue

### DIFF
--- a/docs/website/versioned_docs/version-maintained/manual/getting-started/run-signer-node.md
+++ b/docs/website/versioned_docs/version-maintained/manual/getting-started/run-signer-node.md
@@ -430,7 +430,7 @@ sudo bash -c 'cat > /etc/squid/squid.conf << EOF
 http_port **YOUR_RELAY_LISTENING_PORT**
 
 # ACL for internal IP of your block producer node
-acl block_producer_internal_ip src **YOUR_BLOCK_PRODUCER_INTERNAL_IP**
+acl block_producer_internal_ip src **YOUR_BLOCK_PRODUCER_IP**
 
 # ACL for aggregator endpoint
 acl aggregator_domain dstdomain .mithril.network
@@ -533,7 +533,7 @@ EOF'
 :::
 
 With this configuration, the proxy will:
-- accept incoming traffic originating from the internal IP of the block-producing machine
+- accept incoming traffic originating from the IP of the block-producing machine
 - accept incoming traffic directed to the listening port of the proxy
 - accept incoming HTTPS traffic proxied to `mithril.network` domain hosts
 - anonymize completely the traffic and avoid disclosing any information about the block-producing machine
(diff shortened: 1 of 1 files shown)


jpraynaud (changes requested):
Hi @adrem1, thanks for the PR. 

> No worries if this is deemed irrelevant, I might be the only person to be confused by the terminology "internal IP".

This is a very common expression which is used to reference the IP address of a machine that can only be used in a private network, which is different from the external IP which is available from outside.

Instead of renaming the placeholder, you could probably add a **tip** or **info** block that clarifies the meaning of the **internal IP** and emphasizes the fact that communications with the block producer machine need to remain on the private network?

TrevorBenson:
> Hi @adrem1, thanks for the PR.
> 
> > No worries if this is deemed irrelevant, I might be the only person to be confused by the terminology "internal IP".
> 
> This is a very common expression which is used to reference the IP address of a machine that can only be used in a private network, which is different from the external IP which is available from outside.
> 
> Instead of renaming the placeholder, you could probably add a **tip** or **info** block that clarifies the meaning of the **internal IP** and emphasizes the fact that communications with the block producer machine need to remain on the private network?

I actually agree with @adrem1 in this instance. This is more about security and best practices than a hard requirement that only works on RFC 1918 subnets  (i.e. "internal/private").

An SPO may not be using an RFC1918 subnet, which is traditionally thought of as an internal network/IP but may still have traffic between the block producer, the signer, and/or a mithril relay, fully secured without one. For example, an IPSec VPN (or another protocol) could be used between public IPs, or even entire public ranges/subnets. In this case, the network is a public routable network/IP, often considered an external network/IP, but provides essentially the same functionality as the internal network/IP would.

I agree that a **tip** or **info** block should be used in either case. When using **YOUR_BLOCK_PRODUCER_IP** I would make the block contain a suggestion that the traffic be secured via private networks (i.e. RFC 1918 subnets) or via a VPN tunnel, etc.. I prefer this as it leaves the network architecture design choices up to the SPO while still providing them with the best practices and suggestions to properly secure the communications.

---

This is just my 2 cents on the topic.

jpraynaud:
> I actually agree with @adrem1 in this instance. This is more about security and best practices than a hard requirement that only works on RFC 1918 subnets (i.e. "internal/private").
> 
> An SPO may not be using an RFC1918 subnet, which is traditionally thought of as an internal network/IP but may still have traffic between the block producer, the signer, and/or a mithril relay, fully secured without one. For example, an IPSec VPN (or another protocol) could be used between public IPs, or even entire public ranges/subnets and then dropping all non-IPSec traffic entirely. In this case, the network is a public routable network/IP, often considered an external network/IP, but provides essentially the same functionality as the internal network/IP would.

You're right that there is no hard requirement on the internal subnet. In that case, all the mentions to the internal IP of the block producer in the guide should also be removed (e.g. the comments and name of rules in the relay configuration) :+1:

adrem1:
@jpraynaud and @TrevorBenson thank you both for your input. I have now understood the concern with my wording and will close this PR. 

I will make a new set of changes to reflect your comments, and submit a new PR for your consideration.

Cheers, Stef

jpraynaud:
OK @adrem1 :+1: 
Feel free to re-open this PR and append new commits on it.

PR #1308, opened 2023-10-19
total_stake calculation corrected

### LINKED ISSUE #1306: `CoreVerifier` setup does not compute `total_stake` correctly
## Why
The `Coreverifer` of the `mithril-stm` does not correctly compute the `total_stake` during the `setup`: this can lead to division by zero when the `num-integer-backend`  is used.

## What
The `total_stake` variable is not mutable and is always `0` when returned by the `setup` [function](https://github.com/input-output-hk/mithril/blob/main/mithril-stm/src/stm.rs#L908): 
```rust
pub fn setup(public_signers: &[(VerificationKey, Stake)]) -> Self {
    let total_stake: Stake = 0;
    let mut unique_parties = HashSet::new();
    for signer in public_signers.iter() {
        let (_, overflow) = total_stake.overflowing_add(signer.1);
        if overflow {
            panic!("Total stake overflow");
        }
        unique_parties.insert(MTLeaf(signer.0, signer.1));
    }

    let mut eligible_parties: Vec<_> = unique_parties.into_iter().collect();
    eligible_parties.sort_unstable();
    CoreVerifier {
        eligible_parties,
        total_stake,
    }
}
```

The `rug` backend does not detect a division by zero, probably because it is converted to a float before dividing.

## How

- [ ] Fix the computation of the `total_stake` when creating the `CoreVerifier`
- [ ] Add a test that verifies the computation of `total_stake`

curiecrypt:
## Content

This PR includes the fix for the wrong calculation of the total stake in the setup of a core verifier and a test for total stake calculation.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Issue(s)

Closes #1306

### DIFF
--- a/mithril-stm/src/stm.rs
+++ b/mithril-stm/src/stm.rs
@@ -1634,31 +1634,15 @@ mod tests {
                 _ => unreachable!(),
             }
         }
-    }
 
-    #[test]
-    fn debug_total_stake_calc(){
-        let nparties = 5;
-        let m = 12;
-        let k = 3;
-
-        let params = StmParameters { m, k, phi_f: 0.2 };
-        let mut rng = ChaCha20Rng::from_seed([0u8; 32]);
-
-        let mut public_signers: Vec<(StmVerificationKey, Stake)> = Vec::with_capacity(nparties);
-        let mut initializers: Vec<StmInitializer> = Vec::with_capacity(nparties);
-
-        let stakes = (0..nparties)
-            .map(|_| 1 + (rng.next_u64() % 9999))
-            .collect::<Vec<_>>();
-
-        for stake in stakes {
-            let initializer = StmInitializer::setup(params, stake, &mut rng);
-            initializers.push(initializer.clone());
-            public_signers.push((initializer.verification_key().vk, initializer.stake));
+        #[test]
+        fn test_total_stake_core_verifier(nparties in 2_usize..30,
+                              m in 10_u64..20,
+                              k in 1_u64..5,) {
+            let params = StmParameters { m, k, phi_f: 0.2 };
+            let (_initializers, public_signers) = setup_equal_core_parties(params, nparties);
+            let core_verifier = CoreVerifier::setup(&public_signers);
+            assert_ne!(core_verifier.total_stake, 0, "Total stake is 0.");
         }
-
-        let core_verifier = CoreVerifier::setup(&public_signers);
-        println!("Total stake test: {}", core_verifier.total_stake);
     }
 }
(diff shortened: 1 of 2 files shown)


mithril-stm/src/stm.rs
```diff
@@ -902,13 +902,14 @@ impl CoreVerifier {
...
-        let total_stake: Stake = 0;
+        let mut total_stake: Stake = 3;
         let mut unique_parties = HashSet::new();
         for signer in public_signers.iter() {
-            let (_, overflow) = total_stake.overflowing_add(signer.1);
+            let (res, overflow) = total_stake.overflowing_add(signer.1);
             if overflow {
                 panic!("Total stake overflow");
             }
+            total_stake = res;
```
jpraynaud, line 912: You forgot to add the stake here:
```suggestion
            total_stake += res;
```

mithril-stm/src/stm.rs
```diff
@@ -1633,5 +1634,15 @@ mod tests {
                 _ => unreachable!(),
             }
         }
+
+        #[test]
+        fn test_total_stake_core_verifier(nparties in 2_usize..30,
+                              m in 10_u64..20,
+                              k in 1_u64..5,) {
+            let params = StmParameters { m, k, phi_f: 0.2 };
+            let (_initializers, public_signers) = setup_equal_core_parties(params, nparties);
```
jpraynaud, line 1643: Maybe you could fix the stake of the signers and assert that the result is exactly the value that you expect?

mithril-stm/src/stm.rs
```diff
@@ -902,13 +902,14 @@ impl CoreVerifier {
     ///     * Calculate the total stake of the eligible signers,
     ///     * Sort the eligible signers.
     pub fn setup(public_signers: &[(VerificationKey, Stake)]) -> Self {
-        let total_stake: Stake = 0;
+        let mut total_stake: Stake = 3;
```
jpraynaud, line 905: Do we want to start at `3` for a specific reason?
```suggestion
        let mut total_stake: Stake = 0;
```

curiecrypt: Ahahah! Leftover from debugging the code. :)

mithril-stm/src/stm.rs
curiecrypt: `overflowing_add` function calculates self + rhs and writes the results into `res`.  So, `res` already holds the summation.

jpraynaud (approved):
LGTM :+1:

PR #1341, opened 2023-11-13
Simplify common deps

Alenar:
## Content
This PR reworks `mithril-common` dependencies:
- Move the `APISpec` and the `TestHttpServer` behind features, allowing to make several common dependencies optional (`glob`, `http`, `jsonschema` and `warp`).
- Remove mockall from the dependencies (now used only internally as a dev-dependency).
- Limit tokio features to just what's actually used: from `full` to `fs`, `io-util`, `process`, `rt`, and `sync` for main dependencies + `macros` and `time` for dev-dependencies.
- Remove `slog-scope`: as it's should only be used for binaries, libraries should be able to set their logger freely.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation

## Comments
Less dependency tree for common will simplify wasm adaptation for the client-lib.

### DIFF
--- a/mithril-aggregator/src/tools/mocks.rs
+++ b/mithril-aggregator/src/tools/mocks.rs
@@ -1,0 +1,28 @@
+use async_trait::async_trait;
+use mithril_common::chain_observer::{ChainAddress, ChainObserver, ChainObserverError, TxDatum};
+use mithril_common::crypto_helper::{KESPeriod, OpCert};
+use mithril_common::entities::{Epoch, StakeDistribution};
+use mockall::mock;
+
+mock! {
+    pub ChainObserver {}
+
+    #[async_trait]
+    impl ChainObserver for ChainObserver {
+        async fn get_current_datums(
+            &self,
+            address: &ChainAddress,
+        ) -> Result<Vec<TxDatum>, ChainObserverError>;
+
+        async fn get_current_epoch(&self) -> Result<Option<Epoch>, ChainObserverError>;
+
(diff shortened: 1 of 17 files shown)


mithril-aggregator/src/tools/mocks.rs
```diff
@@ -0,0 +1,34 @@
...
+use mithril_common::crypto_helper::{KESPeriod, OpCert};
+use mithril_common::entities::{Epoch, StakeDistribution};
+use mockall::mock;
+
+mock! {
+    pub ChainObserver {}
+
+    #[async_trait]
+    impl ChainObserver for ChainObserver{
+        /// Retrieve the datums associated to and address
```
dlachaumepalo, line 12:
```suggestion
        /// Retrieve the datums associated to an address
```

dlachaumepalo (approved):

jpraynaud (commented):
We should probably keep the `apispec` in the common, as other nodes (e.g. the new relay) will also need to have access to it.

Alenar:
> We should probably keep the `apispec` in the common, as other nodes (e.g. the new relay) will also need to have access to it.

@jpraynaud I changed my approach: instead of moving it to the aggregator I putted it behind a feature making the removed dependencies optional instead (and did the same thing for the `TestHttpServer` allowing to to make optional the `warp` dependency).

jpraynaud (approved):
LGTM :+1:

Issue #1311, opened 2023-10-20
Light Wallet: Release `mithril-client` library


jpraynaud:
## Why
In order to release the `mithril-client` library to the public, we want to enhance the developer experience and stabilize the API. 
This issue follows up the work done in issue #1141. 

## What
We want to provide:
- A stable API for developers.
- Advanced features that will give fine-grained control on operations provided by the library.
- Make nice and clean documentation for developers.
- Release the library on `crates.io`.

### Design
- [x] How to refactor configuration setup for the client 
- [x] Re-challenge the API to make it more stable 
  - [x] Do we keep client as is or create sub-modules for each entity type?
  - [x] Provide granular functions for advanced usage when (e.g. verify an already downloaded snapshot archive, verify certificate & certificate chain)
  - [x] How do we provide activity feedback (e.g. with events/channels)?
- [x] Prepare API pseudo code 
- [x] Split plan for crates in repository (crates name in repository and API/artifacts name in release, crates.io, npm, ...)
  - both start with the same version (current) 
  - `mithril-client-cli`: client cli
    - `mithril-client{.exe}` for the binary name
  - `mithril-client`: client library
    - `mithril-client` on crates.io
    - `mithril-client` on npm
    - `libmithril_client.{so,a,rlib}` for the library binary name 
    
### Library API implementation
- [x] Rename `mithril-client` to `mithril-client-cli`
- [x] Implement the new API for `mithril-client` library

### Documentation / DevX
- [x] Add an example crate in the repository for a Mithril client (also add a build step in the CI workflow)
- [x] Make nice documentation for home page of the `mithril-client` doc.rs
- [x] Polish/refine/complete the developer doc on website
- [x] Add `mithril-client-cli` line in the table of the [API Reference](https://mithril.network/doc/next/manual/developer-docs/references)

### Prepare publication on `crates.io`

- [x] :warning: Do we need to publish `mithril-common` with `mithril-client` on crates.io? (See [error](https://github.com/input-output-hk/mithril/actions/runs/6548675383/job/17784237335#step:3:238) `error: all dependencies must have a version specified when publishing.` when publishing.
- [x] Activate the new `mithril-client` package in the workflows to publish on crates.io
- [x] Make publication to crates.io sequential by crate?
- [x] Configure files published for `mithril-client` on `crates.io` / Do not publish Cargo.lock file (also fix `mithril-stm`)
- [x] Configure Cargo.toml with docs.rs and repository links (also fix `mithril-stm`)
[...]

PR #1344, opened 2023-11-13
Mithril-client-lib: Gate snapshot download & message computation behind a feature

Alenar:
## Content

This PR gates the snapshot download and snapshot message computation of the mithril-client library behind a `fs` feature. 

This feature is disabled by default because it simpler to enable a feature than to disable a default feature.

The doc is updated to add "Available on **crate feature fs** only" labels, ie:
image

But for those labels to show we need to trick cargo since [this features is still unstable](https://doc.rust-lang.org/unstable-book/language-features/doc-cfg.html).
For that we took inspiration from tokio, a major crate that use this feature, cf their [main toml](https://github.com/tokio-rs/tokio/blob/master/tokio/Cargo.toml#L157) (for how to trick doc rs to add a rustc flag that will be used to enable the feature) and their [dedicated doc macros](https://github.com/tokio-rs/tokio/blob/master/tokio-util/src/cfg.rs).

To test the the labels locally you need to:
1.  remove the `cfg_attr(docsrs` conditions from the `lib.rs`
2. switch to the unstable rust (`rustup default nightly`)
3. build and open the doc: `cargo doc --no-deps -p mithril-client --open --all-features`

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation

## Comments
CI scripts had to be adapted to use the feature to trigger the tests and to include the gated functions in the doc.

## Issue(s)
Relates to #1311

### DIFF
--- a/.github/workflows/actions/publish-crate-package/action.yml
+++ b/.github/workflows/actions/publish-crate-package/action.yml
@@ -11,6 +11,9 @@ inputs:
   api_token: 
     description: crates.io API token.
     required: false
+  publish_args:
+    description: Additional arguments that will be passed to 'cargo publish'
+    required: false
 
 runs:
   using: "composite"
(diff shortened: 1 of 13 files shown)


.github/workflows/actions/publish-crate-package/action.yml
```diff
@@ -36,18 +39,18 @@ runs:
...
+      cargo package -p ${{ inputs.package }} --list
 
   - name: Cargo publish
     if: inputs.dry_run == 'false' && steps.check_version.outputs.should_deploy == 'true'
     shell: bash
     run: |
-      echo "Cargo publish '${{ inputs.package}}' package"
-      cargo publish -p ${{ inputs.package}} --token ${{ inputs.api_token }} --no-verify
+      echo "Cargo publish '${{ inputs.package }}' package"
+      cargo publish -p ${{ inputs.package }} --token ${{ inputs.api_token }} --no-verify ${{ inputs.publish_args }}
```
jpraynaud, line 56: To check if this is needed during publication?

jpraynaud: The `--features` option of the `cargo publish` is linked to the underlying `cargo build` (which we skip here with the `--no-verify` option), so no need to add the set of features here. 

Still, we can keep these `publish_args` in the action :+1:

dlachaumepalo (approved):

ghubertpalo (approved):

PR #1401, opened 2023-12-15
Implement `open_message` timeout in aggregator

### LINKED ISSUE #1387: Implement `open_message` timeout in aggregator
## Why
We want to avoid new eras when we introduce a new signed entity type (which can block the process of certification during a whole epoch if not enough signers signs for it). The certificate production is blocked in this case because the open message never expires. By introducing a delay to sign the open message, when the quorum is not reached after the delay, the open message will be voided.

## What
Implement a mechanism that associates a timeout (or not) to any signed entity type (e.g not timeout for Mithril Stake Distribution and 10 mins for UTxO set). When an open message is opened for longer than its timeout (if exists), the open message is discarded and allows for handling the next open message.

## How

- [x] Add a `get_open_message_timeout` function to `SignedEntityType`
- [x] Update the provider (and add migration)
- [x] Update `CertifierService`
- [x] Update state machine (clean pending certificate)
- [x] Add integration test in aggregator for expired open message
- [x] Adapt the signer (if needed)

jpraynaud:
## Content
This PR includes a mechanism that associates a timeout (or not) to any signed entity type. When an open message is opened for longer than its timeout (if exists), the open message is discarded and thus allows for handling the next open message.

Open message timeouts:
| Signed Entity Type | Timeout |
| --- |:---:|
| `MithrilStakeDistribution` | - |
| `CardanoStakeDistribution` | 600s |
| `CardanoImmutableFilesFull` | - |

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Issue(s)
Closes #1387

### DIFF
--- a/mithril-aggregator/src/runtime/runner.rs
+++ b/mithril-aggregator/src/runtime/runner.rs
@@ -1,6 +1,6 @@
 use anyhow::{anyhow, Context};
 use async_trait::async_trait;
-use slog_scope::{debug, info, warn};
+use slog_scope::{debug, warn};
 use std::{path::Path, path::PathBuf, sync::Arc};
 
 use mithril_common::entities::{
@@ -47,8 +47,8 @@ pub trait AggregatorRunnerTrait: Sync + Send {
     /// Return the current beacon from the chain
     async fn get_beacon_from_chain(&self) -> StdResult<Beacon>;
 
-    /// Retrieves the current non certified open message for a given signed entity type.
-    async fn get_current_non_certified_open_message_for_signed_entity_type(
+    /// Retrieves the current open message for a given signed entity type.
+    async fn get_current_open_message_for_signed_entity_type(
         &self,
         signed_entity_type: &SignedEntityType,
     ) -> StdResult<Option<OpenMessage>>;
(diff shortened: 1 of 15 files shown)


mithril-aggregator/src/services/certifier.rs
```diff
@@ -50,6 +50,11 @@ pub enum CertifierServiceError {
     #[error("Open message for beacon {0:?} already certified.")]
     AlreadyCertified(SignedEntityType),
 
+    /// The open message is expired, no more single signatures may be
+    /// attached to it nor be certified again.
+    #[error("Open message for beacon {0:?} is expired.")]
```
dlachaumepalo, line 55: I'm not sure about the use of "beacon" here, as there is only a "beacon" in Snapshots.
Can we add the notion of "signed_entity_type"?

dlachaumepalo (approved):
LGTM 👍

mithril-aggregator/src/runtime/runner.rs
```diff
@@ -212,16 +188,24 @@ impl AggregatorRunnerTrait for AggregatorRunner {
...
+                }
             }
-            info!(
-                "RUNNER: get_current_non_certified_open_message: open message already certified";
-                "signed_entity_type" => ?signed_entity_type
-            );
+            let protocol_message = self.compute_protocol_message(&signed_entity_type).await.with_context(|| format!("AggregatorRunner can not compute protocol message for signed_entity_type: '{signed_entity_type}'"))?;
+
+            return Ok(Some(
+                        self.create_open_message(&signed_entity_type, &protocol_message)
```
ghubertpalo, line 205: question: wouldn't it be more readable to create a binding with the `create_open_message` result and then return a `Ok(Some(result))` ?

ghubertpalo (approved):
I let a simple comment, otherwise GTM.

mithril-aggregator/src/runtime/runner.rs
jpraynaud: Done :heavy_check_mark:

PR #1463, opened 2024-01-26
Fix building mithril-common with some features and targets combinatory

Alenar:
## Content

This PR fix the building of mithril-common for all it's matrix of features and targets.

Some examples that raise errors right now:
- building with only the `test_tools` features fails because it's need the `ColdKeyGenerator` that's behind the `random` feature.
- building with only the `--all-targets` directive fails horribly (30 errors) because the tests use a lot of tools that are behind the `test_tools` feature.
- building it with only the `fs` and `random` features would fails since the `FakeObserver` is behind the `test_tools` feature that is disabled.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Comments
In order to check more easily if a feature combinatory is failing a new target has been added to the mithril-common makefiles that compute all features combinatory (the "powerset" of features) and run a cargo clippy on them.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3426,7 +3426,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-common"
-version = "0.2.154"
+version = "0.2.155"
 dependencies = [
  "anyhow",
  "async-trait",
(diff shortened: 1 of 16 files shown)


jpraynaud (approved):
LGTM :+1:

dlachaumepalo (approved):

PR #1479, opened 2024-01-31
Implement `CardanoTransactions` artifact routes in aggregator

### LINKED ISSUE #1478: Implement artifact routes in the aggregator
## Why
In order to give the capability to a client to certify a specific list of transactions, the aggregator needs to give access to the artifacts produced for the `CardanoTransactions` signed entity type.

## What
Implement new routes in the aggregator which list and detail produced artifacts for the Cardano transactions.

## How
- [x] Implement artifact routes in aggregator:
  - [x] Implement route `artifact/cardano-transactions` that list artifacts
  - [x] Implement route `artifact/cardano-transaction/{digest}` that gets an artifact
  - [x] Update OpenAPI specs
- [x] Adapt end to end tests to verify that artifacts are produced for Cardano transactions

jpraynaud:
## Content
This PR includes the new routes in the aggregator REST API which give access to the `CardanoTransactions` artifacts.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Issue(s)
Closes #1478

### DIFF
--- a/mithril-aggregator/src/http_server/routes/artifact_routes/cardano_transaction.rs
+++ b/mithril-aggregator/src/http_server/routes/artifact_routes/cardano_transaction.rs
@@ -1,0 +1,302 @@
+use crate::http_server::routes::middlewares;
+use crate::DependencyContainer;
+use std::sync::Arc;
+use warp::Filter;
+
+pub fn routes(
+    dependency_manager: Arc<DependencyContainer>,
+) -> impl Filter<Extract = (impl warp::Reply,), Error = warp::Rejection> + Clone {
+    artifact_cardano_transactions(dependency_manager.clone())
+        .or(artifact_cardano_transaction_by_id(dependency_manager))
+}
+
+/// GET /artifact/cardano-transactions
+fn artifact_cardano_transactions(
+    dependency_manager: Arc<DependencyContainer>,
+) -> impl Filter<Extract = (impl warp::Reply,), Error = warp::Rejection> + Clone {
+    warp::path!("artifact" / "cardano-transactions")
+        .and(warp::get())
+        .and(middlewares::with_http_message_service(dependency_manager))
+        .and_then(handlers::list_artifacts)
+}
+
+/// GET /artifact/cardano-transaction/:id
+fn artifact_cardano_transaction_by_id(
+    dependency_manager: Arc<DependencyContainer>,
+) -> impl Filter<Extract = (impl warp::Reply,), Error = warp::Rejection> + Clone {
+    warp::path!("artifact" / "cardano-transaction" / String)
+        .and(warp::get())
+        .and(middlewares::with_http_message_service(dependency_manager))
+        .and_then(handlers::get_artifact_by_signed_entity_id)
+}
+
+pub mod handlers {
+    use crate::http_server::routes::reply;
+    use crate::services::MessageService;
+
+    use slog_scope::{debug, warn};
+    use std::convert::Infallible;
+    use std::sync::Arc;
+    use warp::http::StatusCode;
+
+    pub const LIST_MAX_ITEMS: usize = 20;
+
+    /// List Cardano Transactions set artifacts
+    pub async fn list_artifacts(
+        http_message_service: Arc<dyn MessageService>,
+    ) -> Result<impl warp::Reply, Infallible> {
+        debug!("⇄ HTTP SERVER: artifacts");
+
+        match http_message_service
+            .get_cardano_transaction_list_message(LIST_MAX_ITEMS)
+            .await
+        {
+            Ok(message) => Ok(reply::json(&message, StatusCode::OK)),
+            Err(err) => {
+                warn!("list_artifacts_cardano_transactions"; "error" => ?err);
+
+                Ok(reply::internal_server_error(err))
+            }
+        }
+    }
+
+    /// Get Artifact by signed entity id
+    pub async fn get_artifact_by_signed_entity_id(
+        signed_entity_id: String,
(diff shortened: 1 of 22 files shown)


openapi.yaml
```diff
@@ -1421,6 +1474,92 @@ components:
...
+            description: Merkle root of the Cardano transactions set
+            type: string
+            format: bytes
+          beacon:
+            $ref: "#/components/schemas/Beacon"
+          created_at:
+            description: Date and time at which the Cardano transactions set was created
+            type: string
+            format: date-time,
+        example:
```
dlachaumepalo, line 1507: In the example, it's supposed to be an array, isn't it?

jpraynaud: Well, this is the example for the array item in that case. 
That's why this is not an array :slightly_smiling_face:

dlachaumepalo: Oops, thanks 🙏

dlachaumepalo: So the name should not be `CardanoTransactionListItemMessage` instead of `CardanoTransactionListMessage`?

dlachaumepalo: And the type `object` instead of `array`?

jpraynaud: No the structure is an array of objects so it's normal that the name is `List` and not `ListItem`.
The object that represents the list item is not named here, this is probably what bothers you.

BTW, I simply followed the nomenclature already in place for the other signed entity types (e.g. see `MithrilStakeDistributionListMessage` definition above in the file)

mithril-aggregator/src/http_server/routes/artifact_routes/cardano_transaction.rs
```diff
@@ -0,0 +1,302 @@
...
+                artifact: record,
+                created_at: DateTime::parse_from_rfc3339("2023-01-19T13:43:05.618857482Z")
+                    .unwrap()
+                    .with_timezone(&Utc),
+            })
+            .collect()
+    }
+
+    #[tokio::test]
+    async fn test_cardano_transactions_get_ok() {
```
dlachaumepalo, line 146: @sfauvel made an interesting remark about this way of testing (used in other routes tests too).
We are not testing the HTTP code result of the request response here.
Which means that even if the query fails (with a code 500), the test would pass because `verify_conformity` just check that the response is valid.

openapi.yaml
dlachaumepalo: I got confused because`CertificateListMessage` and `SnapshotListMessage` have a different description than `MithrilStakeDistributionListMessage` and `CardanoTransactionListMessage`.

mithril-aggregator/src/http_server/routes/artifact_routes/cardano_transaction.rs
jpraynaud: Yes, it's completely true. We should fix all the tests in a different PR.
I'll create an issue for that :+1:

dlachaumepalo (approved):
LGTM 👍

mithril-aggregator/src/http_server/routes/artifact_routes/cardano_transaction.rs
```diff
@@ -0,0 +1,302 @@
...
+            .allow_any_origin()
+            .allow_headers(vec!["content-type"])
+            .allow_methods(vec![Method::GET, Method::POST, Method::OPTIONS]);
+
+        warp::any()
+            .and(warp::path(SERVER_BASE_PATH))
+            .and(routes(dependency_manager).with(cors))
+    }
+
+    pub fn create_signed_entities<T>(
```
Alenar, line 123: This is the second time this method is copied in this module, maybe we could declare it in the module directly behind a `#[cfg(test)]` flag ?

Alenar (approved):
LGTM with a few comments :)

mithril-common/src/test_utils/fake_data.rs
```diff
@@ -218,3 +218,18 @@ pub fn mithril_stake_distributions(total: u64) -> Vec<entities::MithrilStakeDist
...
+        .map(|idx| {
+            entities::CardanoTransactionsCommitment::new(
+                format!("merkleroot-{idx}"),
+                entities::Beacon {
+                    immutable_file_number: idx,
+                    ..beacon()
+                },
+            )
+        })
+        .collect::<Vec<entities::CardanoTransactionsCommitment>>()
```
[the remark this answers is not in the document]
jpraynaud: Done :heavy_check_mark:

mithril-aggregator/src/http_server/routes/artifact_routes/cardano_transaction.rs
jpraynaud: It looks like it's not that easy as the `setup_router` function uses the `routes` function of the sub-modules.
It got complicated very quickly and I couldn't even make it compile because of limitations of trait objects.
Let's keep it like this and give it another try together later :slightly_smiling_face:

mithril-common/src/entities/cardano_transactions_commitment.rs
```diff
@@ -7,26 +7,41 @@ use super::Beacon;
 /// Commitment of a set of Cardano transactions
 #[derive(Clone, Debug, PartialEq, Eq, Default, Serialize, Deserialize)]
 pub struct CardanoTransactionsCommitment {
-    merkle_root: String,
-    beacon: Beacon,
+    /// Hash of the Cardano transactions set
+    pub hash: String,
```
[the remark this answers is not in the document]
jpraynaud: Done :heavy_check_mark:

Issue #1468, opened 2024-01-29
Implement transactions verification in client library


jpraynaud:
## Why
The client library needs to provide a way to generate, verify and certify proofs for a list of transactions.

## What
Implement new API function for verifying a list of proofs generated by the aggregator in the client library.

## How

- [x] Implement transaction verification function in library:
  - [x] Call aggregator to generate a list of proofs for the transaction list
  - [x] Return the list of certified transactions (and list of not certified?)
  - [x] Verify the proofs for all the certified transactions 
  - [x] Retrieve the latest artifact (and its certificate hash)
  - [x] Verify that the Merkle root of the Merkle proofs is signed by the certificate
  - [x] Verify that the certificate chain is valid
- [x] Implement a function that lists artifacts or get latest produced artifact?
- [x] ~~Update Mithril events feedback mechanism~~
- [x] Update documentation
- [x] Add feature flag `unstable`?

## Later
- Add an example in the repository (later, will be done in another PR)

PR #1485, opened 2024-02-01
Certificate hash in transaction set message

Alenar:
## Content

This PR modify the `/proof/cardano-transaction` endpoint so it returns the certificate hash on which the proof merkle root relates to.
This also means that the `ProverService::compute_transactions_proofs` (and the repository behind it) now limit the set of transactions that's they works on. This in order to only works on what have been certified else they could work on transactions that are not in a certificate yet (typically the aggregator is waiting for single signatures).

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation

## Issue(s)
Relates to #1468

### DIFF
--- a/mithril-common/src/entities/cardano_transactions_set_proof.rs
+++ b/mithril-common/src/entities/cardano_transactions_set_proof.rs
@@ -1,7 +1,7 @@
 use crate::crypto_helper::{MKProof, MKTree, MKTreeNode, MKTreeStore, ProtocolMkProof};
 use crate::entities::TransactionHash;
 use crate::messages::CardanoTransactionsSetProofMessagePart;
-use crate::StdResult;
+use crate::{StdError, StdResult};
 
 /// A cryptographic proof of a set of Cardano transactions is included in the global Cardano transactions set
 #[derive(Clone, Debug, PartialEq)]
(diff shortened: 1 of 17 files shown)


openapi.yaml
```diff
@@ -1565,19 +1567,24 @@ components:
       type: object
       additionalProperties: false
       required:
+        - certificate_hash
         - certified_transactions
         - non_certified_transactions
       properties:
+        certificate_hash:
+          description: Hash of the certificate to retrieve
```
Alenar, line 1575: Incorrect comment, the certificate hash is deduced automatically

mithril-common/src/entities/cardano_transactions_set_proof.rs
```diff
@@ -39,6 +41,38 @@ impl CardanoTransactionsSetProof {
...
+            transactions_proof: mktree.compute_proof(&leaves).unwrap(),
+        }
+    }
+}
+
+impl From<CardanoTransactionsSetProof> for CardanoTransactionsSetProofMessagePart {
+    fn from(proof: CardanoTransactionsSetProof) -> Self {
+        Self {
+            transactions_hashes: proof.transactions_hashes,
+            proof: key_encode_hex(proof.transactions_proof).unwrap(),
```
jpraynaud, line 73: Maybe we can implement a `to_hex/from_hex` pair of associated functions to the `CardanoTransactionSetProof`, and implement `TryFrom` to avoid the unwrap?

jpraynaud (approved):
LGTM :+1: 

Just a comment left below

mithril-common/src/entities/cardano_transactions_set_proof.rs
Alenar: I added a `ProtocolMkProof` (aka a `ProtocolKey<MkProof>`) to use the `ProtocolKey::to_json_hex` method. Does that answer your point ?

jpraynaud: The `ProtocolKey` is a good idea :+1: 
Don't we need to get an error instead of a panic if the proof can't be encoded properly?

dlachaumepalo (approved):

PR #1529, opened 2024-02-23
Unified temp dir builder

### LINKED ISSUE #1508: Unify test folder generation
## Why
We want to unify the generation of test folders in order to limit duplication and have a way to handle common problems that might arise (such as socket path too long).

## What
Implement a builder that provide a unified way to create test folders.

## How

- [x] Create a test folder builder
- [x] Add a option on the builder to generate shorter names for cases where path length matters (ie: socket path).

Alenar:
## Content
This PR add an api to generate temp directory in `mithril_common` `test_tools`.

It can be used in two ways:

1. In it's complete form it's a builder pattern, use that to access all of its options:
```rust
let temp_dir = TempDir::new("module", "name")
    .generate_shorter_path()
    .set_short_path_max_len(80)
    .build();
```
2. Using shortcuts methods
```rust
let temp_dir = TempDir::create("module", "name");
let shorter_temp_dir = TempDir::create_with_short_path("module", "name");
```

With short path disabled (the default) the path generated is in the form of
- linux: `/tmp/mithril_test/$module/$name`
- macOS: `/var/folders/_k/7j0m5c_n4g94vgx9gxknp4tm0000gn/T/mithril_test/$module/$name`.

Shorter path are generated by not including a subfolder with the module but instead appending the first 10 characters of the sha256 hash of both the module & name. If even with that the path total length is longer than the `short_path_max_len` then the name is truncated as much as needed.

Example with the temp dir for `pallas_observer::test::get_current_stake_distribution_with_fallback`:
- linux: `/tmp/mithril_test/get_current_stake_distribution_with_fallback_5446f30db1`
- macOS: `/var/folders/_k/7j0m5c_n4g94vgx9gxknp4tm0000gn/T/mithril_test/get_current_stake_5446f30db1`.

In both cases if an a directory already exists with the same path it is removed alongside it's content.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Issue(s)
Closes #1508

### DIFF
--- a/mithril-common/src/chain_observer/cli_observer.rs
+++ b/mithril-common/src/chain_observer/cli_observer.rs
@@ -448,8 +448,7 @@ impl ChainObserver for CardanoCliChainObserver {
     ) -> Result<Option<KESPeriod>, ChainObserverError> {
         let dir = std::env::temp_dir().join("mithril_kes_period");
         fs::create_dir_all(&dir).map_err(|e| ChainObserverError::General(e.into()))?;
-        let opcert_file =
-            std::env::temp_dir().join(format!("opcert_kes_period-{}", opcert.compute_hash()));
+        let opcert_file = dir.join(format!("opcert_kes_period-{}", opcert.compute_hash()));
         opcert
             .to_file(&opcert_file)
             .map_err(|e| ChainObserverError::General(e.into()))?;
--- a/mithril-common/src/test_utils/temp_dir.rs
+++ b/mithril-common/src/test_utils/temp_dir.rs
@@ -1,3 +1,4 @@
+use sha2::{Digest, Sha256};
 use std::path::{Path, PathBuf};
 
 /// A builder of temp directory for tests purpose.
@@ -6,6 +7,7 @@ pub struct TempDir {
     module_name: String,
     name: String,
     enable_short_path: bool,
+    short_path_max_len: usize,
 }
 
 const TEMP_DIR_ROOT_NAME: &str = "mithril_test";
@@ -17,6 +19,8 @@ impl TempDir {
             module_name: module.into(),
             name: name.into(),
             enable_short_path: false,
+            // 90 to have some room for the folder content (in case of restrained length like socket path)
+            short_path_max_len: 90,
         }
     }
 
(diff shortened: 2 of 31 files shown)


sfauvel (approved):

dlachaumepalo (approved):
LGTM 👍

mithril-common/src/chain_observer/cli_observer.rs
```diff
@@ -448,8 +448,7 @@ impl ChainObserver for CardanoCliChainObserver {
     ) -> Result<Option<KESPeriod>, ChainObserverError> {
         let dir = std::env::temp_dir().join("mithril_kes_period");
```
jpraynaud, line 449: We could probably have used the new `TmpDir` tool here?

mithril-common/src/test_utils/temp_dir.rs
```diff
@@ -0,0 +1,232 @@
...
+        self
+    }
+
+    /// Generate the path of the temp directory (no IO operation will be executed)
+    pub fn build_path(&self) -> PathBuf {
+        const SHA_LENGTH: usize = 10;
+        let base_dir = std::env::temp_dir().join(TEMP_DIR_ROOT_NAME);
+
+        // Short path only:
+        // Combined max len should be lower than `self.short_path_max_len` to have some rooms for
```
jpraynaud, line 48:
```suggestion
        // Combined max len should be lower than `self.short_path_max_len` to have some room for
```

mithril-common/src/test_utils/temp_dir.rs
```diff
@@ -0,0 +1,232 @@
...
+        // the folder content.
+        // MacOS temp folder are not in the `/tmp` folder but in a dynamic path adding 45 chars.
```
jpraynaud, line 50:
```suggestion
        // MacOS temp folders are not in the `/tmp` folder but in a dynamic path adding 45 chars.
```

mithril-common/src/test_utils/temp_dir.rs
```diff
@@ -0,0 +1,232 @@
...
+
+            let max_len = self.name.len().min(max_path_len);
+
+            base_dir.join([&self.name[0..max_len], "_", short_sha].concat())
+        } else {
+            base_dir.join(&self.module_name).join(&self.name)
+        }
+    }
+
+    /// Create a directory based of builder configuration in the system temp folder.
```
jpraynaud, line 73:
```suggestion
    /// Create a directory based on the builder configuration in the system temp folder.
```

Also, the expected behavior if the directory already exists could be documented here.

jpraynaud (approved):
LGTM :+1: 

I left few comments/questions below.

mithril-common/src/chain_observer/cli_observer.rs
Alenar: This is production code, the `TmpDir` api was only designed to unify test tmp folders generation. We can generalize it but this will needs some more care for the folders names.

PR #1600, opened 2024-03-27
Aggregate fake aggregator reference json files

### LINKED ISSUE #1594: Provide fake aggregator data in an aggregated form
This will avoid flooding PR with hundreds of data files:
- scraper script export a json file by type of data (or route)
- build scripts reads data from these aggregated json files

Alenar:
## Content

This PR changes the structure of the fake aggregator reference json (in `mithril-test-lab/mithril-aggregator-fake/default_data`).
Before it was structured like this:
- One file per endpoint that listed an entity (a json array)
- One file per entity for the endpoint that show a specific entity (each json named after the entity and its hash)
For a given entity a lot of files could be in the folder, making a lot of noises in a Pull Request when those files where updated.

Now the structure is like this:
- One file per endpoint that listed an entity (a json array)
- One file per endpoint that show a specific entity (a json hmap with the entity hash as key)
Meaning that we only need to include 11 files in a PR when we update them.

## Pre-submit checklist

- Branch
  - [X] Tests are provided (if possible)
  - [X] Commit sequence broadly makes sense
  - [X] Key commits have useful messages
- PR
  - [X] Self-reviewed the diff
  - [X] Useful pull request description
  - [X] Reviewer requested
- Documentation

## Issue(s)
Closes #1594


sfauvel (approved):

mithril-test-lab/mithril-aggregator-fake/scripts/import.sh
```diff
@@ -142,11 +149,35 @@ EOF
...
+    echo "Joining ${name} artifacts into ${dest} …"
+
+    buffer+="{"
+    for filename in "$src/"*; do
+        local key=$(basename "${filename%.*}")
+
+        buffer+=$(echo "$separator\"$key\": " | cat - "$filename")
+        separator=","
+    done
+    buffer+="}"
```
jpraynaud, line 175: You could also have created the JSON file with `jq` here.

jpraynaud (approved):
LGTM :+1:

mithril-test-lab/mithril-aggregator-fake/scripts/import.sh
Alenar: We don't know how to do that with pure jq (the problem is on how to construct an hashmap,), do you have an idea ?

jpraynaud: I'd say that with `--argfile` argument it should work:
```bash
// Write value file
$ echo "{}" | jq --argjson val 123456 '. += {"test_field_1": $val}' > test.json
$ cat test.json
{
  "test_field_1": 123456
}

// Insert file content as a value for a new key
$ echo "{}" | jq --argfile val test.json '. += {"test_field_2": $val}'
{
  "test_field_2": {
    "test_field_1": 123456
  }
}
```

Issue #1591, opened 2024-03-26
Implement incremental storage of Cardano transactions in signer/aggregator


jpraynaud:
## Why
We want to read data from the chain to retrieve only blocks that have not been recorded in the database yet. 
This will allow deploying the signature of the Cardano transactions on the `testing-preview` network.

## What
Implement incremental storage for Cardano transactions:
- Detect the list of immutable file numbers for which transactions are not stored yet
- Store the transactions of the non recorded yet immutable file number
- Use store to retrieve transactions when creating signature and proofs

## How

- [ ] Create new `CardanoTransactionImporter?` that does the work of `TransactionParser` and `TransactionStore` on the aggregator (and temporarily on the signer so we expect code duplication here at first)
- [ ] Implement `CardanoTransactionImporter` detection of  the list of immutable file numbers for which transactions are not stored yet (starting point)
- [ ] Implement `CardanoTransactionImporter` retrieval of blocks given start point
- [ ] Implement `CardanoTransactionImporter` storage of blocks 
- [ ] Create new  `CardanoTransactionRetriever?` retrieval of all blocks to create protocol messages and transactions proof
- [ ] Activate the signature of Cardano transactions on `testing-preview`

PR #1607, opened 2024-04-02
Add and persist `slot_number` and `block_hash` to `CardanoTransaction`

Alenar:
## Content
This PR add two new fields to our `CardanoTransaction` type: `slot_number` and `block_hash`. This will allow comparison between a transaction and their incoming beacon that will include them (and the `block_number` that's already a part of our transaction type).

## Pre-submit checklist

- Branch
- PR
  - [ ] Self-reviewed the diff
  - [ ] Reviewer requested
- Documentation

## Issue(s)
Relates to #1591

### DIFF
--- a/mithril-aggregator/src/database/provider/cardano_transaction.rs
+++ b/mithril-aggregator/src/database/provider/cardano_transaction.rs
@@ -1,6 +1,7 @@
 use mithril_common::{
     entities::{
-        BlockNumber, CardanoDbBeacon, CardanoTransaction, ImmutableFileNumber, TransactionHash,
+        BlockHash, BlockNumber, CardanoDbBeacon, CardanoTransaction, ImmutableFileNumber,
+        SlotNumber, TransactionHash,
     },
     signable_builder::TransactionStore,
     StdResult,
@@ -26,6 +27,12 @@ pub struct CardanoTransactionRecord {
     /// Block number of the transaction
     pub block_number: BlockNumber,
 
+    /// Slot number of the transaction
+    pub slot_number: SlotNumber,
+
+    /// Block hash of the transaction
+    pub block_hash: BlockHash,
+
     /// Immutable file number of the transaction
     pub immutable_file_number: ImmutableFileNumber,
 }
@@ -35,6 +42,8 @@ impl From<CardanoTransaction> for CardanoTransactionRecord {
         Self {
             transaction_hash: transaction.transaction_hash,
             block_number: transaction.block_number,
+            slot_number: transaction.slot_number,
+            block_hash: transaction.block_hash,
             immutable_file_number: transaction.immutable_file_number,
         }
     }
@@ -45,6 +54,8 @@ impl From<CardanoTransactionRecord> for CardanoTransaction {
         CardanoTransaction {
             transaction_hash: other.transaction_hash,
             block_number: other.block_number,
+            slot_number: other.slot_number,
+            block_hash: other.block_hash,
             immutable_file_number: other.immutable_file_number,
         }
     }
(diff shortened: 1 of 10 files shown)


mithril-aggregator/src/database/provider/cardano_transaction.rs
```diff
@@ -141,41 +160,39 @@ impl<'client> InsertCardanoTransactionProvider<'client> {
...
-            .into_iter()
-            .flat_map(|record| {
-                vec![
-                    Value::String(record.transaction_hash),
-                    Value::Integer(record.block_number.try_into().unwrap()),
-                    Value::Integer(record.immutable_file_number.try_into().unwrap()),
-                ]
-            })
-            .collect();
+        // TODO see if we can find another way to do it
```
jpraynaud, line 186: This comment is not very clear, maybe explain what could be improved?

mithril-aggregator/src/database/provider/cardano_transaction.rs
```diff
@@ -141,41 +160,39 @@ impl<'client> InsertCardanoTransactionProvider<'client> {
...
+        let values: StdResult<Vec<Vec<Value>>> =
+            transactions_records.into_iter().map(map_record).collect();
 
-        WhereCondition::new(
+        let values: Vec<Value> = values?.into_iter().flatten().collect();
```
jpraynaud, line 190: This code is less readable than previous one and requires to create 2 vec instead of one.
Why didn't you use `flat_map` here?

mithril-aggregator/src/database/provider/cardano_transaction.rs
```diff
@@ -315,13 +336,14 @@ mod tests {
             .unwrap()
     }
 
+    // TODO Is it really a useful test ?
```
jpraynaud, line 339: Same here, comment is not clear. Could you be more specific?

Same remark for all the below TODOs.

mithril-common/src/signable_builder/cardano_transactions.rs
```diff
@@ -222,6 +232,34 @@ mod tests {
         }
     }
 
+    #[tokio::test]
+    async fn test_compute_merkle_root_order_of_block_range_does_not_matter() {
```
jpraynaud, line 236: Is this test relevant?

mithril-aggregator/src/database/provider/cardano_transaction.rs
Alenar: We can't use a `flat_map` as that it will flatten the result hence the comment, this is the only way that we find in order to have a Result over the Values conversion (or it would still be using unwraps).

jpraynaud: Have you tried a `try_fold`?

mithril-common/src/signable_builder/cardano_transactions.rs
Alenar: The goal of this test is to make more explicit the fact that the order in which the block ranges appears doesn't matter.
This is a fact that we discovered when we adapted the previous tests, we naively multiplied the block number of each transactions by 10 and by doing so the test was red (because the transactions were now in different block ranges and the order of the block ranges doesn't matter here).

jpraynaud (approved):
LGTM :+1:

mithril-aggregator/src/database/provider/cardano_transaction.rs
```diff
@@ -231,28 +246,32 @@ impl CardanoTransactionRepository {
     }
 
     /// Return the [CardanoTransactionRecord] for the given transaction hash.
-    pub async fn get_transaction(
+    pub async fn get_transaction<T: Into<TransactionHash>>(
```
dlachaumepalo, line 249: There are two calls of `get_transaction` that could be modified. If it's the case, same for the repository of the signer node.
- line 476 with `get_transaction("tx-hash-123")`
- line 649 with `get_transaction("tx-hash-000")`

mithril-common/src/signable_builder/cardano_transactions.rs
```diff
@@ -209,7 +219,7 @@ mod tests {
         }
 
         {
-            // Transactions in a different order returns a different merkle root.
+            // In a same block range Transactions in a different order returns a different merkle root.
```
dlachaumepalo, line 222:
```suggestion
            // In a same block range Transactions in a different order return a different merkle root.
```

mithril-common/src/signable_builder/cardano_transactions.rs
```diff
@@ -158,29 +158,39 @@ mod tests {
...
+        let transaction_3 = CardanoTransaction::new("tx-hash-789", 3, 30, "block_hash", 1);
+        let transaction_4 = CardanoTransaction::new("tx-hash-abc", 4, 40, "block_hash", 1);
+
+        for tx in [
+            &transaction_1,
+            &transaction_2,
+            &transaction_3,
+            &transaction_4,
+        ] {
+            assert!(
```
dlachaumepalo, line 173: Is this assertion necessary? As we are verifying the values that are written directly into the test, not computed.
Would a control in the constructor be relevant?

dlachaumepalo (approved):
Just two comments, otherwise LGTM 🚀

mithril-aggregator/src/database/provider/cardano_transaction.rs
sfauvel: Done here and in signer node

mithril-common/src/signable_builder/cardano_transactions.rs
sfauvel: Done

PR #1641, opened 2024-04-18
Handle unparsed blocks in ctx parser

dlachaume:
## Content

This PR includes...

## Pre-submit checklist

- Branch
- PR
  - [ ] Self-reviewed the diff
  - [ ] Reviewer requested
- Documentation

## Issue(s)


.github/workflows/actions/deploy-terraform-infrastructure/action.yml
```diff
@@ -100,6 +100,10 @@ inputs:
   mithril_aggregator_cexplorer_pools_url:
     description: CExplorer url of the list of pools that is used by the mithril aggregator.
     required: false
+  mithril_aggregator_allow_unparsable_block:
+    description: Allow to log the unparsable block errors in the mithril aggregator.
```
jpraynaud, line 104: This is not clear enough: it should say that it logs the error instead of returning it (which means block parsing is not strictly enforced and should be used only in a test environment. 

Can you also make that this is also applied to all other comments/descriptions?

jpraynaud (commented):
Overall looks good :+1:

.github/workflows/actions/deploy-terraform-infrastructure/action.yml
dlachaume: Done ✅. Updated with the same as the one in the code.

mithril-common/src/entities/cardano_network.rs
```diff
@@ -74,6 +74,18 @@ impl CardanoNetwork {
             CardanoNetwork::TestNet(magic_id) => magic_id,
         }
     }
+
+    /// Evaluate the possibility of validating the request to allow unparsable block or not,
+    /// depending on the target Cardano network
```
jpraynaud, line 79: Can you rewrite this one so it's not clear enough IMO?

Also we should probably rename the function to `compute_allow_unparsable_block`?

mithril-common/src/entities/cardano_network.rs
```diff
@@ -74,6 +74,18 @@ impl CardanoNetwork {
...
+    pub fn allow_unparsable_block(&self, allow_unparsable_block: bool) -> StdResult<bool> {
+        let allow_unparsable_block = match self {
+            CardanoNetwork::MainNet => false,
+            CardanoNetwork::TestNet(id) if *id == PREPROD_MAGIC_ID => false,
```
jpraynaud, line 83: As it is blocked for `mainnet` and `preprod` networks, we should probably say that it "Will be ignored in (pre)production networks" in all the comments?

mithril-common/src/entities/cardano_network.rs
dlachaume: Done, is the [comment](https://github.com/input-output-hk/mithril/blob/7467237f72ccd70c012f80f2eeead7621fa1f70c/mithril-common/src/entities/cardano_network.rs#L78) clear enough?

Alenar (approved):

PR #1644, opened 2024-04-20
Fix: MuxBearerClosed socket closed when reading data

### LINKED ISSUE #1632: Mithril Signer Local Error Policy : Error 182 - MuxError
## Context & versions
cardano node version 8.9.1.
mithril-signer v2412.0.
Fresh mithril-signer setup on BP. Squid on Relay.
Pool is correctly registered on Mithril Explorer.

## Actual behavior
Every 60 seconds, only when mithril-signer is running, this error appears in cardano-node process syslog : 

cardano.node.LocalErrorPolicy:Error:182 IP LocalAddress "<PATH TO SOCKET>/socket@58" ErrorPolicyUnhandledApplicationException (MuxError MuxBearerClosed "<socket: 58> closed when reading data, waiting on next header True")

perseus-algol:
Closes #1632.

`client.abort()` causes following error at cardano node `cardano.node.LocalErrorPolicy:Error:182 IP LocalAddress "/socket@58" ErrorPolicyUnhandledApplicationException (MuxError MuxBearerClosed "<socket: 58> closed when reading data, waiting on next header True")`. If remove this call, it works fine. 

When making this edit, I was guided by:
- Sebastian's comment in #1632 - "client should send `MsgRelease` and `MsgDone` after the query"
- Ouroboros network [spec](https://ouroboros-network.cardano.intersectmbo.org/pdfs/network-spec/network-spec.pdf), section 3.12 Local State Query mini-protocol, Figure 3.8: State machine of the Local State Query mini-protocol.
- [Examples](https://github.com/txpipe/pallas/blob/main/examples/n2c-miniprotocols/src/main.rs#L128) from Pallas library, in examples used same type of Client - NodeClient, containing multiplexer inside. And there is no abort call at the end of example. Although in this example we see only [`client.send_release()`](https://github.com/txpipe/pallas/blob/8fde4a5e11e9dab68907f4e6fecd8fa268e60dcf/examples/n2c-miniprotocols/src/main.rs#L81C5-L81C26) and no `send_done` as how is it done in [Mithril](https://github.com/input-output-hk/mithril/blob/820d9377e67c4e05ad3614f7b66b3e15249f5656/mithril-common/src/chain_observer/pallas_observer.rs#L344) and as Sebastian advises.

### DIFF
--- a/mithril-common/src/chain_observer/pallas_observer.rs
+++ b/mithril-common/src/chain_observer/pallas_observer.rs
@@ -379,8 +379,6 @@ impl ChainObserver for PallasChainObserver {
 
         self.post_process_statequery(&mut client).await?;
 
-        client.abort().await;
-
         Ok(Some(Epoch(epoch as u64)))
     }
 
@@ -399,8 +397,6 @@ impl ChainObserver for PallasChainObserver {
 
         self.post_process_statequery(&mut client).await?;
 
-        client.abort().await;
-
         Ok(datums)
     }
 
@@ -412,8 +408,6 @@ impl ChainObserver for PallasChainObserver {
         let stake_distribution = self.get_stake_distribution(&mut client).await?;
 
         self.post_process_statequery(&mut client).await?;
-
-        client.abort().await;
 
         Ok(stake_distribution)
     }
@@ -427,8 +421,6 @@ impl ChainObserver for PallasChainObserver {
         let current_kes_period = self.get_kes_period(&mut client).await?;
 
         self.post_process_statequery(&mut client).await?;
-
-        client.abort().await;
 
         Ok(current_kes_period)
     }
@@ -760,7 +752,6 @@ mod tests {
                 .await
                 .unwrap();
             observer.post_process_statequery(&mut client).await.unwrap();
-            client.abort().await;
             chain_point
         });
 
@@ -784,7 +775,6 @@ mod tests {
                 .await
                 .unwrap();
             observer.post_process_statequery(&mut client).await.unwrap();
-            client.abort().await;
             genesis_config
         });
 
@@ -808,7 +798,6 @@ mod tests {
                 .await
                 .unwrap();
             observer.post_process_statequery(&mut client).await.unwrap();
-            client.abort().await;
             era
         });
 


jpraynaud:
@falcucci what is the purpose of the `client.abort()` in the `PallasChainObserver`? 

Is it safe to remove it or do we need to replace with a different command to gracefully close the connection to the Cardano node?

falcucci:
It remotely shotdown the current task so it might happen instantaneously and interrupt a parallel task that is running or even somehow breakes the message sent by the server if there is something idle to be sent like a block, pending response or even a ping pong message.

I think it's reasonable to replace the approach to close the plexer client (if it's really needed at our use case) to avoid hanged connections only

Maybe we could guarantee we abort the client when and if the mithril instance is killed, so it would be enough.

If this isnt a risk (hanged connections) this pr looks good to me.

perseus-algol:
I'll made some experiments tomorrow regarding graceful shutdown of `muxer` and `demuxer`. I have a suspicion that by calling `abort` we interrupting `tick` method in the middle of message sending/receiving, which can cause `socket closed` error on server side.

ch1bo:
@scarmuega Please see aboves comment on the `tick` insight.

jpraynaud:
@abailly-iohk

abailly-iohk:
This error is raised [here](https://github.com/input-output-hk/ouroboros-network/blob/2402c1342d2d2934a9697071697f5b98ff07fd72/network-mux/src/Network/Mux/Bearer/Socket.hs#L113) in ouroboros-network and is caused by the fact the `MuxBearer` tries to read from the underlying `Socket` which returns an empty `ByteString`.
Tracing the call to the [network](https://hackage.haskell.org/package/network-3.2.0.0) library lead me to the low-level [recv](https://hackage.haskell.org/package/network-3.2.0.0/docs/src/Network.Socket.ByteString.IO.html#recv) function whose documentation says:

```
-- | Receive data from the socket.  The socket must be in a connected
-- state.  This function may return fewer bytes than specified.  If
-- the message is longer than the specified length, it may be
-- discarded depending on the type of socket.  This function may block
-- until a message arrives.
--
-- For TCP sockets, a zero length return value means the peer has
-- closed its half side of the connection.
```

I followed the call tree a bit further down and ultimately it ends in C-land with the [c_recv](https://hackage.haskell.org/package/network-3.2.0.0/docs/src/Network.Socket.Buffer.html#c_recv) function which actually returns 0 if the socket is closed by the peer. There's even a [test](https://github.com/haskell/network/blob/0505db5fe9340dd50ae6de34539da3544b9ef4db/tests/Network/Socket/ByteStringSpec.hs#L140) for that!

So it seems what happens is that the _client_ closes the connection, which leads to this exception being raised in the [Mux](https://github.com/input-output-hk/ouroboros-network/blob/2402c1342d2d2934a9697071697f5b98ff07fd72/network-mux/src/Network/Mux.hs#L413) code which terminates the Mux and all the mini-protocols.

abailly-iohk:
@scamuerga @falcucci Can you point me to the client code?

falcucci:
@abailly-iohk [here](https://github.com/txpipe/pallas/blob/main/pallas-network/src/multiplexer.rs) you can check the client source code from pallas side.

PR #1649, opened 2024-04-22
Use new memory allocator 'jemallocator'

### LINKED ISSUE #1629: Memory leak in Cardano transactions signature/proof
## Why
There is a memory leak in the signature/proof generation of the Cardano transactions. A first investigation shows that it seems to be due to the Merkle tree used to sign that is not released from memory (`~2GB` leak on `testing-preview`)

# What
Investigate and fix the memory leak

## How

- [x] Deactivate the signature of Cardano transactions on `testing-preview`
- [x] Investigate the memory leak
- [x] Fix the problem
- [ ] Reactivate the signature of Cardano transactions on `testing-preview`

jpraynaud:
## Content
This PR includes the activation of a different memory allocator `jemallocator` through a feature. This will likely avoid having memory fragmentation resulting in resident memory of signer/aggregator not being correctly released when signing/proving Cardano transactions.

See https://github.com/input-output-hk/mithril/issues/1629#issuecomment-2069016167 for details.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Issue(s)
Closes #1629

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,8 @@ As a minor extension, we have adopted a slightly different versioning convention
    - `portable` feature now has no effect and should be removed from crate dependencies.
    - Removed it from all other crates (including `mithril-common`).
 
+- Switched memory allocator to `jemallocator` on signer and aggregator to avoid memory fragmentation when signing transactions (which lead to RES memory not being properly returned to the OS).
+
 - Crates versions:
 
 |  Crate  |  Version  |
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3774,6 +3774,7 @@ dependencies = [
  "slog-term",
  "sqlite",
  "thiserror",
+ "tikv-jemallocator",
  "tokio",
 ]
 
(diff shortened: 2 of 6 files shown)


Alenar (commented):
I've some concerns with adding this globally in `mithril-common`:

- I want to reduce the number of features in `mithril-common` (by splitting it up in several projects) so each project can compile exactly what it needs and to allow better build caching.
This would reduce the number of time cargo have to recompile `mithril-common` when developing (since not all project use the same common feature set), enhancing our dev experience.
So adding another feature doesn't go well with that objective.
- Even behind a feature I'm not at ease with adding a feature that change something as structural as the memory allocator on a crate that's published to `crates.io` and is a dependency to another `crates.io` published crates: `mithril-client`.
I think we should just use it directly on our nodes binaries and let users of our `mithril-client` use this allocator themselves if they want to.

Alenar (approved):

dlachaume (approved):
LGTM 👍, tested on Mac arm64.

PR #1652, opened 2024-04-24
Stream blocks during import

### LINKED ISSUE #1646: Stream import of Cardano transactions
## Why
In order to scale for the `mainnet`, we need to stream the import of transactions to avoid loading them all at once on the signer and aggregator.

## What
Have the `BlockScanner` return an iterator on the transactions instead of all the transactions.

## How

- [ ] Implement an iterator on the `BlockScanner`.

Alenar:
## Content

This PR changes the way the blocks are retrieved from the `BlockScanner`. Instead of giving the full list of blocks from the given cardano database it yields now a `BlockStreamer` trait that allow to poll the blocks at a pace given by it's implementation (the `ImmutableBlockStreamer` yield them immutables per immutables).

This reduce the memory load needed during the import process.

The `BlockStreamer` trait has been designed with the transition to the `chain_sync` protocol using pallas in mind. As such it sadly don't implement the rust `Iterator` trait since it needs to be async.

Note: Since `mithril-common/src/cardano_block_scanner.rs` would become quite huge with the new types introduced, it has been promoted to a module directory instead and it's types spread out to dedicated files.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation

## Comments
This PR should be merged after #1650 since its branch is based on that PR code.

## Issue(s)
Closes #1646


jpraynaud (commented):
It looks a bit more complicated to implement that what I expected and it probably means that we will have to handle the back pressure from the incoming stream to control memory consumption :thinking:

mithril-common/src/cardano_block_scanner/scanned_block.rs
```diff
@@ -0,0 +1,80 @@
...
+            multi_era_block.hash().to_string(),
+            multi_era_block.number(),
+            multi_era_block.slot(),
+            immutable_file_number,
+            transactions,
+        )
+    }
+
+    /// Number of transactions in the block
+    pub fn transaction_len(&self) -> usize {
```
dlachaume, line 59: Is it necessary to have this function? IMO, it doesn't make much difference.

dlachaume (approved):
LGTM 👍, just a few suggestions.

mithril-common/src/cardano_block_scanner/dumb_block_scanner.rs
```diff
@@ -0,0 +1,123 @@
...
+impl DumbBlockScanner {
+    /// Factory
+    pub fn new(blocks: Vec<ScannedBlock>) -> Self {
+        Self {
+            blocks: RwLock::new(blocks),
+        }
+    }
+
+    /// Update blocks returned used the streamer constructed by `scan`
+    pub async fn update_blocks(&self, new_blocks: Vec<ScannedBlock>) {
```
dlachaume, line 25: Is this function used anywhere?

jpraynaud (approved):

PR #1672, opened 2024-05-07
Consistent transactions sort

### LINKED ISSUE #1657: Consistent sorting of transactions from `GetCardanoTransactionProvider` in signer and aggregator
## Why
In order to provide consistent order of transactions on the different nodes, we need to rely on `block_number` and `transaction_hash` instead of `rowid` (which can differ from one node to another). Also we want to make sure that the sorting is fast and optimized.

## What
Implement sorting of transactions by `block_number` and `transaction_hash` in signer and aggregator.

## How

- [x] Update `GetCardanoTransactionProvider` in signer and aggregator
- [x] Create a benchmark for retrieval of sorted transactions
- [x] Add composite index on `block_number` and `transaction_hash`
- [x] Truncate table `block_range_root`

Alenar:
## Content
This PR changes the order of retrieval of the Cardano transactions from the database.

Before we used SQLite `rowid` but this order isn't guaranteed to be consistent across node since it depends on the order of insertion in the database.
To solve this issue this PR change the order to the use the transaction `block_number` then its `transaction_hash`.

An composite index on those columns is added to ensure a fast retrieval, plus a benchmark is added to measure the retrieval performances. 

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Issue(s)
Closes #1657

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3452,7 +3452,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-aggregator"
-version = "0.5.5"
+version = "0.5.6"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -3752,7 +3752,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-signer"
-version = "0.2.133"
+version = "0.2.134"
 dependencies = [
  "anyhow",
  "async-trait",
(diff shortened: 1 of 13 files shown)


jpraynaud (approved):
LGTM :+1:

dlachaume (approved):

PR #1674, opened 2024-05-11
feat: add a new `get_current_chain_point` function to fetch chain point information

falcucci:


jpraynaud (approved):
LGTM :+1:

falcucci:
@jpraynaud thanks! I'll leave this as draft until we have pallas released, ping you up when I bump the version.

mithril-common/src/chain_observer/pallas_observer.rs
```diff
@@ -303,6 +313,30 @@ impl PallasChainObserver {
...
+            .acquire(None)
+            .await
+            .map_err(|err| anyhow!(err))
+            .with_context(|| "PallasChainObserver failed to acquire statequery")?;
+
+        let chain_point = self.do_get_chain_point_state_query(statequery).await?;
+
+        let header_hash = match chain_point {
+            Point::Origin => None,
+            Point::Specific(_at_slot, ref hash) => Some(hex::encode(hash)),
```
Alenar, line 328: Just to double check, can you confirm that a block hash computed here is consistent with how we get it when scanning the blocks in the immutable (cf [mithril-common/src/cardano_block_scanner/scanned_block.rs#L50](https://github.com/input-output-hk/mithril/blob/8745b05477e617b158ead8e557c00e8043007b79/mithril-common/src/cardano_block_scanner/scanned_block.rs#L50)).

As far as I can tell the hash in `Point::Specific` is a `Vec<u8>` and the `hash()` of `MultiEraBlock` is a `pallas_crypto::hash::Hash`.
Using `to_string` of `pallas_crypto::hash::Hash` is equivalent to use its Display implementation that do a `hex::encode` as you do here, so it should be fine _(maybe the hash in `Point::Specific` should be a `pallas_crypto::hash::Hash` too, but I don't know if changing this would make sense in your api)_.

falcucci: Yep, they are fully compatible, both are just computed bytes representation and there is no need to use the `Hash` for the `Point` definition. 

As an example, for the chainsync `RollForward` we could use some pseudo code to find an intersect point:

```rust
let block = MultiEraBlock::decode(cbor);
let slot = block.slot();
let hash = block.hash();

pallas_network::miniprotocols::Point::Specific(slot, hash.to_vec());
```

which is the same computed representation you are getting at your scan.

ch1bo:
> @jpraynaud thanks! I'll leave this as draft until we have pallas released, ping you up when I bump the version.

Which PR would that be upstream?

falcucci:
@ch1bo it's this one, I'll actually remove it from draft right now 🙂

Alenar (approved):

PR #1678, opened 2024-05-14
Common Cardano transactions & Block range roots repository

### LINKED ISSUE #1677: Refactor Cardano transactions database layer in persistence for signer and aggregator
## Why
We want to avoid code duplication between our nodes.
We initially thought that the Cardano Transaction and Block range roots code would greatly differ between the aggregator and the signer. For example we thought that the signer would not needs to store the transactions at all but it was far easier to do the import step by storing them first before computing their block range roots.

The only part that differ on the repository is the implementation of a trait needed to retrieve data to compute proof of inclusion for transactions, else everything is the same.

## What
Now that we have some hindsight, we need a common database layer for the Cardano transactions and Block range roots.

## How

- [x] Move the `CardanoTransactionRepository`, its associated records, providers and migrations, the from one of the node to `mithril-persistence`.
- [x] Use this, now shared, code in both nodes.

Alenar:
## Content

In order to reduce code duplication between nodes, this PR move the `CardanoTransactionRepository`, and the underlying providers and records, to `mithril-persistence`.

The only things that's still left in the nodes is the implementations of locally defined traits, but they are mostly pass-through to the shared repository functions.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Comments
This PR also the `SignedEntityHydrator` in persistence to simply `Hydrator` in order to have a shared set of tools to help with common hydration operations. Two existing methods that were doing that locally in the nodes have been migrated here.

## Issue(s)
Closes #1677


jpraynaud (approved):
LGTM :+1:

dlachaume (approved):

PR #1680, opened 2024-05-15
Implement `chainsync` `ChainReader` with `pallas`

### LINKED ISSUE #1590: Retrieve Cardano blocks with `chainsync` in `pallas`
## Why
We want to retrieve blocks from the Cardano node by using the `chainsync` mini protocol instead of reading the immutable files: this will allow us to get data closer to the tip of the chain (e.g. volatile data). This will also enable Mithril node transactions database warm-up in a background thread.

## What
Create a `PallasBlockReader` that can interact with Cardano `chainsync` mini protocol and implements the draft trait `ChainBlockReader`:
```rust
/// The parsed chain block representation
pub type ParsedBlock = pallas::MultiEraBlock;

/// The action that indicates what to do next when scanning the chain
pub enum ChainBlockNextAction {
    /// RollForward event (we are still on the correct fork)
    RollForward {
        /// The next point in the chain to read
        next_point: ChainPoint,
        /// The parsed block 
        block: ParsedBlock,
    },
    /// RollBackward event (we are on an incorrect fork, we need to get back a point to roll forward again)
    RollBackward {
        /// The rollback point in the chain to read (as a new valid point to read from on the main chain, which has already been seen)
        rollback_point: ChainPoint,
    },
}

/// The trait that reads events to either:
/// - read next block on the chain
/// - rollback to another point in case of rollback
/// - do nothing
pub trait ChainBlockReader {
    //// Get next chain block action
    async fn get_next_chain_block(&self, point: &ChainPoint) -> StdResult<Option<ChainBlockNextAction>>;
}
```

## How

- [x] Add a new `chain_reader` module in `mithril-common`
- [x] Define the trait `ChainBlockReader`
- [x] Implement a `PallasBlockReader` for the `ChainBlockReader` trait
- [x] Implement a `FakeBlockReader` for the `ChainBlockReader` trait (that we can feed with appropriate list of `ChainBlockNext` events)

### See
- https://github.com/txpipe/pallas/blob/97a32c9af5cced098809b5ee4eb72ae0be4cddc2/examples/n2c-miniprotocols/src/main.rs#L84
- Oura: worker thread writing to a queue in the background https://github.com/txpipe/oura/blob/29fed274e8dd46948bccf05ca1ed50705d60adf0/src/sources/n2c.rs#L92
- https://ouroboros-network.cardano.intersectmbo.org/pdfs/network-spec/network-spec.pdf

jpraynaud:
## Content
This PR includes an implementation of 
- `ChainReader` trait
- `PallasChainReader` implementation of the `chainsync` mini protocol with `pallas`
- `FakeChainReader` fake implementation 

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Issue(s)
Closes #1590


mithril-common/src/chain_reader/pallas_chain_reader.rs
```diff
@@ -0,0 +1,262 @@
...
+    fn create_temp_dir(folder_name: &str) -> PathBuf {
+        TempDir::create_with_short_path("pallas_chain_observer_test", folder_name)
+    }
+
+    /// Sets up a mock server for related tests.
+    ///
+    /// Use the `intersections` parameter to define exactly how many
+    /// local state queries should be intersepted by the `mock_server`
+    /// and avoid any panic errors.
+    async fn setup_server(socket_path: PathBuf, action: String) -> tokio::task::JoinHandle<()> {
```
jpraynaud, line 159: IMO the test server setup should be given list of `ChainBlockNextAction` and should serve these actions to the client. This would avoid using an action and would help testing more sophisticated scenarios.

Also, the action should probably be an an enum instead of a string.

mithril-common/src/chain_reader/pallas_chain_reader.rs
```diff
@@ -0,0 +1,262 @@
...
+                    fs::remove_file(&socket_path).expect("Previous socket removal failed");
+                }
+
+                let known_point = get_fake_intersection_point();
+                let unix_listener = UnixListener::bind(socket_path.as_path()).unwrap();
+                let mut server = NodeServer::accept(&unix_listener, 10).await.unwrap();
+
+                let chansync_server = server.chainsync();
+
+                // chansync_server receives request next from client, sends rollbackwards
```
jpraynaud, line 172: This comment and the one below do not seem to be describing what's happening the code :thinking:

mithril-common/src/chain_reader/pallas_chain_reader.rs
```diff
@@ -0,0 +1,262 @@
...
+    };
+    use tokio::net::UnixListener;
+
+    use super::*;
+
+    use crate::test_utils::TempDir;
+
+    /// Returns a fake chain point for testing purposes.
+    fn get_fake_chain_point_backwards() -> ChainPoint {
+        ChainPoint::from(Point::Specific(
```
jpraynaud, line 120: Maybe the `PointSpecific` value which is repeated 3 times could be shared somewhere?

mithril-common/src/chain_reader/pallas_chain_reader.rs
```diff
@@ -0,0 +1,262 @@
...
+    }
+
+    /// Creates a new work directory in the system's temporary folder.
+    fn create_temp_dir(folder_name: &str) -> PathBuf {
+        TempDir::create_with_short_path("pallas_chain_observer_test", folder_name)
+    }
+
+    /// Sets up a mock server for related tests.
+    ///
+    /// Use the `intersections` parameter to define exactly how many
```
jpraynaud, line 156: The comment also needs to be updated :wink:

mithril-common/src/chain_reader/pallas_chain_reader.rs
```diff
@@ -0,0 +1,262 @@
...
+        point: &ChainPoint,
+    ) -> StdResult<Option<ChainBlockNextAction>> {
+        let client = self.get_client().await?;
+        let chainsync = client.chainsync();
+
+        chainsync
+            .find_intersect(vec![point.to_owned().into()])
+            .await?;
+
+        let next = match chainsync.has_agency() {
```
jpraynaud, line 92: Do we test both agency cases?

mithril-common/default.profraw
jpraynaud, line 1: What's the purpose of that file?

mithril-common/src/chain_reader/interface.rs
```diff
@@ -0,0 +1,18 @@
...
+use super::ChainBlockNextAction;
+
+/// The trait that reads events to either:
+/// - read next block on the chain
+/// - rollback to another point in case of rollback
+/// - do nothing
+#[async_trait]
+pub trait ChainBlockReader {
+    /// Intersect the point of the chain with the given point
+    async fn intersect_point(&mut self, point: &ChainPoint) -> StdResult<()>;
```
jpraynaud, line 14: @falcucci this trait modification is problematic and it should be reverted IMO: the `intersect_point` is an implementation related function that should exist only in the `PallasChainReader`.

Thus, we should keep to the previous trait definition:
```rust
pub trait ChainBlockReader {
    /// Get the next chain block
    async fn get_next_chain_block(
        &mut self,
        point: &ChainPoint,
    ) -> StdResult<Option<ChainBlockNextAction>>;
}
```

falcucci: in the previous implementation we were intersecting a tip for every next chain block, which should be: you first intersect and then poll the blocks.

> the intersect_point is an implementation related function that should exist only in the PallasChainReader.

why? intersecting the tip isn't a pallas defined rule, it's part of the [chainsync](https://github.com/IntersectMBO/ouroboros-consensus/blob/e9cf1eeb8afce0d44ab4a1284ad8c065a74673a3/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/MiniProtocol/ChainSync/Server.hs#L103) mini protocols consensus

falcucci: furthermore, this change makes it possible to test the `has_no_agency` scenario, which brings us the ability to manipulate the server agency state accordingly with our necessities as you can see [here](https://github.com/txpipe/mithril/blob/adab158c40af399bc6281e1d35936f709c5c1620/mithril-common/src/chain_reader/pallas_chain_reader.rs?plain=1#L323-L325)

jpraynaud: OK, let's rename the function in the trait with `set_chain_point` :+1:

falcucci (approved):
LGTM

Alenar (approved):

mithril-common/src/chain_reader/fake_chain_reader.rs
```diff
@@ -0,0 +1,82 @@
...
+            ChainBlockNextAction::RollBackward {
+                rollback_point: ChainPoint {
+                    slot_number: 1,
+                    block_number: 1,
+                    block_hash: "point-hash-1".to_string(),
+                },
+            },
+        ];
+        let mut chain_reader = FakeChainReader::new(expected_chain_point_next_actions.clone());
+        chain_reader.set_chain_point(&chain_point).await.unwrap();
```
sfauvel, line 72: Is it useful to call this method ? It does nothing on FakeChainReader.
If it's not, we can also remove the `chain_point`variable
If we just want to check that this function can be called without crash, we may create a dedicated test for that

mithril-common/src/chain_reader/fake_chain_reader.rs
```diff
@@ -0,0 +1,82 @@
...
+
+    #[tokio::test]
+    async fn test_get_next_chain_block() {
+        let chain_point = ChainPoint {
+            slot_number: 1,
+            block_number: 1,
+            block_hash: "point-hash-1".to_string(),
+        };
+        let expected_chain_point_next_actions = vec![
+            ChainBlockNextAction::RollForward {
```
sfauvel, line 47: We may create a function that build a chainPoint to hide useless information.
Something like: 
```
        fn buildChainPoint(id: u64) -> ChainPoint {
            ChainPoint {
                slot_number: id,
                block_number: id,
                block_hash: format!("point-hash-{id}"),
            }
        }
        let expected_chain_point_next_actions = vec![
            ChainBlockNextAction::RollForward {
                next_point: buildChainPoint(1),
                parsed_block: ScannedBlock::new("hash-1", 1, 10, 20, Vec::<&str>::new()),
            },
            ChainBlockNextAction::RollForward {
                next_point: buildChainPoint(2),
                parsed_block: ScannedBlock::new("hash-2", 2, 11, 21, Vec::<&str>::new()),
            },
            ChainBlockNextAction::RollBackward {
                rollback_point: buildChainPoint(1),
            },
        ];
```

mithril-common/src/chain_reader/pallas_chain_reader.rs
```diff
@@ -0,0 +1,349 @@
...
+            self.client = Some(self.new_client().await?);
+        }
+
+        self.client
+            .as_mut()
+            .with_context(|| "PallasChainReader failed to get a client")
+    }
+
+    /// Intersects the point of the chain with the given point.
+    async fn intersect_point(&mut self, point: &ChainPoint) -> StdResult<()> {
```
sfauvel, line 53: Why not naming it `instersection_point` ?

mithril-common/src/chain_reader/pallas_chain_reader.rs
```diff
@@ -0,0 +1,349 @@
...
+                let known_point = get_fake_specific_point();
+                let unix_listener = UnixListener::bind(socket_path.as_path()).unwrap();
+                let mut server = NodeServer::accept(&unix_listener, 10).await.unwrap();
+
+                let chainsync_server = server.chainsync();
+
+                chainsync_server.recv_while_idle().await.unwrap();
+
+                chainsync_server
+                    .send_intersect_found(known_point.clone(), Tip(known_point.clone(), 1337))
```
sfauvel, line 208: Is it make sense to create a constant for 1337 ?

mithril-common/src/chain_reader/pallas_chain_reader.rs
```diff
@@ -0,0 +1,349 @@
...
+        match chain_block {
+            ChainBlockNextAction::RollBackward { rollback_point } => {
+                assert_eq!(rollback_point, get_fake_chain_point_backwards());
+            }
+            _ => panic!("Unexpected chain block action"),
+        }
+    }
+
+    #[tokio::test]
+    async fn get_next_chain_block_rolls_forward() {
```
sfauvel, line 271: It's a bit difficult to understand what we try to test.
In one hand, we check the communication between client and server.
In another hand, we check the mapping from the server response to a ChainBlockNextAction.
Could we test this mapping (process_chain_block_next_action) in a another test to simply it ?

sfauvel (approved):
Looks good for me.
Just few remarks.

dlachaume (approved):

mithril-common/src/chain_reader/entity.rs
```diff
@@ -0,0 +1,18 @@
+use crate::{cardano_block_scanner::ScannedBlock, entities::ChainPoint};
+
+/// The action that indicates what to do next when scanning the chain
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum ChainBlockNextAction {
+    /// RollForward event (we are still on the correct fork)
+    RollForward {
+        /// The next point in the chain to read
+        next_point: ChainPoint,
+        /// The parsed chain block
```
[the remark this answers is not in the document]
jpraynaud: It's the parsed block at the current chain point and there is at most one parsed block per chain point: adding the `latest` prefix to the name of the field would not add information or clarity IMO.

mithril-common/src/chain_reader/fake_chain_reader.rs
jpraynaud: Done

mithril-common/src/chain_reader/fake_chain_reader.rs
jpraynaud: Done

mithril-common/src/cardano_block_scanner/scanned_block.rs
```diff
@@ -5,7 +5,7 @@ use crate::entities::{
 };
 
 /// A block scanned from a Cardano database
-#[derive(Debug, Clone, PartialEq)]
+#[derive(Debug, Clone, PartialEq, Eq)]
```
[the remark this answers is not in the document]
jpraynaud: Why would you want to remove it?

mithril-common/src/chain_reader/pallas_chain_reader.rs
```diff
@@ -0,0 +1,349 @@
...
+        let socket_path = create_temp_dir("get_next_chain_block_rolls_forward").join("node.socket");
+        let known_point = Point::Specific(
```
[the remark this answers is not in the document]
jpraynaud: Done

mithril-common/src/chain_reader/pallas_chain_reader.rs
jpraynaud: Done

mithril-common/src/cardano_block_scanner/scanned_block.rs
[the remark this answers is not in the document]
dlachaume: Cause it doesn't seem to be useful

mithril-common/src/chain_reader/pallas_chain_reader.rs
jpraynaud: Maybe `find_intersect_point` is better as the method does not return the intersection point but just computes it?
@falcucci WDYT?

falcucci: I think that `intersect_point` still okay to me, but `find_intersect_point` is also a good name.

is up to you between these two @jpraynaud

jpraynaud: Let's use `find_intersect_point` which describes better what's done under the hood :+1:

mithril-common/src/chain_reader/pallas_chain_reader.rs
jpraynaud: I guess that we try to test that when the `chainsync` server returns a roll forward the `PallasChainReader::get_next_chain_block` function returns a `ChainBlockNextAction::RollForward` correctly formed :thinking: 

Could you elaborate on how we could split the test?

mithril-common/src/chain_reader/pallas_chain_reader.rs
jpraynaud: Done

PR #1685, opened 2024-05-16
Prune cardano transactions in signer after block range roots computation

### LINKED ISSUE #1645: Prune Cardano transactions stored on signer
## Why
In order to provide minimum footprint of the signer nodes on the SPO infrastructure, we want to keep stored only the minimum transactions needed to enable consistent signature.

## What
Prune transactions which are older than the Cardano `k` security parameter (i.e number of blocks after which the blocks are immutable) and which have corresponding block range roots stored in the database.

> [!WARNING]
> The `k` parameter is a genesis parameter that may be different depending on the network. In a first implementation we will fix the parameter given the Cardano network for simplicity, and we will probably use the protocol parameters in a second implementation to get access to the value. The `k` value for `mainnet` is `2160`.

## How

- [x] Implement a decorator for `TransactionImporter` trait that does the pruning of transactions in the signer
- [x] Make prune transactions with block number that is strictly less than `MAX(block_range_root.start) -  k` 
- [x] Make `k` security parameter (which represents the number of blocks to keep) configurable for the signer (with an value with default of `2160`)

Alenar:
## Content
This PR add, to the signer only, a pruning of the cardano transactions after the computation of block ranges roots.

This is done using a decorator wrapping the existing `CardanoTransactionsImporter`.

Two new configuration parameters have been added to signer `Configuration` structure:
- `network_security_parameter`, default to `2160` (mainnet value).
- `enable_transaction_pruning`, default to `true`.

Since both have a default value our signer won't have to update their configurations.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation

## Comments

- As it was the third time that we needed to do a simple query to retrieve a single result from database, a system using a  extension trait on the [sqlite connection](https://docs.rs/sqlite/0.36.0/sqlite/struct.ConnectionThreadSafe.html) has been added to do that simply with reduced boilerplate.
- Some simplification has been done on the configuration system: a leaner way to hydrate default configuration (mithril-signer & mithril-aggregator) and generalized use of the test configuration values (`new_sample(party_id)`, mithril-signer).

## Issue(s)
Closes #1645

### DIFF
--- a/internal/mithril-persistence/src/sqlite/connection_extensions.rs
+++ b/internal/mithril-persistence/src/sqlite/connection_extensions.rs
@@ -1,0 +1,79 @@
+use anyhow::Context;
+use sqlite::{ReadableWithIndex, Value};
+
+use mithril_common::StdResult;
+
+use crate::sqlite::SqliteConnection;
+
+/// Extension trait for the [SqliteConnection] type.
+pub trait ConnectionExtensions {
+    /// Execute the given sql query and return the value of the first cell read.
+    fn query_single_cell<Q: Into<String>, T: ReadableWithIndex>(
+        &self,
+        sql: Q,
+        params: &[Value],
+    ) -> StdResult<T>;
+}
+
+impl ConnectionExtensions for SqliteConnection {
+    fn query_single_cell<Q: Into<String>, T: ReadableWithIndex>(
+        &self,
+        sql: Q,
+        params: &[Value],
+    ) -> StdResult<T> {
+        let sql = sql.into();
+        let mut statement = self.prepare(&sql).with_context(|| {
+            format!(
+                "Prepare query error: SQL=`{}`",
+                &sql.replace('\n', " ").trim()
+            )
+        })?;
+        statement.bind(params)?;
+        statement.next()?;
+        statement
+            .read::<T, _>(0)
+            .with_context(|| "Read query error")
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use sqlite::Connection;
+
+    use super::*;
+
+    #[test]
+    fn test_query_string() {
+        let connection = Connection::open_thread_safe(":memory:").unwrap();
+        let value: String = connection.query_single_cell("select 'test'", &[]).unwrap();
+
+        assert_eq!(value, "test");
+    }
+
+    #[test]
+    fn test_query_max_number() {
+        let connection = Connection::open_thread_safe(":memory:").unwrap();
+        let value: i64 = connection
+            .query_single_cell(
+                "select max(a) from (select 10 a union select 90 a union select 45 a)",
+                &[],
+            )
+            .unwrap();
+
+        assert_eq!(value, 90);
+    }
+
+    #[test]
+    fn test_query_with_params() {
+        let connection = Connection::open_thread_safe(":memory:").unwrap();
+        let value: i64 = connection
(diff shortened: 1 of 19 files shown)


dlachaume (approved):
LGTM 👍

internal/mithril-persistence/src/sqlite/connection_extensions.rs
```diff
@@ -0,0 +1,79 @@
...
+use sqlite::{ReadableWithIndex, Value};
+
+use mithril_common::StdResult;
+
+use crate::sqlite::SqliteConnection;
+
+/// Extension trait for the [SqliteConnection] type.
+pub trait ConnectionExtensions {
+    /// Execute the given sql query and return the value of the first cell read.
+    fn query_single_cell<Q: Into<String>, T: ReadableWithIndex>(
```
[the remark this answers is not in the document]
Alenar: Since it's a low level api that may be used in different kind of scenario I would like to have some flexibility with the input parameters, hence the generic.
This said after re-challenging it I don't need a owned string here but only a ref to a str, so I'm thinking of changing the generics to `AsRef<str>` this avoid an allocation and simplify a bit the implementation:
```diff
 impl ConnectionExtensions for SqliteConnection {
-    fn query_single_cell<Q: Into<String>, T: ReadableWithIndex>(
+    fn query_single_cell<Q: AsRef<str>, T: ReadableWithIndex>(
         &self,
         sql: Q,
         params: &[Value],
     ) -> StdResult<T> {
-        let sql = sql.into();
         let mut statement = self.prepare(&sql).with_context(|| {
             format!(
                 "Prepare query error: SQL=`{}`",
-                &sql.replace('\n', " ").trim()
+                sql.as_ref().replace('\n', " ").trim()
             )
         })?;
         statement.bind(params)?;
```

mithril-signer/src/transactions_importer_with_prune_decorator.rs
```diff
@@ -0,0 +1,130 @@
...
+
+use mithril_common::entities::{BlockNumber, ImmutableFileNumber};
+use mithril_common::signable_builder::TransactionsImporter;
+use mithril_common::StdResult;
+
+/// Cardano transactions pruner
+#[cfg_attr(test, mockall::automock)]
+#[async_trait]
+pub trait TransactionPruner: Send + Sync {
+    /// Prune the transactions older than the given number of blocks (based on the block range root
```
jpraynaud, line 13:
```suggestion
    /// Prune the transactions older than the given number of blocks.
```

mithril-signer/src/transactions_importer_with_prune_decorator.rs
```diff
@@ -0,0 +1,130 @@
...
+    /// stored).
+    async fn prune(&self, number_of_blocks_to_keep: BlockNumber) -> StdResult<()>;
+}
+
+/// A decorator of [TransactionsImporter] that prunes the transactions older than a given number of
+/// blocks after running the import.
+///
+/// If the number of blocks to keep is not provided, no pruning is performed.
+pub struct TransactionsImporterWithPruneDecorator {
```
jpraynaud, line 22: IMO the fact that it is a decorator is an implementation detail which should not appear in the name:
```suggestion
pub struct TransactionsImporterWithPruner {
```

PR #1710, opened 2024-05-23
Remove snapshot command

### LINKED ISSUE #1690: Remove `snapshot` command in client CLI
## Why
We need to remove the deprecated `snapshot` command of the client CLI. It will be released with the next distribution (`2423`).

## What
Remove the command and warn our users that the command will be removed so that they can use the new `cardano-db` command instead.

## How

- [x] Remove the `snapshot` command in the client CLI
- [x] Add an error message when `snapshot` command is used or when a command is not found (but `snapshot` is not displayed anymore in the help)
- [x] Remove the `snapshot` command in the documentation
- [x] Update the minor version of the client CLI as it is a breaking change
- [ ] Warn in advance all users with an announcement in slack/discord

## Later
- Create a blog post when distribution is released

sfauvel:
## Content
This PR remove `snapshot`command from Mithril client cli.

Initial behavior:
image

New behavior:
image

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation

## Issue(s)
Closes #1690


mithril-client-cli/src/main.rs
```diff
@@ -139,6 +139,27 @@ impl Args {
 
         Ok(Logger::root(Arc::new(drain), slog::o!()))
     }
+
+    fn parse_with_decorator(
+        decorator: &dyn Fn(Result<Self, clap::error::Error>) -> Result<Self, clap::error::Error>,
```
jpraynaud, line 144: Can we use a `StdResult` instead?
Also can we alias `clap::error::Error` with `ClapError`?

mithril-client-cli/src/main.rs
```diff
@@ -139,6 +139,27 @@ impl Args {
...
+    }
+
+    fn handle_deprecated_decorator(
+        args_result: Result<Self, clap::error::Error>,
+    ) -> Result<Self, clap::error::Error> {
+        let styles = Args::command().get_styles().clone();
+        Deprecation::handle_deprecated_commands(
+            args_result,
+            styles,
+            vec![DeprecatedCommand::new("snapshot", "cardano-db")],
```
jpraynaud, line 160: This could be part of the configuration of the `parse_with_decorator` function IMO.

jpraynaud (approved):
LGTM :+1: 

I left few comments below.

Alenar (approved):

mithril-client-cli/src/main.rs
sfauvel: Done

mithril-client-cli/src/main.rs
sfauvel: We have made the ClapError alias.
We keep clap Error for the decorator to have the same signature of the decorated function (try_parse).

PR #1727, opened 2024-06-04
Block Number beacon in cardano transaction

### LINKED ISSUE #1697: Sign Cardano transactions with ChainPoint based beacon
## Why
We want to use a new beacon based on a chain point instead of an immutable file number when signing Cardano transactions.

## What
Migrate the `CardanoTransactions` signed entity type to use a `ChainPoint` instead of `ImmutableFileNumber`. A new associated open message is created at every `n` blocks (value of `n` needs to be a configuration parameter to be defined) with a fixed offset `k'` (value of `k'` needs to be a configuration parameter to be defined which represents the unofficial security parameter).

## How

- [x] Update the `CardanoTransactions` signed entity type to use a `ChainPoint` internally
- [x] Update the `TimePoint` to have a `ChainPoint` or `BlockNumber` entry that represents the tip in the state machine of the aggregator
- [x] Compute the signed entity type with `block_number = ((tip.block_number - k') / n) * n`
- [x] Make `k'` and `n` parameters configurable in the aggregator with default values: `k'=3000` and `n=100` (values can be adjusted)
- [x] Make `BlockScanner` of immutable files compute the upper bound on its own (try to get the latest immutable file number?)
- [x] Clean existing conversion between `ImmutableFileNumber` and `BlockNumber` in the database repository
- [x] Adapt signature of `CardanoTransactions` to the new beacon
- [x] Adapt the `CardanoTransactionSnapshot` artifact to support the new beacon
- [x] Clean the artifacts for the Cardano transactions in the aggregator database with a new migration

Alenar:
## Content

This PR changes the beacon used for the `CardanoTransactions` signed entity type:
- before: a `CardanoDbBeacon` (aka: network, epoch, immutable file number)
- now: Epoch and block number tuple.

This changes decouples the production of cardano transactions signatures and proofs from the immutable file number and the cardano node database structure. Allowing us to go closer to the tip of the chain since, with a adapted block streamer, we won't have to wait for a immutable to be complete update our stored transactions.

### In depth

- Add a new parameter to the aggregator to configure the pacing at which the chain should be parsed: `cardano_transactions_signing_config`.
- Update related type:
  -  `CardanoTransactionsSnapshot` artifact: replace `beacon` with `block_number` (note: the epoch can still be fetched if it's wrapped by a `SignedEntity<T>` and is added to the related messages).
  - In `ProtocolMessage` and `CardanoTransactionsProofsMessage`: rename `latest_immutable_file_number` to `latest_block_number`.
- Update services:
  - `BlockScanner` trait: use a `Option<ChainPoint>` for its lower bound and a `BlockNumber` for the upper bound. Previously it used a `ImmutableFileNumber` for both.
  - `CardanoBlockScanner`: ignore the bounds given as parameter and compute them itself (lower bound from db and upper bound take all completed immutables) since it still works using immutables files and there's no way go get the immutable file that contains a given block from the chain.
  - Updated the signable builder for cardano transactions, and the `TransactionsImporter` and `BlockRangeRootRetriever` traits, to use a `BlockNumber` instead of a `ImmutableFileNumber`.
- Cleanup all certificate and signed entities based on `CardanoTransactions` from Aggregator database since their beacons changed.
- Explorer: Updated  to support the new beacon + use dynamic display for certificate `signed_entity` instead of relying on the deprecated `beacon` property.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation

## Comments

This would have been considered a network breaking change on a stable signed entity type.

## Issue(s)
Closes #1697


mithril-client/src/type_alias.rs
```diff
@@ -61,6 +61,6 @@ pub mod common {
         ProtocolParameters,
```
Alenar, line 61: Add `BlockNumber` to the client re-exported type.

mithril-common/src/entities/signed_entity_type.rs
```diff
@@ -41,7 +41,53 @@ pub enum SignedEntityType {
...
+/// Allow to compute the block number to be signed based on the chain tip block number.
+///
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct CardanoTransactionsSigningConfig {
+    /// Number of blocks to wait before taking in account a transaction.
+    pub security_parameter: BlockNumber,
+
+    /// The frequency at which the transactions are signed.
+    ///
+    /// *Note: The step is adjusted to be a multiple of the block range length in order.*
```
Alenar, line 58: Fix this comment :sweat_smile:

openapi.yaml
```diff
@@ -979,10 +979,14 @@ components:
           description: Aggregate verification key (AVK) that will be used to create the next multi signature
           type: string
           format: bytes
+        latest_block_number:
+          description: The latest block number of the Cardano chain
```
Alenar, line 983:
```suggestion
          description: The latest signed block number
```

internal/mithril-persistence/src/database/repository/cardano_transaction_repository.rs
```diff
@@ -127,22 +106,15 @@ impl CardanoTransactionRepository {
             .fetch_collect(InsertBlockRangeRootQuery::insert_many(records)?)
     }
 
-    // TODO: remove this function when the Cardano transaction signature is based on block number instead of immutable number
     /// Get the highest [BlockNumber] of the cardano transactions stored in the database.
```
dlachaume, line 109:
```suggestion
    /// Get the highest [ChainPoint] of the cardano transactions stored in the database.
```

mithril-aggregator/src/artifact_builder/cardano_transactions.rs
```diff
@@ -44,10 +44,10 @@ impl ArtifactBuilder<CardanoDbBeacon, CardanoTransactionsSnapshot>
             .with_context(|| {
                 format!(
                     "Can not compute CardanoTransactionsCommitment artifact for signed_entity: {:?}",
```
dlachaume, line 46: Not a consequence of this PR
```suggestion
                    "Can not compute CardanoTransactions artifact for signed_entity: {:?}",
```

mithril-aggregator/src/services/cardano_transactions_importer.rs
```diff
@@ -60,47 +59,38 @@ impl CardanoTransactionsImporter {
...
-        let rescan_offset = self.rescan_offset.unwrap_or(0);
-        let highest = highest.map(|h| (h + 1).saturating_sub(rescan_offset as u64));
-        Ok(highest)
-    }
-
     async fn parse_and_store_transactions_not_imported_yet(
         &self,
-        from: Option<ImmutableFileNumber>,
-        until: ImmutableFileNumber,
+        from: Option<ChainPoint>,
```
jpraynaud, line 80: Do we really need the full `ChainPoint` here?
```suggestion
        from: Option<BlockNumber>,
```

mithril-aggregator/src/services/cardano_transactions_importer.rs
```diff
@@ -205,8 +195,8 @@ mod tests {
             async fn scan(
               &self,
               dirpath: &Path,
-              from_immutable: Option<ImmutableFileNumber>,
-              until_immutable: ImmutableFileNumber,
+              from: Option<ChainPoint>,
```
jpraynaud, line 198: Same question as above here.

mithril-aggregator/src/services/signed_entity.rs
```diff
@@ -128,13 +129,13 @@ impl MithrilSignedEntityService {
...
-            SignedEntityType::CardanoTransactions(beacon) => Ok(Arc::new(
+            SignedEntityType::CardanoTransactions(_epoch, block_number) => Ok(Arc::new(
                 self.cardano_transactions_artifact_builder
-                    .compute_artifact(beacon.clone(), certificate)
+                    .compute_artifact(block_number, certificate)
                     .await
                     .with_context(|| {
                         format!(
-                            "Signed Entity Service can not compute artifact for entity type: '{signed_entity_type}'"
+                            "Signed Entity Service can not compute artifact for entity type: '{block_number}'"
```
jpraynaud, line 138:
```suggestion
                            "Signed Entity Service can not compute artifact for entity type: '{signed_entity_type}'"
```

mithril-client-cli/src/commands/cardano_transaction/snapshot_list.rs
```diff
@@ -43,8 +42,7 @@ impl CardanoTransactionSnapshotListCommand {
                 .table()
                 .title(vec![
                     "Epoch".cell(),
-                    "Immutable".cell(),
-                    "Network".cell(),
+                    "Block Number".cell(),
```
jpraynaud, line 45: Do we want to remove the `Network` here?

mithril-common/src/entities/signed_entity_type.rs
```diff
@@ -41,7 +41,53 @@ pub enum SignedEntityType {
...
+    CardanoTransactions(Epoch, BlockNumber),
+}
+
+/// Configuration for the signing of Cardano transactions
+///
+/// Allow to compute the block number to be signed based on the chain tip block number.
+///
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct CardanoTransactionsSigningConfig {
+    /// Number of blocks to wait before taking in account a transaction.
```
jpraynaud, line 53:
```suggestion
    /// Number of blocks to discard from the tip of the chain when importing transactions.
```

mithril-common/src/cardano_block_scanner/block_scanner.rs
```diff
@@ -1,25 +1,48 @@
...
 use async_trait::async_trait;
 use slog::{warn, Logger};
 
 use crate::cardano_block_scanner::{BlockScanner, BlockStreamer, ImmutableBlockStreamer};
 use crate::digesters::ImmutableFile;
-use crate::entities::ImmutableFileNumber;
+use crate::entities::{BlockNumber, ChainPoint, ImmutableFileNumber};
 use crate::StdResult;
 
+/// Trait to find the lower bound that should be used the [block scanner][CardanoBlockScanner] when
```
dlachaume, line 12:
```suggestion
/// Trait to find the lower bound that should be used by the [block scanner][CardanoBlockScanner] when
```

mithril-common/src/entities/signed_entity_type.rs
```diff
@@ -41,7 +41,53 @@ pub enum SignedEntityType {
...
+    pub security_parameter: BlockNumber,
+
+    /// The frequency at which the transactions are signed.
```
jpraynaud, line 56:
```suggestion
    /// The number of blocks between signature of the transactions.
```

mithril-common/src/entities/signed_entity_type.rs
```diff
@@ -41,7 +41,53 @@ pub enum SignedEntityType {
...
+            }
+        }
+    }
+
+    /// Compute the block number to be signed based on che chain tip block number.
+    ///
+    /// Given k' = `security_parameter` and n = `step`,
+    /// the latest block number to be signed is computed as *(this use a integer division)*:
+    ///
+    /// **block_number = ((tip.block_number - k') / n) × n**
```
jpraynaud, line 78:
```suggestion
    /// Given k' = `security_parameter` and n = `step`,
    /// the latest block number to be signed is computed as:
    ///
    /// `block_number = ⌊(tip.block_number - k') / n⌋ × n`
    ///
    /// where `⌊x⌋` is the floor function which rounds to the greatest integer less than or equal to `x`.
```

mithril-common/src/messages/cardano_transaction_snapshot.rs
```diff
@@ -2,18 +2,19 @@ use chrono::DateTime;
...
 pub struct CardanoTransactionSnapshotMessage {
     /// Merkle root of the Cardano transactions snapshot
     pub merkle_root: String,
 
-    /// Beacon of the Cardano transactions snapshot
-    pub beacon: CardanoDbBeacon,
+    /// Epoch of the Cardano transactions snapshot
+    pub epoch: Epoch,
+
+    /// ChainPoint of the Cardano transactions snapshot
```
jpraynaud, line 16:
```suggestion
    /// Block number of the Cardano transactions snapshot
```

mithril-common/src/messages/cardano_transaction_snapshot.rs
```diff
@@ -60,14 +63,11 @@ mod tests {
 
     // Test the retro compatibility with possible future upgrades.
     #[test]
-    fn test_v1() {
+    fn test_v2() {
```
jpraynaud, line 66: We can probably leave it as `test_v1` as the feature is still unstable:
```suggestion
    fn test_v1() {
```

mithril-common/src/messages/cardano_transaction_snapshot_list.rs
```diff
@@ -62,19 +65,15 @@ mod tests {
 
     // Test the retro compatibility with possible future upgrades.
     #[test]
-    fn test_v1() {
+    fn test_v2() {
```
jpraynaud, line 68: Same comment as above:
```suggestion
    fn test_v1() {
```

jpraynaud (approved):
Good job! LGTM :rocket: 

I left few comments and typo fixes below.

mithril-aggregator/src/configuration.rs
```diff
@@ -156,6 +156,10 @@ pub struct Configuration {
 
     /// Cardano transactions prover cache pool size
     pub cardano_transactions_prover_cache_pool_size: usize,
+
+    /// Cardano transactions signing configuration
+    #[example = "`{ security_parameter: 3000, step: 90 }`"]
```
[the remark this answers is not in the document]
Alenar: It's the same for all other examples and in [the docusaurus website](https://mithril.network/doc/manual/developer-docs/nodes/mithril-aggregator#configuration-parameters). Maybe there's something to fix separately ?

mithril-aggregator/src/services/cardano_transactions_importer.rs
Alenar: We need the full chain point for the `BlockStreamer` called by this method.

mithril-aggregator/src/configuration.rs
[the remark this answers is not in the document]
jpraynaud: OK, I created this issue https://github.com/input-output-hk/mithril/issues/1729 :+1:

mithril-aggregator/src/services/cardano_transactions_importer.rs
Alenar: We may use a `BlockNumber` but doing so we need to transfer the responsibility to obtain a chain point from a block number to the streamer.

This would be cumbersome since:
- The lower bound is acquired by the importer (that part of its current role), to do so it do a database round-trip.
- We would need to define a trait that take a block number as parameter and return a chain point ... implemented by doing another database round-trip (so two needed instead of one as now).

Another possibility would be to remove altogether the lower bound computation from the importer and transfer it to the scanner.

mithril-client-cli/src/commands/cardano_transaction/snapshot_list.rs
Alenar: There's no network anymore in the message (btw there's no network for the same reason in the list of MithrilStakeDistribution).

mithril-aggregator/src/services/cardano_transactions_importer.rs
jpraynaud: Indeed this shoudl probably be the responsibility of the scanner. T
his can be done in a separate PR :+1:

mithril-client-cli/src/commands/cardano_transaction/snapshot_list.rs
Alenar: I checked and it would be difficult to add back the network in the messages since neither the `SignedEntity<T: Artifact>` nor the CardanoTransactionSnapshot artifact have this property.
We would at least need the adapter to know this property (so build it in the dependency system).

And if this property matters why does it not matter for other signed entities ? If yes we should add it to the `SignedEntity<T>` type.

sfauvel (approved):
Looks good

mithril-client-cli/src/commands/cardano_transaction/snapshot_list.rs
jpraynaud: Let's keep it like this for now :ok_hand:

mithril-aggregator/src/artifact_builder/cardano_transactions.rs
```diff
@@ -104,7 +104,7 @@ mod tests {
         let cardano_transaction_artifact_builder =
             CardanoTransactionsArtifactBuilder::new(Arc::new(mock_prover));
         cardano_transaction_artifact_builder
-            .compute_artifact(CardanoDbBeacon::default(), &certificate)
+            .compute_artifact(12390, &certificate)
```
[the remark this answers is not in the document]
Alenar: Only the certificate matter here, we will modify the tests to make that more obvious.

mithril-common/src/messages/cardano_transaction_snapshot_list.rs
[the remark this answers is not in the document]
Alenar: This part always was clunky as those "versions" only existing in those tests names (there's something to do here).

Those tests needs to verify case such as:
- current version deserialization
- we updated the current version and want another test that check that:
  - we can parse json that contains now removed fields
  - we can parse json that contains renamed fields
  - we can parse json that miss newly added fields (using default values if available, else this would be a breaking change).

PR #1742, opened 2024-06-11
Sqlite auto rollback

### LINKED ISSUE #1741: Automatic rollback on SQL transactions
## Why
Currently the few times that we need to use SQLite transactions we don't handle at all an error in the system, meaning that we don't rollback immediately by rely on the fact that SQLite may rollback the transaction on it's own later.

## What
Design a lightweight `Transaction` system that wrap an SQLite connection and automatically rollback when it goes out of scope if the transaction is still active (an active transaction is a transaction has still yet to be rolled back or committed).

## How

- [x] Add a `Transaction` struct that wrap a SQLite connection and define a light API to do commits or rollbacks.
- [x] Implement `Drop` on the `Transaction` struct that automatically run a rollback if the transaction was still active ().

Alenar:
## Content

This PR add a light layer to easily open, commit, and rollback SQLite transactions.

This is done by adding a `Transaction` struct to `mithril-persistence` that:
- wrap a **borrowed** SQLite connection.
- define three public methods: `begin` (the constructor), `commit` and `rollback`.
- `commit` and `rollback` method move self, meaning that using either of them remove the transaction of the current scope: the compiler won't allow to use several `commit` or `rollback` on the same transaction.
- Implement `Drop` to automatically rollback the transaction when it goes out of scope if it was still active.

Also a `begin_transaction` extension method is added to the sqlite transaction so they can be created without needing to import the new type.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Comments
While this system make using SQLite transactions easier it should not be used on our main SQLite database until we implement a connection pool system instead of sharing a global connection. Else we would be at risk of opening a transaction when one was already opened leading to at minimum an error or at worst a crash.

## Issue(s)
Closes #1741

### DIFF
--- a/internal/mithril-persistence/src/sqlite/transaction.rs
+++ b/internal/mithril-persistence/src/sqlite/transaction.rs
@@ -22,15 +22,15 @@ impl<'a> Transaction<'a> {
 
     /// Commit the transaction.
     pub fn commit(mut self) -> Result<(), sqlite::Error> {
-        self.connection.execute("COMMIT TRANSACTION")?;
         self.is_active = false;
+        self.connection.execute("COMMIT TRANSACTION")?;
         Ok(())
     }
 
     /// Rollback the transaction.
     pub fn rollback(mut self) -> Result<(), sqlite::Error> {
-        self.connection.execute("ROLLBACK TRANSACTION")?;
         self.is_active = false;
+        self.connection.execute("ROLLBACK TRANSACTION")?;
         Ok(())
     }
 }
@@ -38,6 +38,8 @@ impl<'a> Transaction<'a> {
 impl Drop for Transaction<'_> {
     fn drop(&mut self) {
         if self.is_active {
+            // Unwrap should not happen here, otherwise it would mean that we have not handled
+            // correctly the transaction "active" state or that the connection was closed.
             self.connection.execute("ROLLBACK TRANSACTION").unwrap();
         }
     }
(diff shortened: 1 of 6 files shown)


dlachaume (approved):
LGTM 🚀

internal/mithril-persistence/src/sqlite/transaction.rs
```diff
@@ -0,0 +1,174 @@
...
+        Ok(())
+    }
+}
+
+impl Drop for Transaction<'_> {
+    fn drop(&mut self) {
+        if self.is_active {
+            // Unwrap should not happen here, otherwise it would mean that we have not handled
+            // correctly the transaction "active" state or that the connection was closed.
+            self.connection.execute("ROLLBACK TRANSACTION").unwrap();
```
sfauvel, line 43: We could call rollback function here.
Even if it's not necessary to set is_active to false in the drop function, it may help to understand states

internal/mithril-persistence/src/sqlite/transaction.rs
```diff
@@ -0,0 +1,174 @@
...
+        Ok(Self {
+            connection,
+            is_active: true,
+        })
+    }
+
+    /// Commit the transaction.
+    pub fn commit(mut self) -> Result<(), sqlite::Error> {
+        self.is_active = false;
+        self.connection.execute("COMMIT TRANSACTION")?;
```
sfauvel, line 26: We can extract a function that execute and set is_active to false to reduce places where we modify this value.

internal/mithril-persistence/src/sqlite/transaction.rs
Alenar: Sadly we can't call `rollback` here since it move the struct and `drop` only take the struct as a borrowed mut.

Instead I defined a constant for the `ROLLBACK TRANSACTION` instruction.

jpraynaud (approved):
LGTM :+1:

PR #1758, opened 2024-06-17
Signed entity lock mechanism

### LINKED ISSUE #1693: Lock signature of signed entity types during warm-up
## Why
During the warm-up phase of the signer/aggregator we need to have the associated signed entity type locked and not attempting to create signatures.

## What
Implement a lock mechanism which prevents the signed entity type being warmed-up to be signed when a lock is in place. 

## How
- [x] Implement a lock mechanism in common
- [x] Use the lock in the aggregator to block open message generation for the associated signed entity type
- [x] Use the lock in the signer to block the signature for the associated signed entity type

Alenar:
## Content

This PR add a lock mechanism to make our state machines skip a signed entity when needed (ie: when we need to do a costly computation on a signed entity but we still want our state machines to run).

On aggregator the skip is done when it try to create open messages.
On signer the skip is done when it check if it can sign a pending certificate.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation

## Issue(s)
Closes #1693

### DIFF
--- a/mithril-common/src/signed_entity_lock.rs
+++ b/mithril-common/src/signed_entity_lock.rs
@@ -1,0 +1,135 @@
+// characteristics:
+// - non-blocking: thread should not be blocked when trying to access / lock a signed entity
+// - multithreaded: multiple threads should be able to access / lock a signed entity
+
+use std::collections::BTreeSet;
+
+use tokio::sync::RwLock;
+
+use crate::entities::SignedEntityTypeDiscriminants;
+
+struct SignedEntityLock {
+    locked_entities: RwLock<BTreeSet<SignedEntityTypeDiscriminants>>,
+}
+
+impl SignedEntityLock {
+    pub fn new() -> Self {
+        Self {
+            locked_entities: RwLock::new(BTreeSet::new()),
+        }
+    }
+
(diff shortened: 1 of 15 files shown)


mithril-common/src/signed_entity_lock.rs
jpraynaud, line 1: It looks like it's more a `SignedEntityTypeLock` than a `SignedEntityLock`, maybe we should rename it?

mithril-common/src/signed_entity_lock.rs
```diff
@@ -0,0 +1,192 @@
...
+//! threads from modifying the same entity at the same time.
+
+use std::collections::BTreeSet;
+
+use tokio::sync::RwLock;
+
+use crate::entities::SignedEntityTypeDiscriminants;
+
+/// Non-blocking lock mechanism for signed entities to prevent multiple threads from
+/// modifying the same entity at the same time.
```
jpraynaud, line 13: This should probably mention the fact that the runtime will not work on a locked signed entity type, as there is no case yet where the same entity type is accessed by different threads.

jpraynaud (approved):
LGTM :+1:

PR #1763, opened 2024-06-18
Warmup/Preload cardano transactions & block range roots

### LINKED ISSUE #1692: Warmup import Cardano transactions at node startup
## Why
We want to avoid that the activation of the signature of Cardano transactions blocks the network and prevents it to create new certificates. Also we'd like to take advantage of the `2` epochs delay for newcomers in the protocol before they can sign after their first registration.

## What
Implement a warm-up in the signer and aggregator that imports Cardano transactions in a different thread wait node startup.

## How

- [x] Implement the import of the Cardano transactions when the signer/aggregator starts in a thread
  - [x] In the signer
  - [x] In the aggregator 
- [x] Lock signed entity type for signing until all the data are imported
  - [x] In the signer
  - [x] In the aggregator

## Later
- Compute the message to sign on the signer even if the signer is not already eligible to sign it (to extend the warm-up mechanism)
- ~~Launch the warm-up import at regular interval (~1 day)?~~

Alenar:
## Content

This PR includes a preload system for `CardanoTransaction` & `BlockRangeRoot`, it's simply a system that is spawned in a separate thread at aggregator & signer startup that run the transaction importer.
The CardanoTransaction signed entity type is lock during the preload and is released when it finish even if the preload fail.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation

## Comments
We designed this system specifically for the Cardano Transactions import first so we have a rough idea of what's needed before we can consider generalizing it to other tasks that need preload (such as immutable file digester cache or even maybe the import of signers pool tickers).

## Issue(s)
Closes #1692

### DIFF
--- a/mithril-aggregator/src/commands/serve_command.rs
+++ b/mithril-aggregator/src/commands/serve_command.rs
@@ -119,6 +119,16 @@ impl ServeCommand {
         let mut join_set = JoinSet::new();
         join_set.spawn(async move { runtime.run().await.map_err(|e| e.to_string()) });
 
+        // start the cardano transactions preloader
+        let cardano_transactions_preloader = dependencies_builder
+            .create_cardano_transactions_preloader()
+            .await
+            .with_context(|| {
+                "Dependencies Builder can not create cardano transactions preloader"
+            })?;
+        let preload_task =
+            tokio::spawn(async move { cardano_transactions_preloader.preload().await });
+
         // start the HTTP server
         let (shutdown_tx, shutdown_rx) = oneshot::channel();
         let routes = dependencies_builder
@@ -179,6 +189,10 @@ impl ServeCommand {
         // stop servers
         join_set.shutdown().await;
         let _ = shutdown_tx.send(());
+
+        if !preload_task.is_finished() {
+            preload_task.abort();
+        }
 
         info!("Event store is finishing...");
         event_store_thread.await.unwrap();
--- a/mithril-common/src/cardano_transactions_preloader.rs
+++ b/mithril-common/src/cardano_transactions_preloader.rs
@@ -1,0 +1,212 @@
+//! # Cardano Transaction Preloader
+//!
+//! This module provides a preload mechanism for Cardano Transaction signed entity, allowing
+//! to compute in advance the Transactions & Block Range Root to be signed.
+
+use std::sync::Arc;
+
+use anyhow::Context;
+use slog::{debug, trace, Logger};
+
+use crate::chain_observer::ChainObserver;
+use crate::entities::{CardanoTransactionsSigningConfig, SignedEntityTypeDiscriminants};
+use crate::signable_builder::TransactionsImporter;
+use crate::signed_entity_type_lock::SignedEntityTypeLock;
+use crate::StdResult;
+
+/// Preload mechanism for Cardano Transaction signed entity, allowing
+/// to compute in advance the Transactions & Block Range Root to be signed.
+pub struct CardanoTransactionsPreloader {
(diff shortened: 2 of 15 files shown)


dlachaume (approved):

mithril-common/src/cardano_transactions_preloader.rs
```diff
@@ -0,0 +1,212 @@
...
+            cardano_transactions_signing_config,
+            chain_observer,
+            logger,
+        }
+    }
+
+    /// Preload the Cardano Transactions by running the importer up to the current chain block number.
+    pub async fn preload(&self) -> StdResult<()> {
+        debug!(self.logger, "🔥 Preload Cardano Transactions - Started");
+        trace!(self.logger, "🔥 Locking signed entity type"; "entity_type" => "CardanoTransactions");
```
jpraynaud, line 48: Maybe we can put higher log level?
```suggestion
        info!(self.logger, "🔥 Preload Cardano Transactions - Started");
        debug!(self.logger, "🔥 Locking signed entity type"; "entity_type" => "CardanoTransactions");
```

mithril-signer/src/main.rs
```diff
@@ -236,6 +239,10 @@ async fn main() -> StdResult<()> {
         .send(())
         .map_err(|e| anyhow!("Metrics server shutdown signal could not be sent: {e:?}"))?;
 
+    if !preload_task.is_finished() {
```
jpraynaud, line 242: Could we add the `preload_task` to the `join_set` instead?

mithril-aggregator/src/commands/serve_command.rs
```diff
@@ -119,6 +119,16 @@ impl ServeCommand {
...
         join_set.spawn(async move { runtime.run().await.map_err(|e| e.to_string()) });
 
+        // start the cardano transactions preloader
+        let cardano_transactions_preloader = dependencies_builder
+            .create_cardano_transactions_preloader()
+            .await
+            .with_context(|| {
+                "Dependencies Builder can not create cardano transactions preloader"
+            })?;
+        let preload_task =
```
jpraynaud, line 129: Could we add the `preload_task` to the `join_set` instead?

mithril-signer/src/runtime/signer_services.rs
```diff
@@ -294,6 +297,18 @@ impl<'a> ServiceBuilder for ProductionServiceBuilder<'a> {
             cardano_transactions_builder,
         ));
         let metrics_service = Arc::new(MetricsService::new().unwrap());
+        let cardano_transactions_preloader = Arc::new(CardanoTransactionsPreloader::new(
+            signed_entity_type_lock.clone(),
+            transactions_importer.clone(),
+            // Only relevant for the preloader since the state machine get the block number to import
+            // through the pending certificate, should it be configurable ?
```
jpraynaud, line 304: Yes we should configure it, but use some default values on the signer to allow transparent update (the default value will probably change when transactions are imported with low latency).

jpraynaud (approved):
LGTM :+1: 

I left few comments and question below.

mithril-signer/src/main.rs
Alenar: No, that's the error that we did first.

The `join_set` only hold infinite tasks or tasks that wait for a signal (like `ctrl+c`), doing so allow us to wait for the first task of to finish and then to stop all other tasks.
So if we add the `preload_task` to the `join_set` it makes the note stop immediately after the preload_task finish, that's why it must not be included in the `join_set`.

PR #1769, opened 2024-06-19
Import transactions in sequence on signer

### LINKED ISSUE #1766: Import Cardano transactions by sequences of block ranges
## Why
We need to keep a footprint as low as possible on the machine running a Mithril signer. We have noticed that the current process to import the Cardano transactions does not allow an efficient pruning strategy on the signer: the block range roots are computed only once all the transactions are imported. The import mechanism should be able to compute block range roots along the way in order to allow early pruning and overall low disk space requirements.

## What
Compute block range roots as soon as they can be computed by the signer and allow for early pruning on the transaction store.

## How

- [x] ~~During warmup phase on the signer, run the import of transactions with the Transaction importer on a sequence of ranges of block numbers (e.g. every `10` block range root length or `150` blocks)~~
- [x] Implement a decorator on the Transaction importer to import by sequence of block ranges
- [x] Assess the required disk space for a signer node (`~100MB` with `90M` transactions)

Alenar:
## Content

This PR break up in chunk the `mithril-signer` transactions import process, so the transactions pruning is done more frequently, reducing the number of transactions stored at any given time, reducing the disk footprint.

The chunk size is set to `1500` and is configurable.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation

## Issue(s)
Closes #1766

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ As a minor extension, we have adopted a slightly different versioning convention
   - Optimize the performances of the computation of the proof with a Merkle map.
   - Handle rollback events from the Cardano chain by removing stale data.
   - Preload Cardano transactions and Block Range Roots at signer & aggregator startup.
+  - Chunk the Cardano transactions import in `mithril-signer` to reduce disk footprint by running the pruning process more frequently.
 
 - Crates versions:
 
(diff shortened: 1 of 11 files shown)


dlachaume (approved):

jpraynaud (approved):
LGTM :+1:

PR #1771, opened 2024-06-24
Vacuum sqlite database at startup & after transactions preload

Alenar:
## Content

This PR:

- Enable SQLite auto-vacuum for the Cardano transactions databases so they shrink automatically.
- Make `mithril-aggregator` & `mithril-signer` Vacuum their databases at startup (this can take a while so a log message is print).
- Modify preloading of Cardano transaction in `mithril-signer` so it execute a Vacuum after each chunk imported.
We do that since we observed that auto-vacuum isn't executed until the whole import is finished (maybe because the db is a lot busy in this phase) so a manual operation is needed.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation

## Comments

After merging #1766 we observed that the SQLite database file size did not decrease after pruning transactions in signer. This is due to SQLite not automatically reclaiming free pages by itself, we need to use the [`vacuum` operation](https://www.sqlite.org/lang_vacuum.html) to do that.

Note: Vacuum reconstruct the SQLite by creating a new file, that mean that it needs an exclusive lock on the database so we can't use it after every deletion carelessly.

For everyday task we enable [auto-vacuum](https://www.sqlite.org/pragma.html#pragma_auto_vacuum) that makes SQLite handle it by itself but a the cost of greater fragmentation.
Given this drawback we only enable it on Cardano transactions databases since they are the only one were removal of data occurs on large scale, otherwise simply executing Vacuum at startup is enough.

## Issue(s)
Relates to #1766

### DIFF
--- a/.github/workflows/actions/toolchain-and-cache/action.yml
+++ b/.github/workflows/actions/toolchain-and-cache/action.yml
@@ -9,6 +9,9 @@ inputs:
     description: Space seperated list of cargo tools to install
     required: false
     default: ''
+  github-token:
+    description: GitHub token to use for installing cargo tools
+    required: true
 runs:
   using: "composite"
   steps:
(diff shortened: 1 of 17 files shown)


sfauvel (approved):
LGTM
Good job

dlachaume (approved):

jpraynaud (approved):

PR #1779, opened 2024-06-25
Fix #1762: CardanoTransactionSnapshot block number is not certified

### LINKED ISSUE #1762: Transactions of the block number in the artifact are never certified by the prover
## Why
The transactions included in the block number referenced in the artifact (as the last block number signed) are never certified by the prover. Some early investigations have shown that it only the transactions from the penultimate block range are certified. 

## What
Investigate and fix the problem that prevents some transactions to be certified when they are signed.

## How

- [x] Investigate the problem
- [x] Fix the problem

Alenar:
## Content

This PR fixes #1762:

It was due to the request to retrieve the pre-computed block range roots in database using a exclusive bound comparison instead of an inclusive, leading to the exclusion of the last block range root from the retrieved data.

This means that not only the prover was impacted but the protocol message generation for the single & multi signatures too.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Comments
Even if `mithril-signer` and `mithril-aggregator` are not modified by this PR, their behavior is impacted so their version number should be upgraded.

## Issue(s)
Closes #1762

### DIFF
--- a/internal/mithril-persistence/src/database/repository/cardano_transaction_repository.rs
+++ b/internal/mithril-persistence/src/database/repository/cardano_transaction_repository.rs
@@ -146,15 +146,17 @@ impl CardanoTransactionRepository {
             )
     }
 
-    /// Retrieve all the Block Range Roots in database up to the given end block number excluded.
+    /// Retrieve all the Block Range Roots in database up to the given end block number included.
     pub async fn retrieve_block_range_roots_up_to(
         &self,
-        end_block_number: BlockNumber,
+        up_to_or_equal_end_block_number: BlockNumber,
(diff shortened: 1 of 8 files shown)


internal/mithril-persistence/src/database/repository/cardano_transaction_repository.rs
```diff
@@ -149,12 +149,14 @@ impl CardanoTransactionRepository {
     /// Retrieve all the Block Range Roots in database up to the given end block number excluded.
```
jpraynaud, line 149: This comment should be modified as well

jpraynaud (approved):
LGTM :+1:

dlachaume (approved):

Issue #1785, opened 2024-06-27
Cardano transaction importer does not import the last block advertised as certified


jpraynaud:
## Why
The Cardano transaction importer does not import the last block displayed as certified:

- In the aggregator artifact route:
```json
{
  "merkle_root": "05e65f0c58565d1097254ae35432a61a26f06426b2a723ecdac63126bd5d595e",
  "epoch": 493,
  "block_number": 10499370,
  "hash": "877cbbae948ada10c16738d17c53cd98ecf9be43036332a0fff7de5993c99741",
  "certificate_hash": "44c00ffa841ca19529e664f0379481f7afb6e7a76e77ed6fbc413208e81fe0ab",
  "created_at": "2024-06-27T11:00:39.633969066Z"
}
```

- In the aggregator logs:
```json
{"msg":"Compute protocol message for CardanoTransactions at block_number: 10499370","v":0,"name":"slog-rs","level":20,"time":"2024-06-27T10:54:03.200240225Z","hostname":"891e6000d0f1","pid":1}
{"msg":"TransactionsImporter will retrieve Cardano transactions between block_number '10499339' and '10499370'","v":0,"name":"slog-rs","level":20,"time":"2024-06-27T10:54:03.200730418Z","hostname":"891e6000d0f1","pid":1}
{"msg":"ChainReaderBlockStreamer polls next","v":0,"name":"slog-rs","level":20,"time":"2024-06-27T10:54:03.201524327Z","hostname":"891e6000d0f1","pid":1}
{"msg":"ChainReaderBlockStreamer received a RollBackward(127916665)","v":0,"name":"slog-rs","level":20,"time":"2024-06-27T10:54:03.202030767Z","hostname":"891e6000d0f1","pid":1}
{"msg":"ChainReaderBlockStreamer polls next","v":0,"name":"slog-rs","level":20,"time":"2024-06-27T10:54:34.895605277Z","hostname":"891e6000d0f1","pid":1}
{"msg":"TransactionsImporter - computing Block Range Roots","v":0,"name":"slog-rs","level":20,"time":"2024-06-27T10:54:34.94627066Z","hostname":"891e6000d0f1","pid":1,"end_block":10499370,"start_block":10499340}
```

- In the aggregator database:
```bash
+------------------------------------------------------------------+--------------+-----------------------+-------------+------------------------------------------------------------------+
|                         transaction_hash                         | block_number | immutable_file_number | slot_number |                            block_hash                            |
[...]

PR #1795, opened 2024-07-03
Bugfix: Can't generate proof for transactions on the last signed block number

Alenar:
## Content

This PR includes a fix for #1785, there was several problems:

- The computed block number to be signed was not taking in account the fact that block range ends is exclusive, meaning that it overflowed over the next block range, a block range which end after the block number that we want to sign.
  To solve this we simply subtract one to the computation.
- The request that retrieved block range when computing merkle roots excluded the last block range.

Not directly related to the bug: 
- The prover would add leaf to it's merkle tree in some condition, corrupting it as this changed its merkle root so it no longer matcher the one in the certificate.
  This was worrisome since this corrupted merkle tree is returned to a resource pool after use, meaning that a subsequent requests would also yield invalid proofs.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Comments
I'm unsure this totally fix the problem as when investing the logs I could see that, in the transaction importer, they were a gap of minus one between the block number of the transaction we asked it to import and the really imported transactions.
We should oversee the testing preview network after merge to ensure that its fixed or if there's still more work needed.

## Issue(s)
Relates to #1785

### DIFF
--- a/mithril-common/src/crypto_helper/merkle_map.rs
+++ b/mithril-common/src/crypto_helper/merkle_map.rs
@@ -74,7 +74,7 @@ impl<K: MKMapKey, V: MKMapValue<K>> MKMap<K, V> {
                     "MKMap values should be replaced by entry with same root"
                 ));
             }
-            return self.replace(key, value);
+            return self.replace_unchecked(key, value);
         } else {
             let key_max = self.inner_map_values.keys().max();
             if key_max > Some(&key) {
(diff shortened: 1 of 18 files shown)


sfauvel (approved):
LGTM

dlachaume (approved):
LGTM 👍

mithril-common/src/crypto_helper/merkle_map.rs
```diff
@@ -480,6 +491,31 @@ mod tests {
         MKTree::new(&leaves).unwrap()
     }
 
+    fn join_merkle_tree(block_ranges: &[BlockRange]) -> Vec<(BlockRange, MKTree)> {
```
jpraynaud, line 494: The naming of this function is not very clear:
```suggestion
    fn generate_merkle_trees_for_ranges(block_ranges: &[BlockRange]) -> Vec<(BlockRange, MKTree)> {
```

jpraynaud (approved):
LGTM

PR #1796, opened 2024-07-03
Preload Cardano transactions based on Aggregator's Signing Capability

### LINKED ISSUE #1782: Signer preloads transactions when Cardano transactions certification is not activated
## Why
The signer node imports the Cardano transactions, even if this type of data is not activated on the network. The expected behavior is to have the preloading triggered only when the aggregator advertises the Cardano transaction signed entity type.

## What
Trigger the signer preloading of the Cardano transactions only if the aggregator advertises the `CardanoTransactions` type in its root route.

## How

- [ ] TBD

dlachaume:
## Content
This PR includes

## Pre-submit checklist

- Branch
- PR
  - [ ] Self-reviewed the diff
  - [ ] Reviewer requested
- Documentation

## Issue(s)
Closes #1782


mithril-common/src/messages/aggregator_features.rs
Alenar, line 1: Should we add a retro compatibility test like what we did for other messages ?

Alenar (commented):
LGTM :rocket:

Alenar (approved):

mithril-common/src/messages/aggregator_features.rs
dlachaume: Done ✅

mithril-common/src/cardano_transactions_preloader.rs
```diff
@@ -14,6 +15,32 @@ use crate::signable_builder::TransactionsImporter;
 use crate::signed_entity_type_lock::SignedEntityTypeLock;
 use crate::StdResult;
 
+/// CardanoTransactionsPreloaderActivationState
```
jpraynaud, line 18: The name is not very clear, maybe this could be better as it expresses what it is doing:
```suggestion
/// CardanoTransactionsPreloaderChecker
```

Also a sentence is welcome to explain the purpose of the trait, just its name is not enough.

mithril-common/src/cardano_transactions_preloader.rs
```diff
@@ -14,6 +15,32 @@ use crate::signable_builder::TransactionsImporter;
...
+#[async_trait]
+pub trait CardanoTransactionsPreloaderActivationState: Send + Sync {
+    /// Determine if the Cardano Transactions Preloader should preload.
+    async fn is_activated(&self) -> StdResult<bool>;
+}
+
+/// CardanoTransactionsPreloaderActivation
+pub struct CardanoTransactionsPreloaderActivation {
```
jpraynaud, line 26: This type should probably be defined in the aggregator as it is the implementation that it will use.

mithril-signer/src/cardano_transactions_preloader_activation_according_to_aggregator.rs
```diff
@@ -0,0 +1,120 @@
...
+};
+
+use crate::AggregatorClient;
+
+/// CardanoTransactionsPreloaderActivationAccordingToAggregator
+pub struct CardanoTransactionsPreloaderActivationAccordingToAggregator {
+    aggregator_client: Arc<dyn AggregatorClient>,
+}
+
+impl CardanoTransactionsPreloaderActivationAccordingToAggregator {
```
jpraynaud, line 18: The name is a bit cumbersome:
```suggestion
impl CardanoTransactionsPreloaderActivationSigner {
```

So is the file name, maybe this is enough `cardano_transactions_preloader_checker.rs`.

mithril-common/src/cardano_transactions_preloader.rs
```diff
@@ -32,18 +60,28 @@ impl CardanoTransactionsPreloader {
...
+            activation_state,
         }
     }
 
     /// Preload the Cardano Transactions by running the importer up to the current chain block number.
     pub async fn preload(&self) -> StdResult<()> {
+        if !self.is_activated().await? {
+            info!(
+                self.logger,
+                "🔥 Preload Cardano Transactions - Not running, conditions not met"
```
jpraynaud, line 80: Can we replace all the :fire: with a :warning: or :information_source: in the preloader? 
This will help avoid confusion IMHO

Also this log should be removed or kept as a trace.

mithril-common/src/cardano_transactions_preloader.rs
```diff
@@ -105,8 +148,19 @@ mod tests {
         }
     }
 
+    struct CardanoTransactionsPreloaderActivationWithError {}
```
jpraynaud, line 151: This could probably have been mocked instead?

mithril-common/src/cardano_transactions_preloader.rs
```diff
@@ -131,11 +185,57 @@ mod tests {
...
             TestLogger::stdout(),
+            Arc::new(CardanoTransactionsPreloaderActivation::new(true)),
+        );
+
+        preloader.preload().await.unwrap();
+    }
+
+    #[tokio::test]
+    async fn do_not_call_its_inner_importer_when_is_not_activated() {
+        let security_parameter = 542;
```
jpraynaud, line 196: This variable could be inlined here and in tests below.

mithril-common/src/messages/aggregator_features.rs
```diff
@@ -0,0 +1,91 @@
+use std::collections::BTreeSet;
+
+use serde::{Deserialize, Serialize};
+
+use crate::entities::SignedEntityTypeDiscriminants;
+
+/// Message sent by an Aggregator to inform about its features
```
jpraynaud, line 7:
```suggestion
/// Message advertised by an Aggregator to inform about its features
```

openapi.yaml
```diff
@@ -531,7 +531,7 @@ paths:
 
 components:
   schemas:
-    RootMessage:
+    AggregatorFeaturesMessage:
       description: Aggregator public information
```
jpraynaud, line 535: Can you add a bit more information in the description?

mithril-signer/src/runtime/signer_services.rs
```diff
@@ -326,12 +327,16 @@ impl<'a> ServiceBuilder for ProductionServiceBuilder<'a> {
             cardano_transactions_builder,
         ));
         let metrics_service = Arc::new(MetricsService::new().unwrap());
+        let preloader_activation = CardanoTransactionsPreloaderActivationAccordingToAggregator::new(
+            certificate_handler.clone(),
```
jpraynaud, line 331: The name should be changed here:
```suggestion
            aggregator_client.clone(),
```

jpraynaud (commented):
Overall looks good :+1: 

I left few comments below.

mithril-common/src/cardano_transactions_preloader.rs
dlachaume: We believe that this minimalist implementation has the right place here to facilitate its basic use. Every context that need it can simply use it, for instance in the testing contexts in the Signer node.

jpraynaud: It'd be better to stick to the usual way of doing it: implementing Fake/Mock for testing and keeping node specific implementations where they belong.

mithril-common/src/cardano_transactions_preloader.rs
Alenar: The idea with adding emoji is to make the service stand up in the log. So if we use one we should choose it carefully so it's not reused by another service.

Also I think that a info or debug is the right level, trace seems a bit too low since this log won't be spammed and have business related.

Beyond that I rethought about our log usage and ... we probably are using them wrongly.

The main culprit is `slog_scope`, we should only use it for logs not scoped in a service and all services should have a child logger (created using [`Logger::new(&self, values)`](https://docs.rs/slog/latest/slog/struct.Logger.html#method.new)) with a `name` property that identify them (we should also set some property on the root logger).

Basically we should do this:
* At the global logger creation:
```rust
 Logger::root(Arc::new(drain), slog::o!("name" => "aggregator"));
``` 
* When creating a service
```rust
let verifier = Arc::new(MithrilCertificateVerifier::new(
    self.get_logger()?.new(o!("name" => "certificate_verifier")),,
    self.get_certificate_repository().await?,
));
``` 

Doing so our logs instead of looking like this:
```json
{"msg":"EpochService::update_protocol_parameters","v":0,"name":"slog-rs","level":20} 
``` 
Our logs could look like this:
```json
{"msg":"update_protocol_parameters","v":0,"name":"EpochService","level":20} 
```

Alenar: Even better, the [Logger::new(&self, values)](https://docs.rs/slog/latest/slog/struct.Logger.html#method.new) method could be called by the service constructor so nothing change in the builder and the service decide itself the name that it want to use.

jpraynaud: We are on the same page, the :fire: is simply not a very judicious choice in that case: it could be used in case of a crash or something going wrong in the node, which is not the case. :warning: or :information_source: where just suggestions.

I keep thinking that this specific log should not be displayed in info, but debug is probably a good match instead of trace.

And yes, we can probably make some refactoring on the logs as you propose.

Alenar: For the emoji I would prefer something that can be related to "preloading", like: ["⟲"](https://symbl.cc/en/27F2/), ["⟳"](https://symbl.cc/en/27F3/), or ["⤾"](https://symbl.cc/en/293E/).

mithril-signer/src/cardano_transactions_preloader_activation_according_to_aggregator.rs
dlachaume: Done ✅

mithril-common/src/cardano_transactions_preloader.rs
dlachaume: Done ✅

mithril-common/src/cardano_transactions_preloader.rs
dlachaume: Done ✅

mithril-common/src/messages/aggregator_features.rs
dlachaume: Done ✅

openapi.yaml
dlachaume: Done ✅

openapi.yaml
```diff
@@ -532,7 +532,7 @@ paths:
 components:
   schemas:
     AggregatorFeaturesMessage:
-      description: Aggregator public information
+      description: Represents general information about Mithril project, Aggregator public information and signing capabilities
```
jpraynaud, line 535:
```suggestion
      description: Represents general information about Aggregator public information and signing capabilities
```

jpraynaud (approved):
LGTM

mithril-common/src/cardano_transactions_preloader.rs
sfauvel: Done

openapi.yaml
sfauvel: Done

sfauvel (approved):
LGTM

PR #1824, opened 2024-07-12
Fix: Explorer keep previous error until refresh in cardano transaction certification modal

### LINKED ISSUE #1818: Explorer keeps previous Cardano transaction error
## Why

In the Cardano transaction verification modal of the explorer, the error message is displayed even if the transaction is valid:
- if a previous transaction verification returned an error
- the problem is fixed when reloading the page

Image

## What
Fix the error handling in the explorer Cardano transaction verification modal 

## How

- [ ] Fix the explorer Cardano transaction verification modal

Alenar:
## Content

This PR fix #1818: when an error was catch by the Cardano transaction certification modal it was never reset so it would reappear even if the certification was successful, only a refresh would reset it.
This solve this problem by resetting the error to its default state each time the modal is opened.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Comments

This PR also do a temporary fix to the `make upgrade` of the explorer: since it was crashing because of incompatibility between two dependency it fix one of those dependency to a working version (until the problem is solved by those third party then we will go back to the previous behavior of updating to the latest version available).

## Issue(s)

Closes #1818


jpraynaud (approved):
LGTM

PR #1852, opened 2024-07-23
fix: `ChainReader` fails on `find_intersect_point` without agency

### LINKED ISSUE #1836: `ChainReader` is stuck in `Agency is theirs` error
## Why
We have noticed that when we are retrieving the blocks from the chain reader at a depth very close to the tip (e.g. in the e2e test), it can get stuck in the following error: `Agency is theirs`. After investigation, we suspect that the problem comes from the call to the `find_intersect_point` function of the `chainsync` mini-protocol in the `set_chain_point` function: this call should be made only when agency is ours in the mini-protocol.

## What
Fix the `ChainReader` implementation so that it does not trigger the error and just make it wait for a new block from the Cardano node (if exists) by shunting the call to the `find_intersect_point` call.

## How

- [x] Investigate the problem
- [ ] Fix the `ChainReader` implementation

jpraynaud:
## Content
This PR includes a fix to the **Chain Reader**: 
when the **Chain Sync** client does not have agency (i.e. is waiting for the server to callback), the find intersect point call to the **Chain Sync** is not allowed and is now skipped in the `find_intersect_point` function of the **Chain Reader**.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Issue(s)
Closes #1836

### DIFF
--- a/mithril-common/src/chain_reader/pallas_chain_reader.rs
+++ b/mithril-common/src/chain_reader/pallas_chain_reader.rs
@@ -54,9 +54,11 @@ impl PallasChainReader {
         let client = self.get_client().await?;
         let chainsync = client.chainsync();
 
-        chainsync
-            .find_intersect(vec![point.to_owned().into()])
-            .await?;
+        if chainsync.has_agency() {
+            chainsync
+                .find_intersect(vec![point.to_owned().into()])
+                .await?;
+        }
 
         Ok(())
     }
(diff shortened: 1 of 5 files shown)


mithril-common/src/chain_reader/pallas_chain_reader.rs
```diff
@@ -214,7 +216,10 @@ mod tests {
                 chainsync_server.recv_while_idle().await.unwrap();
 
                 if has_agency == HasAgency::No {
+                    println!("Has agency: No");
```
falcucci, line 219: should be these prints be really needed?

falcucci (approved):
left just a comment but to take or leave it.

LGTM

mithril-common/src/chain_reader/pallas_chain_reader.rs
jpraynaud: No, I removed them :+1:

dlachaume (approved):

sfauvel (approved):

PR #1877, opened 2024-08-01
Fix: `ChainReaderBlockStreamer` skips `RollForward` blocks at the tip

### LINKED ISSUE #1875: `dev-mainnet` not signing Cardano transaction after full import
## Why
We have noticed that following the full import of the Cardano transactions in the `dev-mainnet` network, the certification of the transactions stopped after a few successful signatures.

After investigation, it appears that the aggregator of the `dev-mainnet` network is missing a row for a block number in its database (confirmed with the `testing-mainnet` network), which causes the mismatch of signatures:
```
+------------------------------------------------------------------+--------------+-------------+------------------------------------------------------------------+
|                         transaction_hash                         | block_number | slot_number |                            block_hash                            |
+------------------------------------------------------------------+--------------+-------------+------------------------------------------------------------------+
| 58808ad93c3fb337a91bac2cda44f47a06611e1c797d8cf7d06ed31dcec4b2eb | 10641870     | 130823807   | 8ef8d030154695569b7d8a2c0074c5a340b7d4224773dac054877224cbfffd35 |
+------------------------------------------------------------------+--------------+-------------+------------------------------------------------------------------+
```

We have also noticed that the problem occurs only when the import is done very close from the tip of the chain, which is probably the source of the problem. An analysis from the logs of the aggregator showed that the problem is likely due to bounds management in the [ChainReaderBlockStreamer::get_next_chain_block_action](https://github.com/input-output-hk/mithril/blob/7a7396e9f3db91842430e13e290b8eecc5fc3e5b/mithril-common/src/cardano_block_scanner/chain_reader_block_streamer.rs#L111-L112).

## What
Investigate and fix the problem.

## How

- [x] Investigate the problem
- [ ] Fix the problem by fixing the handling of the above threshold of the `ChainReaderBlockStreamer::get_next_chain_block_action`

jpraynaud:
## Content
This PR includes a fix to the `ChainReaderBlockStreamer` which skips `RollForward` blocks at the tip of the chain.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Issue(s)
Closes #1875

### DIFF
--- a/mithril-common/src/cardano_block_scanner/chain_reader_block_streamer.rs
+++ b/mithril-common/src/cardano_block_scanner/chain_reader_block_streamer.rs
@@ -41,8 +41,11 @@ impl BlockStreamer for ChainReaderBlockStreamer {
                 Some(BlockStreamerNextAction::ChainBlockNextAction(
                     ChainBlockNextAction::RollForward { parsed_block },
                 )) => {
+                    let parsed_block_number = parsed_block.block_number;
                     roll_forwards.push(parsed_block);
-                    if roll_forwards.len() >= self.max_roll_forwards_per_poll {
+                    if roll_forwards.len() >= self.max_roll_forwards_per_poll
+                        || parsed_block_number >= self.until
+                    {
                         return Ok(Some(ChainScannedBlocks::RollForwards(roll_forwards)));
                     }
                 }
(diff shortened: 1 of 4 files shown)


mithril-common/src/cardano_block_scanner/chain_reader_block_streamer.rs
```diff
@@ -239,11 +242,19 @@ mod tests {
...
+                    Vec::<&str>::new(),
+                ),
+            },
         ])));
         let mut block_streamer = ChainReaderBlockStreamer::try_new(
-            chain_reader,
+            chain_reader.clone(),
             None,
-            BlockNumber(100),
+            BlockNumber(2),
```
sfauvel, line 257: We no longer test the case where the block number is greater than the last block.
Is this case no longer relevant ?

sfauvel (approved):
LGTM

dlachaume (approved):

mithril-common/src/cardano_block_scanner/chain_reader_block_streamer.rs
jpraynaud: You're right, I added back the test :+1:

Issue #1837, opened 2024-07-17
Mithril signature diffusion with Cardano network layer PoC


jpraynaud:
## Why
We want to create a PoC for implementing the Cardano network decentralized message queue in the Mithril network.
The decentralized message queue is not fully specified and not implemented: this means that we will have to simulate it.

The outcome of the PoC will be:
- Can we implement the n2c mini protocols of the decentralized message queue in the signer and aggregator.
- Assess which type of requirements will be needed when implementing the Cardano decentralized message queue.
- Assess which type of modification will be needed on the long run in the signer and aggregator.

## What
Implement a PoC where signatures are transmitted from a signer to an aggregator through a simulated decentralized message queue. 

Ideally, we want to demonstrate a Mithril devnet fully operating with signature diffusion by the simulated Cardano message queue.

## How

- [x] Abstract interactions:
  - [x] between the signer and the Cardano message queue with a Publisher (sending signatures)
  - [x] between the Cardano message queue and the aggregator with a Consumer (receiving signatures)
- [x] Implement the Consumer in the aggregator:
  - [x] Run the consumer in a new thread
  - [x] The consumer constantly reads the local Unix socket
  - [x] When a new single signature is available, it is send to the Certifier service (`register_single_signature` function)
- [x] Implement the Publisher in the signer:
  - [x] Replace/adapt the aggregator client to send signatures on a local Unix socket
- [x] Simulated Cardano network decentralized message queue with a separate process (new crate)
  - [x] Simulate the Cardano server behavior of the Local Message Submission mini protocol:
    - [x] Open a server on a local Unix socket
    - [x] Upon reception of a call with a signature on the server write to the signature to a channel
  - [x] Simulate the Cardano server behavior of the Local Message Notification mini protocol:
    - [x] Open a server on a local Unix socket
    - [x] Upon reception of a call to retrieve the next signatures read the signature from the channel (if any)

## Questions/Resources
- How to implement a simple Cardano network decentralized message queue simulator? 
  - use `warp` and make it listen on a Unix socket as in https://github.com/masonforest/warp/commit/9b15a710dbe7d6b36c61d04830c2b2989e89a5e6
- How to send HTTP requests on a local Unix socket?
  - Use `curl` with `--unix-socket` option
- Serialize/Deserialize Cbor hex encoded messages (not in first version)???

PR #1902, opened 2024-08-27
POC - Mithril signature diffusion with a "fake Cardano network layer"

Alenar:
## Content

This PR is a proof a concept of a decentralized message queue to diffuse Mithril signatures.

Since the protocol that will be used for signature diffusion through the Cardano node is still on the design phase, we simulate it with the creation of a new binary `signature-network-node`:

- Nodes have an Unix socket were you can register signatures (using http) and pull the available signatures known by the node.
- Nodes have an input directory were json serialized signatures can be put, it will be then added to the node available signatures.
- Nodes share the signatures that are registered to them by writing them as json in their peers input directory.
- Registered signatures can be pulled by calling a `pull-signatures` route on the node unix socket.

`mithril-signer` signature registration is abstracted with a new trait, `SignaturePublisher`, that have two implementations. One is the previous behavior communicating directly to the aggregator using http, the other foward signatures thought a `signature-network-node` unix socket.

`mithril-aggregator` have a new thread that pulls signatures constantly using  the `pull-signatures`  route of a `signature-network-node` unix socket and register them using the `certifier-service` the same way as its http server does.

`mithil-end-to-end` have been adapted to spawn one `signature-network-node` per mithril node.

## Comments

:stop_sign:  **DO NOT MERGE:** :stop_sign: This PR should not be merged as it is intended for experimentation only.

As this is a POC several shortcuts were used:

- `signature-network-node` stores registered signatures indefinitely until their `pull-signatures` route is called, meaning it's possible to make them crash by overloading their memory.
- Tests were written but they are not exhaustive.
- Windows isn't compatible.

## Issue(s)

Relates to #1837


jpraynaud (commented):
Good work :muscle:

PR #1904, opened 2024-08-27
Optimize memory usage of signer for Cardano transactions

### LINKED ISSUE #1903: Optimize memory usage of signer for Cardano transactions
## Why
In order to optimize the **signer memory usage on the SPO infrastructure when signing Cardano transactions**, we need our Merkle tree library to support multiple/custom store implementations optimized for:
- **Proving** in the aggregator: fast but memory intensive (serves both signature and proving)
- **Signing** in the signer: slower but memory optimized (serves only signature)

## What
Implement a Merkle tree store with optimized memory usage in the signer.

## How

- [x] Upgrade the Merkle tree implementation to support multiple stores
- [x] Implement a store for the signer with optimized memory usage based on SQLite
- [x] Benchmark the new signer store
- [ ] Update the signer footprint in documentation

jpraynaud:
## Content
This PR includes the **optimization of the memory usage of the signer** when signing Cardano transactions:
- The capability of the Merkle tree `MKTree` to support multiple `MKTreeStorer` implementations for data storage.
- The implementation of a signing specific Merkle store which is dedicated for the signer and with optimized memory usage.
- The benchmarks of the new signer Merkle store.
- Also some enhancements of the benchmarks of the Merkle tree.

### Signer footprint on the mainnet

The signer memory usage (Resident memory) is **~-67%** less.

Screenshot from 2024-08-27 18-52-00

### Benchmark

The Merkle tree creation takes **~+50%** longer than with the original in memory store.

- **Merkle tree creation for 100,000 leaves**
pdf
- **Merkle tree creation for 1,000,000 leaves**
pdf
- **Merkle tree creation for 10,000,000 leaves**
pdf

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Issue(s)
Closes #1903

### DIFF
--- a/mithril-signer/src/mktree_store_sqlite.rs
+++ b/mithril-signer/src/mktree_store_sqlite.rs
@@ -5,12 +5,7 @@ use std::{
 
 use anyhow::Context;
 use mithril_common::{
-    crypto_helper::{
-        ckb_merkle_mountain_range::{
-            Error as MMRError, MMRStoreReadOps, MMRStoreWriteOps, Result as MMRResult,
-        },
-        Bytes, MKTreeLeafIndexer, MKTreeLeafPosition, MKTreeNode, MKTreeStorer,
-    },
+    crypto_helper::{Bytes, MKTreeLeafIndexer, MKTreeLeafPosition, MKTreeNode, MKTreeStorer},
     StdResult,
 };
 
@@ -96,26 +91,6 @@ impl MKTreeStoreSqlite {
     }
 }
 
-impl MMRStoreReadOps<Arc<MKTreeNode>> for MKTreeStoreSqlite {
-    fn get_elem(&self, pos: u64) -> MMRResult<Option<Arc<MKTreeNode>>> {
-        self.get_element_at_position(pos)
-            .with_context(|| {
-                format!("MKTreeStoreSqlite failed to retrieve element at position {pos}")
-            })
-            .map_err(|e| MMRError::StoreError(e.to_string()))
-    }
-}
-
-impl MMRStoreWriteOps<Arc<MKTreeNode>> for MKTreeStoreSqlite {
-    fn append(&mut self, pos: u64, elems: Vec<Arc<MKTreeNode>>) -> MMRResult<()> {
-        self.insert_elements_from_position(pos, elems)
-            .with_context(|| {
-                format!("MKTreeStoreSqlite failed to insert elements from position {pos}")
-            })
-            .map_err(|e| MMRError::StoreError(e.to_string()))
-    }
-}
-
 impl Clone for MKTreeStoreSqlite {
     fn clone(&self) -> Self {
         unimplemented!("Clone is not implemented for MKTreeStoreSqlite")
@@ -143,6 +118,19 @@ impl MKTreeLeafIndexer for MKTreeStoreSqlite {
 impl MKTreeStorer for MKTreeStoreSqlite {
     fn build() -> StdResult<Self> {
         Self::build()
+    }
+
+    fn get_elem(&self, pos: u64) -> StdResult<Option<Arc<MKTreeNode>>> {
+        self.get_element_at_position(pos).with_context(|| {
+            format!("MKTreeStoreSqlite failed to retrieve element at position {pos}")
+        })
+    }
+
+    fn append(&self, pos: u64, elems: Vec<Arc<MKTreeNode>>) -> StdResult<()> {
+        self.insert_elements_from_position(pos, elems)
+            .with_context(|| {
+                format!("MKTreeStoreSqlite failed to insert elements from position {pos}")
+            })
     }
 }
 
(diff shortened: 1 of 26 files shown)


mithril-signer/src/mktree_store_sqlite.rs
```diff
@@ -0,0 +1,169 @@
...
+    }
+}
+
+impl Clone for MKTreeStoreSqlite {
+    fn clone(&self) -> Self {
+        unimplemented!("Clone is not implemented for MKTreeStoreSqlite")
+    }
+}
+
+impl MKTreeLeafIndexer for MKTreeStoreSqlite {
```
sfauvel, line 125: The MkTreeStoreSqlite is not a MkTreeLeafIndexer.
It would be better to have the trait only for MKTreeStoreInMemory and not allow to call `compute_proof` on a MKTree<MkTreeStoreSqlite>
This is probably a too complicated refactoring which also risks complicating the syntax.
This remains a design problem that we must remain vigilant about.

sfauvel (approved):
LGTM
I just see a design problem but without an obvious solution.

mithril-signer/src/mktree_store_sqlite.rs
```diff
@@ -0,0 +1,169 @@
...
+            elements
+                .into_iter()
+                .enumerate()
+                .fold(vec![], |mut vec, (i, elem)| {
+                    vec.append(&mut vec![
+                        sqlite::Value::Integer((position + i as u64) as i64),
+                        sqlite::Value::Binary((**elem).to_vec()),
+                    ]);
+                    vec
+                });
```
Alenar, line 79: I would prefer to use `flat_map` here as it's a simpler operation than `fold`.
```suggestion
                .flat_map(|(i, elem)| {
                vec![
                    sqlite::Value::Integer((position + i as u64) as i64),
                    sqlite::Value::Binary((**elem).to_vec()),
                ]
            })
            .collect();
```

mithril-signer/src/mktree_store_sqlite.rs
```diff
@@ -0,0 +1,169 @@
...
+/// This store is not suited for proving, it is only used to store the compute the root of a Merkle tree.
+/// This store is slower than the in memory store but uses less memory which is important to minimize the signer footprint on the SPO infrastructure.
+pub struct MKTreeStoreSqlite {
+    inner_store: RwLock<sqlite::ConnectionThreadSafe>,
+}
+
+impl MKTreeStoreSqlite {
+    fn build() -> StdResult<Self> {
+        Ok(Self {
+            inner_store: RwLock::new(Self::create_connection()?),
```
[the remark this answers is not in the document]
jpraynaud: You're right, it's not needed. I removed it in https://github.com/input-output-hk/mithril/pull/1904/commits/8da2bfe8d276e5028c2f0a718c0ab1e6c4dc9e24

mithril-signer/src/mktree_store_sqlite.rs
```diff
@@ -0,0 +1,169 @@
...
+        },
+        Bytes, MKTreeLeafIndexer, MKTreeLeafPosition, MKTreeNode, MKTreeStorer,
+    },
+    StdResult,
+};
+
+/// A Merkle tree store with Sqlite backend
+/// This store is used to store the Merkle tree nodes in a Sqlite database with in memory mode.
+/// This store is not suited for proving, it is only used to store the compute the root of a Merkle tree.
+/// This store is slower than the in memory store but uses less memory which is important to minimize the signer footprint on the SPO infrastructure.
```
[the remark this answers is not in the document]
jpraynaud: Done in https://github.com/input-output-hk/mithril/pull/1904/commits/8da2bfe8d276e5028c2f0a718c0ab1e6c4dc9e24

mithril-signer/src/mktree_store_sqlite.rs
```diff
@@ -0,0 +1,169 @@
...
+                element blob, 
+                primary key (position)
+            )",
+        )?;
+
+        Ok(connection)
+    }
+
+    fn get_element_at_position(&self, position: u64) -> StdResult<Option<Arc<MKTreeNode>>> {
+        let inner_store = self.inner_store.read().unwrap();
```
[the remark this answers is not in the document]
jpraynaud: It's not needed anymore as I removed the usage of `RwLock` in https://github.com/input-output-hk/mithril/pull/1904/commits/8da2bfe8d276e5028c2f0a718c0ab1e6c4dc9e24.
Otherwise, I'd rather have modified the code so that it properly returns an error in case of poisoning of the lock.

mithril-signer/src/mktree_store_sqlite.rs
jpraynaud: Done in https://github.com/input-output-hk/mithril/pull/1904/commits/8da2bfe8d276e5028c2f0a718c0ab1e6c4dc9e24

Alenar (approved):

mithril-signer/src/mktree_store_sqlite.rs
jpraynaud: You're right, I think that's a limitation of the design. 
However, the `MKTreeLeafIndexer` is only used when creating proofs which is not happening in the signer node.
In any case, I'll try to find a way to circumvent this problem in a different PR :+1:

PR #1909, opened 2024-08-30
Optional signed message in registered signatures

### LINKED ISSUE #1899: Add message signed in signature HTTP messages
## Why
In order to authenticate signature messages (which can arrive in advance to the aggregator, i.e. for an open message which is not known yet by the aggregator), we need to know the message which was used to create the Mithril individual signature.

## What
Add the message used to create the Mithril individual signature in the `RegisterSignatureMessage` as an optional field.

## How

- [ ] Add an optional field to the `RegisterSignatureMessage`
- [x] Make the signer add the message used to sign in the `RegisterSignatureMessage`

Alenar:
## Content

This PR add an optional field to single signatures: `signed_message`.

In order to decentralize signatures orchestration we will need to buffer incoming single signatures when the Aggregator has yet to create an open message for the signed entity type.
But in that case we won't be able to some validation to reject immediately malicious signatures from bad actors. 
This new field will allow us to do an integrity check of incoming signatures, by checking their validity against their signed message and the current stake distribution, .

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Issue(s)

Closes #1899

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3552,7 +3552,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-aggregator"
-version = "0.5.57"
+version = "0.5.58"
 dependencies = [
  "anyhow",
  "async-trait",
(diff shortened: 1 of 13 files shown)


jpraynaud (approved):
LGTM

PR #1940, opened 2024-09-19
Enhance `mithril-common` logging

Alenar:
## Content

This PR add a `src` tag with the component name on all components in `mithril-common` that have a logger. This change allow us to remove all prefixes that were previously used to tag a log source.

This have the added benefit to add the source component to logs that previously did not use prefixes.

IE a log that previously written like this:
```json
{"msg":"⟳ Preload Cardano Transactions - Started","v":0,"name":"slog-rs","level":30}
{"msg":"⟳ Preload Cardano Transactions - Finished","v":0,"name":"slog-rs","level":30}
{"msg":"⟳ Next Preload Cardano Transactions will start in 10 s","v":0,"name":"slog-rs","level":30}
{"msg":"ChainReaderBlockStreamer polls next","v":0,"name":"slog-rs","level":20,"pid":80105}
``` 
Have their prefixes supersede by a `src` tag like this:
```json
{"msg":"Started","v":0,"name":"slog-rs","level":30,"src":"CardanoTransactionsPreloader"}
{"msg":"Locking signed entity type","v":0,"name":"slog-rs","level":20,"src":"CardanoTransactionsPreloader","entity_type":"CardanoTransactions"}
{"msg":"Releasing signed entity type","v":0,"name":"slog-rs","level":20,"src":"CardanoTransactionsPreloader","entity_type":"CardanoTransactions"}
{"msg":"polls next","v":0,"name":"slog-rs","level":20,"src":"ChainReaderBlockStreamer"}
```

And logs that did not use prefix now have the `src` tag to identify the source component, ie:
```json
{"msg":"Verifying certificate","v":0,"name":"slog-rs","level":20,"src":"MithrilCertificateVerifier","certificate_signed_entity_type":"MithrilStakeDistribution(Epoch(11))","certificate_epoch":"Epoch(11)","certificate_previous_hash":"","certificate_hash":"632455d1d184b380acfb5ba71233a82b216266bc1dcebee312039ccd2a0d9ef5"}
```

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Comments

Also this PR remove the dev dependency to `slog_scope` from `mithril-common` has we now have test tools to construct Logger when a test need one.


mithril-common/src/cardano_block_scanner/chain_reader_block_streamer.rs
```diff
@@ -111,7 +111,7 @@ impl ChainReaderBlockStreamer {
             from,
             until,
             max_roll_forwards_per_poll,
-            logger,
+            logger: logger.new(slog::o!("src" => "ChainReaderBlockStreamer")),
```
sfauvel, line 114: To avoid to repeat the struct name (that is a common mistake), it's possible to use `std::any::type_name::<Self>()`.
The name is prefixed by the module name so, we have to remove it if we want only the name.
 We can add a `.split("::").last().unwrap()` for example even it's probably not the best solution

sfauvel (approved):
LGTM
Good job, that will help to filter logs

jpraynaud (approved):

dlachaume (approved):

mithril-common/src/cardano_block_scanner/chain_reader_block_streamer.rs
Alenar: Done using an extension trait :).

PR #1942, opened 2024-09-20
Refactor: signable builder services compute full protocol message

### LINKED ISSUE #1941: Refactor signable builder services to compute full protocol message in signer/aggregator
## Why
We have some duplicated code when the signature of the aggregate verification key is appended to the signed protocol message. Also this operation is done in the runners of the nodes whereas it should be the responsibility of the signable builder services in the signer and aggregator.

## What
Make the signable builder compute the full protocol message on its own in the signer and aggregator

## How

- [x] Refactor the signable builder service to compute the full protocol message on its own in the aggregator
- [x] Refactor the signable builder service to compute the full protocol message on its own in the signer

jpraynaud:
## Content

This PR includes a refactoring of the **signable builder service** so that it computes the next aggregate verification key part of the protocol message instead of delegating this responsibility to the state machines runners in signer and aggregator.

A new `SignableSeedBuilder` trait has been created which is responsible for computing part of the protocol message by the `SignableBuilder` service to enrich it.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Issue(s)
Closes #1941


mithril-signer/src/services/signable_seed_builder.rs
```diff
@@ -0,0 +1,187 @@
...
+        let mut protocol_message = ProtocolMessage::new();
+        let epoch_service = self.epoch_service.read().await;
+        let epoch = (*epoch_service).epoch_of_current_data()?;
+        let next_signer_retrieval_epoch = epoch.offset_to_next_signer_retrieval_epoch();
+        let next_protocol_initializer = self
+            .protocol_initializer_store
+            .get_protocol_initializer(next_signer_retrieval_epoch)
+            .await?
+            .ok_or_else(|| {
+                anyhow!("protocol_initializer at epoch {next_signer_retrieval_epoch}")
```
dlachaume, line 57:
```suggestion
                anyhow!("cannot get protocol_initializer at epoch {next_signer_retrieval_epoch}")
```

dlachaume (approved):
LGTM 👍

mithril-aggregator/src/services/signable_seed_builder.rs
```diff
@@ -0,0 +1,85 @@
...
+        Self { epoch_service }
+    }
+}
+
+#[async_trait]
+impl SignableSeedBuilder for SignableSeedBuilderService {
+    /// Compute seed protocol message
+    async fn compute_seed_protocol_message(&self) -> StdResult<ProtocolMessage> {
+        let epoch_service = self.epoch_service.read().await;
+        let mut protocol_message = ProtocolMessage::new();
```
sfauvel, line 36: I wonder if this service should only return a ProtocolMessagePartValue and let the building of ProtocolMessage to the SignableBuilderService.
This would avoid to return a ProtocolMessage that is not fully completed.

sfauvel (approved):
LGTM
I have some questions about responsabilities of the services

mithril-aggregator/src/services/signable_seed_builder.rs
Alenar, line 1: It's always difficult to draw the line between what's a service or not.

In my opinion this is not a service but a subpart of the signable builder service ... which is in common.

Maybe we could add a `signable_builder` subfolder to both aggregator & signer `servicse` module and move this file here and rename it to remove the 'service' in its name ? This way it would be tied to the signable builder.

Btw I was thinking of doing the same for other components used by a specific service, like the aggregator multi signer that could be moved to the `services/certifier` module.

mithril-signer/src/store/protocol_initializer_store.rs
```diff
@@ -4,8 +4,12 @@ use tokio::sync::RwLock;
 use mithril_common::{crypto_helper::ProtocolInitializer, entities::Epoch, StdResult};
 use mithril_persistence::store::{adapter::StoreAdapter, StorePruner};
 
+#[cfg(test)]
+use mockall::automock;
+
 type Adapter = Box<dyn StoreAdapter<Key = Epoch, Record = ProtocolInitializer>>;
 
+#[cfg_attr(test, automock)]
```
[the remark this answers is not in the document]
jpraynaud: It's better like this indeed

mithril-signer/src/services/signable_seed_builder.rs
jpraynaud: Done

mithril-signer/src/services/signable_seed_builder.rs
```diff
@@ -0,0 +1,187 @@
...
+            ) -> StdResult<&'static Option<CardanoTransactionsSigningConfig>>;
+
+            /// Check if a signer is included in the current stake distribution
+            fn is_signer_included_in_current_stake_distribution(
+                &self,
+                party_id: PartyId,
+                protocol_initializer: ProtocolInitializer,
+            ) -> StdResult<bool>;
+        }
+    }
```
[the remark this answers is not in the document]
jpraynaud: Yes, it was a bit of a pain to do. I moved the implementation back to the signer epoch service module :+1:

mithril-common/src/signable_builder/interface.rs
```diff
@@ -16,13 +16,25 @@ pub trait Artifact: Debug + Send + Sync {
...
+        seed_protocol_message: ProtocolMessage,
+    ) -> StdResult<ProtocolMessage>;
+}
+
+/// SignableSeedBuilder is a trait for building a seed protocol message
+#[cfg_attr(test, automock)]
+#[async_trait]
+pub trait SignableSeedBuilder: Send + Sync {
+    /// Compute seed protocol message
+    async fn compute_seed_protocol_message(&self) -> StdResult<ProtocolMessage>;
```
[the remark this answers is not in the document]
jpraynaud: You're right, this makes more sense. I have opted for the second option which will avoid duplicating code when signing the protocol parameters depending on the era :+1:

mithril-aggregator/src/services/signable_seed_builder.rs
jpraynaud: Done, see https://github.com/input-output-hk/mithril/pull/1942#discussion_r1773415634 :+1:

mithril-common/src/signable_builder/interface.rs
```diff
@@ -16,13 +16,25 @@ pub trait Artifact: Debug + Send + Sync {
...
 pub trait SignableBuilder<U>: Send + Sync
 where
     U: Beacon,
 {
     /// Compute a protocol message
-    async fn compute_protocol_message(&self, beacon: U) -> StdResult<ProtocolMessage>;
+    async fn compute_protocol_message(
+        &self,
+        beacon: U,
+        seed_protocol_message: ProtocolMessage,
```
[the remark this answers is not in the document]
jpraynaud: Done

mithril-aggregator/src/services/signable_seed_builder.rs
jpraynaud: Good idea, done :+1:

Alenar (approved):
LGTM :fire:

mithril-common/src/signable_builder/signable_builder_service.rs
```diff
@@ -105,38 +139,69 @@ mod tests {
...
     }
 
+    mock! {
+        SignableSeedBuilderImpl { }
+
+        #[async_trait]
+        impl SignableSeedBuilder for SignableSeedBuilderImpl {
+            async fn compute_next_aggregate_verification_key_protocol_message_part_value(&self) -> StdResult<ProtocolMessagePartValue>;
+        }
+    }
```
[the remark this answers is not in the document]
jpraynaud: It's not needed actually, the automatic mock is enough: I removed this implementation :+1:

Issue #1925, opened 2024-09-09
Signer computes what to sign on its own


jpraynaud:
## Why
The signer needs to compute the next message to sign independently and without relying on the pending certificate of the aggregator.

## What
Replace the usage of the Pending Certificate in the signer with its own computation of which is the next message to sign.

## How

- [x] Adapt the mechanism of the aggregator to compute the next signed entity type to sign
- [x] Store the signed entity type signed persistently when it has been successfully (when signature is sent)
- [x] Prune or keep last 2 epochs in the store?
- [x] Remove all pending certificate code in the signer
- [x] Deprecate the `/pending-certificate` route in OpenAPI

Issue #1923, opened 2024-09-09
Retrieve custom signing configurations with epoch settings in signer


jpraynaud:
## Why
The signer need to get access to the Cardano transactions signing configuration which are currently a configuration of the aggregator.

## What 
Provide the custom signing configurations in the `/epoch-settings` for the current and next epochs.
In this first phase, we want to use the configuration parameters of the aggregator as the source of the signing configurations (which can change in the middle of an epoch).

## How

- [x] Add the `current_cardano_transactions_signing_config` field in the epoch settings
- [x] Add the `next_cardano_transactions_signing_config` field in the epoch settings
- [x] Store these configurations in the epoch service of the signer when there is an epoch transition
- [x] Update Open API specifications

## Later
- Use these configurations to allow the signer to sign when he has transitioned to a new epoch and aggregator has not yet transitioned (if a buffered signer registration is implemented in the aggregator)

PR #1970, opened 2024-10-01
Signer self compute signed entity

Alenar:
## Content

This PR makes the `mithril-signer` compute what to sign on its own instead of relying on the aggregator `CertificatePending`.

To make this work the signer:
1.  still needs some data from the aggregator to create a `SignedEntityConfig`:
    - The list of allowed signed entity type discriminant, that we can get from the aggregator `/` route.
    - The cardano transaction signing configuration, that's exposed in the aggregator `/epoch-settings` route since #1923
1. use the `SignedEntityConfig` to list all possible `SignedEntityType` for its current time point.
1. filter that list of `SignedEntityType` to remove: locked signed entities, previously signed entities.

Both of the data needed from the aggregator are stored in the signer `EpochService`, meaning that they're fetched once per epoch.
This also mean that if an aggregator is restarted with change to its configured allowed discriminant the signers will take in account this change either at the next epoch or after a restart (since that configuration change is immediately advertised by the aggregator and not constant for an epoch).

Filtering out the previously signed entities mean that we need to persist them, to do so this PR add a new table in the signer database.
Since we don't need to keep those data in the long term (they are mostly useful for the current epoch), stale data of past epoch can be pruned based on the already existing `store_retention_limit` parameter.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation

## Issue(s)

Relates to #1925

### DIFF
--- a/mithril-signer/src/database/record/signed_beacon_record.rs
+++ b/mithril-signer/src/database/record/signed_beacon_record.rs
@@ -5,7 +5,7 @@ use mithril_common::entities::{Epoch, SignedEntityType};
 use mithril_persistence::database::Hydrator;
 use mithril_persistence::sqlite::{HydrationError, Projection, SqLiteEntity};
 
-use crate::services::BeaconToSign;
+use crate::entities::BeaconToSign;
 
 /// Database record of a beacon signed by the signer
 #[derive(Debug, Clone, PartialEq)]
--- a/mithril-signer/src/database/record/signed_beacon_record.rs
+++ b/mithril-signer/src/database/record/signed_beacon_record.rs
@@ -1,0 +1,154 @@
+use chrono::{DateTime, Utc};
+use sqlite::Row;
+
+use mithril_common::entities::{Epoch, SignedEntityType};
+use mithril_persistence::database::Hydrator;
+use mithril_persistence::sqlite::{HydrationError, Projection, SqLiteEntity};
+
+use crate::services::BeaconToSign;
+
+/// Database record of a beacon signed by the signer
+#[derive(Debug, Clone, PartialEq)]
+pub struct SignedBeaconRecord {
+    /// The epoch when the beacon was issued
+    pub epoch: Epoch,
+
+    /// The signed entity type to sign
+    pub signed_entity_type: SignedEntityType,
+
+    /// Datetime when the beacon was initiated
+    pub initiated_at: DateTime<Utc>,
+
+    /// Datetime when the beacon was signed
+    pub signed_at: DateTime<Utc>,
+}
+
+impl From<BeaconToSign> for SignedBeaconRecord {
+    fn from(beacon: BeaconToSign) -> Self {
+        Self {
+            epoch: beacon.epoch,
+            signed_entity_type: beacon.signed_entity_type,
+            initiated_at: beacon.initiated_at,
+            signed_at: Utc::now(),
+        }
+    }
+}
+
+#[cfg(test)]
+impl SignedBeaconRecord {
+    /// Create a fake `SignedBeaconRecord` for testing purposes
+    ///
+    /// The dates fields are set to constant values to make it easier to compare.
+    pub(crate) fn fake(epoch: Epoch, signed_entity_type: SignedEntityType) -> Self {
+        let initiated_at = DateTime::<Utc>::default();
+        Self {
+            epoch,
+            signed_entity_type,
+            initiated_at,
+            signed_at: initiated_at + chrono::TimeDelta::minutes(3),
+        }
+    }
+
+    pub(crate) fn fakes(records: &[(Epoch, Vec<SignedEntityType>)]) -> Vec<Self> {
+        records
+            .iter()
+            .flat_map(|(epoch, signed_entity_types)| {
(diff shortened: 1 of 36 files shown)


mithril-signer/src/database/repository/signed_beacon_repository.rs
```diff
@@ -0,0 +1,312 @@
...
+
+        let to_filter = all_signed_entity_type_for(&time_point);
+        let available_entities = repository
+            .filter_out_already_signed_entities(to_filter.clone())
+            .await
+            .unwrap();
+
+        let expected: Vec<SignedEntityType> = to_filter
+            .iter()
+            .filter(|entity| !signed_beacons.contains(entity))
```
sfauvel, line 244: This code is the same as the one in the code of filter_out_already_signed_entities.
Is this test is really relevant if we use the same code in both sides ?

mithril-signer/src/database/record/signed_beacon_record.rs
```diff
@@ -0,0 +1,154 @@
...
+        };
+        let signed_beacon = SignedBeaconRecord {
+            epoch: Epoch(3),
+            signed_entity_type: SignedEntityType::MithrilStakeDistribution(Epoch(4)),
+            initiated_at,
+            signed_at: initiated_at + chrono::TimeDelta::minutes(2),
+        };
+
+        assert_eq!(signed_beacon, beacon_to_sign);
+        assert_eq!(beacon_to_sign, signed_beacon);
```
sfauvel, line 152: We don't check that `eq` can return false

mithril-signer/src/services/certifier.rs
```diff
@@ -0,0 +1,374 @@
...
+                .unwrap();
+            let next_beacon_to_sign = certifier_service.get_beacon_to_sign().await.unwrap();
+
+            if let Some(beacon) = next_beacon_to_sign {
+                assert!(
+                    !all_signed_beacons.contains(&beacon),
+                    "Beacon should not repeat"
+                );
+                assert!(
+                    SignedEntityTypeDiscriminants::from(
```
sfauvel, line 267: We can try to split this test in two to see if it will be simplest to understand
One to check only the order and another one for the rest.

sfauvel (approved):
Good work !
We have to check if doc need to be updated

dlachaume (approved):

mithril-signer/src/services/upkeep_service.rs
```diff
@@ -42,16 +56,40 @@ impl SignerUpkeepService {
...
             main_db_connection,
             cardano_tx_connection_pool,
             signed_entity_type_lock,
+            pruning_tasks,
+            store_retention_limit,
             logger,
         }
     }
 
+    async fn execute_pruning_tasks(&self, current_epoch: Epoch) -> StdResult<()> {
```
jpraynaud, line 73: I'm not sure that we need to support multiple pruning tasks here. 
Also this design does not seem very adapted if the epoch thresholds are different from one task to another.

jpraynaud (approved):
LGTM :+1: 

I left few suggestions and comments below.

mithril-test-lab/mithril-end-to-end/src/mithril/signer.rs
```diff
@@ -55,6 +55,7 @@ impl Signer {
                 signer_config.pool_node.db_path.to_str().unwrap(),
             ),
             ("DATA_STORES_DIRECTORY", &data_stores_path),
+            ("STORE_RETENTION_LIMIT", "10"),
```
jpraynaud, line 58: Maybe we should use a bigger retention to facilitate debug investigations? 
```suggestion
            ("STORE_RETENTION_LIMIT", "50"),
```

mithril-signer/src/services/upkeep_service.rs
Alenar: Good point for the threshold, I will move the computation to the tasks themself and only pass the current epoch.

For the multiple tasks I was thinking we could move pruning on our current tasks to that kind of design, this would avoid doing it every-time we save data.

mithril-test-lab/mithril-end-to-end/src/mithril/signer.rs
Alenar: The goal is to find a balance between doing no pruning at all in the e2e and not doing them too much. 10 seems to strike that balance since e2e usually start producing data about epoch 6~7 and finish around epoch 20~22.

mithril-signer/src/database/repository/signed_beacon_repository.rs
Alenar: You're right, i changed the test to use hard coded lists.

mithril-test-lab/mithril-end-to-end/src/mithril/signer.rs
jpraynaud: OK, maybe `15` would be a better value in that case. 
But as we can change this value easily if needed, let's keep `10` for now :+1:

PR #1972, opened 2024-10-01
Deprecate certificate pending

### LINKED ISSUE #1925: Signer computes what to sign on its own

Alenar:
## Content

This PR deprecate the `CertificatePendingMessage` both in rust code and in the openapi. It also remove the remaining certificate pending code in `mithril-signer`.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Comments

This is a follow up on #1970, so it must be merged after it.

## Issue(s)

Closes #1925


mithril-common/src/messages/certificate_pending.rs
```diff
@@ -1,10 +1,15 @@
...
 use serde::{Deserialize, Serialize};
 
 use crate::entities::{CardanoDbBeacon, Epoch, ProtocolParameters, SignedEntityType};
 use crate::messages::SignerMessagePart;
 
 /// Structure to transport [crate::entities::CertificatePending] data.
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+#[deprecated(
+    since = "0.4.61",
+    note = "Exist only for backward-compatibility, will be removed in the future"
```
jpraynaud, line 11:
```suggestion
    note = "Exists only for backward-compatibility, will be removed in the future"
```

jpraynaud (approved):
LGTM

mithril-signer/src/services/aggregator_client.rs
```diff
@@ -214,36 +205,6 @@ impl AggregatorClient for AggregatorHTTPClient {
         }
     }
```
sfauvel, line 207: The /certificate_pending still exist in openapi.yaml even if it was deprecated. 
Is it normal to remove all tests ? What happen when the route is called ?

sfauvel (approved):
LGTM

dlachaume (approved):

mithril-common/src/test_utils/apispec.rs
```diff
@@ -433,12 +443,12 @@ components:
     fn test_validate_ok_when_request_without_body_and_expects_response() {
         APISpec::from_file(&APISpec::get_default_spec_file())
             .method(Method::GET.as_str())
-            .path("/certificate-pending")
+            .path("/")
```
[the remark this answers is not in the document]
Alenar: Agree with that :+1: , as you say we should do a refactoring PR to solve that problem.

PR #1979, opened 2024-10-07
Signer certifier service sign & send signatures

### LINKED ISSUE #1945: Refactor Certifier service to sign and send signatures in signer
## Why
Following the implementation of #1925, we want to give to the Certifier service of the signer the responsibility to sign and send the messages.

## What
Refactor the Certifier service with its new responsibilities: sign and send individual signatures.

## How

- [x] Implement sign individual signatures in Certifier service
- [x] Implement send individual signatures in Certifier service

Alenar:
## Content

This PR refactor the `mithril-signer` in order to move the sign & send signatures responsibility from the state machine runner to the signer `Certifier` service.

One of the main concern for this refactor was to minimize the number of new dependencies that needed to be add to the certifier.
Simply moving the code from the runner to the certifier would have result in adding the following dependencies:
- `epoch_service` and `protocol_initializer_store` and `single_signer`: to issue the single signature.
- `certificate_handler`: to send the signatures but we would only need one method of it's whole api.

### Avoiding `protocol_initializer_store` dependency

The protocol initializer of the current epoch is now stored on the epoch service.
It's stored as an option since it's only available if a signer is registered in the Mithril stake distribution of the current epoch.
This is used in the next refactor.

### Refocus single signer service

The `single_signer` service should be able to issue single signatures without needing parameters that must be computed with other dependencies.
To do so the `compute_single_signatures` now only take a protocol message as parameter, the `protocol_initializer` and the list of `signers_with_stake` are fetched from the epoch service.

The `compute_aggregate_verification_key` was only used by the `SignerSignableSeedBuilder` and need not the current protocol_initializer but the next one, a data not available in the epoch service.
I choose to remove this method from the single signer api and instead allow the `SignerSignableSeedBuilder`to do the computation by using the lower level tools since it already had the signers & protocol initializer needed.

### Dependency inversion for signature publishing

By adding a `SignaturePublisher` trait defined for the certifier needs. It's automatically implement it for every `AggregatorClient` trait implemetors.

### Additional refactor

Since the epoch service now hold all data needed for the `can_signer_sign_current_epoch` function, it's code is moved from the signer state machine runner directly to the epoch service.

## Pre-submit checklist

- Branch
- PR
  - [ ] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation

## Issue(s)
Closes #1945

### DIFF
--- a/mithril-signer/src/services/epoch_service.rs
+++ b/mithril-signer/src/services/epoch_service.rs
@@ -16,6 +16,7 @@ use mithril_persistence::store::StakeStorer;
 use crate::dependency_injection::EpochServiceWrapper;
 use crate::entities::SignerEpochSettings;
 use crate::services::SignedEntityConfigProvider;
+use crate::store::ProtocolInitializerStorer;
 use crate::RunnerError;
 
 /// Errors dedicated to the EpochService.
@@ -31,7 +32,7 @@ pub enum EpochServiceError {
 pub trait EpochService: Sync + Send {
     /// Inform the service a new epoch has been detected, telling it to update its
     /// internal state for the new epoch.
-    fn inform_epoch_settings(
+    async fn inform_epoch_settings(
         &mut self,
         epoch_settings: SignerEpochSettings,
         allowed_discriminants: BTreeSet<SignedEntityTypeDiscriminants>,
@@ -42,6 +43,11 @@ pub trait EpochService: Sync + Send {
 
     /// Get next protocol parameters used in next epoch (associated with the actual epoch)
     fn next_protocol_parameters(&self) -> StdResult<&ProtocolParameters>;
+
+    /// Get the protocol initializer for the current epoch if any
+    ///
+    /// `None` if the signer can't sign for the current epoch.
+    fn protocol_initializer(&self) -> StdResult<&Option<ProtocolInitializer>>;
 
     /// Get signers for the current epoch
     fn current_signers(&self) -> StdResult<&Vec<Signer>>;
@@ -79,6 +85,7 @@ pub trait EpochService: Sync + Send {
 struct EpochData {
     epoch: Epoch,
     next_protocol_parameters: ProtocolParameters,
+    protocol_initializer: Option<ProtocolInitializer>,
     current_signers: Vec<Signer>,
     next_signers: Vec<Signer>,
     allowed_discriminants: BTreeSet<SignedEntityTypeDiscriminants>,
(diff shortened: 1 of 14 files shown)


mithril-signer/src/services/epoch_service.rs
```diff
@@ -419,11 +509,106 @@ mod tests {
...
+            });
+
+        let can_sign_current_epoch = epoch_service
+            .can_signer_sign_current_epoch(fixtures.signers()[0].party_id.clone())
+            .unwrap();
+        assert!(!can_sign_current_epoch);
+    }
+
+    #[test]
+    fn signer_cant_sign_if_it_dont_use_stored_protocol_initializer() {
```
jpraynaud, line 541: This test is not very clear: are we trying to test that a protocol initializer created by a different pool can not sign?

Also small typo:
```suggestion
    fn signer_cant_sign_if_it_doesnt_use_stored_protocol_initializer() {
```

jpraynaud (approved):
LGTM :+1:

mithril-signer/src/services/epoch_service.rs
```diff
@@ -308,6 +308,51 @@ impl SignedEntityConfigProvider for SignerSignedEntityConfigProvider {
     }
 }
 
+#[cfg(test)]
+impl MithrilEpochService {
+    /// `TEST ONLY` - Create a new instance of the service using dumb dependencies.
```
dlachaume, line 313: I think you can remove all the `TEST ONLY` comments since the functions are implemented with the `#[cfg(test)]` attribute.

mithril-signer/src/services/epoch_service.rs
```diff
@@ -419,11 +509,106 @@ mod tests {
...
             .is_signer_included_in_current_stake_distribution(
                 party_id,
-                protocol_initializer_not_included
+                &protocol_initializer_not_included
             )
             .unwrap());
     }
 
+    #[test]
+    fn signer_cant_sign_if_no_protocol_initializer_are_stored() {
```
sfauvel, line 518: Those 4 tests about protocol initializer may be into a sub-module.
We also may have `test_protocol_initializer_is_available_after_register_epoch_settings_call_if_in_store()` in this module

dlachaume (approved):
LGTM 👍

mithril-signer/src/services/epoch_service.rs
```diff
@@ -419,11 +509,106 @@ mod tests {
...
+            });
+
+        let can_sign_current_epoch = epoch_service
+            .can_signer_sign_current_epoch(fixtures.signers()[0].party_id.clone())
+            .unwrap();
+        assert!(!can_sign_current_epoch);
+    }
+
+    #[test]
+    fn signer_can_sign_if_in_current_signers_and_use_the_stored_protocol_initializer() {
```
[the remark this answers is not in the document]
Alenar: I used `MithrilEpochService::new_with_dumb_dependencies` test method which does exactly that.

mithril-signer/src/services/epoch_service.rs
Alenar: I grouped the four tests together under a 'can_signer_sign_current_epoch' umbrella, but since this name doesn't fit the last test that you suggested to move I kept it where it was.

mithril-signer/src/services/epoch_service.rs
Alenar: I renamed the test to `cant_sign_if_stored_protocol_initializer_belong_to_another_party`.

What this test aim to check is the `&& s.verification_key == protocol_initializer.verification_key().into()` part that's in the `is_signer_included_in_current_stake_distribution` private method.

Issue #1981, opened 2024-10-07
Enhance readability of node logs


jpraynaud:
## Why
We need to enhance the readability of node logs by adding a key in the JSON log output with the name of the caller (instead of having the name in the log message).

## What
Enhance the readability of logs in the signer, aggregator, client(?) and relay.

## How

- [ ] Remove global loggers: 
   - [ ] signer
   - [ ] aggregator
   - [ ] client
   - [ ] relay
 - [ ] Add a `src` key for the caller when logging:
   - [ ] signer
   - [ ] aggregator
   - [ ] client
   - [ ] relay
 - [ ] Update the name of the node when logging: (https://docs.rs/slog-bunyan/2.5.0/slog_bunyan/fn.with_name.html)
   - [ ] signer
   - [ ] aggregator
   - [ ] client
   - [ ] relay

PR #1992, opened 2024-10-09
Enhance logs in mithril-signer and mithril-persistence

Alenar:
## Content

This PR refactor the log usage in `mithril-signer` and `mithril-persistence`:
- A dedicated child logger is used for each service
- Each child logger is tagged with its service name in order to identify quickly a log source without the need of a prefix in the log message.
- Remove `slog_scope` to enforce the usage of a name tagged child logger (only kept in the integration tests).
- Change the name property in all logs from the default `slog-rs` to `mithril-signer`.
- Move away from the log message into a property huge structure (ie: some logs joined errors into their message, those errors have been moved to a "error" property).

Also this PR enhance errors issued by the signer `AggregatorClient` when there's an http error:
- the status code canonical name (ie: "bad request" for a 400) is prepend to the error message. This add some context when there's no body in the http response (previously there were only the back trace). 
- if a json body is available in the http response, it tries to deserialize it as either a `ClientError` or `ServerError` (returned by the aggregator with 4xx and 5xx errors respectively). There's fallback in case of unknown format.

Also this PR fix an issue when registering a signature: if the aggregator returned a 202 (Accepted) status code it was handled as an error, making the state machine fail its cycle.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Issue(s)

Relates to #1981


mithril-signer/src/runtime/error.rs
```diff
@@ -40,39 +41,37 @@ impl RuntimeError {
...
+                Some(err) => crit!(logger, "{self}"; "nested_error" => ?err),
+            },
+        }
+    }
 }
 
 impl std::error::Error for RuntimeError {}
 
 impl Display for RuntimeError {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
```
sfauvel, line 63: We may rename the variable 'f'

sfauvel (approved):
LGTM

dlachaume (approved):
LGTM 🚀

jpraynaud (approved):
LGTM :+1:

PR #1998, opened 2024-10-11
Sfauvel/1990/refactor prometheus metrics using struct

sfauvel:
## Content

Create a new crate `internal/mithril-metrics` that provide tooling to simplify the creation of a metric service. 

This PR includes the creation of a new crate that provide a MetricServer and a macro to simply declare a new metric service.

## Pre-submit checklist

- Branch
- PR
  - [ ] Self-reviewed the diff
  - [ ] Reviewer requested
- Documentation

## Issue(s)


internal/mithril-metric/src/commons.rs
```diff
@@ -0,0 +1,494 @@
...
+    pub fn new(logger: Logger, name: &str, help: &str) -> StdResult<Self> {
+        let counter = MetricCounter::create_metric_counter(name, help)?;
+        Ok(Self {
+            logger,
+            name: name.to_string(),
+            counter: Box::new(counter),
+        })
+    }
+
+    pub fn record(&self) {
```
jpraynaud, line 148:
```suggestion
    pub fn increment(&self) {
```

internal/mithril-metric/src/commons.rs
```diff
@@ -0,0 +1,494 @@
...
+    pub fn new(logger: Logger, name: &str, help: &str) -> StdResult<Self> {
+        let gauge = MetricGauge::create_metric_gauge(name, help)?;
+        Ok(Self {
+            logger,
+            name: name.to_string(),
+            gauge: Box::new(gauge),
+        })
+    }
+
+    pub fn record(&self, epoch: Epoch) {
```
jpraynaud, line 191:
```suggestion
    pub fn record::<T: Into<f64>>(&self, value: T) {
```

internal/mithril-metric/src/commons.rs
jpraynaud, line 1: Maybe `common` is not an explicit enough name for this module.
Also this module should probably be split in multiple (sub)modules.

internal/mithril-metric/src/commons.rs
```diff
@@ -0,0 +1,494 @@
...
+///     ),
+///     gauge_example: MetricGauge(
+///         "custom_gauge_example_name",
+///         "Example of a gauge metric"
+///     )
+/// );
+///
+/// build_metrics_service!(
+///     MetricsService,
+///     counter_example: MetricCounter("Example of a counter metric"),
```
jpraynaud, line 37: IMO we should always create a `Metric` with a given metric name to avoid side effects if e.g. we rename the field in the service.

internal/mithril-metric/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+license = { workspace = true }
+repository = { workspace = true }
+
+[lib]
+crate-type = ["lib", "cdylib", "staticlib"]
+
+[dependencies]
+anyhow = "1.0.86"
+axum = "0.7.5"
+mithril-common = { path = "../../mithril-common", features = ["full"] }
```
dlachaume, line 17: Do we need to enable the feature full?

sfauvel: I also remove it from `tokio`

internal/mithril-metric/src/commons.rs
sfauvel: With this generic, when we want to record an epoch we need to dereference to u64 and cast the value
`my_gauge.record(*value as f64);`

We probably also want to return a f64 from `get()` function. In that case, the caller need to cast the result and then build an Epoch: `Epoch(my_gauge.get() as u64)`

Is it ok for you ?

internal/mithril-metric/Makefile
```diff
@@ -0,0 +1,19 @@
...
+test:
+	${CARGO} test
+
+check:
+	${CARGO} check --release --all-features --all-targets
+	${CARGO} clippy --release --all-features --all-targets
+	${CARGO} fmt --check
+
+doc:
+	${CARGO} doc --no-deps --open --features full
```
Alenar, line 19: There's no feature on this new crate :).
```suggestion
	${CARGO} doc --no-deps --open
```

mithril-signer/src/metrics/service.rs
```diff
@@ -1,278 +1,147 @@
...
+        "Number of successful runtime cycles since startup on a Mithril signer node"
+    ),
+    runtime_cycle_total_since_startup_counter:MetricCounter(
+        "mithril_signer_runtime_cycle_total_since_startup",
+        "Number of runtime cycles since startup on a Mithril signer node"
+    )
+
+);
+
+// TODO these functions are kept while changes are made for all uses
```
Alenar, line 47: Is this TODO still relevant ?

mithril-signer/Cargo.toml
```diff
@@ -23,9 +23,11 @@ config = "0.14.0"
 hex = "0.4.3"
 mithril-common = { path = "../mithril-common", features = ["full"] }
 mithril-doc = { path = "../internal/mithril-doc" }
+mithril-metric = { path = "../internal/mithril-metric" }
 mithril-persistence = { path = "../internal/mithril-persistence" }
 openssl = { version = "0.10.66", features = ["vendored"], optional = true }
 openssl-probe = { version = "0.1.5", optional = true }
+paste = "1.0.15"
```
Alenar, line 30: I think this dependency can be removed from `mithril-signer` since it doesn't use it.

internal/mithril-metric/src/helper.rs
```diff
@@ -0,0 +1,233 @@
...
+/// let service = MetricsService::new(TestLogger::stdout()).unwrap();
+/// service.get_counter_example().record();
+/// service.get_gauge_example().record(Epoch(12));
+#[macro_export]
+macro_rules! build_metrics_service {
+    ($service:ident, $($metric_attribute:ident:$metric_type:ident($name:literal, $help:literal)),*) => {
+        paste::item! {
+            /// Metrics service which is responsible for recording and exposing metrics.
+            pub struct $service {
+                registry: Registry,
```
Alenar, line 31: You should use fully qualified path inside macro so their user don't need to import the dependencies used by the macro.
```suggestion
                registry: prometheus::Registry,
```

internal/mithril-metric/src/helper.rs
```diff
@@ -0,0 +1,233 @@
...
+                $(
+                    $metric_attribute: $metric_type,
+                )*
+            }
+
+            impl $service {
+                /// Create a new MetricsService instance.
+                pub fn new(logger: Logger) -> StdResult<Self> {
+
+                    let registry = Registry::new();
```
Alenar, line 41:
```suggestion
                    let registry = prometheus::Registry::new();
```

Alenar (approved):
LGTM, great works !

Usage looks really nice in the signer.

dlachaume (approved):
LGTM 👍

internal/mithril-metric/src/lib.rs
```diff
@@ -0,0 +1,13 @@
+//! metrics module.
```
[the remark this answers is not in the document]
jpraynaud: Yes, this should be added in this crate as well :+1:

internal/mithril-metric/src/metric.rs
```diff
@@ -0,0 +1,139 @@
...
+    pub fn new(logger: Logger, name: &str, help: &str) -> StdResult<Self> {
+        let gauge = MetricGauge::create_metric_gauge(name, help)?;
+        Ok(Self {
+            logger,
+            name: name.to_string(),
+            gauge: Box::new(gauge),
+        })
+    }
+
+    pub fn record<T: Into<f64> + Copy>(&self, value: T) {
```
jpraynaud, line 82: I'm not sure that `Copy` is mandatory for this generic type but it can be fixed another time.

internal/mithril-metric/src/metric.rs
```diff
@@ -0,0 +1,139 @@
...
+            self.logger,
+            "Set '{}' gauge value to {}",
+            self.name,
+            value.into()
+        );
+        self.gauge.set(value.into());
+    }
+
+    pub fn get(&self) -> f64 {
+        self.gauge.get().round()
```
jpraynaud, line 93: What is the real need to round the value here?

mithril-signer/src/metrics/service.rs
```diff
@@ -1,278 +1,147 @@
...
+    ),
+    signature_registration_total_since_startup_counter:MetricCounter(
+        "mithril_signer_signature_registration_total_since_startup",
+        "Number of signature registrations since startup on a Mithril signer node"
+    ),
+    signature_registration_success_last_epoch_gauge:MetricGauge(
+        "mithril_signer_signature_registration_success_last_epoch",
+        "Latest epoch at which signature successfully registered on a Mithril signer node"
+    ),
+    // Runtime cycle metrics
```
jpraynaud, line 35: This comment could be removed.

jpraynaud (approved):
LGTM :+1:

internal/mithril-metric/src/commons.rs
sfauvel: Removing this possibility, I also simplify build_metrics_service macro.

mithril-signer/Cargo.toml
sfauvel: It need it to use macro.

internal/mithril-metric/src/server.rs
```diff
@@ -13,7 +13,9 @@ use tokio::sync::oneshot::Receiver;
 use mithril_common::logging::LoggerExtensions;
 use mithril_common::StdResult;
 
-use crate::MetricsService;
+pub trait MetricsServiceExporter {
```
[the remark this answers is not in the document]
sfauvel: Done

internal/mithril-metric/src/metric.rs
sfauvel: Done

mithril-signer/src/metrics/service.rs
sfauvel: Done

internal/mithril-metric/src/commons.rs
sfauvel: Renamed to `helper`

mithril-signer/Cargo.toml
Alenar: Even if you use fully qualified path for paste inside the `build_metrics_service` ?

internal/mithril-metric/src/lib.rs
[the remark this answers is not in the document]
sfauvel: Done

mithril-signer/src/metrics/service.rs
sfauvel: I removed all obsolete function and associated tests

PR #1999, opened 2024-10-11
Feat: implement certificate chain builder for tests

jpraynaud:
## Content

This PR includes the implementation of a **Certificate Chain Builder for tests**:
- This part of the code is now more readable and maintainable
- The certificate chains allows alteration of the genesis and standard certificates with closures
- The computation of the number of signers per epoch is now configurable with a closure

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

### DIFF
--- a/mithril-common/src/test_utils/certificate_chain_builder.rs
+++ b/mithril-common/src/test_utils/certificate_chain_builder.rs
@@ -19,6 +19,9 @@ pub type GenesisCertificateProcessorFunc =
 pub type StandardCertificateProcessorFunc =
     dyn Fn(Certificate, &CertificateChainBuilderContext) -> Certificate;
 
+/// Total signers per epoch processor function type. For tests only.
+pub type TotalSignersPerEpochProcessorFunc = dyn Fn(Epoch) -> usize;
+
 /// Context used while building a certificate chain. For tests only.
 pub struct CertificateChainBuilderContext<'a> {
     pub index_certificate: usize,
@@ -31,7 +34,7 @@ pub struct CertificateChainBuilderContext<'a> {
 
 /// A builder for creating a certificate chain. For tests only.
 ///
-/// # Example usage for building a fully valid certificate chain
+/// # Simple example usage for building a fully valid certificate chain
 ///
 /// ```
 ///     use mithril_common::crypto_helper::ProtocolParameters;
@@ -45,10 +48,10 @@ pub struct CertificateChainBuilderContext<'a> {
 ///     assert_eq!(5, certificate_chain.len());
 /// ```
 ///
-/// # Example usage for building an adversarial certificate chain
+/// # More complex example usage for building a fully valid certificate chain
 ///
 /// ```
-///     use mithril_common::entities::Epoch;
+///     use std::cmp::min;
 ///     use mithril_common::crypto_helper::ProtocolParameters;
 ///     use mithril_common::test_utils::CertificateChainBuilder;
 ///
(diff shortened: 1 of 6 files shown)


mithril-common/src/test_utils/certificate_chain_builder.rs
```diff
@@ -0,0 +1,752 @@
...
+            "Certificates per epoch must be greater than 0"
+        );
+        assert!(
+            total_certificates >= certificates_per_epoch,
+            "Total certificates must be greater or equal to certificates per epoch"
+        );
+        // The total number of epochs in the sequence is the total number of certificates divided by
+        // the number of certificates per epoch plus two (one for the genesis epoch and one to compute the next fixtures of the last epoch)
+        let total_epochs_in_sequence =
+            (total_certificates - 1).div_ceil(certificates_per_epoch) + 2;
```
dlachaume, line 275: I understand the `+2` part, but why `(total_certificates - 1)`?

dlachaume (approved):
LGTM 🚀, thanks for the tool, it will make testing cases on the certificate chain much easier 🙏.

Just a comment regarding the exposure of data in the `certificate_chain_builder` module.
It seems that we could set the following traits to private visibility:
- `GenesisCertificateProcessorFunc`
- StandardCertificateProcessorFunc
- TotalSignersPerEpochProcessorFunc

The `CertificateChainBuilderContext` struct could also switch to private by adding a `#[cfg(test)]` attribute on the `with_genesis_certificate_processor` and `with_standard_certificate_processor` functions.

Do you think it could be a good idea?

mithril-common/src/test_utils/certificate_chain_builder.rs
```diff
@@ -0,0 +1,752 @@
...
+            multi_signature.into(),
+        );
+
+        certificate
+    }
+
+    // Returns the chained certificates in reverse order
+    // The latest certificate of the chain is the first in the vector
+    fn compute_chained_certificates(&self, certificates: Vec<Certificate>) -> Vec<Certificate> {
+        let mut certificates_chained: Vec<Certificate> = Vec::new();
```
sfauvel, line 402: Handle the first certificate add complexity to understand what is done.
It may be easier to understand if we create a function to update hash of the ̀Certificate`.
In the method, we compute twice the hash a certificate. We probably can reuse the hash of the previous_certificate instead of recompute it.

I propose you something like (I would like to avoid cloning certificate but I'm not sure it's possible):
```
   fn update_previous_hash(certificate: &Certificate, previous_hash: String) -> Certificate {
        let mut certificate_new = certificate.clone();
        certificate_new.previous_hash = previous_hash.to_string();
        certificate_new.hash = certificate_new.compute_hash();
        certificate_new
    }

    let mut certificates_chained: Vec<Certificate> = Vec::new();
    certificates
        .iter()
        .fold("".to_string(), |previous_hash, certificate| {
            let certificate_new = update_previous_hash(certificate, previous_hash);
            let new_hash = certificate_new.hash.clone();
            certificates_chained.push(certificate_new);

            new_hash
        });
        ``

**Review by sfauvel — APPROVED** [review]
LGTM

**Review by Alenar — APPROVED** [review]
LGTM

File: mithril-common/src/test_utils/certificate_chain_builder.rs (hunk 1 above)
Comment (jpraynaud) [reply mithril-common/src/test_utils/certificate_chain_builder.rs:271]: This is because the genesis epoch will have only one certificate.
I enhanced the readability of the formula and I gave some more details in the explanation comment :+1:

File: mithril-common/src/test_utils/certificate_chain_builder.rs [hunk 3]
```diff
@@ -0,0 +1,752 @@
...
+            .enumerate()
+            .for_each(|(i, certificate)| {
+                let mut certificate_new = certificate.clone();
+                if i > 0 {
+                    if let Some(previous_certificate) = certificates_chained.get(i - 1) {
+                        certificate_new.previous_hash = previous_certificate.compute_hash();
+                    }
+                } else {
+                    certificate_new.previous_hash = "".to_string();
+                }
```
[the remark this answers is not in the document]
jpraynaud: Yes, it would be better indeed. It will be fixed in a slightly different way with the implementation of this proposition: https://github.com/input-output-hk/mithril/pull/1999#discussion_r1802998682 :+1:

jpraynaud:
> Just a comment regarding the exposure of data in the `certificate_chain_builder` module. It seems that we could set the following traits to private visibility:
> 
> * `GenesisCertificateProcessorFunc`
> * `StandardCertificateProcessorFunc`
> * `TotalSignersPerEpochProcessorFunc`

You're right. Done :heavy_check_mark: 

> The `CertificateChainBuilderContext` struct could also switch to private by adding a `#[cfg(test)]` attribute on the `with_genesis_certificate_processor` and `with_standard_certificate_processor` functions.
> 
> Do you think it could be a good idea?

The usage of  `#[cfg(test)]` is not possible as we will use this builder in other crates. 
Also the `CertificateChainBuilderContext` should keep its field public as they give valuable information when altering the certificate chain

mithril-common/src/test_utils/certificate_chain_builder.rs
jpraynaud: Absolutely, thanks for the proposal! 
I updated the implementation of the certificate chaining in https://github.com/input-output-hk/mithril/pull/1999/commits/c223548733d72af7094486246d299e7f63e3d5ee :+1:

PR #2002, opened 2024-10-11
Enhance logs construction and usage in aggregator

Alenar:
## Content

This PR refactor the log usage in `mithril-aggregator`:
- A dedicated child logger is used for each service
- Each child logger is tagged with its service name in order to identify quickly a log source.
- Remove `slog_scope` to enforce the usage of a name tagged child logger (only kept in the integration tests).
- Change the name property in all logs from the default `slog-rs` to `mithril-signer`.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Comments

As this PR is already quite big, a second PR will come that will pass over all logs in the aggregator to move away from the log message huge structure  into properties, to remove prefixes in message as now all logs are tagged with their source, and to enforce a common style & structure.

## Issue(s)

Relates to #1981

### DIFF
--- a/mithril-aggregator/src/dependency_injection/builder.rs
+++ b/mithril-aggregator/src/dependency_injection/builder.rs
@@ -98,6 +98,9 @@ pub struct DependenciesBuilder {
     /// Configuration parameters
     pub configuration: Configuration,
 
+    /// Application root logger
+    pub root_logger: Logger,
+
     /// SQLite database connection
     pub sqlite_connection: Option<Arc<SqliteConnection>>,
 
@@ -237,9 +240,10 @@ pub struct DependenciesBuilder {
 
 impl DependenciesBuilder {
     /// Create a new clean dependency builder
-    pub fn new(configuration: Configuration) -> Self {
+    pub fn new(root_logger: Logger, configuration: Configuration) -> Self {
         Self {
             configuration,
+            root_logger,
             sqlite_connection: None,
             sqlite_connection_cardano_transaction_pool: None,
             stake_store: None,
@@ -303,7 +307,7 @@ impl DependenciesBuilder {
         sqlite_file_name: &str,
         migrations: Vec<SqlMigration>,
     ) -> Result<SqliteConnection> {
-        let logger = self.get_logger()?;
+        let logger = self.root_logger();
         let connection_builder = match self.configuration.environment {
             ExecutionEnvironment::Test
                 if self.configuration.data_stores_directory.to_string_lossy() == ":memory:" =>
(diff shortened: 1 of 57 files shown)


mithril-aggregator/src/store/epoch_settings_storer.rs
```diff
@@ -37,7 +36,6 @@ pub trait EpochSettingsStorer: Sync + Send {
         for epoch_offset in 0..=2 {
             let epoch = current_epoch + epoch_offset;
             if self.get_epoch_settings(epoch).await?.is_none() {
-                debug!("Handle discrepancies at startup of epoch settings store, will record epoch settings from the configuration for epoch {epoch}: {epoch_settings_configuration:?}");
```
jpraynaud, line 40: Do we want to remove this log?

jpraynaud (approved):
LGTM :+1:

mithril-aggregator/src/dependency_injection/builder.rs
```diff
@@ -1598,6 +1621,13 @@ impl DependenciesBuilder {
     }
 }
 
+#[cfg(test)]
+impl DependenciesBuilder {
+    pub(crate) fn new_for_test(configuration: Configuration) -> Self {
```
sfauvel, line 1626: I prefer a name that explains what is does rather than a name that says who should use it or in which context. It's just an opinion, the opposite can also be justified.
Something like : new_with_logs_to_stdout

mithril-aggregator/src/tools/signer_importer.rs
```diff
@@ -84,10 +95,6 @@ pub trait SignersImporterPersister: Sync + Send {
 #[async_trait]
 impl SignersImporterPersister for SignerStore {
     async fn persist(&self, signers: HashMap<PartyId, Option<PoolTicker>>) -> StdResult<()> {
-        info!(
```
sfauvel, line 87: Is it wanted to remove this ?

sfauvel (approved):
Good work
Just a doubt about a function name

mithril-aggregator/src/store/epoch_settings_storer.rs
Alenar: It has not been removed but moved to where it was called in the builder. I can't keep it here without a global log system such as `slog_scope`.

mithril-aggregator/src/tools/signer_importer.rs
Alenar: It has not been removed but displaced to where it was called, in the `run` method above,  since I don't have access to the logger here.

dlachaume (approved):
LGTM 🚀

PR #2012, opened 2024-10-16
Refactor client cli command parameters

Alenar:
## Content

This PR change how parameters are passed to client-cli commands: Instead of providing a `ConfigBuilder` from the config crate a `CommandContext` new struct is passed so we can provide more than just configuration parameters to sub-commands.

This change is needed in order to remove `slog_scope` so we can share the main logger defined in the bin `main`.

Two others changes are included:
- Introduce a `SharedArgs` struct so to define in one place arguments shared between all commands (only the `--json` option for now). It's flattened so this doesn't change the api.
- Minimize use of the config crate by instead using a "source" system with the `ConfigParameters`. Using both crate to fetch the same thing (parameters) was quite confusing, this is changed so config is only used for its capacity of reading multiple files formats and to provide the initial hasmap of parameters.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3662,7 +3662,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-client-cli"
-version = "0.9.15"
+version = "0.9.16"
 dependencies = [
  "anyhow",
  "async-trait",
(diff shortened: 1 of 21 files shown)


dlachaume (approved):
LGTM 🚀

jpraynaud (approved):
LGTM :+1:

PR #2032, opened 2024-10-21
Fix: computation of the chaining of the certificates in tests

jpraynaud:
## Content

This PR includes a fix to the chaining computation of the certificate chain in tests:

The previous certificate is either:
- the first certificate of the previous epoch if the certificate is the first of the epoch
- otherwise the first certificate of the epoch

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

### DIFF
--- a/mithril-common/src/certificate_chain/certificate_verifier.rs
+++ b/mithril-common/src/certificate_chain/certificate_verifier.rs
@@ -524,7 +524,7 @@ mod tests {
     #[tokio::test]
     async fn test_verify_certificate_chain_ok() {
         let total_certificates = 15;
-        let certificates_per_epoch = 2;
+        let certificates_per_epoch = 1;
         let (fake_certificates, genesis_verifier) =
             setup_certificate_chain(total_certificates, certificates_per_epoch);
         let mut mock_certificate_retriever = MockCertificateRetrieverImpl::new();
@@ -551,7 +551,7 @@ mod tests {
     #[tokio::test]
     async fn test_verify_certificate_chain_ko() {
         let total_certificates = 15;
-        let certificates_per_epoch = 2;
+        let certificates_per_epoch = 1;
         let (mut fake_certificates, genesis_verifier) =
             setup_certificate_chain(total_certificates, certificates_per_epoch);
         let index_certificate_fail = (total_certificates / 2) as usize;
--- a/mithril-common/src/certificate_chain/certificate_verifier.rs
+++ b/mithril-common/src/certificate_chain/certificate_verifier.rs
@@ -524,7 +524,7 @@ mod tests {
     #[tokio::test]
     async fn test_verify_certificate_chain_ok() {
         let total_certificates = 15;
-        let certificates_per_epoch = 1;
+        let certificates_per_epoch = 2;
         let (fake_certificates, genesis_verifier) =
             setup_certificate_chain(total_certificates, certificates_per_epoch);
         let mut mock_certificate_retriever = MockCertificateRetrieverImpl::new();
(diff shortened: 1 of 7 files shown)


mithril-common/src/test_utils/certificate_chain_builder.rs
```diff
@@ -760,32 +789,66 @@ mod test {
...
+        let certificate_7 = create_fake_certificate(Epoch(4), 7);
+        let certificates = vec![
+            certificate_1,
+            certificate_2,
+            certificate_3,
+            certificate_4,
+            certificate_5,
+            certificate_6,
+            certificate_7,
         ];
```
Alenar, line 815: Maybe we could use a more compact code since those variables are not used afterward, I don't think that we loose any readability since the index is still provided (maybe giving the index as the first parameter would be a little more clear).
```suggestion
        let certificates = vec![
            create_fake_certificate(Epoch(1), 1),
            create_fake_certificate(Epoch(2), 2),
            create_fake_certificate(Epoch(2), 3),
            create_fake_certificate(Epoch(3), 4),
            create_fake_certificate(Epoch(4), 5),
            create_fake_certificate(Epoch(4), 6),
            create_fake_certificate(Epoch(4), 7),
        ];
```

mithril-common/src/test_utils/certificate_chain_builder.rs
```diff
@@ -419,29 +419,58 @@ impl<'a> CertificateChainBuilder<'a> {
...
+            .iter()
+            .rev()
+            .filter(|c| {
+                if is_certificate_first_of_epoch {
+                    // The previous certificate of the first certificate of an epoch
+                    // is the first certificate of the previous epoch
+                    c.epoch == certificate.epoch.previous().unwrap()
+                } else {
+                    // The previous certificate of not the first certificate of an epoch
+                    // is the first certificate of the epoch
```
dlachaume, line 456: It didn’t seem clear to me, do you think this is better?
```suggestion
                    // The previous certificate that is not the first certificate of an epoch
                    // is the first certificate of the current epoch
```

dlachaume (approved):

mithril-common/src/test_utils/certificate_chain_builder.rs
```diff
@@ -760,32 +789,66 @@ mod test {
...
+            certificate_chained_2.previous_hash,
+            certificate_chained_1.hash
+        );
+        assert_eq!(
+            certificate_chained_3.previous_hash,
+            certificate_chained_2.hash
+        );
+        assert_eq!(
+            certificate_chained_4.previous_hash,
+            certificate_chained_2.hash
```
sfauvel, line 839: I don't understand why we compare previous_hash from 4th with hash from 2nd (idem with 7th and 5th).
If it's an error in test, there is  probably because all hashes are equals and it's a problem to really test values.

sfauvel (approved):
LGTM
Something strange in tests.

mithril-common/src/test_utils/certificate_chain_builder.rs
jpraynaud: I compacted the code as you propose :+1:

mithril-common/src/certificate_chain/certificate_verifier.rs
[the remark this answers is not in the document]
jpraynaud: I have put back 2 certificates per epoch. In order to make the test work, as the sequence of mock calls was not working properly (with multiple certificates per epoch there is not anymore exactly one call per certificate made by the retriever), I added a fake implementation of the certificate retriever.

PR #2049, opened 2024-10-24
Refactor ChainPoint conversion from pallas Point

### LINKED ISSUE #2037: Refactor `ChainPoint` conversion from pallas `Point`
## Why
The conversion from pallas `Point` representation to `ChainPoint` is problematic as it introduces a `BlockNumber` with value `0`.
This means that the conversionis not safe and should be done to another type which does not hold a `BlockNumber`.

## What
Refactor the conversion from `Point` to `RawCardanoPoint` instead of `ChainPoint`

## How

- [x] Refactor the conversion from `Point` to `RawCardanoPoint` instead of `ChainPoint`

Alenar:
## Content

This PR introduce a new type, `RawCardanoPoint`, that is used in place of our `ChainPoint` in the transaction import process.

The `RawCardanoPoint` doesn't have a block number unlike `ChainPoint`, this allow a lossless conversion from and to the pallas point used in their chainsync implementation, avoiding the creation of `ChainPoint` with a block number of `0`.

Also this PR changes the `ScannedBlock` block hash to be a bytes array as this reduce the number of conversion needed when working with pallas types as this type is only used in conjunction with those types (`RawCardanoPoint` also use a bytes array for the same reason).

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation

## Issue(s)

Closes #2037

### DIFF
--- a/mithril-common/src/cardano_block_scanner/scanned_block.rs
+++ b/mithril-common/src/cardano_block_scanner/scanned_block.rs
@@ -1,14 +1,13 @@
 use pallas_traverse::MultiEraBlock;
+use std::fmt::{Debug, Formatter};
 
-use crate::entities::{
-    BlockHash, BlockNumber, CardanoTransaction, ChainPoint, SlotNumber, TransactionHash,
-};
+use crate::entities::{BlockNumber, CardanoTransaction, ChainPoint, SlotNumber, TransactionHash};
 
 /// A block scanned from a Cardano database
-#[derive(Debug, Clone, PartialEq)]
+#[derive(Clone, PartialEq)]
 pub struct ScannedBlock {
     /// Block hash
-    pub block_hash: BlockHash,
+    pub block_hash: Vec<u8>,
     /// Block number
     pub block_number: BlockNumber,
     /// Slot number of the block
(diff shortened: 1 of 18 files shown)


sfauvel (approved):
LGTM

dlachaume (approved):
LGTM 🚀

mithril-common/src/cardano_block_scanner/scanned_block.rs
```diff
@@ -19,11 +18,11 @@ pub struct ScannedBlock {
 
 impl ScannedBlock {
     /// Scanned block factory
-    pub fn new<T: Into<TransactionHash>, U: Into<BlockHash>>(
-        block_hash: U,
+    pub fn new<BlkHash: Into<Vec<u8>>, TxHash: Into<TransactionHash>>(
```
jpraynaud, line 21: It's not very clear that `BlkHash`and `TxHash` are generic types:
```suggestion
    pub fn new<BLOCK_HASH: Into<Vec<u8>>, TX_HASH: Into<TransactionHash>>(
```

jpraynaud (approved):
LGTM :+1:

PR #2065, opened 2024-10-30
Upgrade mithril explorer to Next 15

Alenar:
## Content

This PR upgrade mithril explorer main dependencies, NextJs, to its latest major version.

There's some caveat:
* Vercel, the NextJs editor, recommend to use React 19 that's still a release candidate, but it's not compatible with `@reduxjs/toolkit` so React 18.3 is kept.
* Version 15 of `eslint-config-next` is compatible with eslint 9 but still ask for version `8.57.1` in its package json, leading to warnings when using `npm install`.
  We should keep using eslint `8.57` until this is solved.
* Vercel say that Turbopack, its webpack replacement made with rust, is  ready for use in dev, but when we use it the app crash because it can't import properly the wasm library.
  We should keep using webpack until this is solved.
  Note: To test this tool, add `--turbopack` to the `dev` scripts in  package.json.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

### DIFF
--- a/mithril-explorer/Makefile
+++ b/mithril-explorer/Makefile
@@ -52,8 +52,9 @@ clean:
 	rm package-lock.json || true
 
 upgrade: clean install
-	# Note: eslint versions 9 or up are not compatible with versions 14.2+ of eslint-config-next
-	# Todo: switch to eslint-config-next@latest when a compatible version is released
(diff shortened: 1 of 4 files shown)


jpraynaud (approved):
LGTM

dlachaume (approved):
LGTM 👍

Issue #1957, opened 2024-09-26
Remove `network` field from `CardanoDbBeacon`


jpraynaud:
## Why
The `CardanoDbBeacon` holds a `network` field which should be removed as this information is not necessary to operate the Mithril network.

## What
Remove the `network` field in the `CardanoDbBeacon` **without breaking-change**.

## How

- [ ] Adapt the computation of the hash in the `compute_hash` of the `CardanoImmutableDigester` to retrieve the network value:
  - [ ] from the configuration in the signer
  - [ ] from the configuration in the aggregator
  - [ ] from the certificate metadata in the client
  - [ ] Golden test the computation of the digest to make sure there is no breaking change
- [ ] Remove the `network` field in the `CardanoDbBeacon`
- [ ] :warning: Recompute certificate chain to be scheduled with next distribution

PR #2081, opened 2024-11-04
Streamline configurations usage in aggregator

Alenar:
## Content

This PR changes the way the `Configuration` structure is used by the aggregator: except for one case (for tests), it's now only used in the dependency builder.

The `Configuration` struct contains un-validated, raw, values that must be transformed using an error prone operation. We should strictly limit its usages to the dependency builder to limit such errors to the application startup.

The main user of the `Configuration` beside the dependency container is the HTTP Server. In order to provide for its need validated configuration key the following refactor was done:

1. A new structure is introduced to hold validated configuration keys: `RouterConfig`
2. Another new structure, `RouterState`, is added to hold the dependency container (to access services) and the `RouterConfig`
3. **All** routes takes a `RouterState` as parameter instead of directly the `DependencyContainer`
4. For routes that needs a config value a middleware is added to extract the value needed: `middlewares::extract_config`

Beside the http server only few direct usages of the `Configuration` struct existed and this PR reduced them to only one case. This remaining case is not trivial to address, as such I think it needs further thinking.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Issue(s)

Relates to #1957

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3552,7 +3552,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-aggregator"
-version = "0.5.98"
+version = "0.5.99"
 dependencies = [
  "anyhow",
  "async-trait",
--- a/mithril-aggregator/Cargo.toml
+++ b/mithril-aggregator/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mithril-aggregator"
-version = "0.5.98"
+version = "0.5.99"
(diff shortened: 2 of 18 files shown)


sfauvel (approved):
LGTM

dlachaume (approved):
LGTM 👍

jpraynaud (approved):
LGTM :+1:

PR #2090, opened 2024-11-05
Simplify tests

Alenar:
## Content

This PR includes simplifications for tests:

- signer `mithril-signer/src/services/aggregator_client.rs`:
  - use a shared `setup_server_and_client` function to create the client in tests.
- aggregator `mithril-aggregator/src/services/message.rs`:
  - Create the `MessageService` directly instead of relying on the dependency builder, this allow finer control of its construction and avoid importing unnecessary context
  - Use a builder pattern for a simple service construction, to limit the boilerplate in tests
  - Group tests by artifacts types, to make them easier to navigate
- aggregator http server:
  - Use dummy message directly instead of relying entities conversion using adapters
- aggregator adapters:
  - remove all "test only" adapters since all their usages have been removed with changes above  

Those changes make easier to write the tests needed for #1957.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Issue(s)

Relates to #1957


sfauvel (approved):

jpraynaud (approved):
LGTM :+1:

PR #2103, opened 2024-11-08
Refactor pruning with upkeep service

sfauvel:
## Content

This PR includes...

## Pre-submit checklist

- Branch
- PR
  - [ ] Self-reviewed the diff
  - [ ] Reviewer requested
- Documentation

## Issue(s)


internal/mithril-persistence/src/sqlite/connection_extensions.rs
```diff
@@ -30,6 +30,12 @@ pub trait ConnectionExtensions {
     fn fetch_collect<Q: Query, B: FromIterator<Q::Entity>>(&self, query: Q) -> StdResult<B> {
         Ok(self.fetch(query)?.collect::<B>())
     }
+
+    /// Apply a query that do not return data to the database(ie: insert, delete, ...).
```
dlachaume, line 34:
```suggestion
    /// Apply a query that do not return data from the database(ie: insert, delete, ...).
```

internal/mithril-persistence/src/sqlite/connection_extensions.rs
```diff
@@ -115,4 +123,56 @@ mod tests {
 
         assert_eq!(value, 45);
     }
+
+    #[test]
+    fn test_apply_execute_the_query() {
```
dlachaume, line 128: Would it be relevant to ensure correct behavior with `update` and `delete` queries as well?

dlachaume (approved):
LGTM 👍

mithril-aggregator/src/services/upkeep.rs
```diff
@@ -9,21 +9,31 @@ use std::sync::Arc;
...
 
 /// Define the service responsible for the upkeep of the application.
 #[cfg_attr(test, mockall::automock)]
 #[async_trait]
 pub trait UpkeepService: Send + Sync {
     /// Run the upkeep service.
-    async fn run(&self) -> StdResult<()>;
+    async fn run(&self, epoch: Epoch) -> StdResult<()>;
+}
+/// Define the task responsible for pruning a datasource below a certain epoch threshold.
```
jpraynaud, line 28:
```suggestion
}

/// Define the task responsible for pruning a datasource below a certain epoch threshold.
```

internal/mithril-persistence/src/sqlite/connection_extensions.rs
sfauvel: Done

internal/mithril-persistence/src/sqlite/connection_extensions.rs
sfauvel: I'm not sure that does much. The apply function only executes the query whatever it is.
We can add this to the test but it will make it more complex. 
Do you think this is useful?

dlachaume: I see your point and I'm fine keeping it as is. 👍

mithril-aggregator/src/dependency_injection/builder.rs
```diff
@@ -133,8 +133,8 @@ pub struct DependenciesBuilder {
     /// Verification key store.
     pub verification_key_store: Option<Arc<dyn VerificationKeyStorer>>,
 
-    /// Epoch settings storer.
-    pub epoch_settings_storer: Option<Arc<dyn EpochSettingsStorer>>,
+    /// Epoch settings store.
+    pub epoch_settings_store: Option<Arc<EpochSettingsStore>>,
```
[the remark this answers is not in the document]
sfauvel: This is because of the `build_upkeep_service` that use `get_epoch_settings_store` to retrieve the store from the `epoch_settings_store`. 
If it's the trait `EpochSettingsStorer`, it' could not be use as a `EpochPruningTask` even if the trait `EpochSettingsStorer` is a `EpochPruningTask`.
From what I understood, it's a rust limitation with `upcast` and requires adding specific upcast implementations.
https://stackoverflow.com/questions/73003800/how-does-the-trait-upcast-coercion-workaround-work-trait-method-for-getting-an
This seemed acceptable to me because this is only internal variable and functions of the `aggregator` (why `get_epoch_settings_store` is public ?). 
This is also what we do with `SignerStore` and `CertificatePendingStore`.
Do you think we need to rework this part?

mithril-aggregator/src/services/upkeep.rs
sfauvel: Done

mithril-aggregator/src/dependency_injection/builder.rs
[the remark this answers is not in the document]
jpraynaud: I'm not sure that the trait casting is a problem :thinking: 

I have created this [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=5011d623d14b15f8e42f60430f115a93) that works well by using a `Storer` and casting it to a `Pruner`:
```rust

trait Pruner {
    fn prune(&self){
        println!("prune")
    }
}

trait Storer: Pruner {
    fn store(&self){
        println!("store")
    }
}

struct MyStore;

impl Pruner for MyStore {}
impl Storer for MyStore {}

fn do_something_with_pruner(pruner: &dyn Pruner) {
    println!("do_something_with_pruner");
    pruner.prune();
}

fn main(){
    let store = MyStore{};
    store.store();
    store.prune();
    
    do_something_with_pruner(&store)
}
```

And outputs:
```
store
prune
do_something_with_pruner
prune
```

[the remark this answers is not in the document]
sfauvel: Yes but if your `store` variable is a Arc<dyn Storer> it doesn't work.
You can verify it forcing the type: `let store: Arc<dyn Storer> = Arc::new(MyStore {});`

This is what the `get_epoch_settings_store` returned (`̀Result<Arc<EpochSettingsStore>>`)
I haven't found a simple solution to this problem.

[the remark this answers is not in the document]
jpraynaud: It actually works by creating a reference from the `Arc` (with `as_ref()`) as in this [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=c0ded17a351fc4e51cce7fc7c32cb58d).
Let me know if that helps :+1:

jpraynaud (approved):
LGTM :+1:

PR #2110, opened 2024-11-13
Remove network from `CardanoDbBeacon`

### LINKED ISSUE #1957: Remove `network` field from `CardanoDbBeacon`

Alenar:
## Content

> [!WARNING]
> This PR changes the certificate hashes computation when their signed entity type is `CardanoImmutableFilesFull`, the [`recompute-certificates-hash`](https://github.com/input-output-hk/mithril/tree/main/docs/runbook/recompute-certificates-hash) tool must be run in order to avoid breaking the chain.

This PR remove the `network` field from the `CardanoDbBeacon` struct. With the groundwork done in  #2097 no additional adaptation was needed.

With this field removal some tools and  services that used it only to produce `CardanoDbBeacon` were simplified:

- signed entity config structure: used to provide conversion from timepoints to signed entities
- epoch services from both the aggregator and signer: used to provide signed entity config struct

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation

## Comments

Some network field still persist in serialized `CardanoDbBeacon` in database:

- Aggregator:
  - `open_message.beacon`: existing network won't matter as they won't be mapped when read and since open messages are pruned regularly they will disappear by themself.
  - `certificate.beacon`:  existing network won't matter as they won't be mapped when read, plus certificate hashes will be recomputed for all certificates that hold a `CardanoDbBeacon` meaning that all those certificates will be re-inserted in DB, eliminating the old network field.
  - `signed_entity.beacon`: for previously stored `CardanoImmutableFilesFull` entities the network will remain but should not matter as it won't be read when de-serialized.
  - `signed_entity.artifact`: for previously stored `CardanoImmutableFilesFull` entities the artifact a `network` field will remain in serialized `Snapshot`, this should not matter as it won't be read when de-serialized.
- Signer:
  - `signed_beacon.beacon`: existing network won't matter as they won't be mapped when read and since signed beacon are pruned regularly they will disappear by themself.

Question: Should the database be cleaned of those remaining data with a migration ? Especially for signed_entities as those data won't be removed.

## Issue(s)

Closes #1957

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,11 @@ As a minor extension, we have adopted a slightly different versioning convention
 
 - Implement the `/status` route on the aggregator's REST API to provide information about its current status.
 
+- Deprecate the `network` field in all messages that contain a `CardanoDbBeacon` field.
+
+- **BREAKING** Remove the `network` field from the internal `CardanoDbBeacon` struct, Certificates of types `CardanoImmutableFilesFull`
+  must have their hashes recomputed in order to stay valid (procedure detailed in the [`recompute-certificates-hash`](./docs/runbook/recompute-certificates-hash) runbook).
+
 - Crates versions:
 
 | Crate | Version |
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3588,7 +3588,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-aggregator"
-version = "0.5.110"
+version = "0.5.111"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -3700,7 +3700,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-client-cli"
-version = "0.10.2"
+version = "0.10.3"
 dependencies = [
  "anyhow",
  "async-trait",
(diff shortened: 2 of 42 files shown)


sfauvel (approved):
LGTM

dlachaume (approved):
LGTM 👍

jpraynaud (approved):
LGTM :rocket:

PR #2128, opened 2024-11-18
Display aggregator status information in explorer

Alenar:
## Content

This PR upgrade the explorer UI, superseding the `Epoch Settings` block with a `Aggregator Status` pulling data from the new status route.

image

This new block is responsive across 5 breakpoints:
[Screencast from 2024-11-18 16-48-49.webm](https://github.com/user-attachments/assets/bf28bd80-1d22-4841-95c6-27ea2b6e69d6)

It can be collapsed if needed at any breakpoints:
[Screencast from 2024-11-18 16-50-05.webm](https://github.com/user-attachments/assets/db2538d4-b13c-4e1c-91e3-fdc54abcd8d6)

If the status route isn't available on the target aggregator the UI fallback to the previous `Epoch Settings`:
[Screencast from 2024-11-18 16-52-01.webm](https://github.com/user-attachments/assets/664098dd-2207-47bf-a952-0c3c9161a3e4)

## Pre-submit checklist

- Branch
  - [ ] Tests are provided (if possible)
  - [x] Crates versions are updated (if relevant)
  - [ ] CHANGELOG file is updated (if relevant)
  - [x] Commit sequence broadly makes sense
  - [x] Key commits have useful messages
- PR
  - [x] No clippy warnings in the CI
  - [x] Self-reviewed the diff
  - [x] Useful pull request description
  - [x] Reviewer requested

## Issue(s)

Closes #2073

PR #2130, opened 2024-11-18
fix(aggregator): properly handle `404` errors in http server

Alenar:
## Content

This PR fix the handling of `404` errors in the aggregator http server.

In case of a `404` the custom rejection middleware  immediatly emit a rejection _(this custom rejection was added to set the appropriate status code for API Version Mistmatch errors)_.
This means that no other middleware would be applied nor our custom headers.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Comments

This problem cascade to the browser javascript `fetch` api, the missing headers make the api immediately throw an error thus making impossible [to handle the `404` in a idiomatic way](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#checking_response_status).

In this example the status code can't be checked in the `then` block since a exception is raised _(note: the exception doesn't contains the status code)_.
```javascript
fetch(aggregatorEndpoint)
  .then((response) => {
    if (response.status === 404) {
      // do something with it
    } else {
      return response.status === 200 ? response.json() : {};
    }
  })
  .catch((error) => {
    console.error("Fetch status error:", error);
  });
``` 

This problems made more cumbersome the fallback to `Epoch Setting` mechanism used in PR #2128, we had to handle it in the `catch` block without the ability to check that it was indeed a `404` or not.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3588,7 +3588,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-aggregator"
-version = "0.5.112"
+version = "0.5.113"
 dependencies = [
  "anyhow",
  "async-trait",
--- a/mithril-aggregator/Cargo.toml
+++ b/mithril-aggregator/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mithril-aggregator"
-version = "0.5.112"
+version = "0.5.113"
 description = "A Mithril Aggregator server"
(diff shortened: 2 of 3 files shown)


dlachaume (approved):

jpraynaud (approved):
LGTM :+1:

PR #2137, opened 2024-11-22
Also bump versions of js packages in upgrade version script

Alenar:
## Content

This PR add to the upgrade version scrip in the devbook the capacity to bump js packages versions and rename it from `upgrade-crates-and-openapi-versions` to simply `bump-versions`.

It works the following way:

1. Detect all `package.json` files in the project, excluding those auto-generated or from imported packages
1. If there was a change compared to the `main` branch in a `package.json` directory
   - _In dry run_ only print the updated version
   - _else_ use `npm version patch` to bump the patch version (with `--no-git-tag-version` to handle the commit our-self)
1. Append the modified versions to the commit message

> [!NOTE]
> `docs/website/` is excluded as we don't want to bump its version if its articles or blog posts changed.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation

### DIFF
--- a/docs/devbook/bump-versions/bump_versions.sh
+++ b/docs/devbook/bump-versions/bump_versions.sh
@@ -149,7 +149,7 @@ fi
 
 if [ true = $DRY_RUN ]
 then
-    echo -e "${ORANGE}warning${RESET}: script is run in dry mode. To execute the changes, run ${GREEN}./update_crate_versions.sh --run${RESET}"
+    echo -e "${ORANGE}warning${RESET}: script is run in dry mode. To apply the changes, run ${GREEN}$0 --run${RESET}"
 else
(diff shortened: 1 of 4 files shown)


docs/devbook/bump-versions/bump_versions.sh
```diff
@@ -99,14 +146,20 @@ else
   then
     UPDATED_CRATES="\n${UPDATED_CRATES}"
   fi
+  UPDATED_PACKAGE_JSONS="$(echo "$PACKAGE_JSON_FILES" | xargs git diff "$COMMIT_REF" | grep -E "^[\+\-] *\"version\": \"[0-9\.]+\"|name\": " | tr '\n' ' ' | sed -r "s/ *\"name\": \"([a-z@\-]*\/)?([a-z\-]+)\", -  \"version\": \"([0-9\.]+)\", \+  \"version\": \"([0-9\.]+)\", ?/* [js] \2 from \`\3\` to \`\4\`\n/g")"
```
sfauvel, line 149: We may put some explanation somewhere (here or at the beginning of the script)
It could be just a short example of what we have as input and what we want as output to help understanding what we are doing.

sfauvel (approved):
LGTM
Just one remark

docs/devbook/bump-versions/bump_versions.sh
Alenar: :heavy_check_mark: I've added a comment, see few lines above, that contains the transformation that we want to do for those messages.

jpraynaud (approved):

PR #2146, opened 2024-11-28
Feat: implement signable builder for incremental Cardano DB

### LINKED ISSUE #2122: Implement signable builder for Incremental Cardano DB
## Why
In order to certify the Cardano database and provide differential restoration, we need to implement the signable builders for the `CardanoDatabase` signed entity type.

## What
Implemement the signable builder for the `CardanoDatabase` signed entity type.

## How

- [x] Activate optionally the `CardanoDatabase` signed entity type in the aggregator 
- [x] Adjust the timeout on the `CardanoDatabase` signed entity type to `1800s`
- [x] Implement the signable builder:
  - [x] Create a Merkle tree with leaves `(immutable_digest)` (same computation for the digest as in `CardanoImmutableFilesFull`)
  - [x] Use the Merkle root as the message to sign
- [x] ~~Create a boilerplate artifact builder?~~
- [x] Activate the `CardanoDatabase` signed entity type in the e2e test

dlachaume:
## Content

This PR includes...

## Pre-submit checklist

- Branch
- PR
  - [ ] Self-reviewed the diff
  - [ ] Reviewer requested
- Documentation

## Issue(s)

Closes #2122


mithril-common/src/digesters/cardano_immutable_digester.rs
```diff
@@ -49,99 +84,125 @@ impl ImmutableDigester for CardanoImmutableDigester {
...
+        let immutables_to_process =
+            list_immutable_files_to_process(dirpath, beacon.immutable_file_number)?;
+        info!(self.logger, ">> compute_merkle_tree"; "beacon" => #?beacon, "nb_of_immutables" => immutables_to_process.len());
+        let computed_immutables_digests = self.process_immutables(immutables_to_process).await?;
+
+        let leaves: Vec<_> = computed_immutables_digests
+            .digests
+            .into_iter()
+            .map(|digest| MKTreeNode::new(digest.into()))
+            .collect();
```
jpraynaud, line 134: This conversion could probably be avoided here and left to the constructor of the Merkle tree which takes whatever is convertible into a MKTreeNode: `fn new<T: Into<MKTreeNode> + Clone>(leaves: &[T]) -> StdResult<Self>`.

mithril-common/src/digesters/cardano_immutable_digester.rs
```diff
@@ -49,99 +84,125 @@ impl ImmutableDigester for CardanoImmutableDigester {
...
                 let data = hex::encode(entry.compute_raw_hash::<Sha256>()?);
-                hasher.update(&data);
+                digests.push(data.clone());
                 new_cached_entries.push((entry.filename.clone(), data));
             }
             Some(digest) => {
-                hasher.update(digest);
+                digests.push(digest.to_string());
             }
         };
```
jpraynaud, line 207: This could maybe be written in a more functional way?

mithril-common/src/digesters/cardano_immutable_digester.rs
```diff
@@ -49,99 +84,125 @@ impl ImmutableDigester for CardanoImmutableDigester {
...
+                found_number: Some(last_immutable_file.number),
+                db_dir: dirpath.to_owned(),
+            })
+        }
+        Some(_) => Ok(immutables),
     }
 }
 
-fn compute_hash(
+fn compute_immutables_digests(
```
jpraynaud, line 185: Maybe this function could be unit tested as the `ComputedImmutablesDigests` struct is not order safe.

IMO a better structure for the `ComputedImmutablesDigests` would be:
```rust
struct ComputedImmutablesDigests {
    entries: BTreeMap<ImmutableFile, HexEncodedDigest>,
    new_cached_entries: Vec<ImmutableFileName>,
}
```

mithril-common/src/digesters/cardano_immutable_digester.rs
```diff
@@ -502,7 +655,48 @@ mod tests {
     }
 
     #[tokio::test]
-    async fn cache_read_failure_dont_block_computation() {
+    async fn merkle_tree_computation_is_quicker_with_a_full_cache() {
```
jpraynaud, line 658: This test is probably not necessary (very redundant) and not that relevant after all :thinking:

mithril-common/src/digesters/immutable_digester.rs
```diff
@@ -78,4 +86,8 @@ pub enum ImmutableDigesterError {
     /// Error raised when the digest computation failed.
     #[error("Digest computation failed")]
     DigestComputationError(#[from] io::Error),
+
+    /// Error raised when the merkle tree computation failed.
```
jpraynaud, line 90:
```suggestion
    /// Error raised when the Merkle tree computation failed.
```

mithril-common/src/signable_builder/cardano_database.rs
```diff
@@ -0,0 +1,145 @@
...
+            _dirpath: &Path,
+            _beacon: &CardanoDbBeacon,
+        ) -> Result<MKTree<MKTreeStoreInMemory>, ImmutableDigesterError> {
+            let leaves = create_leaves_from_digests(self.digests.clone());
+
+            Ok(MKTree::new(&leaves).unwrap())
+        }
+    }
+
+    fn create_leaves_from_digests(digests: Vec<String>) -> Vec<MKTreeNode> {
```
jpraynaud, line 114: Same remark as above, this function should not be necessary.

mithril-common/src/digesters/cardano_immutable_digester.rs
dlachaume: I initially thought it was a good way to demonstrate the behavior for each distinct process from start to finish, even if the cache reading process is shared.
I understand and I agree, the test is redundant, so I will remove it 👍

mithril-common/src/digesters/cardano_immutable_digester.rs
dlachaume: WDYT about this implementation:

```
fn compute_immutables_digests(
    logger: Logger,
    entries: BTreeMap<ImmutableFile, Option<HexEncodedDigest>>,
) -> ComputedImmutablesDigestsResult {
    let mut new_cached_entries = Vec::new();
    let mut progress = Progress {
        index: 0,
        total: entries.len(),
    };

    let digests: Result<Vec<String>, io::Error> = entries
        .iter()
        .enumerate()
        .map(|(ix, (entry, cache))| {
            let digest = match cache {
                None => {
                    let data = hex::encode(entry.compute_raw_hash::<Sha256>()?);
                    new_cached_entries.push((entry.filename.clone(), data.clone()));
                    data
                }
                Some(digest) => digest.to_string(),
            };

            if progress.report(ix) {
                info!(logger, "Hashing: {progress}");
            }

            Ok(digest)
        })
        .collect();

    Ok(ComputedImmutablesDigests {
        digests: digests?,
        new_cached_entries,
    })
}
```

mithril-common/src/digesters/cardano_immutable_digester.rs
dlachaume: In this implementation, we iterate through each value in the `Vec` in the order the elements were inserted. The test `can_compute_hash_of_a_hundred_immutable_file_trio` ensures that the computed hash remains identical to what was previously computed.

I agree with your suggestion to modify the `ComputedImmutablesDigests` struct to use a `BTreeMap`, as it enforces a defined order on `ImmutableFile`. I'll apply this change 👍

dlachaume: Done ✅

mithril-common/src/digesters/cardano_immutable_digester.rs
dlachaume: Done ✅

mithril-common/src/signable_builder/cardano_database.rs
dlachaume: Done ✅

mithril-common/src/digesters/cardano_immutable_digester.rs
dlachaume: @Alenar proposed this refactoring solution which address both this and the comment above: https://github.com/input-output-hk/mithril/pull/2146/commits/7953b8f68184363f34164f753d821435f64dd006

mithril-common/src/digesters/cardano_immutable_digester.rs
```diff
@@ -1,20 +1,25 @@
...
 use sha2::{Digest, Sha256};
 use slog::{debug, info, warn, Logger};
 use std::{collections::BTreeMap, io, path::Path, sync::Arc};
 
-/// Result of a cache computation, contains the digest and the list of new entries to add
+/// Result of a cache computation, contains the list of immutable digests and the list of new entries to add
 /// to the [ImmutableFileDigestCacheProvider].
-type CacheComputationResult =
-    Result<([u8; 32], Vec<(ImmutableFileName, HexEncodedDigest)>), io::Error>;
+type ComputedImmutablesDigestsResult = Result<ComputedImmutablesDigests, io::Error>;
```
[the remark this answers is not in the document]
dlachaume: Done ✅

mithril-common/src/digesters/immutable_digester.rs
```diff
@@ -78,4 +97,8 @@ pub enum ImmutableDigesterError {
     /// Error raised when the digest computation failed.
     #[error("Digest computation failed")]
     DigestComputationError(#[from] io::Error),
+
+    /// Error raised when the Merkle tree computation failed.
+    #[error("Merkle tree computation failed")]
+    MerkleTreeComputationError(anyhow::Error),
```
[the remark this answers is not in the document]
dlachaume: Done ✅

mithril-common/src/digesters/cardano_immutable_digester.rs
dlachaume: The code above is outdated. It's better to refer to the comment below instead.

mithril-common/src/signable_builder/cardano_immutable_full_signable_builder.rs
```diff
@@ -83,6 +84,14 @@ mod tests {
         ) -> Result<String, ImmutableDigesterError> {
             Ok(format!("immutable {}", beacon.immutable_file_number))
         }
+
+        async fn compute_merkle_tree(
+            &self,
+            _dirpath: &Path,
+            _beacon: &CardanoDbBeacon,
+        ) -> Result<MKTree<MKTreeStoreInMemory>, ImmutableDigesterError> {
+            unimplemented!()
```
jpraynaud, line 93: Just a suggestion to avoid the use of `unimplemented!` here, we could add the `process_immutables` function to the `ImmutableDigester` trait. This would allow to do a blanket implementation for both `compute_digest` and `compute_merkle_tree`.

jpraynaud (approved):
LGTM :rocket:

mithril-common/src/signable_builder/cardano_immutable_full_signable_builder.rs
Alenar: I don't think that's a good idea:
- `process_immutables` is an internal method, making it available in the public API would also force to make public internal types (`ComputedImmutablesDigests` in this case).
- The `ImmutableDigester` is used [on the public api of the `mithril-client` lib](https://docs.rs/mithril-client/latest/mithril_client/struct.MessageBuilder.html#method.with_immutable_digester) (to allow usage of other digesters by third party developers, as we do for other traits), so we should keep its API surface lean.
- `compute_merkle_tree` and `compute_digest` are not side effects free: they both use the logger and need to update the cache, two things only available on the structure that implement the `ImmutableDigester` trait.

dlachaume: Approach to avoid `unimplemented!()` by using a pattern builder: https://github.com/input-output-hk/mithril/pull/2146/commits/ac1e197213d00f04b3a5f448e10bbe5075b55f7b

mithril-common/src/digesters/cardano_immutable_digester.rs
```diff
@@ -40,6 +41,58 @@ impl CardanoImmutableDigester {
             logger: logger.new_with_component_name::<Self>(),
         }
     }
+
+    async fn process_immutables(
+        &self,
+        immutables: Vec<ImmutableFile>,
+    ) -> Result<ComputedImmutablesDigests, ImmutableDigesterError> {
+        let cached_values = match self.cache_provider.as_ref() {
```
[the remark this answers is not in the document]
dlachaume: Done in this commit: https://github.com/input-output-hk/mithril/pull/2146/commits/2209f530546af06ecf506f7bde83594d0dbdb735

mithril-common/src/signable_builder/cardano_immutable_full_signable_builder.rs
jpraynaud: I think on the opposite that there is no problem to even have the digester trait with a single function `compute_immutable_digests` that returns a `ComputedImmutablesDigests` which is responsible to compute both the digest and the Merkle tree. Apart form the fact that it would introduce a breaking change in the client, it would be even easier for third party developers to use our API.  The cache update should also probably be done by this function as computing the Merkle tree and the digest should be stateless operations.

OK with the builder approach to remove the usage of `unimplemented!` :+1:

jpraynaud (approved):

Issue #1484, opened 2024-02-01
Handle cache for certificate verification in WASM client


jpraynaud:
## Why
We want the client library to provide faster certificate verification in the explorer and also to allow for cancellation of an in progress verification.

## What
Provide a list of already signed transactions to the client so that it does not verify them or handle a dedicated browser store for that purpose. It's worth mentioning that once a certificate is verified and cache, all the remaining certificate chain is considered valid.

## How

- [ ] Handle cache for certificates verification
- [ ] Adapt explorer to take advantage of these new capabilities
- [ ] Update documentation

roynakakawa:
@jpraynaud I believe this issue is done ? can we close it ?

jpraynaud:
> [@jpraynaud](https://github.com/jpraynaud) I believe this issue is done ? can we close it ?

@roynakakawa yes we can close it 👍

PR #2166, opened 2024-12-10
Experimental: Cache in client certificate verification (lib and wasm)

Alenar:
## Content

This PR add a experimental cache system to the certificate chain verification of the client library.

This cache have two implementation:
- One purely in-memory
- One only for the client-wasm that store its value in the browser local storage.

### Client library

- Add a `CertificateVerifierCache` trait which define how to: get one cached value, add one value to a cache, reset all cached values.
  - Only available using the `unstable` feature
  - This cache only contains the certificate hash (key) and the certificate parent hash (value).
- Refactor the certificate chain validation process to use the cache:
  - Only available using the `unstable` feature
  - If a certificate hash is cached it's considered valid and the verification continue with the parent hash
  - Certificate are downloaded only if not cached
  - The certificate given as a starting point and the genesis certificate are always validated without the cache
  - If validated without the cache the certificate hash and parent hash are then stored in the cache
  - Genesis certificate are never stored in the cache
  - Providing a cache system is optional
- Add the `MemoryCertificateVerifierCache` struct:
   - Only available using the `unstable` feature
   - implement `CertificateVerifierCache` using a in-memory HashMap
   - have configurable expiration delay for its cached values
- Change the `certificate_client` module from a single file to a folder with the following organization:
  - `api.rs`: define the main `CertificateClient` struct and the public traits that extends it (such as the `CertificateVerifierCache`).
  - `fetch.rs`: implementation and tests of Certificate data fetching (get, list)
  - `verify.rs`: implementation and tests of Certificate chain verification
  - `verify-cache`: a submodule directory to regroup all cache implementations that are part of the client library
- Add a new feedback event: `CertificateValidatedFromCache`

### Client Wasm library

- Add the `LocalStorageCertificateVerifierCache` struct:
   - implement `CertificateVerifierCache` using [browsers local storage](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage)
   - for simplicity each value are stored in a separate key in the local storage (this have the side-effect of flooding the local storage with values, before it only had the 'Explorer_state' value).
     We could instead store them in a single key but that would mean serializing and/or de-serializing the whole cache at each access, alternatively it could be stored in-memory with a synchronization system.
- Add two new clients options:
  - `enable_certificate_chain_verification_cache`: default false, when used conjointly with `unstable` it load the `LocalStorageCertificateVerifierCache`.
  - `certificate_chain_verification_cache_duration_in_seconds`: modify cache item expiration duration, default to `604800` (one week).
- When trying to use the `LocalStorageCertificateVerifierCache` and the local storage is not available on the current platform (such as with nodejs) a warning is logged in the console and the cache system remain disabled.

### Explorer

- Enable `LocalStorageCertificateVerifierCache` 
- Certificate validated using the cache use a specific label 
- Add a clickable link at the end of the verification process to clear the cache.

### Demo

Here's a screencast of the cache in action:

[Screencast from 2024-12-12 16-30-18.webm](https://github.com/user-attachments/assets/c969607b-7fe5-4edf-ae7d-47e929c91255)

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation

## Comments

As the security of this cache system is quite low (see comment below), this should be only see as experimental and not used in production.

## Issue(s)

Relates to #1484


mithril-client-wasm/src/certificate_verification_cache.rs
```diff
@@ -0,0 +1,441 @@
...
+                .await
+                .unwrap();
+
+            let cached = cache
+                .get_cached_value("hash")
+                .expect("Cache should have been populated");
+
+            assert_eq!(1, cache.len());
+            assert_eq!("parent", cached.previous_hash);
+            assert!(cached.expire_at - now >= expiration_delay);
```
sfauvel, line 305: `now` it's a bit confusing. 
We expect that `expire_at - now` is less than `expiration_delay`
`now` it's the time before we add the cache.
Note: If we want to be more precise, it's possible to make 2 assertions comparing with time before assertion and one with time after assertion.

mithril-client-wasm/src/certificate_verification_cache.rs
```diff
@@ -0,0 +1,441 @@
...
+
+            assert_eq!(2, cache.len());
+            assert_eq!(
+                Some("another_parent".to_string()),
+                cache.get_previous_hash("another_hash").await.unwrap(),
+                "Existing but not updated value should not have been altered, content: {:#?}, now: {:?}",
+                local_storage_content(), now,
+            );
+            assert_eq!("updated_parent", updated_value.previous_hash);
+            assert!(updated_value.expire_at - now >= expiration_delay);
```
sfauvel, line 360: I'm not sure that we prove that the `expire_at` was updated here. 
We may check that the value has change.

mithril-client-wasm/src/client_wasm.rs
```diff
@@ -88,15 +91,64 @@ impl MithrilClient {
...
+        let certificate_verifier_cache = if client_options.unstable
+            && client_options.enable_certificate_chain_verification_cache
+        {
+            let cache = Self::build_certifier_cache(
+                aggregator_endpoint,
+                TimeDelta::seconds(
+                    client_options.certificate_chain_verification_cache_duration_in_seconds as i64,
+                ),
+            );
+            if let Some(cache) = &cache {
```
sfauvel, line 108: Isn't possible to make this treatment on the result variable `certificate_verifier_cache` ? 
We can then bring closer `client_builer` declaration and this treatment.

mithril-client/src/certificate_client/verify.rs
```diff
@@ -0,0 +1,495 @@
...
+                CertificateToVerify::Downloaded {
+                    certificate: certificate.clone().try_into()?,
+                },
+            )
+            .await?;
+
+        loop {
+            match current_certificate {
+                None => break,
+                Some(next) => {
```
sfauvel, line 212: No need to ensure the absence of infinite loop ?

mithril-client/src/client.rs
```diff
@@ -20,8 +22,13 @@ use crate::snapshot_client::SnapshotClient;
 use crate::snapshot_downloader::{HttpSnapshotDownloader, SnapshotDownloader};
 use crate::MithrilResult;
 
+#[cfg(target_family = "wasm")]
+const fn certificate_chain_verification_cache_duration_in_seconds_default() -> u32 {
+    604800
```
sfauvel, line 27: We may make more explicit that it is 1 week.

sfauvel (approved):
LGTM

mithril-client-wasm/src/client_wasm.rs
Alenar: We could yes !
One thing that I wonder is: should the `with_certificate_verifier_cache` of the builder be able to take an option ? In a way we could think of this as a possibility to disable the cache at the client api level and this would hugely simplify the construction of the builder here (since it would avoid to have to store the builder in a variable).

mithril-client/src/certificate_client/verify.rs
Alenar: We did not check it either before (and don't check it fully in the common verifier either as we only check that a certificate don't loop to itself), but this also bother me.

mithril-client-wasm/src/certificate_verification_cache.rs
```diff
@@ -0,0 +1,441 @@
...
+    }
+}
+
+#[cfg(test)]
+pub(crate) mod test_tools {
+    use std::collections::HashMap;
+
+    use super::*;
+
+    /// `Test only` Return the raw content of the local storage
```
dlachaume, line 165: Are `Test only` comments here and below mandatory since there's a `#[cfg(test)]` on the module?
```suggestion
    /// Return the raw content of the local storage
```

mithril-client-wasm/src/client_wasm.rs
```diff
@@ -88,15 +91,64 @@ impl MithrilClient {
...
+    }
+
+    fn build_certifier_cache(
+        aggregator_endpoint: &str,
+        expiration_delay: TimeDelta,
+    ) -> Option<Arc<dyn CertificateVerifierCache>> {
+        if web_sys::window().is_none() {
+            web_sys::console::warn_1(
+                &"Can't enable certificate chain verification cache: window object is not available\
+                    (are you running in a browser environment ?)"
```
dlachaume, line 135:
```suggestion
                    (are you running in a browser environment?)"
```

dlachaume (approved):
LGTM 👍
Just some minor suggestions.

mithril-client-wasm/src/client_wasm.rs
```diff
@@ -88,15 +91,64 @@ impl MithrilClient {
...
+                    .into(),
+            );
+            return None;
+        }
+
+        web_sys::console::warn_1(
+            &"Certificate chain verification cache is enabled.\n\
+            This feature is experimental and will be heavily modified or removed in the \
+            future as its security implications are not fully understood."
```
jpraynaud, line 144:
```suggestion
            &"Danger: the certificate chain verification cache is enabled.\n\
            This feature is highly experimental and insecure, and it must not be used in production."
```

mithril-client/src/certificate_client/verify.rs
jpraynaud: Good point. Let's do that in another PR :+1:

mithril-client/src/feedback.rs
```diff
@@ -98,6 +98,13 @@ pub enum MithrilEvent {
         /// The validated certificate hash
         certificate_hash: String,
     },
+    /// A individual certificate of a chain have been validated.
+    CertificateValidatedFromCache {
```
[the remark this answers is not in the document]
Alenar: :+1:, renamed to `CertificateFetchedFromCache`.

mithril-client/src/certificate_client/verify_cache/memory_cache.rs
```diff
@@ -0,0 +1,337 @@
...
+use crate::certificate_client::CertificateVerifierCache;
+use crate::MithrilResult;
+
+pub type CertificateHash = str;
+pub type PreviousCertificateHash = str;
+
+/// A in-memory cache for the certificate verifier.
+pub struct MemoryCertificateVerifierCache {
+    expiration_delay: TimeDelta,
+    cache: RwLock<HashMap<String, CachedCertificate>>,
```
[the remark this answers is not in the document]
Alenar: The defined alias isn't on `String` but on `str`, that's why I did not reuse it here :sweat_smile:, I could add another alias but this would add confusion with the existing alias on `str`.

mithril-client/src/certificate_client/api.rs
```diff
@@ -0,0 +1,83 @@
...
+}
+
+#[cfg(feature = "unstable")]
+/// API that defines how to cache certificates validation results.
+#[cfg_attr(test, mockall::automock)]
+#[cfg_attr(target_family = "wasm", async_trait(?Send))]
+#[cfg_attr(not(target_family = "wasm"), async_trait)]
+pub trait CertificateVerifierCache: Sync + Send {
+    /// Store a validated certificate hash and its parent hash in the cache.
+    async fn store_validated_certificate(
```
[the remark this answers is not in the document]
Alenar: I will change the API to take a `entities::Certificate` so the name will be relevant (this will also fit with a refactor where we would move this system to common).

mithril-client/src/certificate_client/verify.rs
```diff
@@ -0,0 +1,495 @@
...
+            Ok(Some(CertificateToVerify::ToDownload {
+                hash: previous_hash,
+            }))
+        } else {
+            self.verify_not_cached_certificate(certificate_chain_validation_id, certificate)
+                .await
+        }
+    }
+
+    async fn verify_not_cached_certificate(
```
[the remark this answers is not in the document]
Alenar: Yes this notion is important, I may use a different name but the very purpose of this method is to not use the cache.

mithril-client/src/client.rs
Alenar: Renamed it to `one_week_in_seconds`

mithril-client-wasm/src/certificate_verification_cache.rs
Alenar: Renamed those variables to `start_time`.

mithril-client-wasm/src/certificate_verification_cache.rs
```diff
@@ -0,0 +1,441 @@
...
+            .map_err(|err| anyhow!("Error deserializing cached certificate: {err:?}"))
+    }
+
+    fn cache_key(&self, certificate_hash: &CertificateHash) -> String {
+        format!("{}{}", self.cache_key_prefix, certificate_hash)
+    }
+}
+
+fn open_local_storage() -> MithrilResult<Option<Storage>> {
+    let window = window()
```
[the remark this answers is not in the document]
Alenar: I changed this api to not return an option but instead raise an error if the local storage is missing. This should not be problematic since when building this cache we verify beforehand that the local storage is available and don't build the cache plus log an warning if it's not.

mithril-client-wasm/src/certificate_verification_cache.rs
Alenar: Even if you need to be in a `test` context, those methods are available at the crate level (as other tests can, and do, use some of them) so I put those as a reminder but I don't mind removing them.

mithril-client-wasm/src/certificate_verification_cache.rs
```diff
@@ -0,0 +1,441 @@
...
+                            Ok(None)
+                        } else {
+                            Ok(Some(cached.previous_hash))
+                        }
+                    }
+                    None => Ok(None),
+                }
+            }
+        }
+    }
```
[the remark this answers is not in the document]
Alenar: With the changes on the `local_storage` method this code was simplified.

mithril-client/src/certificate_client/verify_cache/memory_cache.rs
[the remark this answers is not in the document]
dlachaume: Alright👌! 
And what do you think about renaming `CachedCertificate` to `CachedParentCertificate`?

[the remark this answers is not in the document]
Alenar: I would rather not, I think that keeping it as a `CachedCertificate` with a `parent_hash` field would sustain changes more easily.
For example if we added the avk inside this structure that would be the avk of the certificate, not of its parent, so this naming would conflict with the new intent that we would like to express.

mithril-client/src/certificate_client/api.rs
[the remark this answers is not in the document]
Alenar: After trying this change It's more difficult that I expected. The `mithril-client` library don't expose directly or with an alias the `entities::Certificate`, so using it here would mean either:

- third party implementers must add a dependency to `mithril-common`, as we would need to do in `mithril-client-wasm`
- we expose the entity in the client lib, adding a confusion with the already exposed `CertificateMessage` (exposed with the `MithrilCertificate` alias)

Another option would be to add a new type like `CertificateToCache` which would contains the properties that a cache system implementers could use.

The best option in my opinion is to keep it as is (with an potential rename) or to add a new type.

mithril-client/src/certificate_client/verify_cache/memory_cache.rs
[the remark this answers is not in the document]
dlachaume: Okay, that works for me. Thanks for the explanation 😊.

mithril-client-wasm/src/certificate_verification_cache.rs
dlachaume: I'm fine with that. Let's keep it like this 👍.

mithril-common/src/test_utils/certificate_chain_builder.rs
```diff
@@ -82,6 +82,20 @@ impl<'a> CertificateChainBuilderContext<'a> {
...
+pub enum CertificateChainingMethod {
+    /// `default` Chain certificates to the 'master' certificate of the epoch or if it's the 'master'
+    /// certificate, chain it to the 'master' certificate of the previous epoch.
+    ///
+    /// The 'master' certificate of an epoch is the first certificate of the epoch.
+    #[default]
+    ToMasterCertificate,
+
+    /// Chain certificates to their direct predecessor.
+    ToDirectPredecessor,
```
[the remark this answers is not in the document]
Alenar: I renamed the variant to `Sequential`.

mithril-client/src/certificate_client/verify.rs
```diff
@@ -196,22 +204,37 @@ impl CertificateVerifier for MithrilCertificateVerifier {
...
-            .verify_not_cached_certificate(
-                &certificate_chain_validation_id,
-                CertificateToVerify::Downloaded {
-                    certificate: certificate.clone().try_into()?,
-                },
-            )
-            .await?;
+        // Validate certificates without cache until we cross an epoch boundary
+        // This is necessary to ensure that the AVK of the current epoch was included
+        // in a certificate in the previous epoch.
```
jpraynaud, line 209:
```suggestion
        // Validate certificates without cache until we cross an epoch boundary
        // This is necessary to ensure that the AVK chaining is correct
```

mithril-client/src/certificate_client/verify.rs
```diff
@@ -470,14 +520,14 @@ mod tests {
                 .build();
             let last_certificate_hash = chain.first().unwrap().hash.clone();
 
-            // All certificates are cached except the last and the genesis (we always fetch the both)
+            // All certificates are cached except the last two (to cross an epoch boundary) and the genesis (we always fetch the both)
```
jpraynaud, line 523:
```suggestion
            // All certificates are cached except the last two (to cross an epoch boundary) and the genesis
```

jpraynaud (approved):

dlachaume (approved):

CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ As a minor extension, we have adopted a slightly different versioning convention
 
 - Build and publish both a `stable` version (for release networks) and an `unstable` version (for testing networks) of the explorer.
 
+- **UNSTABLE** Implement a **EXPERIMENTAL** caching mechanism in client certificate verification to improve performance.
```
jpraynaud, line 14: I'm not convinced that we want this experimental feature to be in the CHANGELOG

Issue #2151, opened 2024-12-02
Implement artifacts builder for Incremental Cardano DB


jpraynaud:
## Why
We need to create artifacts for the Incremental Cardano Database certification. We want the artifact structure to be versatile enough to support new type of storage (e.g. BitTorrent, IPFS, ...)

## What
Implement the artifact builder for the Incremental Cardano Database with the target JSON structure:

```json
{
  "merkle_root": "0506e5e1f7df9314ebdbb532768a5addc2931004c5cd8fd8cd464090f95e46e4",
  "beacon": {
    "network": "mainnet",
    "epoch": 525,
    "immutable_file_number": 6550
  },
  "certificate_hash": "04219eb7a66c37f0d917160b8bdc617dc4ff911344f2e72bdb22c1316cbe8d71",
  "total_db_size_uncompressed": 52470081414,
  "created_at": "2024-12-02T04:36:40.811764301Z",
  "locations": {
    "digests": [
      {
        "type": "aggregator",
        "uri": "https://aggregator.release-mainnet.api.mithril.network/aggregator/artifact/cardano-database/digests"
      },
      {
        "type": "cloud_storage",
        "uri": "https://storage.googleapis.com/cdn.aggregator.release-mainnet.api.mithril.network/mainnet-e525-i6550.digests.tar.zst"
      }
    ],
    "immutables": [
      {
        "type": "cloud_storage",
        "uri": {
                "Template": "https://storage.googleapis.com/cdn.aggregator.release-mainnet.api.mithril.network/mainnet-e525-i6550.{immutable_file_number}.tar.zst"
        }
      },
      {
        "type": "bit_torrent",
        "uri": {
                "Template": "https://storage.googleapis.com/cdn.aggregator.release-mainnet.api.mithril.network/mainnet-e525-i6550.{immutable_file_number}.tar.zst"
        }
      }
    ],
    "ancillary": [
      {
        "type": "cloud_storage",
        "uri": "https://storage.googleapis.com/cdn.aggregator.release-mainnet.api.mithril.network/mainnet-e525.ancillary.tar.zst"
      }
    ]
  },
  "compression_algorithm": "zstandard",
  "cardano_node_version": "10.1.2"
}
```

And the target design:

![Image](https://github.com/user-attachments/assets/d586d695-39a1-4029-b19d-f68df72d0ba6)

## How

- [x] Scaffold the artifact builder for the Incremental Cardano Database with minimum implementation
[...]

PR #2189, opened 2024-12-20
feat: extend `Snapshotter` to archive only specific files and directories

dlachaume:
## Content

This PR includes an update to the `Snapshotter` that allows it to create archives from specific directories and files.
It’s a preparatory step for the creation of archives from ancillary files.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation

## Issue(s)

Relates to #2151

### DIFF
--- a/mithril-aggregator/src/snapshotter.rs
+++ b/mithril-aggregator/src/snapshotter.rs
@@ -3,7 +3,7 @@ use flate2::Compression;
 use flate2::{read::GzDecoder, write::GzEncoder};
 use slog::{info, warn, Logger};
 use std::fs::{self, File};
-use std::io::{self, Read, Seek, SeekFrom};
+use std::io::{self, Read, Seek, SeekFrom, Write};
 use std::path::{Path, PathBuf};
 use std::sync::RwLock;
 use tar::{Archive, Entry, EntryType};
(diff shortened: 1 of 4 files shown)


mithril-aggregator/src/snapshotter.rs
```diff
@@ -37,6 +44,64 @@ impl From<ZstandardCompressionParameters> for SnapshotterCompressionAlgorithm {
...
+            } else if entry_path.is_file() {
+                let mut file = File::open(entry_path.clone())?;
+                tar.append_file(entry, &mut file).with_context(|| {
+                    format!(
+                        "Can not add file: '{}' to the archive",
+                        entry_path.display()
+                    )
+                })?;
+            } else {
+                return Err(anyhow!(
```
jpraynaud, line 96: What would it be in that case? Do we want to raise an error or skip?

jpraynaud (approved):
LGTM :+1:

Alenar (approved):
LGTM :rocket: 

I had a sightly different idea regarding the snapshotter, but your design may work as well and is flexible enough to be adapted if needed.

_What I had in mind was to let the snapshotter compute itself the subset of files that must be included in a ancillary archive, the only parameter to pass would be the immutable to include._

mithril-aggregator/src/snapshotter.rs
dlachaume: It seemed more coherent to us to leave the responsibility for handling errors to the caller.

jpraynaud: So if any of the entries is neither a file nor a directory (then what would it be?), the archive creation is doomed.
Is this really the behavior that we expect?

dlachaume: The error message has been reworded to be more relevant: https://github.com/input-output-hk/mithril/pull/2189/files#diff-54bfe4d87d887f1cae0f5ce311f07e0bfb726b285085fbea640d9689c48149c4R97

PR #2198, opened 2025-01-03
Fix: `CardanoDatabase` artifact identifier collisions

### LINKED ISSUE #2197: `CardanoDatabase` artifact identifiers have collisions
## Why
We have noticed that the `CardanoDatabase` artifact can have collisions on their identifiers. This happens when the same immutable file number is snapshot on two different epochs. This is also a source of flakiness in the CI.

## What
Replace the identifier of the `CardanoDatabase` artifact to a `hash` computed from the `merkle_root` and the `epoch`

## How

- [x] Implement a new `hash` identifier for the `CardanoDatabase` artifact
- [x] Update the HTTP routes
- [x] Update the OpenAPI specs

jpraynaud:
## Content

This PR includes a **fix to avoid collisions of the `CardanoDatabase` artifacts** when a snapshot is taken for the same immutable file number at two different epochs.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Issue(s)
Closes #2197

### DIFF
--- a/mithril-common/src/entities/cardano_database.rs
+++ b/mithril-common/src/entities/cardano_database.rs
@@ -1,5 +1,6 @@
 use semver::Version;
 use serde::{Deserialize, Serialize};
+use sha2::{Digest, Sha256};
 use strum::EnumDiscriminants;
 
 use crate::{
@@ -10,6 +11,9 @@ use crate::{
 /// Cardano database snapshot.
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
 pub struct CardanoDatabaseSnapshot {
+    /// Unique hash of the Cardano database snapshot.
+    pub hash: String,
+
     /// Merkle root of the Cardano database snapshot.
     pub merkle_root: String,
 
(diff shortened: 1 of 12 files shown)


dlachaume (approved):
LGTM 👍

dlachaume (approved):

mithril-common/src/entities/cardano_database.rs
```diff
@@ -106,6 +122,109 @@ pub struct ArtifactsLocations {
...
+                merkle_root: "mk-root-456".to_string(),
+                beacon: CardanoDbBeacon::new(123, 98),
+                ..dummy()
+            }
+            .compute_hash()
+        );
+    }
+
+    #[test]
+    fn compute_hash_returns_different_hash_with_different_beacon() {
```
sfauvel, line 199: Return different hash with different beacon epoch but return same hash with different immutable file.
We may create 2 tests to give them a more explicit name

sfauvel (approved):
LGTM

mithril-common/src/entities/cardano_database.rs
jpraynaud: Done :+1:

PR #2236, opened 2025-01-20
Fix: support trailing whitespace in protocol key files

jpraynaud:
## Content

This PR includes a **fix to the deserialization of the protocol keys to support trailing whitespace in their files**.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

### DIFF
--- a/mithril-common/src/crypto_helper/types/protocol_key.rs
+++ b/mithril-common/src/crypto_helper/types/protocol_key.rs
@@ -50,7 +50,7 @@ where
 
     /// Create an instance from a JSON hex representation
     pub fn from_json_hex(hex_string: &str) -> StdResult<Self> {
-        let key = key_decode_hex::<T>(hex_string).with_context(|| {
+        let key = key_decode_hex::<T>(hex_string.trim()).with_context(|| {
             format!(
                 "Could not deserialize a ProtocolKey from JSON hex string. Inner key type: {}",
                 type_name::<T>()
@@ -218,6 +218,8 @@ macro_rules! impl_codec_and_type_conversions_for_protocol_key {
 
 #[cfg(test)]
 mod test {
+    use std::io::Write;
+
     use crate::{crypto_helper::ProtocolKey, test_utils::fake_keys};
     use mithril_stm::stm::StmVerificationKeyPoP;
     use serde::{Deserialize, Serialize};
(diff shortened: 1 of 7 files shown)


mithril-common/src/crypto_helper/types/protocol_key.rs
```diff
@@ -285,4 +287,25 @@ mod test {
 
         assert_eq!(expected_key, read_key);
     }
+
+    #[test]
+    fn can_read_a_verification_key_from_file_with_trailing_whitespaces() {
+        let expected_key: ProtocolKey<StmVerificationKeyPoP> = VERIFICATION_KEY.try_into().unwrap();
+        let key_path = std::env::temp_dir().join("can_read_and_write_to_file_a_verification_key");
```
dlachaume, line 294: Wouldn't it be more robust to use the dedicated temporary directory creation tooling?
```suggestion
        let key_path = TempDir::create("protocol_key", "can_read_a_verification_key_from_file_with_trailing_whitespaces");
```

But I see that it is not used in the previous tests, maybe it's not relevant.

mithril-common/src/crypto_helper/types/protocol_key.rs
```diff
@@ -285,4 +287,25 @@ mod test {
...
+
+        expected_key
+            .write_json_hex_to_file(&key_path)
+            .expect("Writing to file should not fail");
+        let mut key_file = std::fs::OpenOptions::new()
```
sfauvel, line 299: I find that there are a lot of error messages that are not directly related to the test and that make reading more complex.
I would have preferred something linke: 
```
    #[test]
    fn can_read_a_verification_key_from_file_with_trailing_whitespaces_2() {
        fn add_space_to_file(key_path: &Path) {
            let mut key_file = std::fs::OpenOptions::new()
                .append(true)
                .open(&key_path)
                .unwrap();
            key_file.write_all(b"\n").unwrap();
        }

        let expected_key: ProtocolKey<StmVerificationKeyPoP> = VERIFICATION_KEY.try_into().unwrap();
        let key_path = std::env::temp_dir().join("can_read_and_write_to_file_a_verification_key");

        expected_key.write_json_hex_to_file(&key_path).unwrap();
        add_space_to_file(&key_path).unwrap();
        let read_key =
            ProtocolKey::<StmVerificationKeyPoP>::read_json_hex_from_file(&key_path).unwrap();

        assert_eq!(expected_key, read_key);
    }
```
If something goes wrong, it may be less explicit, but the reason won't be related to what we're trying to test, so it doesn't seem worrying to me.

sfauvel (approved):
LGTM
Just a suggestion on test

mithril-common/src/crypto_helper/types/protocol_key.rs
```diff
@@ -50,7 +50,7 @@ where
 
     /// Create an instance from a JSON hex representation
     pub fn from_json_hex(hex_string: &str) -> StdResult<Self> {
-        let key = key_decode_hex::<T>(hex_string).with_context(|| {
+        let key = key_decode_hex::<T>(hex_string.trim()).with_context(|| {
```
Alenar, line 53: Is this a problem with this method of with `key_decode_hex` ?

mithril-common/src/crypto_helper/types/protocol_key.rs
```diff
@@ -285,4 +287,25 @@ mod test {
 
         assert_eq!(expected_key, read_key);
     }
+
+    #[test]
+    fn can_read_a_verification_key_from_file_with_trailing_whitespaces() {
```
Alenar, line 292: Since the fix doesn't involve files in any-ways should the test use files ?

mithril-common/src/crypto_helper/types/protocol_key.rs
jpraynaud: Done, also for remaining tests not using the correct tooling

mithril-common/src/crypto_helper/types/protocol_key.rs
jpraynaud: Good idea, done

mithril-common/src/crypto_helper/types/protocol_key.rs
jpraynaud: The problem occurred when creating a key file with a copy paste, this is why it's tested here.

mithril-common/src/crypto_helper/types/protocol_key.rs
jpraynaud: You're right, I fixed the problem in the 'key_decode_hex' function instead :+1:

Alenar (approved):

PR #2241, opened 2025-01-22
refactor(aggregator): better url sanitization process by using a value object

Alenar:
## Content

This PR introduce a `SanitizedUrlWithTrailingSlash` struct to avoid redundant call to the url sanitizer as it guarantee that the url that it hold is already sanitized.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

### DIFF
--- a/mithril-aggregator/src/tools/url_sanitizer.rs
+++ b/mithril-aggregator/src/tools/url_sanitizer.rs
@@ -1,10 +1,70 @@
 use anyhow::{anyhow, Context};
 use reqwest::Url;
+use std::fmt::{Display, Formatter};
+use std::ops::Deref;
 
 use mithril_common::StdResult;
 
+/// A sanitized URL, guaranteed to have a trailing slash and no empty segments
+///
+/// This type is meant to be used as a base path to produce resources path, for example:
+/// `https://example.xy/download/` can be joined with `artifact/file.zip` to produce a download link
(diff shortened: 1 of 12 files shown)


jpraynaud (approved):
LGTM :+1:

mithril-aggregator/src/tools/url_sanitizer.rs
```diff
@@ -1,10 +1,67 @@
...
 
 use mithril_common::StdResult;
 
+/// A sanitized URL, guaranteed to have a trailing slash and no empty segments
+#[derive(Debug, Clone, PartialOrd, PartialEq, Eq, Ord, Hash)]
+pub struct SanitizedUrl {
+    internal_url: Url,
+}
+
+impl SanitizedUrl {
```
sfauvel, line 14: This structure represent a Uri that is not a resource. 
We may named it to reflect that. 
`SanitizedUrl` doesn't mean much. This gives the impression of having a valid Url

sfauvel (approved):
LGTM
It's a good thing to have a typed Url but the name should be improve

mithril-aggregator/src/tools/url_sanitizer.rs
Alenar: We share this concern but could not find a name that would fit that description.

Since we could not find anything better we renamed it to `SanitizedUrlWithTrailingSlash` with hope we found a better name in the future.

PR #2245, opened 2025-01-24
split snapshotter

Alenar:
## Content

This PR makes the snapshotter code more manageable by:

- moving it to services so it's no longer uncategorized at the source route
- split the module into several files to limit amount of code to navigate and make it more cohesive.

This PR also simplify local uploaders constructors by removing their StdResult (no longer needed since #2241)

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested


mithril-aggregator/src/services/snapshotter/appender.rs
```diff
@@ -0,0 +1,67 @@
...
+/// Define multiple ways to append content to a tar archive.
+pub(super) trait TarAppender {
+    fn append<T: Write>(&self, tar: &mut tar::Builder<T>) -> StdResult<()>;
+}
+
+pub(super) struct AppenderDirAll {
+    pub db_directory: PathBuf,
+}
+
+impl TarAppender for AppenderDirAll {
```
sfauvel, line 19: We may move some tests to this file, especially those with particular cases (same files, sub folder in a folder already added, ...)

sfauvel (approved):
LGTM
Just one remark on the file that contain tests for TarAppender

jpraynaud (approved):
LGTM :+1:

Issue #2234, opened 2025-01-20
Incremental Cardano DB artifacts production enhancements


jpraynaud:
## Why
We have noticed few optimizations and enhancements that need to be addressed with the artifacts produced for the Incremental Cardano DB.

## What
Implement the optimizations and enhancements:
- Ancillary archive file should be deleted after upload sequence
- URLs are not correctly formed in the artifact locations
- Immutable files archives are removed at node startup

## How

- [x] Delete the ancillary archive after the upload sequence of the ancillary builder
- [x] URLs are not correctly formed in the artifact locations
  - [x] Address is incorrect for the template of the immutable archives in the local uploader (should be `http://0.0.0.0:8080/aggregator/cardano_database_download/immutable/{immutable_file_number}.tar.zst`)
  - [x] Same for the ancillary archives
  - [x] Same for the digests archives
  - [x] The Cardano immutable files full and Cardano Database files should be in separate directories (`cardano-immutable-files-full` and `cardano-database`)
  - [x] Rename the `get_server_url` in the configuration of the aggregator to `get_local_server_url`
  - [x] Add a new optional configuration (with fall back on the `get_server_url`) in the aggregator for the (external) aggregator endpoint url which is used to create the url of the digests route of the aggregator in the digests builder.
  - [x] Adapt the infrastructure to provide this new configuration
  - [x] Adapt the documentation with this new configuration
- [x] Immutable files archives are removed at node startup (Pending snapshot folder is completely cleared at node startup when immutables files should be kept )
  - [x] Move files to a `final` subfolder of `snapshot_directory`
  - [x] Artifact builder verifies if the archive exists in the final folder. If not, ask the snapshotter to create the archive
  - [x] Expose the files of the final folder with the file uploader
- [x] Activate the `CardanoDatabase` certification in `testing-preview`

PR #2256, opened 2025-01-28
Permanent storage of immutables trio archives

Alenar:
## Content

This PR makes permanent the storages of immutable trio archives, created by the `ImmutableArtifactBuilder` (part of the `CardanoDatabaseArtifactBuilder`).
Previously they were deleted at each startup since they were stored in the ongoing snapshot directory of the snapshotter.

### Details

- feat: Add a `immutables_storage_dir` to the `ImmutableArtifactBuilder`
  - In production this directory target  `{snapshot_directory}/cardano_database/immutable`, said `{snapshot_directory}/cardano_database` is exposed for file download on the `/cardano-database-download` route
- feat: Move archives created by the snapshotter used by the `ImmutableArtifactBuilder` to the `immutables_storage_dir`
- feat: Add a capacity to `LocalUploader` to only return the location without doing file move to avoid double move.
- test: Add a `FakeSnapshotter` that creates real but empty files in a test directory that it control, usefull to tests behavior files operations are executed on files returned by a snapshotter (a thing that the existing `DumbSnapshotter` don't handle).
- test: fix creations of multiples sub-directories in the `mithril-aggregator` crate directory when running tests.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Issue(s)

Relates to #2234


mithril-aggregator/src/artifact_builder/cardano_database_artifacts/immutable.rs
```diff
@@ -558,46 +533,53 @@ mod tests {
...
+            fs::File::create(work_dir.join(file_name)).unwrap();
+
+            assert_eq!(
+                builder.retrieve_existing_snapshot_archive(file_name),
+                Some(work_dir.join(file_name))
+            );
+        }
+
         #[test]
         fn return_all_archives_but_not_rebuild_archives_already_compressed() {
```
jpraynaud, line 561: It looks like we don't test anymore if the existing archives are recreated in this test.

mithril-aggregator/src/services/snapshotter/test_doubles.rs
jpraynaud, line 1: This was not a problem to have them in separate modules IMHO

mithril-aggregator/src/configuration.rs
```diff
@@ -251,7 +251,9 @@ impl Configuration {
             public_server_url: None,
             run_interval: 5000,
             db_directory: PathBuf::new(),
-            snapshot_directory: PathBuf::new(),
+            // Note: this is a band-aid solution to avoid IO operations in the `mithril-aggregator`
+            // crate directory. This should be fixed in the future.
```
jpraynaud, line 255: Maybe you can add few details about the problems not solved (e.g. collisions)?

jpraynaud (approved):
LGTM :+1:

mithril-aggregator/src/services/snapshotter/test_doubles.rs
Alenar: We made it that way to avoid cluttering the production code with the tests tooling (as it would have taken a third of the directory space), plus even merged this file is still lean.

sfauvel (approved):
LGTM

mithril-aggregator/src/artifact_builder/cardano_database_artifacts/immutable.rs
Alenar: I updated the test so it writes a specific content to the files before running the builder then check that those contents did not change.

mithril-aggregator/src/configuration.rs
Alenar: done

PR #2257, opened 2025-01-28
Do not delete an already existing archive on error

### LINKED ISSUE #2215: Enhance Snapshotter to avoid file deletion on error if already exists
## Why
We need to support some edge case of the Snapshotter. Some existing files could be deleted or compromised if the snapshotting is asked multiple times on the same file.

## What
Implement a mechanism to replace an existing snapshotted file only if the snapshotting is successful.

## How

- [ ] Create the snapshot in a temporary file (same name with a `.tmp` suffix)
- [ ] If the snapshotting is successful rename the temporary file to the final file name otherwise delete the temporary file

sfauvel:
## Content

If an archive exists, it must not be deleted if an error occurs when trying to create an archive with the same name.
We replace an existing snapshotted file only if the snapshotting is successful.

This PR include:
- Creation of the snapshot in a temporary file (same name with a .tmp suffix)
- Renaming the temporary file to the final file name, if the snapshotting is successful 
- Deleting the temporary file in any case, even if this file already exist before snapshotting

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation

## Issue(s)

Closes #2215

### DIFF
--- a/mithril-aggregator/src/services/snapshotter/compressed_archive_snapshotter.rs
+++ b/mithril-aggregator/src/services/snapshotter/compressed_archive_snapshotter.rs
@@ -108,7 +108,7 @@ impl CompressedArchiveSnapshotter {
     fn snapshot<T: TarAppender>(&self, filepath: &Path, appender: T) -> StdResult<OngoingSnapshot> {
         let temporary_archive_path = self
             .ongoing_snapshot_directory
-            .join(Path::new(&format!("{}.tmp", filepath.display())));
+            .join(filepath.with_extension("tmp"));
         let archive_path = self.ongoing_snapshot_directory.join(filepath);
         if let Some(archive_dir) = archive_path.parent() {
             fs::create_dir_all(archive_dir).with_context(|| {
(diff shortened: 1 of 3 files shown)


mithril-aggregator/src/services/snapshotter/compressed_archive_snapshotter.rs
```diff
@@ -114,6 +114,8 @@ impl CompressedArchiveSnapshotter {
     }
 
     fn snapshot<T: TarAppender>(&self, filepath: &Path, appender: T) -> StdResult<OngoingSnapshot> {
+        let temporary_archive_path =
```
jpraynaud, line 117: Maybe this could be done more elegantly with `with_extension`?

jpraynaud (approved):
LGTM :+1:

mithril-aggregator/src/services/snapshotter/compressed_archive_snapshotter.rs
sfauvel: I tried to use it but `Path::with_extension` replace the extension by the new one, ("archive.tar.gz" become "archive.tar.tmp").
There is also a `Path::with_added_extension` that just add the extension but it's `unstable`

dlachaume (approved):
LGTM

Alenar (approved):

PR #2279, opened 2025-02-05
Chore: upgrade dependencies

Alenar:
## Content

This PR **upgrades the repository dependencies to their latest version** by running the [upgrade script from the runbook](https://github.com/input-output-hk/mithril/tree/main/docs/runbook/upgrade-repository-dependencies).

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Comments

Some problems had to be fixed:

- Some makefiles recipes had their name changed, change that was not reflected in the upgrade script
- `mithril-client` failed to compile to wasm: as its `uuid` dependency now requires its `js` features for wasm (this features already existed but we could compile to wasm without it before, that's not the case anymore)

One problem is remaining:
`rand_core` released it's [new major version last week](https://github.com/rust-random/rand/releases/tag/0.9.0) (version 0.9) but it contains breaking changes to their apis, changes that are yet to be reflected on another of our dependency ([ed25519-dalek](https://crates.io/crates/ed25519-dalek)).
Since this incompatibility can't be resolved yet, the upgrade script was run with the `--compatible` option in order to keep the current version of `rand_core` (version `0.6.4`).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -59,11 +59,11 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e89da841a80418a9b391ebaea17f5c112ffaaa96f621d2c285b5174da76b9011"
 dependencies = [
  "cfg-if",
- "getrandom",
+ "getrandom 0.2.15",
  "once_cell",
  "serde",
  "version_check",
- "zerocopy",
+ "zerocopy 0.7.35",
 ]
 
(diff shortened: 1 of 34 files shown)


sfauvel (approved):

jpraynaud (approved):
LGTM

Issue #2222, opened 2025-01-13
Flakiness in e2e tests in CI


jpraynaud:
## Why
We have noticed some flakiness in the e2e tests in the CI which fails quite often.

### Minimum expected mithril stake distribution epoch not reached : XX < YY
- https://github.com/input-output-hk/mithril/actions/runs/12159186537/job/33909010778
- https://github.com/input-output-hk/mithril/actions/runs/12159406584/job/33910097011#step:6:589
- https://github.com/input-output-hk/mithril/actions/runs/12398254570/job/34610766498#step:6:661
- https://github.com/input-output-hk/mithril/actions/runs/12398387893/job/34611232337#step:6:665
- https://github.com/input-output-hk/mithril/actions/runs/12398387893/job/34612511048#step:6:948
- https://github.com/input-output-hk/mithril/actions/runs/12668778598/job/35305158660#step:6:276
- https://github.com/input-output-hk/mithril/actions/runs/12706153806/job/35419162536#step:6:420
- https://github.com/input-output-hk/mithril/actions/runs/12751941085/job/35540460712#step:6:513
- https://github.com/input-output-hk/mithril/actions/runs/12752388690/job/35542004023#step:6:402
- https://github.com/input-output-hk/mithril/actions/runs/12764751348/job/35577771541#step:6:329
- https://github.com/input-output-hk/mithril/actions/runs/12793153128/job/35665568667#step:6:369
- https://github.com/input-output-hk/mithril/actions/runs/12873215847/job/35890464988#step:6:608
- https://github.com/input-output-hk/mithril/actions/runs/12873247216/job/35890553161#step:6:867
- https://github.com/input-output-hk/mithril/actions/runs/13068848697/job/36466494633#step:6:834

### `mithril-aggregator genesis bootstrap` exited with code: 1 
- https://github.com/input-output-hk/mithril/actions/runs/12140713467/job/33851741099#step:6:991
- https://github.com/input-output-hk/mithril/actions/runs/12178789405/job/33970064206#step:6:533
- https://github.com/input-output-hk/mithril/actions/runs/12231609594/job/34115506322#step:6:583
- https://github.com/input-output-hk/mithril/actions/runs/12871164504/job/35884122045#step:6:548
- https://github.com/input-output-hk/mithril/actions/runs/12873215847/job/35890464322#step:6:552
- https://github.com/input-output-hk/mithril/actions/runs/12906035647/job/35986780846#step:6:554

## What
[...]

PR #2303, opened 2025-02-12
Fix mithril-aggregator genesis bootstrap flakiness in e2e tests

Alenar:
## Content

This PR fix a flakiness in mithril-end-to-end tests that manifest with a failure when bootstrapping the genesis certificate.

#### Summary

In a successful run:

1.  the stake distribution that we try to fetch (for the actual epoch, without offset) isn't immediately available since when we register them we do so with an offset of +1.
2. we don't see stake distribution unavailability errors because ... the epoch settings also isn't available because  the ones we are fetching in `EpochService::inform_epoch` are with offset -1, 0, +1 and the one we register in `handle_discrepancies_at_startup` are with offset 0, 1, 2.
3. the errors solve themself after epoch change, then the stake distribution registered in the previous epoch is now available and the epoch settings registered by `handle_discrepancies_at_startup` too.
4. The state machine will still have it cycles fails but after the call to `update_epoch_settings`, allowing the genesis bootstrap to run since it was what was missing.

In failing run:
1. There's an epoch change between `handle_discrepancies_at_startup` and the first state machine cycle
2. This time there's the three epoch settings available because of the epoch change
3. But there's still no stake distribution available for the actual epoch so `EpochService::inform_epoch` will fail, making it missing calls to `update_epoch_settings`.
4. After an epoch change the epoch settings for epoch +1, needed by `EpochService::inform_epoch`, is not available since `update_epoch_settings` was not called, making each subsequent cycles fails.

#### Details
<details><summary>Details</summary>
<p>

> :notebook: **Note** 
> Logs extracted from the following failed run: https://github.com/input-output-hk/mithril/actions/runs/12178789405/job/33970064206#step:6:533 

1. The genesis bootstrap fails with the following error:

```
Error: genesis-tools: initialization error

Caused by:
    Missing protocol parameters for epoch 10
``` 

2. The protocol parameters are part of the epoch settings that are inserted:
-  at startup when building the epoch settings store by running `handle_discrepancies_at_startup`: this fills epochs settings for the actual epoch plus the two following.
- by the state machine `update_epoch_settings` when transitioning from `idle` to `ready` state after calling to `EpochService::inform_epoch`

3. We can see epoch settings inserted by `handle_discrepancies_at_startup` in the database at epoch 5, 6 and 7 plus 10, 11, and 12 (when the bootstrap command was run). This means that `handle_discrepancies_at_startup` was run at epoch 5 and 10.

4. In the logs we can see that between the run of `handle_discrepancies_at_startup` at epoch 5 and the first state machine cycle, the epoch changed.

5. When reading the aggregator logs we see that the state machine got **every cycle for the epoch 6** an error from `EpochService::inform_epoch`:
```
Epoch service failed to obtain the stake distribution for epoch: 6

Caused by:
    The stake distribution for epoch Epoch(6) is not available.
``` 

6. This error is raised at the end of `EpochService::inform_epoch` when it tries to obtains the total cardano spo and stakes which is based on stakes distribution stored previously.
In turn the state machine never get to `update_epoch_settings` since this call is done after `inform_epoch`.

7. As a consequence of being unable to run `update_epoch_settings` the error raised at each cycle change after epoch change (to epoch 7). It fails earlier in the function when trying to obtains the epoch settings for signer registration epoch (+1):
```
Epoch service could not obtain signer registration epoch settings for epoch 8

Stack backtrace:
    0: anyhow::error::<impl core::convert::From<E> for anyhow::Error>::from
    1: mithril_aggregator::services::epoch_service::MithrilEpochService::get_epoch_settings::{{closure}}
``` 
Afterward the state machine cycle will always fails for this same reason and won't be able to be ready for genesis bootstrap.
</p>
</details> 

#### Unavailability of the stake distribution
 
> [!IMPORTANT]
> There's a missing offset of -1 (signer retrieval epoch) when fetching the stake distribution for total spo and stake in epoch service.
> This is because the stake distribution that should be used is the one associated with the **current** signers (those who can sign), which use that offset.

This is normal, when we first start an aggregator the stake distribution that we register won't be usable before two epochs.
This means that we can't reliably provide total spo/stake in the epoch service, those have to be `Option` to take in account this gap when there's no stake distribution usable.

Making this change solve the problems since the `inform_epoch` won't fail anymore, but the aggregator status route have to be adapted.

####  Unavailability of the epoch settings

This is a minor issue as this don't stop the aggregator from bootstrapping a genesis certificate as long as we wait for one epoch after start.
But this makes the signers calls to `/epoch-settings` fails, delaying when they can register and filling their logs with errors.

By also filling epoch settings with the retrieval offset in `handle_discrepancies_at_startup` we can avoid this problem. Also we should fills epoch settings for 4 epochs to avoid the edge case when there's a epoch change between `handle_discrepancies_at_startup` and the first state machine cycle.

### Changes

For `mithril-aggregator`:

- `EpochService`: 
  - fix incorrect epoch for fetching the stake distribution used to compute total cardano stake and spo
  - total cardano stake and spo are now `Option` since they are not available before two epochs when starting an aggregator for the first time
- `/status` route: use 0 for  total cardano stake and spo when the values returned by the Epoch Service are none
- `handle_discrepancies_at_startup`:
  - called with an offset of `-1` so data are available for `EpochService::inform_epoch`
  - register epoch settings for `4` epochs instead of `3` so `EpochService::inform_epoch` won't fails if there's an epoch change between calls to `handle_discrepancies_at_startup` in the aggregator builder and the first state machine cycle.

For `mithril-explorer`:
- Properly handle a total of 0 when computing percentages.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Issue(s)

Relates to #2222

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3642,7 +3642,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-aggregator"
-version = "0.7.2"
+version = "0.7.3"
 dependencies = [
  "anyhow",
  "async-trait",
--- a/mithril-aggregator/Cargo.toml
+++ b/mithril-aggregator/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mithril-aggregator"
-version = "0.7.2"
+version = "0.7.3"
 description = "A Mithril Aggregator server"
 authors = { workspace = true }
 edition = { workspace = true }
--- a/mithril-aggregator/src/dependency_injection/builder/enablers/epoch.rs
+++ b/mithril-aggregator/src/dependency_injection/builder/enablers/epoch.rs
@@ -10,7 +10,7 @@ impl DependenciesBuilder {
         let epoch_settings_storer = self.get_epoch_settings_store().await?;
         let chain_observer = self.get_chain_observer().await?;
         let era_checker = self.get_era_checker().await?;
-        let stake_distribution_service = self.get_stake_distribution_service().await?;
+        let stake_store = self.get_stake_store().await?;
         let epoch_settings = self.configuration.get_epoch_settings_configuration();
         let allowed_discriminants = self
             .configuration
@@ -23,7 +23,7 @@ impl DependenciesBuilder {
                 verification_key_store,
                 chain_observer,
                 era_checker,
-                stake_distribution_service,
+                stake_store,
             ),
             allowed_discriminants,
             self.root_logger(),
--- a/mithril-aggregator/src/dependency_injection/builder/support/stores.rs
+++ b/mithril-aggregator/src/dependency_injection/builder/support/stores.rs
@@ -120,6 +120,12 @@ impl DependenciesBuilder {
                 message: "cannot build aggregator runner: no epoch returned.".to_string(),
                 error: None,
             })?;
+        let retrieval_epoch = current_epoch
+            .offset_to_signer_retrieval_epoch()
+            .map_err(|e| DependenciesBuilderError::Initialization {
+                message: format!("cannot create aggregator runner: failed to offset current epoch '{current_epoch}' to signer retrieval epoch."),
+                error: Some(e.into()),
+            })?;
 
         {
             // Temporary fix, should be removed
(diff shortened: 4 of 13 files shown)


jpraynaud (approved):
LGTM :fire:

PR #2322, opened 2025-02-19
Fix BLST bindings: Error handling for infinite values of sigs and vks

### LINKED ISSUE #2321: BLST Rust bindings: Aggregation fails if vk or sig lists include identity elements
## Context & versions
In the C implementation of the pippenger function, [it](https://github.com/supranational/blst/blob/165ec77634495175aefd045a48d3469af6950ea4/src/multi_scalar.c#L11C32-L11C37) mentions that;
```
/*
 * Infinite point among inputs would be devastating. Shall we change it?
 */
```
Meaning that C function breaks if you provide the identity of the group among the inputs. It does not make sense to add it, as the identity scaled and added will not alter the answer, but it does imply that this function is unsafe without a proper check that the inputs do not include zero.
Now, the weird thing is, the rust bindings for blst do not make this check.
And indeed, if you add [this](https://github.com/perturbing/bench-msm/blob/perturbing/test-msm-inf-point/src/main.rs) identity to the inputs, the result of running this little test is that the msm indeed returns infinity and not the expected unchanged non-zero point:

## Steps to reproduce
[These lines in /mithril-stm/src/multi_sig.rs](https://github.com/input-output-hk/mithril/blob/main/mithril-stm/src/multi_sig.rs#L382-L383) are impacted:
```
let grouped_vks = p2_affines::from(transmuted_vks.as_slice());
let grouped_sigs = p1_affines::from(transmuted_sigs.as_slice());
```

If the `transmuted_{vks/sigs}` contains the identity element, both `p2_affines::from()` and `p1_affines::from()` will return the identity element (which may trivially satisfy conditions like signatures checks).

## Actual behavior
If the `transmuted_{vks/sigs}` contains the identity element, both `p2_affines::from()` and `p1_affines::from()` convert all elements in the lists to identity elements.

## Expected behavior
There should be no identity element in the signature and verification key list. Even if there is any, functions should not convert all elements to identity elements.

curiecrypt:
## Content
If there is any identity element in the following vectors:
```
let transmuted_vks: Vec<blst_p2> = vks.iter().map(vk_from_p2_affine).collect();
let transmuted_sigs: Vec<blst_p1> = signatures.iter().map(sig_to_p1).collect();
```
The content of the following vectors in `mithril-stm/src/multi_sig.rs`:
```
let grouped_vks = p2_affines::from(transmuted_vks.as_slice());
let grouped_sigs = p1_affines::from(transmuted_sigs.as_slice());
```
Becomes vectors full of identity elements.

This PR includes the changes to avoid having an identity element in signature and verification lists.

In `mithril-stm/src/multi_sig.rs`:
- `Signature::verify` function is updated: If signature is an infinity value, it returns an error.
- `VerificationKeyPoP::check` function is updated: If the verification key is an infinity value, it returns an error.
- `test_infinity_sig` test is added.
- `test_infinity_vk` test is added.
- `test_keyreg_with_infinity_vk` test is added.

In `mithril-stm/src/error.rs`:
- `MultiSignatureError` is updated to cover 
    - `SignatureInfinity`
    - `VerificationKeyInfinity`
- `impl From<MultiSignatureError> for StmSignatureError` is updated.
- `impl<D: Digest + FixedOutput> From<MultiSignatureError> for StmAggregateSignatureError<D>` is updated.
- `impl From<MultiSignatureError> for CoreVerifierError` is updated.
- `pub(crate) fn blst_err_to_mithril` is updated.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [ ] Reviewer requested
- Documentation

## Issue(s)
Closes #2321

### DIFF
--- a/mithril-stm/src/multi_sig.rs
+++ b/mithril-stm/src/multi_sig.rs
@@ -654,7 +654,9 @@ mod tests {
             let sk = SigningKey::gen(&mut ChaCha20Rng::from_seed(seed));
             let vk = VerificationKey::from(&sk);
             let sig = sk.sign(&msg);
-            assert!(sig.verify(&msg, &vk).is_ok());
+
+            let result = sig.verify(&msg, &vk);
+            assert!(result.is_ok(), "verify {result:?}");
         }
 
         #[test]
@@ -664,7 +666,9 @@ mod tests {
             let vk1 = VerificationKey::from(&sk1);
             let sk2 = SigningKey::gen(&mut rng);
             let fake_sig = sk2.sign(&msg);
-            assert!(fake_sig.verify(&msg, &vk1).is_err());
+
+            let result = fake_sig.verify(&msg, &vk1);
+            assert_eq!(result, Err(MultiSignatureError::SignatureInvalid(fake_sig)));
         }
 
         #[test]
@@ -732,7 +736,8 @@ mod tests {
                 mvks.push(vk);
             }
 
-            assert!(Signature::verify_aggregate(&msg, &mvks, &sigs).is_ok());
+            let result = Signature::verify_aggregate(&msg, &mvks, &sigs);
+            assert!(result.is_ok(), "Aggregate verification failed {result:?}");
         }
 
         #[test]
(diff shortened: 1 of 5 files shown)


mithril-stm/src/multi_sig.rs
```diff
@@ -748,14 +753,32 @@ mod tests {
...
             let vk = VerificationKey::from(&sk);
             let vk_bytes = vk.to_bytes();
-            let vk2 = VerificationKey::from_bytes(&vk_bytes).unwrap();
-            assert_eq!(vk, vk2);
+            let result = VerificationKey::from_bytes(&vk_bytes);
+            match result {
+                Ok(vk_from_bytes) => {
+                    assert_eq!(vk, vk_from_bytes);
+                }
+                Err(MultiSignatureError::SerializationError) => {
```
jpraynaud, line 764: I'm wondering why there is this new case that is tested as the `SerializationError` already existed before this PR? I have the same question for all the other modified tests below.

mithril-stm/src/multi_sig.rs
```diff
@@ -749,13 +811,23 @@ mod tests {
...
+                    }
+                    Err(MultiSignatureError::AggregateSignatureInvalid) => {
+                        println!("Aggregation failed.");
+                    }
+                    _ => unreachable!(),
+                }
             }
-            assert!(Signature::batch_verify_aggregates(&batch_msgs, &batch_vk, &batch_sig).is_ok());
+            let batch_verify_aggregates = Signature::batch_verify_aggregates(&batch_msgs, &batch_vk, &batch_sig);
+            assert!(batch_verify_aggregates.is_ok(), "{batch_verify_aggregates:?}");
```
jpraynaud, line 830: This should be also rolled back as we have discussed.

mithril-stm/src/multi_sig.rs
```diff
@@ -287,10 +291,14 @@ impl From<&SigningKey> for ProofOfPossession {
...
-            Some(*self),
-        )
+        match self.0.validate(true) {
+            Ok(_) => blst_err_to_mithril(
+                self.0.verify(false, msg, &[], &[], &mvk.0, false),
+                Some(*self),
+                None,
+            ),
+            Err(e) => blst_err_to_mithril(e, Some(*self), None),
+        }
```
jpraynaud, line 301: Maybe this would be more readable:
```suggestion
        blst_err_to_mithril(
            self.0.validate(true).map_or_else(
                |e| e,
                |_| self.0.verify(false, msg, &[], &[], &mvk.0, false),
            ),
            Some(*self),
            None,
        )
```

mithril-stm/src/multi_sig.rs
```diff
@@ -643,19 +654,69 @@ mod tests {
             let sk = SigningKey::gen(&mut ChaCha20Rng::from_seed(seed));
             let vk = VerificationKey::from(&sk);
             let sig = sk.sign(&msg);
-            assert!(sig.verify(&msg, &vk).is_ok());
+
+            let result = sig.verify(&msg, &vk);
+            assert!(result.is_ok(), "verify {result:?}");
```
jpraynaud, line 659: This should be enough to see that the result is `Ok`. If not the `unwrap` will help identify the triggered error:
```suggestion
            sig.verify(&msg, &vk).unwrap();
```

jpraynaud (approved):
LGTM :+1:

Issue #2342, opened 2025-02-26
Create a `Code ADR` record in the repository


jpraynaud:
## Why
We want to offload implementation related decisions from the website to the repository with a new `ADR.md` file at its root. This will facilitate the understanding of some portions of the code, to get an history of specific decisions that were made and why they were made.

## What
Implement a "Code ADR" in the repository.

## How

- [x] Create an ADR that explains the decision to create a Code ADR
- [x] Create a skeleton for the `DEV-ADR.md` file in the repository
- [x] Create the first entry of the `DEV-ADR.md` file with the fix of #2325
- [x] Update the PR template with checkboxes in 'Documentation' for:
  - New Dev ADR created
  - No new TODO introduced

PR #2343, opened 2025-02-26
Code ADR for mithril-common

Alenar:
## Content

This PR add a "code ADR" file to the `mithril-common` project in order to details project specific decisions that would be to specific to fit our website ADR blog.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Issue(s)

Relates to #2342

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -13,7 +13,7 @@ This PR includes...
   - [ ] Commit sequence broadly makes sense
   - [ ] Key commits have useful messages
 - PR
-  - [ ] No clippy warnings in the CI
+  - [ ] All check jobs of the CI have succeeded
   - [ ] Self-reviewed the diff
   - [ ] Useful pull request description
   - [ ] Reviewer requested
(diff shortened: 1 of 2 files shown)


mithril-common/ADR.md
```diff
@@ -0,0 +1,61 @@
+# Mithril Common Architecture Decision Records
+
+## 1. Record Architecture Decisions
+
+date: 2025-02-26
+status: Accepted
```
sfauvel, line 6: Is it really useful to have a status ? 
Will there ever be anything other than `Accepted` ?

mithril-common/ADR.md
```diff
@@ -0,0 +1,61 @@
...
+
+To keep things simple, we will store these ADRs in a single file, and use a simple format to keep them readable.
+
+### Consequences
+
+See Michael Nygard's article, linked above.
+
+---
+
+## 2. Remove Artifacts serialization support when compiling to WebAssembly
```
sfauvel, line 26: What do you think about reverse order and add new decisions at the beginning ?

sfauvel (approved):

mithril-common/ADR.md
Alenar: I agree that it have a low value, the only one that I see is when an entry supersede a previous one as it allow us to mark the previous entry with a link to the new one.
I will keep it for now, we may remove them in the future if this case can be handled differently.

dlachaume (approved):

jpraynaud (approved):
LGTM :+1:

PR #2351, opened 2025-03-05
Feat: implement aggregator `slave` signer registration mode

### LINKED ISSUE #2334: Implement a `slave` mode for the signer registration in the aggregator
## Why
We want to implement a slave mode in the aggregator to support centralized signer registrations available to multiple aggregators.

## What
Implement a slave mode which does not accept signer registrations and retrieves them from the master aggregator of the network.

## How 
 
- [x] Add a master aggregator endpoint optional parameter to the aggregator
- [x] Create a a new trait `SignerRegistrationVerifier`:
  - [x] Which verifies that a signer registration message is valid
  - [x] Which is a dependency of the `MithrilSignerRegisterer`
- [x] The signer registration route of the aggregator is disabled in slave mode:
  - [x] Build a new `MithrilSignerRegistererSlave` and inject it in the DI (if the aggregator is slave or use the existing one)
  - [x] `MithrilSignerRegistererSlave` always returns an error on `register_signer` call
- [x] The slave aggregator pulls the next signer registrations from the epoch settings route of the master (handle unavailability of this information on the master):
  - [x] From the `IDLE` state, check that the master aggregator has changed of epoch before trying to transition to `READY` 
  - [x] In the transition attempt from `IDLE` to `READY`, fetch the signer registration from the epoch settings of the master:
    - [x] Use the `SignerRegistrationVerifier` to check that the fetched signer registrations are valid
    - [x] Record the `next_signers` in the signer registration store at `current_epoch - 1`

## Later
- Add a startup phase for slave:
    - Use the `current_signers` if there is no signer registration available for the epoch  at `current_epoch`
    - Fetch the certificate chain of the master

### LINKED ISSUE #2335: Add integration test for mode for the signer registration in the aggregator
## Why
We need to add an integration test implemented for the `slave` mode of the aggregator for signer registration.

## What
Implement an integration test for the `slave` mode of the aggregator for signer registration.

## How

- [x] Implement an integration test which verifies:
  - [x] that the signer registration correctly is fetched from a `master` aggregator
  - [x] that the `slave` aggregator can create a valid certificate chain?

jpraynaud:
## Content

This PR includes the implementation of a **`slave` signer registration mode in the aggregator**.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Issue(s)
Closes #2334
Closes #2335


mithril-aggregator/src/services/aggregator_client.rs
```diff
@@ -0,0 +1,518 @@
...
+    pub fn prepare_request_builder(&self, request_builder: RequestBuilder) -> RequestBuilder {
+        let request_builder = request_builder
+            .header(
+                MITHRIL_API_VERSION_HEADER,
+                self.api_version_provider
+                    .compute_current_version()
+                    .unwrap()
+                    .to_string(),
+            )
+            .header(MITHRIL_SIGNER_VERSION_HEADER, env!("CARGO_PKG_VERSION"));
```
Alenar, line 195: We should send another header or remove it.

mithril-aggregator/src/services/aggregator_client.rs
Alenar, line 1: Later, maybe we should think about a shared tool, wrapping reqwest to simplify http calls and error handling for our case. At least between signer & aggregator, maybe even client.

mithril-aggregator/src/runtime/runner.rs
```diff
@@ -264,6 +276,54 @@ impl AggregatorRunnerTrait for AggregatorRunner {
...
+            .ok_or(anyhow::anyhow!(
+                "Slave aggregator did not return any stake distribution"
+            ))?;
+        self.dependencies
+            .signer_synchronizer
+            .synchronize_signers(registration_epoch, &next_signers, &stake_distribution)
+            .await?;
+
+        Ok(())
+    }
```
Alenar, line 325: Would it be better to implement those two behaviours in the synchroniser directly ?

mithril-aggregator/src/dependency_injection/builder/mod.rs
```diff
@@ -147,8 +149,26 @@ pub struct DependenciesBuilder {
...
+    pub signer_synchronizer: Option<Arc<dyn SignerSynchronizer>>,
+
+    /// Signer registration verifier
+    pub signer_registration_verifier: Option<Arc<dyn SignerRegistrationVerifier>>,
+
+    /// Signer registration round opener service
+    pub signer_registration_round_opener: Option<Arc<dyn SignerRegistrationRoundOpener>>,
+
+    /// Signer registration pruning task
+    pub signer_registration_pruning_task: Option<Arc<dyn EpochPruningTask>>,
```
Alenar, line 171: Would it be better to not store the pruning task in the builder but instead make `get_signer_registration_pruning_task` return the correct task source directly ? They should already be stored in another field of the builder anyway.

mithril-aggregator/src/commands/serve_command.rs
```diff
@@ -53,6 +53,14 @@ pub struct ServeCommand {
...
     #[clap(long)]
     metrics_server_port: Option<u16>,
+
+    /// Master aggregator endpoint
+    ///
+    /// This is the endpoint of the aggregator that will be used to fetch the latest epoch settings
+    /// and store the signer registrations when the aggregator is running in a slave mode.
+    /// If this is not set, the aggregator will run in a master mode.
+    #[clap(long)]
+    pub master_aggregator_endpoint: Option<String>,
```
Alenar, line 63: does this field needs to be `pub` ? I don't even remember why there's some pub fields in this structure.

mithril-aggregator/src/dependency_injection/builder/enablers/misc.rs
```diff
@@ -51,4 +56,31 @@ impl DependenciesBuilder {
 
         Ok(self.message_service.as_ref().cloned().unwrap())
     }
+
+    /// build an [AggregatorClient]
+    pub async fn build_master_aggregator_client(&mut self) -> Result<Arc<dyn AggregatorClient>> {
+        let master_aggregator_endpoint = &self
+            .configuration
+            .master_aggregator_endpoint
+            .clone()
```
Alenar, line 65: Maybe using `cloned` would avoid the needs to use `to_owned` bellow.

mithril-aggregator/src/signer_registration/slave.rs
```diff
@@ -0,0 +1,345 @@
...
+        let stake_distribution = fixture.stake_distribution();
+
+        signer_registerer
+            .synchronize_signers(registration_epoch, &signers, &stake_distribution)
+            .await
+            .expect_err("synchronize_signers should fail");
+    }
+
+    #[tokio::test]
+    async fn mock_prune_epoch_older_than_threshold() {
```
Alenar, line 303:
```suggestion
    async fn prune_epoch_older_than_threshold() {
```

mithril-aggregator/src/signer_registration/slave.rs
```diff
@@ -0,0 +1,345 @@
...
+            Arc::new(signer_registration_verifier),
+            retention_limit,
+        );
+
+        let current_epoch = Epoch(4) + PROTOCOL_INITIALIZER_PRUNE_EPOCH_THRESHOLD;
+        signer_registerer.prune(current_epoch).await.unwrap();
+    }
+
+    #[tokio::test]
+    async fn mock_without_threshold_nothing_is_pruned() {
```
Alenar, line 327:
```suggestion
    async fn without_threshold_nothing_is_pruned() {
```

mithril-aggregator/tests/create_certificate_slave.rs
```diff
@@ -0,0 +1,395 @@
...
+            StakeDistributionParty::from_signers(fixture.signers_with_stake()).as_slice(),
+            fixture.compute_and_encode_avk(),
+            SignedEntityType::MithrilStakeDistribution(Epoch(1)),
+            ExpectedCertificate::genesis_identifier(Epoch(1)),
+        )
+    );
+
+    comment!("Epoch: 2");
+
+    comment!("Master: change the epoch");
```
Alenar, line 127: Most of the file comments repeat what the code do but doesn't say why we are doing those steps, this make this already long test even more difficult to understand :sweat_smile:.
Detailing for each epochs what we want to test would already be a huge help.

Alenar (approved):
LGTM

mithril-aggregator/src/commands/serve_command.rs
jpraynaud: You're right, it's not needed to have it `pub`. I fixed it and all other occurences in the `ServeCommand` :+1:

mithril-aggregator/src/dependency_injection/builder/enablers/misc.rs
jpraynaud: `cloned()` is specific to iterators, but using `to_owned()` in the first place avoids the problem :+1:

mithril-aggregator/src/signer_registration/slave.rs
jpraynaud: Done

mithril-aggregator/src/signer_registration/slave.rs
jpraynaud: Done

mithril-aggregator/src/services/aggregator_client.rs
jpraynaud: Good catch! I used a new `MITHRIL_AGGREGATOR_VERSION_HEADER` header instead :+1:

mithril-aggregator/tests/create_certificate_slave.rs
jpraynaud: You're right, I added explanations about what we want to test at each epoch for both the master and slave aggregators :+1:

mithril-aggregator/src/dependency_injection/builder/mod.rs
jpraynaud: Good idea, done :+1:

mithril-aggregator/tests/create_certificate_slave.rs
```diff
@@ -0,0 +1,385 @@
...
+        snapshot_directory: TempDir::create("aggregator-integration", "create_certificate_slave"),
+        master_aggregator_endpoint: Some(master_aggregator_http_server.url()),
+        ..master_configuration
+    };
+    let mut slave_tester = RuntimeTester::build(start_time_point, slave_configuration).await;
+
+    comment!("Epoch 1:");
+    comment!("- the master aggregator bootstraps its genesis certificate");
+    comment!("- the slave aggregator synchronizes signers from the master aggregator");
+    comment!("- the slave aggregator stays in Idle state with an error as it doesn't have a genesis certificate yet");
```
Alenar, line 67: As `comment!` uses are numbered, it would be better to avoid doing multiple call in a row but only one instead _(same for the other occurrences)_.


```suggestion
    comment!("Epoch 1:
    - the master aggregator bootstraps its genesis certificate
    - the slave aggregator synchronizes signers from the master aggregator
    - the slave aggregator stays in Idle state with an error as it doesn't have a genesis certificate yet"
    );
```

Doing so will output this in the terminal (with no capture):
```
Mar 06 18:39:32.219 DEBG COMMENT 01 💬 Epoch 1:
    - the master aggregator bootstraps its genesis certificate
    - the slave aggregator synchronizes signers from the master aggregator
    - the slave aggregator stays in Idle state with an error as it doesn't have a genesis certificate yet
``` 

For reference, keeping it unchanged output the following:
```
Mar 06 18:39:32.219 DEBG COMMENT 02 💬 Epoch 1:
Mar 06 18:39:32.219 DEBG COMMENT 03 💬 - the master aggregator bootstraps its genesis certificate
Mar 06 18:39:32.219 DEBG COMMENT 04 💬 - the slave aggregator synchronizes signers from the master aggregator
Mar 06 18:39:32.219 DEBG COMMENT 05 💬 - the slave aggregator stays in Idle state with an error as it doesn't have a genesis certificate yet
```

jpraynaud: Done

mithril-common/src/entities/epoch.rs
```diff
@@ -78,6 +81,11 @@ impl Epoch {
         *self + Self::CARDANO_STAKE_DISTRIBUTION_SNAPSHOT_OFFSET
     }
 
+    /// Apply the [recording offset][Self::SIGNER_MASTER_SYNCHRONIZATION_OFFSET] to this epoch
+    pub fn offset_to_master_synchronization_epoch(&self) -> Self {
+        *self + Self::SIGNER_MASTER_SYNCHRONIZATION_OFFSET
+    }
+
```
dlachaume, line 88: Does this function and the constant declaration need to be in `mithril-common`?
I'm not sure, but I feel like it will only be used in the aggregator context, right?

mithril-aggregator/src/signer_registration/slave.rs
```diff
@@ -0,0 +1,345 @@
...
+        let stake_distribution = fixture.stake_distribution();
+
+        signer_registerer
+            .open_registration_round(registration_epoch, stake_distribution)
+            .await
+            .expect("signer registration round opening should not fail");
+    }
+
+    #[tokio::test]
+    async fn close_registration_always_succeeds() {
```
sfauvel, line 188: suggestion: `close_registration_always_succeeds_on_slave`

mithril-aggregator/src/signer_registration/slave.rs
```diff
@@ -0,0 +1,345 @@
...
+        );
+
+        signer_registerer
+            .close_registration_round()
+            .await
+            .expect("signer registration round opening should not fail");
+    }
+
+    #[tokio::test]
+    async fn register_signer_always_fails() {
```
sfauvel, line 206: suggestion: `register_signer_always_fails_on_slave`

mithril-aggregator/src/signer_registration/slave.rs
```diff
@@ -0,0 +1,345 @@
...
+            verification_key_store.clone(),
+            Arc::new(MockSignerRecorder::new()),
+            Arc::new(MockSignerRegistrationVerifier::new()),
+            None,
+        );
+        let registration_epoch = Epoch(1);
+        let fixture = MithrilFixtureBuilder::default().with_signers(1).build();
+        let stake_distribution = fixture.stake_distribution();
+
+        signer_registerer
```
sfauvel, line 181: We may group all tests on methods that do nothing in a same test (`open_registration_round`, `close_registration_round` and `register_signer`)

mithril-aggregator/tests/create_certificate_slave.rs
```diff
@@ -0,0 +1,385 @@
...
+    };
+    let master_configuration = Configuration {
+        protocol_parameters: protocol_parameters.clone(),
+        signed_entity_types: Some(
+            [
+                SignedEntityTypeDiscriminants::CardanoTransactions.to_string(),
+                SignedEntityTypeDiscriminants::CardanoDatabase.to_string(),
+            ]
+            .join(","),
+        ),
```
dlachaume, line 43: Is this specific configuration necessary because we don't verify certificate production on the slave aggregator for these types of data?
Should we also verify that the slave aggregator doesn't produce certificates if it has a different configuration from the aggregator? For instance, the master aggregator is configured for `CardanoTransactions` and `CardanoDatabase` and the slave aggregator only for `CardanoTransactions`?

dlachaume (approved):
LGTM 👍

mithril-aggregator/src/runtime/runner.rs
```diff
@@ -264,6 +276,54 @@ impl AggregatorRunnerTrait for AggregatorRunner {
...
+            .master_aggregator_client
+            .retrieve_epoch_settings()
+            .await?
+        {
+            time_point.epoch == master_epoch_settings.epoch
+        } else {
+            false
+        };
+
+        Ok(is_slave_aggregator_at_same_epoch_as_master)
```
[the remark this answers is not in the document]
jpraynaud: Yes this is more readable :+1:

mithril-aggregator/src/runtime/runner.rs
```diff
@@ -264,6 +276,54 @@ impl AggregatorRunnerTrait for AggregatorRunner {
             .await
     }
 
+    async fn is_slave_aggregator_at_same_epoch_as_master(
+        &self,
+        time_point: &TimePoint,
+    ) -> StdResult<bool> {
+        let is_slave_aggregator_at_same_epoch_as_master = if let Some(master_epoch_settings) = self
```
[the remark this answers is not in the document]
jpraynaud: Yes, it's more readable. I implemented a suggestion from @dlachaume here which is similar: https://github.com/input-output-hk/mithril/pull/2351#discussion_r1984864468

mithril-common/src/entities/epoch.rs
jpraynaud: Yes, it's better to keep in the `Epoch` implementation otherwise a trait is necessary and would be cumbersome.

mithril-aggregator/tests/create_certificate_slave.rs
jpraynaud: You're right this configuration is not necessary so I removed it. 
Otherwise, there is no need to repeat the unit tests related to signed entity configuration in the integration tests.

mithril-aggregator/src/commands/serve_command.rs
```diff
@@ -121,6 +129,15 @@ impl Source for ServeCommand {
                 Value::new(Some(&namespace), ValueKind::from(metrics_server_port)),
             );
         }
+        if let Some(master_aggregator_endpoint) = self.master_aggregator_endpoint.clone() {
+            result.insert(
+                "master_aggregator_endpoint".to_string(),
```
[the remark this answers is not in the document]
jpraynaud: We have used a string in all the other nodes even though this is not perfect. 
I'd rather fix this problem and the simplification of the aggregator configuration in separate issues.

mithril-aggregator/src/signer_registration/slave.rs
```diff
@@ -0,0 +1,345 @@
...
+        services::EpochPruningTask,
+        signer_registration::{
+            MockSignerRecorder, MockSignerRegistrationVerifier, SignerSynchronizer,
+        },
+        store::MockVerificationKeyStorer,
+        MithrilSignerRegistrationSlave, SignerRegisterer, SignerRegistrationRoundOpener,
+    };
+
+    #[tokio::test]
+    async fn open_registration_always_succeeds() {
```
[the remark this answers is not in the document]
jpraynaud: The test is in the `slave` module so suffixing with `slave` is not much more explicit.

mithril-aggregator/src/signer_registration/slave.rs
jpraynaud: The test is in the `slave` module so suffixing with `slave` is not much more explicit.

mithril-aggregator/src/signer_registration/slave.rs
jpraynaud: The test is in the `slave` module so suffixing with `slave` is not much more explicit.

mithril-aggregator/src/signer_registration/slave.rs
jpraynaud: Done

mithril-aggregator/src/dependency_injection/builder/mod.rs
```diff
@@ -249,7 +269,12 @@ impl DependenciesBuilder {
             snapshotter: None,
             certificate_verifier: None,
             genesis_verifier: None,
-            mithril_registerer: None,
+            mithril_signer_registration_master: None,
```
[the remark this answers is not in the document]
jpraynaud: I'd rather do a thorough refactoring of the dependency injection of the aggregator in a separate PR.

mithril-aggregator/src/runtime/state_machine.rs
```diff
@@ -388,439 +405,575 @@ mod tests {
...
-        assert_eq!("idle".to_string(), runtime.get_state());
+        #[tokio::test]
+        pub async fn idle_new_epoch_detected_and_master_has_transitioned_to_epoch() {
+            let mut runner = MockAggregatorRunner::new();
+            let time_point = TimePoint::dummy();
+            let new_time_point = TimePoint {
+                epoch: time_point.epoch + 1,
+                ..time_point.clone()
+            };
+            let new_time_point_clone = new_time_point.clone();
```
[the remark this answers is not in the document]
jpraynaud: I agree, but as we have already discussed the state machine design leads to these readability problems.
A full refactoring is really needed IMO.

Issue #2384, opened 2025-03-24
Simplify the configuration of commands in aggregator


jpraynaud:
## Why
The configuration of the commands of the aggregator is sub-optimal as it uses the configuration of the serve command. This leads to poor user experience with the commands where the configuration of the server command needs to be specified when only a very small subset of parameters.

## How
Simplify the configuration of the commands by requiring only the mandatory configuration parameters.

## What

- [ ] Create a `Configuration` trait which:
  - [ ]  Has a function for each pub field of the existing configuration
  - [ ]  With a blanket implementation that panics 
- [ ] The configuration of the `DependenciesBuilder` is `Arc<dyn Configuration>`
- [ ] Update the current dependency container to use the `Configuration` trait
- [ ] Create a specific configuration for each command (which implements the `Configuration` trait)
- [ ] Create a specific dependency container for each command (already the case for the `genesis` command)
- [ ] The `DependenciesBuilder` has functions to create the specific dependency containers
- [ ] Remove the `configuration` field from the `DependencyContainer`
- [ ] Rename the `DependencyContainer` to `ServeDependencyContainer`
- [ ] Update the documentation:
  - [ ] Update the aggregator node page
  - [ ] Update the `Recompute certificate hash` runbook and remove the config file download

PR #2432, opened 2025-04-22
Refactor: Simplify the configuration of genesis command in aggregator

dlachaume:
## Content

This PR introduces a specific configuration for the aggregator's `genesis` command to avoid relying on the `serve` configuration.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation
  - [x] No new TODOs introduced

## Issue(s)

Relates to #2384

### DIFF
--- a/mithril-aggregator/src/commands/genesis_command.rs
+++ b/mithril-aggregator/src/commands/genesis_command.rs
@@ -8,7 +8,9 @@ use slog::{debug, Logger};
 
 use mithril_common::{
     chain_observer::ChainObserverType,
-    crypto_helper::{ProtocolGenesisSecretKey, ProtocolGenesisSigner},
+    crypto_helper::{
+        ProtocolGenesisSecretKey, ProtocolGenesisSigner, ProtocolGenesisVerificationKey,
+    },
     entities::{HexEncodedGenesisSecretKey, HexEncodedGenesisVerificationKey},
     StdResult,
 };
@@ -46,9 +48,6 @@ pub struct GenesisCommandConfiguration {
     /// Directory to store aggregator data (Certificates, Snapshots, Protocol Parameters, ...)
     #[example = "`./mithril-aggregator/stores`"]
     pub data_stores_directory: PathBuf,
-
-    /// Genesis verification key
-    pub genesis_verification_key: HexEncodedGenesisVerificationKey,
 }
 
 impl ConfigurationSource for GenesisCommandConfiguration {
@@ -87,10 +86,6 @@ impl ConfigurationSource for GenesisCommandConfiguration {
 
     fn data_stores_directory(&self) -> PathBuf {
         self.data_stores_directory.clone()
-    }
-
-    fn genesis_verification_key(&self) -> HexEncodedGenesisVerificationKey {
-        self.genesis_verification_key.clone()
     }
 
     fn store_retention_limit(&self) -> Option<usize> {
@@ -201,6 +196,10 @@ pub struct ImportGenesisSubCommand {
     /// Signed Payload Path
     #[clap(long)]
     signed_payload_path: PathBuf,
+
+    /// Genesis Verification Key
+    #[clap(long)]
+    genesis_verification_key: HexEncodedGenesisVerificationKey,
 }
 
 impl ImportGenesisSubCommand {
@@ -232,7 +231,10 @@ impl ImportGenesisSubCommand {
             .await
             .with_context(|| "genesis-tools: initialization error")?;
         genesis_tools
-            .import_payload_signature(&self.signed_payload_path)
+            .import_payload_signature(
+                &self.signed_payload_path,
+                &ProtocolGenesisVerificationKey::from_json_hex(&self.genesis_verification_key)?,
+            )
             .await
             .with_context(|| "genesis-tools: import error")?;
         Ok(())
@@ -352,10 +354,6 @@ mod tests {
 
     #[tokio::test]
     async fn create_container_does_not_panic() {
-        let genesis_verification_key = ProtocolGenesisSigner::create_deterministic_signer()
-            .create_verifier()
-            .to_verification_key();
-
         let config = GenesisCommandConfiguration {
             cardano_cli_path: None,
             cardano_node_socket_path: PathBuf::new(),
@@ -363,7 +361,6 @@ mod tests {
             network: "devnet".to_string(),
             chain_observer_type: ChainObserverType::Fake,
             data_stores_directory: temp_dir!().join("stores"),
-            genesis_verification_key: genesis_verification_key.to_json_hex().unwrap(),
         };
         let mut dependencies_builder =
             DependenciesBuilder::new(TestLogger::stdout(), Arc::new(config));
--- a/mithril-aggregator/src/commands/genesis_command.rs
+++ b/mithril-aggregator/src/commands/genesis_command.rs
@@ -23,7 +23,7 @@ use crate::{
 pub struct GenesisCommandConfiguration {
     /// Cardano CLI tool path
     #[example = "`cardano-cli`"]
-    pub cardano_cli_path: PathBuf,
+    pub cardano_cli_path: Option<PathBuf>,
 
     /// Path of the socket used by the Cardano CLI tool
     /// to communicate with the Cardano node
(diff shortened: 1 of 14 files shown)


mithril-aggregator/src/commands/genesis_command.rs
```diff
@@ -1,18 +1,130 @@
...
+    /// Path of the socket used by the Cardano CLI tool
+    /// to communicate with the Cardano node
+    #[example = "`/tmp/cardano.sock`"]
+    pub cardano_node_socket_path: PathBuf,
+
+    /// Cardano Network Magic number
+    ///
+    /// useful for TestNet & DevNet
+    #[example = "`1097911063` or `42`"]
+    pub network_magic: Option<u64>,
```
jpraynaud, line 40: Do we really need to connect to the Cardano node to operate genesis commands?

mithril-aggregator/src/commands/genesis_command.rs
```diff
@@ -1,18 +1,130 @@
...
+
+    /// Cardano network
+    #[example = "`testnet` or `mainnet` or `devnet`"]
+    network: String,
+
+    /// Cardano chain observer type
+    pub chain_observer_type: ChainObserverType,
```
jpraynaud, line 47: Same comment here

mithril-aggregator/src/commands/genesis_command.rs
```diff
@@ -1,18 +1,130 @@
...
+
+    /// Directory of the Cardano node store.
+    pub db_directory: PathBuf,
+
+    /// Directory to store aggregator data (Certificates, Snapshots, Protocol Parameters, ...)
+    #[example = "`./mithril-aggregator/stores`"]
+    pub data_stores_directory: PathBuf,
```
jpraynaud, line 54: Why do we need 2 different directories for the database?

mithril-aggregator/src/commands/genesis_command.rs
```diff
@@ -1,18 +1,130 @@
...
+    /// Genesis verification key
+    pub genesis_verification_key: HexEncodedGenesisVerificationKey,
+
+    /// Protocol parameters
+    #[example = "`{ k: 5, m: 100, phi_f: 0.65 }`"]
+    pub protocol_parameters: ProtocolParameters,
+
+    /// Cardano transactions signing configuration
+    #[example = "`{ security_parameter: 3000, step: 120 }`"]
+    pub cardano_transactions_signing_config: CardanoTransactionsSigningConfig,
```
jpraynaud, line 65: We shouldn't need these parameters in the genesis configuration

mithril-aggregator/src/commands/genesis_command.rs
```diff
@@ -1,18 +1,130 @@
...
+
+    /// Directory of the Cardano node store.
+    pub db_directory: PathBuf,
+
+    /// Directory to store aggregator data (Certificates, Snapshots, Protocol Parameters, ...)
+    #[example = "`./mithril-aggregator/stores`"]
+    pub data_stores_directory: PathBuf,
+
+    /// Genesis verification key
+    pub genesis_verification_key: HexEncodedGenesisVerificationKey,
```
jpraynaud, line 57: This information is probably needed only when signing or bootstrapping

mithril-aggregator/src/commands/genesis_command.rs
jpraynaud, line 1: The commands documentation should also be updated I guess

sfauvel (approved):
LGTM

mithril-aggregator/src/commands/genesis_command.rs
dlachaume: Yes, we need to connect to the Cardano node to retrieve the current time point.
The time point epoch is required to fetch the list of signers and the protocol parameters used to compute the genesis AVK.

mithril-aggregator/src/commands/genesis_command.rs
dlachaume: We need the observer type for the same reason as above.

mithril-aggregator/src/commands/genesis_command.rs
dlachaume: The `db_directory` relates to the Cardano database directory with the immutable files. It's required to build the immutable observer used by the ticker service.
The `data_stores_directory ` contains the aggregator's SQLite database, which is needed to establish the SQL connection.

mithril-aggregator/src/commands/genesis_command.rs
dlachaume: I agree, that's the point I was raising regarding `handle_discrepancies_at_startup`.
Both configurations are required when building the `EpochSettingsStore` dependency, as `handle_discrepancies_at_startup` is invoked at that point.

mithril-aggregator/src/commands/genesis_command.rs
jpraynaud: Can we use the chain observer directly to avoid this entanglement with immutable files which are really not needed for genesis?

mithril-aggregator/src/commands/genesis_command.rs
jpraynaud: Maybe using the `EpochSettingsStorer` is overkill and we just need use a new trait that retrieves the protocol parameters for an epoch

mithril-aggregator/src/commands/genesis_command.rs
jpraynaud: This probably means that multiple configurations are needed for the sub commands

mithril-aggregator/src/commands/genesis_command.rs
dlachaume: Thanks for your suggestion!
Indeed, I was able to replace the ticker service dependency with the chain observer directly. This allowed to remove the `db_directory` configuration, since only the epoch from the time point was actually being used.

mithril-aggregator/src/commands/genesis_command.rs
jpraynaud: I'm not sure that the `cardano_cli_path` is needed. 
Maybe we can use only the `PallasChainObserver` here as well

mithril-aggregator/src/commands/genesis_command.rs
```diff
@@ -1,18 +1,122 @@
...
+
+    /// Cardano Network Magic number
+    ///
+    /// useful for TestNet & DevNet
+    #[example = "`1097911063` or `42`"]
+    pub network_magic: Option<u64>,
+
+    /// Cardano network
+    #[example = "`testnet` or `mainnet` or `devnet`"]
+    network: String,
```
jpraynaud, line 44: I think that we can compute the name of the network from the network magic id

dlachaume: We can compute the network magic id from its code only for `mainnet`, `testnet`, `preview`, and `preprod`. For `devnet` and `private` networks, we need both the code and the magic id to create our `CardanoNetwork` type.

jpraynaud: But we can do the opposite and compute the network name from the network magic

dlachaume: As discussed, it's not possible, so we'll keep both in the configuration.

mithril-aggregator/src/commands/genesis_command.rs
dlachaume: Thanks! As discussed, I’ve added a `ProtocolParametersRetriever` trait, which isolates the responsibility of retrieving protocol parameters from the database. This allowed us to remove these two configuration elements, which were not suited for this purpose.

mithril-aggregator/src/commands/genesis_command.rs
dlachaume: As discussed, since the genesis bootstrap command is executed from the aggregator's stress test using a mock of cardano-cli as the chain observer implementation, we currently can't enforce the use of `PallasChainObserver`.
`cardano_cli_path` configuration has been made optional for the genesis command.

jpraynaud (approved):
LGTM

Alenar (approved):

mithril-aggregator/src/commands/genesis_command.rs
dlachaume: Done ✅

PR #2440, opened 2025-04-25
refactor: shared workspace dependencies

### LINKED ISSUE #2430: Accelerate Rust compilation time with workspace dependencies
## Why
We want to reduce the compilation times in the repository and we have noticed that some common dependencies to our crates are rebuilt too frequently: small modifications in the code can lead to high compilation time.

## What
Declare common dependencies in the workspace to avoid recompiling them too frequently.

## How

- [ ] Identify the common dependencies (e.g. serde, clap, tokio, ...)
- [ ] In the `Cargo.toml` file of the workspace, add the common dependencies and their common features (use the intersection)
- [ ] Assess that compile times are lower
- [ ] Adapt the dependency upgrade:
  - [ ] script
  - [ ] runbook?

Alenar:
## Content

This PR tries to reduce build time, especially when switching between crates, by moving the most critical or shared dependencies to the workspace directly.

It also:
- Remove the build script from `mithril-stm` as everything now compile to wasm without needing to gate `batch_verify`.
  - Since we don't wan't to touch stm code and still always enable the gated code, this implies to set by default `batch-verify-aggregates`
  - `batch-verify-aggregates` is now deprecated and cfg gate that use it should be removed, then we will be able to remove the feature
- Refactor `mithril-build-script` open api code generators: to separate the part where we find the open api spec files and the part that generate code based on them
- `mithril-common`:
  - Refactor build script: rerun only if a openapi spec file changed
  - Make `serde_yaml` optional: it's only needed for `apispec` test tool and this crate is **[deprecated](https://github.com/dtolnay/serde-yaml)**  ( :warning: we should find an alternative for our few usages)

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation
  - [x] No new TODOs introduced

## Comments

- devbook dependency upgrade script is not impacted as it bump all crates patch version indiscriminately
- Even with those changes `mithril-stm` always recompile if we alternate between aggregator and signer, more work is needed or maybe a tool like [cargo-hakari](https://github.com/guppy-rs/guppy/tree/main/tools/cargo-hakari)

## Issue(s)

Closes #2430


turmelclem (approved):
LGTM

dlachaume (approved):
LGTM

mithril-common/Cargo.toml
```diff
@@ -47,7 +47,7 @@ kes-summed-ed25519 = { version = "0.2.1", features = [
     "serde_enabled",
     "sk_clone_enabled",
 ] }
-mithril-stm = { path = "../mithril-stm", version = "0.3", default-features = false }
+mithril-stm = { path = "../mithril-stm", version = "0.3", default-features = false, features = ["batch-verify-aggregates"] }
```
jpraynaud, line 50: Do we need to use the `batch-verify-aggregates` here as it is default now?

mithril-stm/CHANGELOG.md
```diff
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## 0.3.yy (dd-mm-2025)
```
jpraynaud, line 8: This needs to be updated before merging :slightly_smiling_face:

mithril-stm/CHANGELOG.md
```diff
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
...
+
+- Remove the build script and deprecate `batch-verify-aggregate` feature as the code behind this feature is now
+  compatible with WASM.
```
jpraynaud, line 11:
```suggestion
- Removed the build script and deprecated `batch-verify-aggregate` feature as the code behind this feature is now
compatible with WASM.
```

jpraynaud (approved):
LGTM

mithril-common/Cargo.toml
Alenar: Since we disable default-feature to allow backend switch: yes, we need to repeat the feature here since the code is still gated behind it.

sfauvel (approved):
LGTM

PR #2465, opened 2025-05-06
Set correct package name for mithril-client-cli

noonio:
Now it matches the binary name as defined in the cargo, so `nix run ...` commands work correctly.

To test this, without this change the following command fails (run in a checkout of this repo):

```
nix run .#mithril-client-cli
```

After, it succeeds :)


jpraynaud (approved):
LGTM

dlachaume (approved):
LGTM, thanks!

PR #2476, opened 2025-05-13
Test: add Cardano node startup check to `Mithril Client multi-platform test` workflow

### LINKED ISSUE #2471: Add a Cardano node startup check step in the manual client test workflow
## Why

We want to test that the Cardano node can start without failure when its database is restored with Cardano DB v1 and Cardano DB v2 (with ancillary and without ancillary download). To do so, we need to adapt the [`Mithril Client multi-platform test`](https://github.com/input-output-hk/mithril/actions/workflows/test-client.yml) workflow.

## What
Add steps to test the Cardano node success startup when restored from a Cardano DB snapshot.

## How

- [x] Support restoration with and without ancillary (use a matrix):
  - [x] Cardano DB v1
  - [x] Cardano DB v2
- [x] Check that the Cardano node starts successfully (Linux only and client binary only):
  - [x] Read and run the docker command printed in the client and check that `Started opening Immutable DB` is in the log (within 15 seconds)
  - [x] With ancillary, read and run the docker command printed in the client and check that `Chain extended, new tip` is in the log (within 30 seconds)
  - [x] Cardano DB v1
  - [x] Cardano DB v2

dlachaume:
## Content

This PR includes...

## Pre-submit checklist

- Branch
- PR
  - [ ] Self-reviewed the diff
  - [ ] Reviewer requested
- Documentation
  - [ ] No new TODOs introduced

## Issue(s)

Closes #2471


.github/workflows/test-client.yml
```diff
@@ -167,9 +158,20 @@ jobs:
           echo "CDB_SNAPSHOT_DIGEST=$(./mithril-client --origin-tag CI cardano-db snapshot list --json | jq -r '.[0].digest')" >> $GITHUB_ENV
 
       - name: Cardano Database Snapshot / download & restore latest
+        id: cardano-db-download
         shell: bash
         working-directory: ./bin
-        run: ./mithril-client ${{ steps.prepare.outputs.debug_level }} --origin-tag CI cardano-db download $CDB_SNAPSHOT_DIGEST $INCLUDE_ANCILLARY
+        run: |
+          OUTPUT=$(./mithril-client ${{ steps.prepare.outputs.debug_level }} --origin-tag CI cardano-db download $CDB_SNAPSHOT_DIGEST ${{ matrix.extra_args }})
```
jpraynaud, line 165: You could write the output of the command in a file and then use this file as input of the verification command. This will avoid flooding the logs with all the content printed by the client CLI. This will also allow to remove all the lines needed to capture the output with the `GITHUB_OUTPUT` below.

.github/workflows/scripts/verify-cardano-db-restoration.sh
```diff
@@ -0,0 +1,78 @@
...
+echo "$DOCKER_CMD"
+
+DOCKER_CMD_DETACHED="${DOCKER_CMD/docker run/docker run -d}"
+echo "Running Docker command in detached mode:"
+echo "$DOCKER_CMD_DETACHED"
+
+CONTAINER_ID=$(eval "$DOCKER_CMD_DETACHED")
+
+FOUND_IMMUTABLE_DB_LOG=false
+echo "Waiting up to 15 seconds for 'Started opening Immutable DB'..."
```
jpraynaud, line 31: The timeout could be set as a variable.

.github/workflows/scripts/verify-cardano-db-restoration.sh
```diff
@@ -0,0 +1,78 @@
...
+
+FOUND_CHAIN_TIP_LOG=false
+echo "Waiting up to 30 seconds for 'Chain extended, new tip'..."
+for ((i=1; i<=30; i++)); do
+  if docker logs "$CONTAINER_ID" 2>&1 | grep -q "Chain extended, new tip"; then
+    FOUND_CHAIN_TIP_LOG=true
+    break
+  fi
+  sleep 1
+done
```
jpraynaud, line 67: You could create a function to use here and above as the code is very similar.

Alenar (approved):
LGTM, but I've a question: does the check against the cardano node is quick enough when running without the ancillary ? In that case the node must reconstruct the ledger and I fear that the added overhead make it too slow (plus maybe only checking with ancillary is enough ?).

.github/workflows/test-client.yml
dlachaume: Thanks, that's a good idea. I've made the change.

.github/workflows/scripts/verify-cardano-db-restoration.sh
dlachaume: You're right. It's done, thanks.

.github/workflows/test-client.yml
```diff
@@ -243,13 +239,19 @@ jobs:
         if: steps.aggregator_capability_unix.outputs.cardano_database_v2_enabled == 'true' || steps.aggregator_capability_windows.outputs.cardano_database_v2_enabled == 'true'
         shell: bash
         working-directory: ./bin
-        run: ./mithril-client ${{ steps.prepare.outputs.debug_level }} --unstable --origin-tag CI cardano-db-v2 download $CARDANO_DATABASE_V2_SNAPSHOT_HASH $INCLUDE_ANCILLARY
+        run: ./mithril-client ${{ steps.prepare.outputs.debug_level }} --unstable --origin-tag CI cardano-db-v2 download $CARDANO_DATABASE_V2_SNAPSHOT_HASH ${{ matrix.extra_args }} > cdb-v2-download-output.txt
```
jpraynaud, line 242: Maybe we can also add the error logs in the output file?

jpraynaud (approved):
LGTM :+1:

.github/workflows/test-client.yml
dlachaume: Done ✅

dlachaume:
> LGTM, but I've a question: does the check against the cardano node is quick enough when running without the ancillary ? In that case the node must reconstruct the ledger and I fear that the added overhead make it too slow (plus maybe only checking with ancillary is enough ?).

You're absolutely right, without the ancillary, the ledger state reconstruction is time-consuming. That's why we've implemented a two verification steps:

**Without ancillary**: We only check that the node starts opening the Immutable DB. This ensures that the necessary files are present to initiate the process. This check is quick and completes within the first 15 seconds.

**With ancillary**: In addition to the previous check, we verify that the node has received a new valid block, confirming that it's actively running and extending its chain.

turmelclem (approved):
LGTM

PR #2475, opened 2025-05-13
Aggregator: Re-implement the CloudBackendUploader with gcloud-storage crate

Alenar:
## Content

This PR re-implement the GCP Storage `CloudBackendUploader` using [`gcloud-storage`](https://crates.io/crates/gcloud-storage), replacing the previous implementation that was using [`cloud-storage`](https://crates.io/crates/cloud-storage).

`cloud-storage` have not seen any update in three years and, since it make external requests, can pose security risks. The new `gcloud-storage` is more "bare metal" but is actively maintained.

### Changes

- Reorganize the `gcp_uploader` module:
  - promoted to a directory
  - renamed to `cloud_uploader` as it's the backend that ties to a specific provider
  - separated the backend from other the part
- Extract the `get_location` from the legacy uploader to the `CloudRemotePath` (renamed to `to_gcloud_storage_location`) since we need to reuse it in the new uploader
- Rename the previous uploader to `GcpBackendUploaderLegacy` then delete it and remove the `cloud-storage` dependency
- Add a new `CloudBackendUploader` using `gcloud-storage`: it should be functionally equivalent (I made extra care to ensure the same api paths are called with the same parameters), but since there's no easy way to unit test it we will have to do it directly on `testing-preview`

## Pre-submit checklist

- Branch
  - [x] Tests are provided (if possible)
  - [x] Crates versions are updated (if relevant)
  - [ ] CHANGELOG file is updated (if relevant)
  - [x] Commit sequence broadly makes sense
  - [x] Key commits have useful messages
- PR
  - [x] All check jobs of the CI have succeeded
  - [x] Self-reviewed the diff
  - [x] Useful pull request description
  - [x] Reviewer requested
- Documentation
  - [x] No new TODOs introduced

## Comments

If the new implementation crash on `testing-preview` and can't be easily fixed, revert the commit named `chore(aggregator): remove legacy gcp storage uploader backend` and rewire the `GcpBackendUploaderLegacy` in dependency injection.

## Issue(s)

Relates to #2460

Issue #2460, opened 2025-05-05
Re-implement the `GCPBackendUploader` with `cloud-storage` crate in aggregator


jpraynaud:
## Why

The current `cloud-storage` crate that we use for uploading files to GCP is unmaintained. We want to replace is with another crate which has better support: `gcloud-storage`.

## What

Implement a new version of the `file_uploader/gcp_uploader` in the aggregator with the `gcloud-storage` crate.

## How

- [x] Rename `gcp_uploader` module  to `gcp_uploader_legacy` and `GcpBackendUploader` to `GcpBackendUploaderLegacy`
- [x] Create a new `gcp_uploader`module and `GcpBackendUploader` with `gcloud-storage` crate with the same features
- [x] Update infra and documentation if the configuration has changed
- [x] Deprecate the `gcp_uploader_legacy` module
- [x] Once the new uploader is deployed on `testing-preview` and works as expected, remove the `gcp_uploader_legacy` module

PR #2480, opened 2025-05-14
aggregator: fix setting public access to files uploaded to GCP storage

Alenar:
## Content

This PR try to fix the upload of files to GCP storages and enhance logging of failures when a retried operation fails.

The upload of files to GCP after merging #2475 fails when applying the ACL to make the file public but the associated error is not log (when only see that it failed after retrying x time).
After investigating the internal difference between the new and old backend crate we found a difference: the previous crate was enforcing itself [GCP url encoding rules](https://cloud.google.com/storage/docs/request-endpoints#encoding) but the new crate doesn't.
This PR align the behavior of the new implementation with the previous one by enforcing manually the encoding rule.

### Changes

- Add the source error of the last failure when all retries fails in `mithril-aggregator` file upload and `mithril-client`  file download
- Encode object name when sending the acl insert request to GCP

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation
  - [x] No new TODOs introduced

## Issue(s)

Relates to #2460

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3793,6 +3793,7 @@ dependencies = [
  "mithril-signed-entity-preloader",
  "mockall",
  "paste",
+ "percent-encoding",
  "rayon",
  "regex",
  "reqwest",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3769,7 +3769,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-aggregator"
-version = "0.7.48"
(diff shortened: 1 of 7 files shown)


jpraynaud (approved):
LGTM :fire:

dlachaume (approved):

turmelclem (approved):

PR #2502, opened 2025-05-21
feat: cleanup unexpected files in immutable folder after download

### LINKED ISSUE #2429: Cleanup unexpected immutable files in archive in client
## Why
We want to avoid restoring immutable files which have not been certified by Mithril but are in the artifact archive: the maximum immutable file number restored should be lower than the target immutable file in the beacon.

## What
Add a cleanup phase of the immutable files in the client.

## How

- [x] Identify the immutable files which have been downloaded and should not be part of the archive:
  - [x] Cardano DB v1
  - [x] Cardano DB v2
- [x] Remove those files if any
  - [x] Cardano DB v1 
  - [x] Cardano DB v2
- [x] Display a warning message explaining that some files were unexpected and deleted
  - [x] Cardano DB v1 
  - [x] Cardano DB v2
- [x] Run the cleanup phase even if an error occurs or alternatively remove the target directory?
  - [x] Cardano DB v1 
  - [x] Cardano DB v2
- [x] Make sure that the files are not kept in the Trash?

Alenar:
## Content

This PR add a two phase tool in the `mithril-client` that:

1. compute the expected state of the immutable folder after download
2. remove any file or directory that is not in the expectation
3. report offenders in a warning log

The removal is executed even if the download fail.

### caveats

- The tool can be tricked in some specific case, i.e: download of a Range of immutable, if a downloaded immutable archive contains a immutable below the downloaded range it won't be identified
- The tool enforce the current directory structure of the Cardano node immutable directory: if the node change it's immutable naming scheme this tool will need to be adapted, else it will remove the files with the new scheme
- The reported offenders are only in a warning log, if library users don't enable the logs they won't see the message

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation
  - [x] No new TODOs introduced

## Issue(s)

Closes #2429

### DIFF
--- a/mithril-client/src/snapshot_client.rs
+++ b/mithril-client/src/snapshot_client.rs
@@ -120,8 +120,7 @@ use crate::file_downloader::{DownloadEvent, FileDownloader};
 #[cfg(feature = "fs")]
 use crate::utils::create_bootstrap_node_files;
 #[cfg(feature = "fs")]
-use crate::utils::AncillaryVerifier;
-use crate::utils::ANCILLARIES_NOT_SIGNED_BY_MITHRIL;
+use crate::utils::{AncillaryVerifier, ANCILLARIES_NOT_SIGNED_BY_MITHRIL};
 use crate::{MithrilResult, Snapshot, SnapshotListItem};
 
 /// Error for the Snapshot client
--- a/mithril-client/src/snapshot_client.rs
+++ b/mithril-client/src/snapshot_client.rs
@@ -120,7 +120,9 @@ use crate::file_downloader::{DownloadEvent, FileDownloader};
 #[cfg(feature = "fs")]
 use crate::utils::create_bootstrap_node_files;
 #[cfg(feature = "fs")]
-use crate::utils::{AncillaryVerifier, ANCILLARIES_NOT_SIGNED_BY_MITHRIL};
+use crate::utils::{
+    AncillaryVerifier, UnexpectedDownloadedFileVerifier, ANCILLARIES_NOT_SIGNED_BY_MITHRIL,
+};
 use crate::{MithrilResult, Snapshot, SnapshotListItem};
 
 /// Error for the Snapshot client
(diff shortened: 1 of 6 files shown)


mithril-client/src/utils/unexpected_downloaded_file_verifier.rs
```diff
@@ -0,0 +1,511 @@
+//! Toolbox to verify if unexpected files are included when downloading and unpacking
```
dlachaume, line 1: Just a suggestion: I noticed there are quite a few comments throughout the file, some of which explain business rules. Would it make sense to centralize them, maybe at the top of the file?

For example:
- devnet vs. other networks
- with vs. without ancillary files

dlachaume (approved):
LGTM 👍

mithril-client/src/utils/unexpected_downloaded_file_verifier.rs
```diff
@@ -0,0 +1,511 @@
...
+            immutable_files_range_to_expect: compute_immutable_files_range_to_expect(
+                network_kind,
+                include_ancillary,
+                last_downloaded_immutable_file_number,
+            ),
+            logger: logger.clone(),
+        }
+    }
+
+    /// Compute the expected state of the folder after download finish
```
jpraynaud, line 59:
```suggestion
    /// Compute the expected state of the folder after download completed
```

jpraynaud (approved):
LGTM :+1:

mithril-client/src/snapshot_client.rs
```diff
@@ -260,16 +267,48 @@ impl SnapshotClient {
...
+            let include_ancillary = false;
+            let expected_files_after_download = UnexpectedDownloadedFileVerifier::new(
+                target_dir,
+                &snapshot.network,
+                include_ancillary,
+                snapshot.beacon.immutable_file_number,
+                &self.logger
+            )
+            .compute_expected_state_after_download()
+            .await?;
```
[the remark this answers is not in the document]
Alenar: I would like to, but it's not practical because that would mean returning the value from both the `Ok` and `Err` variant or creating a specialized result, else I would not be able to run the cleanup on error.

turmelclem (approved):

PR #2520, opened 2025-05-26
fix: missing `00000` immutable files with Cardano DB v2

### LINKED ISSUE #2519: Cardano DB v2 restoration is missing `00000` immutables
## Why
The immutable files `00000` are not available when being restored by the Mithril client with the Cardano DB v2 certification. This causes the failure of the Cardano node at startup.

## What
Make the Cardano DB v2 certification fully support all the immutable files.

## How

- [x] Investigate the issue
- [x] Fix the issue

jpraynaud:
## Content

This PR includes a fix on **Cardano DB v2 which can not restore a valid full Cardano database**:
- embed the `00000` immutable files in the snapshot archive files
- download the `00000` immutable files in the client

## Pre-submit checklist

- Branch
- PR
  - [ ] Self-reviewed the diff
  - [ ] Reviewer requested
- Documentation
  - [ ] No new TODOs introduced

## Issue(s)
Closes #2519


mithril-aggregator/src/artifact_builder/cardano_database.rs
```diff
@@ -227,7 +227,7 @@ mod tests {
         }
 
         let immutable_artifact_builder = {
-            let number_of_immutable_file_loaded = beacon.immutable_file_number;
+            let number_of_immutable_file_loaded = beacon.immutable_file_number + 1;
```
Alenar, line 230: Adding a light comment explaining that this `+1` comes from the fact that we include the `00000` immutable would be appreciated.

Alenar (approved):
LGTM

I think you can simplify the tests by shifting the previous upper bound immutable to keep the same number of checks.

PR #2546, opened 2025-06-03
Record usage of client type in metrics

### LINKED ISSUE #2494: Record usage of client type in metrics
## Why
We want to record the usage of the different types of Mithril clients in the metrics that we record.

## What
Provide automatic recording of the type of Mithril client used to call the aggregator in the metrics.

## How

- [x] Create a list of authorized client types:
  - [x] Fully hard-coded: `CLI`, `WASM`, `LIBRARY`, `NA`
- [x] In the client library:
  - [x] Set a specific option for the client type:
    - [x] defaults to `LIBRARY`
    - [x] ~~use a an enum for the values (undocumented with `#[doc(hidden)]`)~~
    - [x] create a `with_client_type` function in the `ClientBuilder`
  - [x] Add a new `mithril-client-type` HTTP header to the calls made to aggregator with the value of this option
- [x] In the client CLI, use the client type `CLI` when building the client
- [x] In the client WASM, use the client type `WASM` when building the client
- [x] In the aggregator, add a middleware to the HTTP routes for computing the client type:
  - [x] If it is not set or unknown, then fallback to `NA`
  - [x] Use the received client type in the metrics recording inside the handlers of the routes 
  - [x] Update `EventMessage` with the client type when needed
- [x] In the aggregator, extend the metrics:
  - [x] Support recording of a metric with a client type
  - [x] Update all the calls to the metrics to support an optional client type
  - [x] In the state machine, the calls to the metrics should not be modified (as there is no client type in that case)
- [x] Adapt the `UsageReporter` to send the client type in the event stored in the event store
- [x] Create a new migration in the event store database of the aggregator and add a new `metrics_per_day_and_client_type` view

## Later

- Adapt the Grafana dashboard

turmelclem:
## Content

Add a second dimensions in multiple metrics called client_type for some existing metrics incremented by http server.

- mithril-client-cli call http server with "CLI" value
- mithril-client wasm call http server with "WASM" value
- by default mithril library use "LIBRARY" value
- mithril-aggregator have been enhanced to support metrics with multiple dimensions 
- usage_reporter have been enhanced by populating origin_tag and client_type in dedicated attribute instead of concatenation mecanism

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation
  - [x] No new TODOs introduced

## Issue(s)

Closes #2494

### DIFF
--- a/mithril-aggregator/Cargo.toml
+++ b/mithril-aggregator/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mithril-aggregator"
-version = "0.7.56"
+version = "0.7.57"
 description = "A Mithril Aggregator server"
 authors = { workspace = true }
 edition = { workspace = true }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3819,7 +3819,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-aggregator"
-version = "0.7.56"
+version = "0.7.57"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -3909,7 +3909,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-client"
-version = "0.12.9"
+version = "0.12.10"
 dependencies = [
  "anyhow",
  "async-recursion",
@@ -3943,7 +3943,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-client-cli"
-version = "0.12.7"
+version = "0.12.8"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -3977,7 +3977,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-client-wasm"
-version = "0.9.0"
+version = "0.9.1"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -3995,7 +3995,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-common"
-version = "0.5.34"
+version = "0.5.35"
 dependencies = [
  "anyhow",
  "async-trait",
--- a/examples/client-wasm-nodejs/package-lock.json
+++ b/examples/client-wasm-nodejs/package-lock.json
@@ -16,7 +16,7 @@
     },
     "../../mithril-client-wasm": {
       "name": "@mithril-dev/mithril-client-wasm",
-      "version": "0.9.0",
+      "version": "0.9.2",
       "license": "Apache-2.0"
     },
     "node_modules/@mithril-dev/mithril-client-wasm": {
--- a/examples/client-wasm-web/package-lock.json
+++ b/examples/client-wasm-web/package-lock.json
@@ -20,7 +20,7 @@
     },
     "../../mithril-client-wasm": {
       "name": "@mithril-dev/mithril-client-wasm",
-      "version": "0.9.0",
+      "version": "0.9.2",
       "license": "Apache-2.0"
     },
     "node_modules/@discoveryjs/json-ext": {
--- a/mithril-client-cli/Cargo.toml
+++ b/mithril-client-cli/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mithril-client-cli"
-version = "0.12.7"
+version = "0.12.8"
 description = "A Mithril Client"
 authors = { workspace = true }
 edition = { workspace = true }
--- a/mithril-client-wasm/Cargo.toml
+++ b/mithril-client-wasm/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mithril-client-wasm"
-version = "0.9.0"
+version = "0.9.1"
 description = "Mithril client WASM"
 authors = { workspace = true }
 edition = { workspace = true }
--- a/mithril-client-wasm/ci-test/package-lock.json
+++ b/mithril-client-wasm/ci-test/package-lock.json
@@ -21,7 +21,7 @@
     },
     "..": {
       "name": "@mithril-dev/mithril-client-wasm",
-      "version": "0.9.1",
+      "version": "0.9.2",
       "license": "Apache-2.0"
     },
     "node_modules/@discoveryjs/json-ext": {
--- a/mithril-client-wasm/package.json
+++ b/mithril-client-wasm/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@mithril-dev/mithril-client-wasm",
-  "version": "0.9.1",
+  "version": "0.9.2",
   "description": "Mithril client WASM",
   "license": "Apache-2.0",
   "collaborators": [
--- a/mithril-client/Cargo.toml
+++ b/mithril-client/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mithril-client"
-version = "0.12.9"
+version = "0.12.10"
 description = "Mithril client library"
 authors = { workspace = true }
 edition = { workspace = true }
--- a/mithril-common/Cargo.toml
+++ b/mithril-common/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mithril-common"
-version = "0.5.34"
+version = "0.5.35"
 description = "Common types, interfaces, and utilities for Mithril nodes."
 authors = { workspace = true }
 edition = { workspace = true }
--- a/mithril-explorer/package-lock.json
+++ b/mithril-explorer/package-lock.json
@@ -36,7 +36,7 @@
     },
     "../mithril-client-wasm": {
       "name": "@mithril-dev/mithril-client-wasm",
-      "version": "0.9.1",
+      "version": "0.9.2",
       "license": "Apache-2.0"
     },
     "node_modules/@adobe/css-tools": {


mithril-client/src/client.rs
```diff
@@ -343,6 +346,14 @@ impl ClientBuilder {
...
             headers.insert(MITHRIL_ORIGIN_TAG_HEADER.to_string(), origin_tag);
         }
+        if let Some(client_type) = self.client_type.clone() {
+            headers.insert(MITHRIL_CLIENT_TYPE_HEADER.to_string(), client_type);
+        } else if !headers.contains_key(MITHRIL_CLIENT_TYPE_HEADER) {
+            headers.insert(
+                MITHRIL_CLIENT_TYPE_HEADER.to_string(),
+                "LIBRARY".to_string(),
+            );
+        }
```
jpraynaud: This would probably be a bit more readable:
```suggestion
headers.insert(MITHRIL_CLIENT_TYPE_HEADER.to_string(), self.client_type.clone().unwrap_or_else("LIBRARY".to_string()));
```

mithril-client/src/client.rs
```diff
@@ -410,6 +421,12 @@ impl ClientBuilder {
         self
     }
 
+    /// Set the client type.
+    pub fn with_client_type(mut self, client_type: Option<String>) -> Self {
```
jpraynaud, line 427: Do we really need to consume a `mut self` or `self` could be enough?

mithril-aggregator/Cargo.toml
```diff
@@ -72,6 +72,7 @@ zstd = { version = "0.13.3", features = ["zstdmt"] }
 tikv-jemallocator = { version = "0.6.0", optional = true }
 
 [dev-dependencies]
+bytes = "1.10.1"
```
jpraynaud: Is it really needed? We could probably use the `Bytes` from warp instead (`warp::hyper::body::Bytes`)

mithril-aggregator/src/metrics/mod.rs
```diff
@@ -3,4 +3,5 @@
 
 mod service;
 
+pub use service::MetricMap;
 pub use service::MetricsService;
```
jpraynaud:
```suggestion
pub use service::{MetricMap, MetricsService};
```

mithril-aggregator/src/metrics/service.rs
```diff
@@ -146,20 +147,54 @@ impl MetricsService {
...
             .collect()
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+
+/// A struct to represent a metric with a Label/Value map and a counter.
+pub struct MetricMap {
+    /// key (labels) and value
+    pub key_value: HashMap<String, String>,
```
jpraynaud: IMO this name is not explicit enough , maybe `label_value_map` is better?

Also can we create alias of the `String` types in the Hashmap to make it more readable?

mithril-aggregator/src/metrics/service.rs
```diff
@@ -146,20 +147,54 @@ impl MetricsService {
...
             })
             .collect()
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+
+/// A struct to represent a metric with a Label/Value map and a counter.
+pub struct MetricMap {
+    /// key (labels) and value
```
jpraynaud: Can you make sure that doc comments are properly capitalized here and others?

mithril-aggregator/src/metrics/service.rs
```diff
@@ -146,20 +147,54 @@ impl MetricsService {
...
                 )
             })
             .collect()
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+
+/// A struct to represent a metric with a Label/Value map and a counter.
+pub struct MetricMap {
```
jpraynaud: Maybe this name is a bit too generic?

mithril-aggregator/src/services/usage_reporter.rs
```diff
@@ -47,24 +50,32 @@ impl UsageReporter {
...
+        metrics_before: Option<&HashMap<String, MetricMap>>,
+        metrics_after: &HashMap<String, MetricMap>,
+    ) -> HashMap<String, MetricMap> {
         metrics_after
             .iter()
-            .map(|(name, value)| {
-                let last_value = metrics_before.and_then(|m| m.get(name)).unwrap_or(&0);
-                let delta: i64 = (*value as i64) - (*last_value as i64);
-                (name.clone(), delta)
+            .filter_map(|(name, value)| {
```
jpraynaud: It's a bit hard to understand what `name` and `value` represent here.

mithril-aggregator/src/services/usage_reporter.rs
```diff
@@ -47,24 +50,32 @@ impl UsageReporter {
...
+                let last_value = metrics_before.and_then(|m| m.get(name));
+                let delta = match last_value {
+                    Some(last) => (value.counter as i64) - (last.counter as i64),
+                    None => value.counter as i64,
+                };
+                if delta != 0 {
+                    let mut delta_metric = value.clone();
+                    delta_metric.counter = delta as u32; // Use abs() to avoid negative u32
```
jpraynaud: You can probably use saturating sub for that.

mithril-aggregator/src/services/usage_reporter.rs
```diff
@@ -84,13 +95,22 @@ impl UsageReporter {
 
         self.last_reported_metrics = metrics;
 
-        for (name, origin_values) in delta {
-            for (origin, value) in origin_values {
+        for (name, metrics_map) in delta {
+            for (_origin, metric_map) in metrics_map {
```
jpraynaud: I think `origin` is not an appropriate name as this is not what is represented here

mithril-client/src/client.rs
turmelclem: this approach does not check if we already have headers populated by function `with_options`.
with your suggestion, if we use `with_options` to set client_type header, it will  be overwritten  to "LIBRARY" and make following test fail: `compute_http_headers_with_options_containing_client_type_returns_client_type`

jpraynaud: I'm not convinced that we need to verify if there is already a value set for the `MITHRIL_CLIENT_TYPE_HEADER`: it should probably be overwritten.

mithril-client/src/client.rs
turmelclem: @jpraynaud  something like that ? 
```rust 
    pub fn with_client_type(self, client_type: Option<String>) -> Self {
        Self {
            client_type,
            ..self
        }
    }
```
i coult do so on each `with_...` functions

mithril-client/src/client.rs
turmelclem: i'm ok that `with_client_type` have to overwrite client_type header set by `with_options`

@jpraynaud but, here, if we have populated `client_type` header through `with_options` only, it will be overwritten by "LIBRARY" either if `with_client_type` is not called 
that mean, `client_type` header can not be set through "with_options" method. is that ok ?  its not consistent with `origin_tag`.

mithril-client/src/client.rs
jpraynaud: Yes :+1:

mithril-aggregator/src/services/usage_reporter.rs
turmelclem: yes and since _origin is not used, I replace it with : 
```rust
for metric_map in metrics_map.values() {
```

mithril-client/src/client.rs
```diff
@@ -417,21 +450,22 @@ impl ClientBuilder {
...
 }
 
 #[cfg(test)]
 mod tests {
     use super::*;
 
+    fn default_headers() -> HashMap<String, String> {
+        HashMap::from([(
+            MITHRIL_CLIENT_TYPE_HEADER.to_string(),
+            "LIBRARY".to_string(),
```
dlachaume: Perhaps it would be better to define it as a `const`, such as `DEFAULT_CLIENT_TYPE`.

mithril-client/src/client.rs
```diff
@@ -495,4 +524,111 @@ mod tests {
...
+                ("Key".to_string(), "Value".to_string()),
+                (
+                    MITHRIL_CLIENT_TYPE_HEADER.to_string(),
+                    "CLIENT_TYPE".to_string()
+                )
+            ])
+        );
+    }
+
+    //TODO FIX THIS BEHAVIOUR
```
dlachaume: Is the comment still relevant?

mithril-aggregator/src/metrics/service.rs
```diff
@@ -146,20 +147,59 @@ impl MetricsService {
...
             })
             .collect()
     }
 }
 
+type LabelName = String;
+type LabelValue = String;
+#[derive(Debug, Clone, PartialEq, Eq)]
+
+/// MetricLabelValueMap represent a metric with a LabelName/LabelValue map and a counter.
```
dlachaume: I recommend using the types with brackets in the documentation. That way, if you ever rename it, you won’t risk forgetting to update the doc comment.
```suggestion
/// [MetricLabelValueMap] represent a metric with a [LabelName]/[LabelValue] map and a counter.
```

mithril-aggregator/src/metrics/service.rs
```diff
@@ -175,20 +215,41 @@ mod tests {
         metric_b.increment_by(12);
 
         let export = metrics_service.export_metrics_map();
-        assert_eq!(5, export[&metric_a.name()][""]);
-        assert_eq!(12, export[&metric_b.name()][""]);
+        assert_eq!(5, export[&metric_a.name()][""].clone().to_owned().counter);
+        assert_eq!(12, export[&metric_b.name()][""].clone().to_owned().counter);
```
dlachaume:
```suggestion
        assert_eq!(5, export[&metric_a.name()][""].counter);
        assert_eq!(12, export[&metric_b.name()][""].counter);
```

mithril-aggregator/src/metrics/service.rs
```diff
@@ -198,11 +259,11 @@ mod tests {
         metric_a.increment_by(5);
 
         let export = metrics_service.export_metrics_map();
-        assert_eq!(5, export[&metric_a.name()][""]);
+        assert_eq!(5, export[&metric_a.name()][""].clone().to_owned().counter);
```
dlachaume:
```suggestion
        assert_eq!(5, export[&metric_a.name()][""].counter);
```

mithril-aggregator/src/metrics/service.rs
```diff
@@ -198,11 +259,11 @@ mod tests {
...
 
         metric_a.increment();
         let export = metrics_service.export_metrics_map();
-        assert_eq!(6, export[&metric_a.name()][""]);
+        assert_eq!(6, export[&metric_a.name()][""].clone().to_owned().counter);
```
dlachaume:
```suggestion
        assert_eq!(6, export[&metric_a.name()][""].counter);
```

mithril-aggregator/src/metrics/service.rs
```diff
@@ -211,7 +272,7 @@ mod tests {
         let metric_a = metrics_service.get_runtime_cycle_total_since_startup();
 
         let export = metrics_service.export_metrics_map();
-        assert_eq!(0, export[&metric_a.name()][""]);
+        assert_eq!(0, export[&metric_a.name()][""].clone().to_owned().counter);
```
dlachaume:
```suggestion
        assert_eq!(0, export[&metric_a.name()][""].counter);
```

mithril-aggregator/src/metrics/service.rs
```diff
@@ -175,20 +215,41 @@ mod tests {
...
-        metric_a.increment_by(&["TOKEN_B"], 12);
+        metric_a.increment_by(&["A", "B"], 5);
+        metric_a.increment_by(&["1", "2"], 12);
 
         let export = metrics_service.export_metrics_map();
-        assert_eq!(5, export[&metric_a.name()]["TOKEN_A"]);
-        assert_eq!(12, export[&metric_a.name()]["TOKEN_B"]);
+
+        let metric_a_token_a_b = export[&metric_a.name()]["B,A"].clone();
+        let metric_a_token_1_2 = export[&metric_a.name()]["2,1"].clone();
```
dlachaume, line 232:
```suggestion
        let metric_a_token_a_b = export[&metric_a.name()]["B,A"];
        let metric_a_token_1_2 = export[&metric_a.name()]["2,1"];
```

mithril-aggregator/src/http_server/routes/certificate_routes.rs
```diff
@@ -190,23 +194,27 @@ mod tests {
         let metric = dependency_manager
             .metrics_service
             .get_certificate_detail_total_served_since_startup();
-        let initial_counter_value_with_tag = metric.get(&["TEST"]);
-        let initial_counter_value_with_unknown_tag = metric.get(&["UNKNOWN"]);
+        let initial_counter_value_with_tag = metric.get(&["TEST", "CLI"]);
+        let initial_counter_value_with_unknown_tag = metric.get(&["UNKNOWN", "UNKNOWN"]);
```
Alenar: The test would be stronger with differentiated values (+ adapt usage below):
```suggestion
        let initial_counter_value_with_unknown_tag = metric.get(&["UNKNOWN_ORIGIN", "UNKNOWN_CLIENT_TYPE"]);
```

mithril-client-cli/src/commands/mod.rs
```diff
@@ -17,6 +17,7 @@ use mithril_client::{ClientBuilder, MithrilResult};
 
 use crate::configuration::ConfigParameters;
 
+const CLIENT_TYPE_CLI: &str = "CLI";
 /// Shared arguments for all commands
```
Alenar:
```suggestion
const CLIENT_TYPE_CLI: &str = "CLI";

/// Shared arguments for all commands
```

mithril-client/src/client.rs
```diff
@@ -343,49 +346,65 @@ impl ClientBuilder {
...
-        mut self,
+        self,
         aggregator_client: Arc<dyn AggregatorClient>,
     ) -> ClientBuilder {
-        self.aggregator_client = Some(aggregator_client);
-        self
+        Self {
+            aggregator_client: Some(aggregator_client),
+            ..self
+        }
```
Alenar: This realocate a value, I think that previous behavior using `mut` is better (same for usages below).

mithril-aggregator/src/services/usage_reporter.rs
```diff
@@ -344,39 +382,45 @@ mod tests {
...
-            assert_eq!(
-                expected,
-                UsageReporter::compute_metrics_delta_with_label(&metrics_before, &metrics_after)
-            );
-        }
+        //     assert_eq!(
+        //         expected,
+        //         UsageReporter::compute_metrics_delta_with_label(&metrics_before, &metrics_after)
+        //     );
+        // }
```
Alenar: Should this test be reinstated ?

mithril-aggregator/src/services/usage_reporter.rs
```diff
@@ -386,32 +430,63 @@ mod tests {
...
+                    &[("origin_1", build_metric_with_origin("CLI", u32::MAX))]
+                )]),
                 UsageReporter::compute_metrics_delta_with_label(&metrics_at_0, &metrics_at_max)
             );
-            assert_eq!(
-                build_hashmap(&[("metric_a", &[("origin_1", -(u32::MAX as i64))])]),
-                UsageReporter::compute_metrics_delta_with_label(&metrics_at_max, &metrics_at_0)
-            );
+
+            // TODO Does this test make sense? It seems to be testing a negative delta which is not typical for metrics since we can only increment them.
```
Alenar: If the code allow it, yes I think we should reinstate it, this ensure that we can handle extreme cases such as external data manipulation or case such as metrics system "rollback".

mithril-client/src/client.rs
turmelclem: what do you think @jpraynaud ? ( @Alenar https://github.com/input-output-hk/mithril/pull/2546#discussion_r2126110030 )

mithril-aggregator/src/services/usage_reporter.rs
turmelclem: that was my interrogation, that mean we thinks metrics can be negative, (but technically, only be incremented).
since delta of before/after metrics rebuild a map, attribute `counter` from `MetricLabelValueMap` will have to be unsigned (i32)

mithril-aggregator/src/services/usage_reporter.rs
turmelclem: ok, good point

mithril-client/src/client.rs
jpraynaud: It's not a problem as it is only used once to build the client.

Otherwise, you can do the mutation of the consumed self in the function.

mithril-aggregator/src/services/usage_reporter.rs
turmelclem: fixed by  switching to i32 to allow negative delta metrics

mithril-client/src/client.rs
turmelclem: do we agree to keep function with self mutable ? @jpraynaud @Alenar

dlachaume (approved):

mithril-client/src/client.rs
turmelclem: as discussed, we keep the self arg mutable

PR #2572, opened 2025-06-13
feat: revise OpenAPI compatibility enforcement

### LINKED ISSUE #2535: Loose enforcement of OpenAPI compatibility
## Why
Currently if the minor version of the OpenAPI is bumped, all the previous versions of the signer, client and aggregator will fail to retrieve information from an updated aggregator. This is due to the aggregator which strictly enforces the compatibility of the versions of OpenAPI between the nodes.

## What
Replace the strict OpenAPI verification mechanism with a loose one.

## How

- [x] Do not reject a request made by a client running a possibly incompatible version of the OpenAPI in the aggregator
- [x] In the aggregator client display a warning message when versions looks incompatible:
  - [x] In the signer
  - [x] In the aggregator (when follower)
  - [x] In the client library

dlachaume:
## Content

This PR removes the HTTP response status code `412 Precondition Failed` returned by the aggregator on OpenAPI version mismatch and adds a client-side warning log when a mismatch is detected.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation
  - [x] No new TODOs introduced

## Issue(s)

Closes #2535

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ As a minor extension, we have adopted a slightly different versioning convention
 
 - `cardano_database_v2` in the `mithril-client` library is now stable
 
+- Support for loose enforcement of OpenAPI compatibility: a warning is displayed when a call to an aggregator may be incompatible.
+
 - Crates versions:
 
 | Crate | Version |
--- a/mithril-aggregator/src/services/aggregator_client.rs
+++ b/mithril-aggregator/src/services/aggregator_client.rs
@@ -22,7 +22,7 @@ use crate::message_adapters::FromEpochSettingsAdapter;
 const JSON_CONTENT_TYPE: HeaderValue = HeaderValue::from_static("application/json");
 
 const API_VERSION_MISMATCH_WARNING_MESSAGE: &str =
-    "OpenAPI API version mismatch, please update your Mithril aggregator node.";
+    "OpenAPI version may be incompatible, please update your Mithril node to the latest version.";
 
 /// Error structure for the Aggregator Client.
 #[derive(Error, Debug)]
@@ -195,28 +195,30 @@ impl AggregatorHTTPClient {
 
     /// Check API version mismatch and log a warning if the leader aggregator's version is more recent.
     fn warn_if_api_version_mismatch(&self, response: &Response) {
-        let leader_aggregator_version_in_header = response
+        let leader_version = response
             .headers()
             .get(MITHRIL_API_VERSION_HEADER)
             .and_then(|v| v.to_str().ok())
             .and_then(|s| Version::parse(s).ok());
 
-        if let Some(leader_aggregator_version) = leader_aggregator_version_in_header {
-            let aggregator_version = match self.api_version_provider.compute_current_version() {
-                Ok(version) => version,
-                Err(error) => {
-                    error!(self.logger, "Failed to compute the current aggregator API version"; "error" => error.to_string());
-                    return;
-                }
-            };
+        let follower_version = self.api_version_provider.compute_current_version();
 
-            if aggregator_version.lt(&leader_aggregator_version) {
+        match (leader_version, follower_version) {
+            (Some(leader), Ok(follower)) if follower < leader => {
                 warn!(self.logger, "{}", API_VERSION_MISMATCH_WARNING_MESSAGE;
-                    "leader aggregator version" => leader_aggregator_version.to_string(),
-                    "aggregator version" => aggregator_version.to_string()
+                    "leader_aggregator_version" => %leader,
+                    "aggregator_version" => %follower,
                 );
             }
-        };
+            (Some(_), Err(error)) => {
+                error!(
+                    self.logger,
+                    "Failed to compute the current aggregator API version";
+                    "error" => error.to_string()
+                );
+            }
+            _ => {}
+        }
     }
 }
 
@@ -517,8 +519,7 @@ mod tests {
         fn version_provider_with_open_api_version<V: Into<String>>(
             version: V,
         ) -> APIVersionProvider {
-            let mut version_provider =
-                APIVersionProvider::new(Arc::new(DummyApiVersionDiscriminantSource::default()));
+            let mut version_provider = version_provider_without_open_api_version();
             let mut open_api_versions = HashMap::new();
             open_api_versions.insert(
                 "openapi.yaml".to_string(),
@@ -555,17 +556,17 @@ mod tests {
         ) {
             assert!(log_inspector.contains_log(API_VERSION_MISMATCH_WARNING_MESSAGE));
             assert!(log_inspector.contains_log(&format!(
-                "leader aggregator version={}",
+                "leader_aggregator_version={}",
                 leader_aggregator_version.into()
             )));
             assert!(log_inspector
-                .contains_log(&format!("aggregator version={}", aggregator_version.into())));
+                .contains_log(&format!("aggregator_version={}", aggregator_version.into())));
         }
 
         #[test]
         fn test_logs_warning_when_leader_aggregator_api_version_is_newer() {
-            let leader_aggregator_version = "1.0.0";
-            let aggregator_version = "0.0.999";
+            let leader_aggregator_version = "2.0.0";
+            let aggregator_version = "1.0.0";
             let (logger, log_inspector) = TestLogger::memory();
             let version_provider = version_provider_with_open_api_version(aggregator_version);
             let mut client = setup_client("whatever");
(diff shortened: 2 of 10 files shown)


mithril-client/src/aggregator_client.rs
```diff
@@ -453,6 +407,35 @@ impl AggregatorHTTPClient {
 
         AggregatorClientError::RemoteServerTechnical(anyhow!("{server_error}"))
     }
+
+    /// Check API version mismatch and log a warning if the aggregator's version is more recent.
+    async fn warn_if_api_version_mismatch(&self, response: &Response) {
```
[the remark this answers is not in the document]
dlachaume: I thought about deduplicating the code, but I was not sure, especially given that the previous `handle_api_error` function was already duplicated across crates.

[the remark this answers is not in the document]
dlachaume: I thought about deduplicating the code, but I preferred not to, especially given that the previous handle_api_error function was already duplicated across clients.

mithril-aggregator/src/services/aggregator_client.rs
```diff
@@ -201,20 +193,30 @@ impl AggregatorHTTPClient {
...
+                Err(error) => {
+                    error!(self.logger, "Failed to compute the current aggregator API version"; "error" => error.to_string());
+                    return;
+                }
+            };
+
+            if aggregator_version.lt(&leader_aggregator_version) {
+                warn!(self.logger, "{}", API_VERSION_MISMATCH_WARNING_MESSAGE;
+                    "leader aggregator version" => leader_aggregator_version.to_string(),
+                    "aggregator version" => aggregator_version.to_string()
```
Alenar: We usually avoid spaces in logs keys, and I think that you don't need to use `.to_string()` since `Version` implement Display.
_(same comment for similar logs on other crates)_
```suggestion
                    "leader_aggregator version" => &leader_aggregator_version,
                    "aggregator_version" => &aggregator_version
```

mithril-aggregator/src/services/aggregator_client.rs
```diff
@@ -514,4 +506,184 @@ mod tests {
...
+            client.api_version_provider = Arc::new(version_provider);
+            client.logger = logger;
+            let response = build_fake_response_with_header(MITHRIL_API_VERSION_HEADER, version);
+
+            client.warn_if_api_version_mismatch(&response);
+
+            assert!(!log_inspector.contains_log(API_VERSION_MISMATCH_WARNING_MESSAGE));
+        }
+
+        // TODO: Not sure about this test, it doesn't make much sense from a business perspective.
```
Alenar: it makes sense for me since in the future there may be leader aggregators that are not operated by us.

mithril-aggregator/src/services/aggregator_client.rs
```diff
@@ -20,6 +21,9 @@ use crate::message_adapters::FromEpochSettingsAdapter;
 
 const JSON_CONTENT_TYPE: HeaderValue = HeaderValue::from_static("application/json");
 
+const API_VERSION_MISMATCH_WARNING_MESSAGE: &str =
+    "OpenAPI API version mismatch, please update your Mithril aggregator node.";
```
jpraynaud: Here and for all other occurrences, you can use this message:
```suggestion
    "OpenAPI version may be incompatible, please update your Mithril node to the latest version.";
```

mithril-aggregator/src/services/aggregator_client.rs
```diff
@@ -201,20 +193,30 @@ impl AggregatorHTTPClient {
...
-        }
+    /// Check API version mismatch and log a warning if the leader aggregator's version is more recent.
+    fn warn_if_api_version_mismatch(&self, response: &Response) {
+        let leader_aggregator_version_in_header = response
+            .headers()
+            .get(MITHRIL_API_VERSION_HEADER)
+            .and_then(|v| v.to_str().ok())
+            .and_then(|s| Version::parse(s).ok());
+
+        if let Some(leader_aggregator_version) = leader_aggregator_version_in_header {
```
jpraynaud: Maybe it would be a bit more readable to match the tuple `(leader_version,follower_version)`?

mithril-aggregator/src/services/aggregator_client.rs
```diff
@@ -514,4 +506,184 @@ mod tests {
...
     }
+
+    mod warn_if_api_version_mismatch {
+        use std::collections::HashMap;
+
+        use mithril_common::test_utils::MemoryDrainForTestInspector;
+
+        use super::*;
+
+        fn version_provider_with_open_api_version<V: Into<String>>(
```
jpraynaud, line 519: Maybe you can build this one from the version without OpenAPI version?

mithril-aggregator/src/services/aggregator_client.rs
```diff
@@ -514,4 +506,184 @@ mod tests {
...
+                leader_aggregator_version.into()
+            )));
+            assert!(log_inspector
+                .contains_log(&format!("aggregator version={}", aggregator_version.into())));
+        }
+
+        #[test]
+        fn test_logs_warning_when_leader_aggregator_api_version_is_newer() {
+            let leader_aggregator_version = "1.0.0";
+            let aggregator_version = "0.0.999";
```
jpraynaud: You could assert that the versions order is correct here. Also maybe use versions `2.0.0` and `1.0.0` instead.

mithril-aggregator/src/services/aggregator_client.rs
jpraynaud: +1

mithril-client/src/aggregator_client.rs
```diff
@@ -996,4 +881,182 @@ mod tests {
...
+            );
+            client.logger = logger;
+            let response = build_fake_response_with_header(MITHRIL_API_VERSION_HEADER, version);
+
+            client.warn_if_api_version_mismatch(&response).await;
+
+            assert!(!log_inspector.contains_log(API_VERSION_MISMATCH_WARNING_MESSAGE));
+        }
+
+        // TODO: Not sure about this test, it doesn't make much sense from a business perspective.
```
jpraynaud: This test should be kept as well as in the future some aggregators could be running an older version.

mithril-aggregator/src/services/aggregator_client.rs
dlachaume: I've updated everything except Mithril client, since it's not a node. What do you think of this message:
`OpenAPI version may be incompatible, please update Mithril client library to the latest version.`

mithril-aggregator/src/services/aggregator_client.rs
dlachaume: Do you think this would be a bit clearer?
```rust
        let follower_version = self.api_version_provider.compute_current_version();

        match (leader_version, follower_version) {
            (Some(leader), Ok(follower)) if follower < leader => {
                warn!(self.logger, "{}", API_VERSION_MISMATCH_WARNING_MESSAGE;
                    "leader_aggregator_version" => %leader,
                    "aggregator_version" => %follower,
                );
            }
            (Some(_), Err(error)) => {
                error!(
                    self.logger,
                    "Failed to compute the current aggregator API version";
                    "error" => error.to_string()
                );
            }
            _ => {}
        }
 ```

Comment (jpraynaud) [reply mithril-aggregator/src/services/aggregator_client.rs]: Yes, it's better like this :+1:

File: mithril-aggregator/src/services/aggregator_client.rs (hunk 7 above)
Comment (dlachaume) [reply mithril-aggregator/src/services/aggregator_client.rs]: I'm good with using versions `2.0.0` and `1.0.0` instead 👍.
I'm not sure to understand the need for asserting the version ordering, could you explain?

File: mithril-aggregator/src/services/aggregator_client.rs (hunk 5 above)
Comment (dlachaume) [reply mithril-aggregator/src/services/aggregator_client.rs]: Thanks 🎉

File: mithril-aggregator/src/services/aggregator_client.rs (hunk 2 above)
Comment (dlachaume) [reply mithril-aggregator/src/services/aggregator_client.rs]: Done, thanks!

File: mithril-aggregator/src/services/aggregator_client.rs (hunk 6 above)
Comment (dlachaume) [reply mithril-aggregator/src/services/aggregator_client.rs:519]: You're right, done ✅

File: mithril-aggregator/src/services/aggregator_client.rs (hunk 7 above)
Comment (dlachaume) [reply mithril-aggregator/src/services/aggregator_client.rs]: You mean adding the assertion right after the declarations like this?
```rust
    assert!(
        Version::parse(leader_aggregator_version).unwrap()
            > Version::parse(aggregator_version).unwrap()
    );
```

jpraynaud: It'd better to make sure that the order we have is the one we expect (but you would need to casts strings to `SemVer` first to be able to compare versions).

dlachaume: Done ✅

CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ As a minor extension, we have adopted a slightly different versioning convention
 
 - `cardano_database_v2` in the `mithril-client` library is now stable
 
+- Loose enforcement of OpenAPI compatibility: the clients (signer node, aggregator follower, and client library) now logs a warning if the aggregator’s API version is more recent.
```
jpraynaud:
```suggestion
- Support for loose enforcement of OpenAPI compatibility: a warning is displayed when a call to an aggregator may be incompatible.
```

PR #2607, opened 2025-06-25
filip(fix): force re-render to initialise scrollY framer motion hooks

fstoqnov-iohk:
Add useEffects/stateful variables to enforce re-render and initialise useScroll framer motion hook which is causing issues on main site


jpraynaud (approved):
LGTM

PR #2612, opened 2025-06-26
filip(fix): rewrite scroll-based animation with pure css/js to exclude framer

fstoqnov-iohk:


jpraynaud (approved):

docs/website/src/css/custom.css
```diff
@@ -302,3 +302,21 @@ a.menu__link {
   perspective: 500px;
   transform-style: preserve-3d;
 }
+
+.grid-item {
```
Alenar: This class name is not specific enough, as such it may collides with existing or upcoming css. Using something like `why-mithril-grid-item` or `timeline-grid-item` would avoid the issue.

Issue #2534, opened 2025-06-02
Synchronize certificate chain of follower aggregator


jpraynaud:
## Why
We want the follower aggregator to retrieve an existing certificate chain from the leader aggregator when it does not have a valid certificate chain (empty or invalid). We need the mechanism to be able to detect that a re-genesis has occurred on the leader chain.

## What
Implement a synchronization mechanism for the follower certificate chain from the leader certificate chain.

## How

- [x] Synchronize the certificate chain from the leader when it is empty or invalid:
  - [x] Add a route to get the latest genesis certificate in the aggregator `/certificate/genesis`
    - [x] Extend the `MessageService` to retrieve the last genesis certificate (the most recent genesis certificate if multiple in the store) 
    - [x] The route must be declared before the  `/certificate/{certificate_hash}` route in the warp router declaration
    - [x] Implement a `certificate_certificate_genesis` handler (increment the metric `certificate_detail_total_served_since_startup`)
    - [x] Update OpenAPI
  - [x] Create a new `RemoteCertificateRetriever` trait in the aggregator (see code below)
  - [x] Implement the `RemoteCertificateRetriever` trait with reqwest (by pulling the aggregator certificate routes)
  - [x] Create a new `CertificateChainSynchronizer` trait that fetches a full certificate chain from the leader aggregator
  - [x] Implement the `RemoteCertificateChainSynchronizer` for the `CertificateChainSynchronizer` trait
  - [x] When the synchronization has taken place, the `RemoteCertificateChainSynchronizer` creates an entry on the open message for the `MithrilStakeDistribution` of the current epoch and mark it has certified (does not happen in case of error)
- [x] In the state machine of the follower aggregator, when transitioning from IDLE to READY, run the synchronization of the certificate chain:
  - [x] If the certificate chain in invalid
  - [x] Or if the remote certificate genesis has changed
  - [x] With the code snippet in `async fn try_transition_from_idle_to_ready` (see code below)
- [x] Add an integration test
- [x] Update the e2e test so that follower aggregators do not bootstrap their own genesis certificate

```rust
pub trait RemoteCertificateRetriever: Sync + Send {
    /// Get [Certificate] details
    async fn get_certificate_details(
        &self,
        certificate_hash: &str,
    ) -> Result<Certificate, CertificateRetrieverError>;
   
   // Get latest certificate
  async fn get_latest_certificate_details(
        &self,
    ) -> Result<Certificate, CertificateRetrieverError>;

[...]

PR #2615, opened 2025-06-27
test:enhance certificate chain builder and mock build

Alenar:
## Content

This PR enhance the certificate chain builder and add propose a tooling to create mock in a standardized way.

### Changes

- Rework `CertificateChainBuilder`:
  - Now `.build()` return a `CertificateChainFixture` that's contains the same data as the previous returned tuple (the chain ordered from latest to genesis and the genesis verifier)
  - The addition of `CertificateChainFixture` allow to attach utility methods related to the produced certificates:
    - `genesis_certificate(&self)`: shortcut to get the generated genesis
    - `chain_reversed`: return a copy of the generated chain reversed
    - `certificate_path_to_genesis`: return a list with the given certificate hash  and all it's previous certificates, useful when testing certificate chain validation
  - Additionally `CertificateChainFixture` implement:
    -  `Deref` and `DerefMut` to `[Certificate]` allowing to avoid directly operating on the stored `certificates_chained` list most of the time.
    - `impl<C: From<Certificate>> From<CertificateChainFixture> for Vec<C>` to allow conversion of the full list to a list of another type that can be converted from `Certificate` (mostly useful in db layer when handling `CertificateRecord`)
- Add `mock_extensions` module with a `MockBuilder` tool that allow to create mock instance with a block dedicated to the mock configuration, avoiding pollution of the test code, exemple:
```rust
let mock = MockBuilder::<MockType>::configure(|mock| {
    mock.expect_foo().returning(|| Bar::new());
});
``` 

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation
  - [x] No new TODOs introduced

## Issue(s)

Relates to #2534

### DIFF
--- a/mithril-common/src/test_utils/certificate_chain_builder.rs
+++ b/mithril-common/src/test_utils/certificate_chain_builder.rs
@@ -1,19 +1,20 @@
+use std::cmp::min;
+use std::collections::{BTreeSet, HashMap};
+use std::iter::repeat_n;
+use std::ops::{Deref, DerefMut};
+use std::sync::Arc;
+
 use crate::{
     certificate_chain::CertificateGenesisProducer,
     crypto_helper::{
         ProtocolAggregateVerificationKey, ProtocolClerk, ProtocolGenesisSigner,
         ProtocolGenesisVerifier, ProtocolParameters,
     },
-    entities::{Certificate, CertificateSignature, Epoch, ProtocolMessage, ProtocolMessagePartKey},
+    entities::{
+        CardanoDbBeacon, Certificate, CertificateMetadata, CertificateSignature, Epoch,
+        ProtocolMessage, ProtocolMessagePartKey, SignedEntityType,
+    },
     test_utils::{fake_data, MithrilFixture, MithrilFixtureBuilder, SignerFixture},
-};
-
-use crate::entities::CertificateMetadata;
-use std::{
-    cmp::min,
-    collections::{BTreeSet, HashMap},
-    iter::repeat_n,
-    sync::Arc,
 };
 
 /// Genesis certificate processor function type. For tests only.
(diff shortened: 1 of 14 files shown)


mithril-common/src/test_utils/certificate_chain_builder.rs
```diff
@@ -96,6 +97,72 @@ pub enum CertificateChainingMethod {
...
+}
+
+impl CertificateChainFixture {
+    /// Return the genesis certificate of this chain
+    pub fn genesis_certificate(&self) -> &Certificate {
+        &self.certificates_chained[self.certificates_chained.len() - 1]
+    }
+
+    /// Return a copy of the chain but with reversed order (from genesis to last)
+    pub fn chain_reversed(&self) -> Vec<Certificate> {
```
dlachaume: What do you think about renaming this to `chain_from_genesis`? I find it more domain-specific and precise.
```suggestion
    pub fn chain_from_genesis(&self) -> Vec<Certificate> {
```

dlachaume (approved):
LGTM 🚀

mithril-common/src/test_utils/certificate_chain_builder.rs
jpraynaud: IMO it is more an iteration over a reversed chained (from oldest to newest), maybe `reversed_chain` is a better naming?

mithril-common/src/test_utils/certificate_chain_builder.rs
```diff
@@ -96,6 +97,72 @@ pub enum CertificateChainingMethod {
...
+        let mut hash_to_search = certificate_hash.as_ref().to_string();
+
+        while let Some(certificate) = self
+            .certificates_chained
+            .iter()
+            .find(|c| c.hash == hash_to_search)
+        {
+            subchain.push(certificate.clone());
+            hash_to_search = certificate.previous_hash.clone();
+        }
```
jpraynaud: It does not looks like a very efficient way to filter compute the path by  constantly re-scanning the whole list of certificates. Taking advantage of the fact that the list is sorted and detecting a change in epoch has likely a much better complexity.

mithril-common/src/test_utils/mock_extensions.rs
```diff
@@ -0,0 +1,61 @@
...
+    /// ```
+    /// use mithril_common::test_utils::mock_extensions::MockBuilder;
+    /// # #[derive(Default)] struct MockType {};
+    ///
+    /// // from the builder generic
+    /// let mock = MockBuilder::<MockType>::configure(|mock| {});
+    ///
+    /// // or from the closure parameter
+    /// let mock = MockBuilder::configure(|mock: &mut MockType| {});
+    /// ```
```
jpraynaud, line 28: It sounds like the way generic types and their inference work in Rust. Do we need to keep it?

jpraynaud (approved):
LGTM

mithril-common/src/test_utils/mock_extensions.rs
Alenar: You're right, I wrote this because of the frustration of not being able to make rust infer the type automatically from the context (even `let mock: Arc<MockType> = MockBuilder::configure(|mock| {});` doesn't work :cry: ).

Still this doc provides a light example on how to use the `MockBuilder`. I can reformulate it, keep it like that, or delete it, which do you prefer ?

mithril-common/src/test_utils/certificate_chain_builder.rs
Alenar: I renamed it `reversed_chain`

mithril-common/src/test_utils/mock_extensions.rs
jpraynaud: OK, let's keep it :+1:

mithril-common/src/test_utils/certificate_chain_builder.rs
Alenar: I reworked it to only scan the list once taking advantage of the list order as you suggested.

PR #2633, opened 2025-07-04
feat: add support for Mithril era transition in clients

### LINKED ISSUE #2619: Clients support Mithril era transition
## Why
We want the client to be able to support a Mithril era switch and thus avoid breaking changes when the client implementation is era specific. The client is not connected to the Cardano chain where the era markers are registered and it does not need to deal with hot switch from an era to another (when the signer and aggregator do).

## What
Align the era used by the client on the era advertised by the aggregator to which it is connected.

## How

- [ ] Leverage the existing era advertised by the aggregator in the `/status` route
- [ ] Compute the current supported era with an `EraFetcher` that can be used to create era-specific code:
  - [ ] Create the `EraFetcher` trait in the client library
  - [ ] Implement an `EraFetcher` which calls the aggregator with the aggregator client in the client library
  - [ ] Support the case where the fetched era is not known in the availablr at build time
  - [ ] Implement an `EraFetcher` which has a forced value and that can be used when the client CLI specifies a `--era` parameter
- [ ] Create a small prototype with a new era to make sure the system works as expected? 
- [ ] Update the documentation if needed
- [ ] Update the ADR if needed

dlachaume:
## Content

This PR includes changes to the `mithril-client` library and CLI to ease and reduce friction during Mithril era transitions.

### mithril-client library updates:
* **Add `MithrilEraClient` struct:** client wrapper around an `EraFetcher` to improve extensibility.
* **Add `EraFetcher` trait:**  defines an interface to retrieve the current Mithril era, returning a `FetchedEra`.
* **Introduce `FetchedEra` struct:** wraps a raw era string and provides conversion to a `SupportedEra`.
* **Implement `AggregatorHttpEraFetcher`:** `EraFetcher` implementation that retrieves the era from the aggregator’s `/status` endpoint.
* **Update `Client` API**: allows setting a custom `MithrilEraClient` and exposes the conversion to `SupportedEra`.

### mithril-client CLI updates:
* **Add `--era` global CLI argument:** allows users to override the era manually.
* **Implement `ForcedEraFetcher`:** a static `EraFetcher` that always returns the CLI-provided era.

### mithril-client WASM updates:
* **Add `fetch_current_mithril_era` function**

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation
  - [x] No new TODOs introduced

## Issue(s)

Closes #2619


docs/website/adr/004-mithril-network-update-strategy.md
```diff
@@ -92,3 +92,9 @@ sequenceDiagram
     Chain->>Old Node: Era change at Epoch XX
     Old Node->>User: 💀 unsupported Era, quit.
 ```
+
+### Client-side Era Awareness
```
jpraynaud:
```suggestion
### Client-side era awareness
```

docs/website/root/manual/develop/nodes/mithril-client.md
```diff
@@ -244,6 +244,8 @@ Options:
           Enable unstable commands
       --origin-tag <ORIGIN_TAG>
           Request origin tag
+      --era <ERA>
+          Override the current Mithril era with a specific one
```
jpraynaud:
```suggestion
          Override the Mithril era
```

mithril-client-cli/src/utils/forced_era_fetcher.rs
```diff
@@ -0,0 +1,22 @@
+use async_trait::async_trait;
+
+use mithril_client::{EraFetcher, FetchedEra, MithrilResult};
+
+pub struct ForcedEraFetcher {
```
jpraynaud, line 8: Some tests are likely missing in this sub module.

mithril-client/src/era_fetcher.rs
```diff
@@ -0,0 +1,233 @@
...
+                FetchedEra {
+                    era: SupportedEra::Pythagoras.to_string(),
+                }
+            },
+            fetched_era
+        );
+    }
+
+    #[tokio::test]
+    async fn returns_error_if_response_is_invalid_json() {
```
jpraynaud:
```suggestion
    async fn fetch_current_era_returns_error_if_response_is_invalid_json() {
```

mithril-client/src/era_fetcher.rs
```diff
@@ -0,0 +1,233 @@
...
+        let era_fetcher = AggregatorHttpEraFetcher::new(mock_aggregator_client);
+
+        era_fetcher
+            .fetch_current_era()
+            .await
+            .expect_err("Expected an error due to invalid JSON response");
+    }
+
+    #[tokio::test]
+    async fn returns_error_if_mithril_era_field_is_missing() {
```
jpraynaud:
```suggestion
    async fn fetch_current_era_returns_error_if_mithril_era_field_is_missing() {
```

mithril-client-cli/src/main.rs
```diff
@@ -91,6 +91,11 @@ pub struct Args {
     /// Request origin tag
     #[clap(long, global = true)]
     origin_tag: Option<String>,
+
+    /// Override the current Mithril era with a specific one
```
jpraynaud:
```suggestion
    /// Override the Mithril era
```

docs/website/adr/004-mithril-network-update-strategy.md
```diff
@@ -92,3 +92,9 @@ sequenceDiagram
...
+
+The Mithril client library and its consumers (e.g., Mithril client CLI) do not directly connect to a Cardano node and therefore cannot read Era Activation Markers from the blockchain themselves. Instead, they rely on the aggregator to determine the current Mithril era.
+
+To achieve this, the client fetches the current era from the aggregator's /status endpoint. This information represents the aggregator's reading of the most recent Era Activation Marker observed on-chain.
```
jpraynaud:
```suggestion
The Mithril clients do not have access to a Cardano node and therefore can not read the Era Activation Markers stored on chain. As a consequence, they rely on the current era run by the aggregator, by using the era advertised by its `/status` route.
```

mithril-client-cli/src/utils/forced_era_fetcher.rs
dlachaume: Apart from testing that the struct is instantiated with the value passed as a parameter, do you see any other cases?

mithril-client/src/client.rs
```diff
@@ -105,6 +106,7 @@ pub struct Client {
     cardano_database_client: Arc<CardanoDatabaseClient>,
     cardano_transaction_client: Arc<CardanoTransactionClient>,
     cardano_stake_distribution_client: Arc<CardanoStakeDistributionClient>,
+    era_fetcher: Arc<dyn EraFetcher>,
```
Alenar: I feel that we are missing a layer here, other clients are all concrete type that use configurable traits (in the builder) and we are making an exception.

I don't know if we ever will add more era features in the `mithril-client` but if we do so it would be better to have a `EraClient` struct that hold a `EraFetcher` trait even if this mean only doing an indirection for now.
This would give us a natural entry point to add more era features has users would only have to retrieve the existing era client to use the new methods (i.e. `client.era().do_new_feature();`).

To resume the changes:

1. rename `era_fetcher` module to `era`
2. add a `EraClient` struct in the `era` module, it's constructed from the existing `EraFetcher` and only have the `fetch_current` method (no need to repeat `era` in the name, as we do for other clients)
3. change the `Client` struct so it holds a `EraClient` instead of a `EraFetcher`

> [!NOTE]
> Maybe we want to name it `MithrilEraClient` / `client.mithril_era()` to make a clear distinction with the cardano era's.

Alenar (approved):
LGTM, but I've a concern on how era features will evolve in the client, see the comment below.

mithril-client/src/client.rs
dlachaume: Thank you for your feedback, I completely agree with this approach and I think `MithrilEraClient` would be a good name, as you suggested.

mithril-client-cli/src/utils/forced_era_fetcher.rs
dlachaume: I've added a test to ensure `fetch_current_era()` returns the value initially passed to `ForcedEraFetcher`.

mithril-client/src/client.rs
dlachaume: Done ✅

CHANGELOG.md
```diff
@@ -27,6 +27,8 @@ As a minor extension, we have adopted a slightly different versioning convention
 
 - Add pre-built Linux ARM binaries in the distribution for the signer, client CLI, and aggregator.
 
+- Support for Mithril era fetching in the in the `mithril-client` library and add a new `--era` global argument in the CLI to override the current Mithril era.
```
jpraynaud:
```suggestion
- Support for Mithril era transition in the client library, CLI and WASM.
```

Issue #2640, opened 2025-07-15
Implement a common aggregator client


jpraynaud:
## Why
We have some code that is duplicated for the aggregator client in all the crates that need to communicate with an aggregator (aggregator, signer, client). We want to remove the duplicated code for easier maintenance.

## What
Create a new internal crate for supporting the communications with an aggregator and publish it.

## How

- [ ] Create a new internal `mithril-aggregator-client` crate
- [ ] Abstract the requests to be executed by the client with a trait instead of the existing enums (as what has been done in the [`Query`](https://github.com/input-output-hk/mithril/blob/e52b52aa53672634d5367d194654ee25ba28f657/internal/mithril-persistence/src/sqlite/query.rs#L8) trait
- [ ] The client API must not have a breaking change: we need to support a mapping from `AggregatorRequest` enum variants to the new trait implementations
- [ ] Publish manually the first version of the crate on crates.io + setup crate owner (@jpraynaud)

Alenar:
# Existing features across clients

| Feature | Mithril-client (lib) | Mithril-Aggregator | Mithril-Signer |
|--------|:------:|:------:|:------:|
| custom header | ✔️ | ❌  | ❌ |
| feat | ✔️ | ❌  | |
| feat | ✔️ | ❌  | |
| feat | ✔️ | ❌  | |
| feat | ✔️ | ❌  | |
| feat | ✔️ | ❌  | |
| feat | ✔️ | ❌  | |

PR #2661, opened 2025-07-31
add shared aggregator client

Alenar:
## Content

This PR add a new `mithril-aggregator-client` internal crate that define a api allowing to fetch and send data through Http to a Mithril Aggregator.

This is a first step that define the internal mechanism but all available queries are yet to be implemented in order to replace existing clients in the aggregator, signer and client lib.

### Details

- new `mithril-aggregator-client` internal crate
  - Define a `AggregatorClient` struct that expose one method: `send`
  - This send method can take any object that implement the, crate internal, `AggregatorQuery` trait
  - The `AggregatorClient` is meant to be built using a builder pattern accessible from either `AggregatorClient::builder` or directly though the builder: `AggregatorClientBuilder::new(aggregator_url)`
  - The client have all features that are available on our three builders:
    - timeout
    - relay endpoint proxy
    - aggregator url sanitation
    - mithril api version header 
    - api version mismatch warning 
    - configurable additional header
    - error analysis
    - mithril specific 450 http code error handling (registration round not opened)
  - One query is defined as example, including tests: `CertificateDetailsQuery` to retrieve a certificate with a given hash or the latest genesis
- `mithril-common` - make `ApiVersionProviderTestExtensions` that defines three test utilities: `update_open_api_versions` (test method moved from `ApiVersionProvider` public api), `new_with_default_version({Version})`, and `new_failing`

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation
  - [ ] No new TODOs introduced

### Todo

A TODO is added because of an `unwrap` in `AggregatorClient.send`:
```rust
// Todo: error handling ? Reuse the version in `warn_if_api_version_mismatch` ?
let current_api_version = self.api_version_provider.compute_current_version().unwrap();
```
This unwrap is a strict copy from all three existing aggregator client behavior, this raise a question if a mithril api version, and by extension sending the `mithril-api-version` header, is mandatory or not. Especially since the behavior of `warn_if_api_version_mismatch`, that is used afterward when handling the response, in all three clients is to not crash if an error is raised when computing the version.

### Api design choice

The choice have been made of making the `AggregatorQuery` trait internal to the crate, this means childs crates **can't define custom queries**.
The idea behind this choice is that the public api of the crate avoid exposing types from third party, non-mithril, crates, but the `AggregatorQuery` mechanism need to pass around a `reqwest::Response`, so making this trait public would be exposing a "leaking" abstraction. 

We may decide instead to take that upfront if we think that defining queries on a child crate is useful, but since all messages that are available from a mithril aggregator are defined in the shared `mithril-common` crate right now there's no difficulty to define all queries in the `mithril-aggregator-client` crate.

## Issue(s)

Relates to #2640

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3928,7 +3928,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-aggregator"
-version = "0.7.77"
+version = "0.7.78"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -4201,7 +4201,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-common"
-version = "0.6.12"
+version = "0.6.13"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -4411,7 +4411,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-signer"
-version = "0.2.262"
+version = "0.2.263"
 dependencies = [
  "anyhow",
  "async-trait",
--- a/mithril-aggregator/Cargo.toml
+++ b/mithril-aggregator/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mithril-aggregator"
-version = "0.7.77"
+version = "0.7.78"
 description = "A Mithril Aggregator server"
 authors = { workspace = true }
 edition = { workspace = true }
--- a/mithril-common/Cargo.toml
+++ b/mithril-common/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mithril-common"
-version = "0.6.12"
+version = "0.6.13"
 description = "Common types, interfaces, and utilities for Mithril nodes."
 authors = { workspace = true }
 edition = { workspace = true }
--- a/mithril-signer/Cargo.toml
+++ b/mithril-signer/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mithril-signer"
-version = "0.2.262"
+version = "0.2.263"
 description = "A Mithril Signer"
 authors = { workspace = true }
 edition = { workspace = true }


README.md
```diff
@@ -80,6 +80,8 @@ This repository consists of the following parts:
 - [**Mithril signer**](./mithril-signer): the node of the **Mithril network** responsible for producing individual signatures that are collected and aggregated by the **Mithril aggregator**.
 
 - [**Internal**](./internal): the shared tools and API used by **Mithril** crates.
+  - [**Mithril aggregator client**](./internal/mithril-aggregator-client): a client to request data from a Mithril Aggregator, used by **Mithril network** nodes.
```
dlachaume: Maybe mention the library?
```suggestion
  - [**Mithril aggregator client**](./internal/mithril-aggregator-client): a client to request data from a Mithril Aggregator, used by **Mithril network** nodes and Mithril client library.
```

internal/mithril-aggregator-client/src/builder.rs
```diff
@@ -0,0 +1,122 @@
...
+        url
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn enforce_trailing_slash_for_aggregator_url() {
```
turmelclem, line 116: also test when url already end with a slash that the url is not modified ?

internal/mithril-aggregator-client/src/error.rs
```diff
@@ -0,0 +1,254 @@
...
+    #[error("json parsing failed")]
+    JsonParseFailed(#[source] StdError),
+
+    #[error("invalid endpoint")]
+    InvalidEndpoint(#[source] StdError),
+
+    /// No signer registration round opened yet
+    #[error("a signer registration round is not opened yet, please try again later")]
+    RegistrationRoundNotYetOpened(#[source] StdError),
+}
```
dlachaume: It might be worth standardizing the case of the error messages and starting each one with a capital letter.

internal/mithril-aggregator-client/src/client.rs
```diff
@@ -0,0 +1,587 @@
...
+                    > Version::parse(client_version).unwrap()
+            );
+
+            client.send(TestGetQuery).await.unwrap();
+
+            assert_api_version_warning_logged(&log_inspector, aggregator_version, client_version);
+        }
+
+        #[tokio::test]
+        async fn test_client_post_log_warning_if_api_version_mismatch() {
```
turmelclem: do we need to test api_version_mismatch on both GET and POST method ?

PR #2642, opened 2025-07-16
Enhance STM library readability - Golden tests

curiecrypt:
## Content

This PR includes **"golden" (de)serialization tests for several core types in the `mithril-stm` crate, ensuring that their JSON representations remain stable and correct**:
- [x] BLS multi signature:
  - [x] Proof of possession
  - [x] Signature
  - [x] Signing key
  - [x] Verification key
- [x] Participant:
  - [x] Initializer
- [x] Single signature:
  - [x] Signature
  - [x] Signature with registered party
- [x] Aggregate signature:
  - [x] Concatenation proof     

> [!NOTE]
> This is a preliminary work that will help avoid introducing breaking changes in the protocol when the fields of the structures are renamed.

## Pre-submit checklist

- Branch
  - [x] Tests are provided (if possible)
  - [x] Crates versions are updated (if relevant)
  - [x] CHANGELOG file is updated (if relevant)
  - [x] Commit sequence broadly makes sense
  - [x] Key commits have useful messages
- PR
  - [x] All check jobs of the CI have succeeded
  - [x] Self-reviewed the diff
  - [x] Useful pull request description
  - [x] Reviewer requested

## Issue(s)
Closes #2568

PR #2689, opened 2025-09-11
feat: support multiple proof systems in STM

### LINKED ISSUE #2680: Support multiple aggregation proof systems in STM library
## Why
The current STM library supports only the Concatenation proof system. In the future, we will need to support other proof systems (e.g. ALBA, SNARK, ...). This is why we want to evolve the STM library to seamlessly support multiple aggregation proof systems which can exist on the same certificate chain (e.g. to support the transition from a proof system to another).

## What
Extend the STM library to seamlessly support multiple aggregation proof systems.

## How

- [x] Refactor the STM library and upgrade the aggregate signature to an enum:
  - [x] The first variant is the `Concatenation` proof system
  - [x] Add a `Future` proof system and gate it behind the `future_proof_system` feature
- [x] Support the (de)serialization of the different variants in a common bytes encoding that can be extended (using proof  encoding prefix byte)
- [x] Support backward compatibility of the aggregate signatures (stored in the certificate chain and verified by non updated client/aggregator):
  - [x] Encoding an aggregate signature with the concatenation proof uses the `untagged` serde directive for backward compatibility
  - [x] All other variants of the aggregate signature do not use the `untagged` serde directive
  - [x] Remove `bincode` (de)serialization tests (golden tests are implemented in https://github.com/input-output-hk/mithril/pull/2642)?
- [x] Support for the selection of the proof system operated by an aggregator
- [x] Adapt e2e test
- [x] Adapt CI workflow
- [x] Adapt infrastructure
- [x] Update GitHub environment parameters (add `AGGREGATOR_AGGREGATE_SIGNATURE_TYPE`):
  - [x] `dev-preview`
  - [x] `dev-follower-preview`
  - [x] `dev-mainnet`
  - [x] `testing-preview` 
- [x] Run benchmarks and make sure there are no performance regression
- [x] Update documentation
- [x] Run successful backward compatibility test

jpraynaud:
## Content

This PR includes **support for multiple proof systems in the STM library**:

- Adds `AggregateSignatureType` enum with initial support for Concatenation proof system
- Refactors `AggregateSignature` from struct to enum to support multiple proof system implementations (a new proof is now implemented in a dedicated sub-module of the `proof` sub-module)
- Update of aggregator, e2e tests, infrastructure and CI to include the configuration of the aggregate signature type

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation
  - [x] No new TODOs introduced

## Issue(s)
Closes #2680


mithril-stm/src/aggregate_signature/mod.rs
```diff
@@ -356,9 +365,6 @@ mod tests {
 
             let bytes = initializer.to_bytes();
             assert!(Initializer::from_bytes(&bytes).is_ok());
-
-            let bytes = bincode::serde::encode_to_vec(&initializer, bincode::config::legacy()).unwrap();
-            assert!(bincode::serde::decode_from_slice::<Initializer,_>(&bytes, bincode::config::legacy()).is_ok())
```
turmelclem, line 361: does using the new Concatenation signature type have impact to check bytes ?

mithril-stm/src/aggregate_signature/signature.rs
```diff
@@ -1,210 +1,218 @@
...
-    /// * the lottery is indeed won by each one of them,
-    /// * the merkle tree path is valid,
-    /// * the aggregate signature validates with respect to the aggregate verification key
-    ///   (aggregation is computed using functions `MSP.BKey` and `MSP.BSig` as described in Section 2.4 of the paper).
+    /// Verify an aggregate signature
     pub fn verify(
         &self,
         msg: &[u8],
-        avk: &AggregateVerificationKey<D>,
+        avk: &AggregateVerificationKey<D>, //TODO: replace with a general verification key for the proof system
```
turmelclem: is it something to do now ? or a future ticket to track ?

turmelclem (approved):
LGTM 👍

mithril-stm/src/aggregate_signature/mod.rs
jpraynaud: These conversions were removed because they don't provide any value apart from testing the `bincode` library (also its serde implementation is not exhaustive and not compatible with some serde attributes such as `untagged`). There are golden tests implemented in another PR https://github.com/input-output-hk/mithril/pull/2642 that will check the JSON output done by serde).

mithril-stm/src/aggregate_signature/signature.rs
```diff
@@ -1,210 +1,218 @@
...
         msg: &[u8],
         avk: &AggregateVerificationKey<D>,
         parameters: &Parameters,
     ) -> Result<(), StmAggregateSignatureError<D>> {
-        let msgp = avk.get_merkle_tree_batch_commitment().concatenate_with_message(msg);
-        let (sigs, vks) = self.preliminary_verify(msg, avk, parameters)?;
-
-        BlsSignature::verify_aggregate(msgp.as_slice(), &vks, &sigs)?;
-        Ok(())
+        match self {
```
curiecrypt, line 99: A future proof system could require a different set of inputs to be verified. Is there a way to handle this case without causing a backward compatibility issue?

mithril-stm/src/aggregate_signature/signature.rs
```diff
@@ -1,210 +1,218 @@
...
+            AggregateSignature::Concatenation(proof) => Some(proof),
+            #[cfg(feature = "future_proof_system")]
+            AggregateSignature::Future => None,
         }
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
curiecrypt, line 200: Should we also include unit tests/proptests for some of the functions?
Even if we updated the tests in `mod.rs`, it might be worth trying all possible outputs of the functions, WDYT?

mithril-stm/src/aggregate_signature/mod.rs
```diff
@@ -21,13 +24,13 @@ mod tests {
...
     use rand_core::{RngCore, SeedableRng};
 
-    use crate::bls_multi_signature::BlsVerificationKey;
     use crate::merkle_tree::MerkleBatchPath;
     use crate::{
         AggregateSignature, AggregationError, BasicVerifier, Clerk, CoreVerifierError, Initializer,
         KeyRegistration, Parameters, Signer, SingleSignature, SingleSignatureWithRegisteredParty,
         Stake,
     };
+    use crate::{AggregateSignatureType, bls_multi_signature::BlsVerificationKey};
```
curiecrypt: Is there a specific reason for importing `AggregateSignatureType` here? Can we also do the following?
```rust
    use crate::bls_multi_signature::BlsVerificationKey;
    use crate::merkle_tree::MerkleBatchPath;
    use crate::{
        AggregateSignature, AggregateSignatureType, AggregationError, BasicVerifier, Clerk,
        CoreVerifierError, Initializer, KeyRegistration, Parameters, Signer, SingleSignature,
        SingleSignatureWithRegisteredParty, Stake,
    };
```

mithril-stm/src/aggregate_signature/clerk.rs
```diff
@@ -109,6 +91,7 @@ impl<D: Digest + Clone + FixedOutput> Clerk<D> {
     ///
     /// It returns an instance of `AggregateSignature`.
     #[deprecated(since = "0.5.0", note = "Use `aggregate_signatures` instead")]
```
curiecrypt, line 93: Do we need to update this deprecation message as well?

mithril-stm/src/aggregate_signature/mod.rs
```diff
@@ -459,28 +459,36 @@ mod tests {
...
-                    for index in sig_reg.sig.indexes.iter_mut() {
-                       *index %= clerk.params.k - 1
+                if let AggregateSignature::Concatenation(concatenation_proof) = aggr {
+                    for sig_reg in concatenation_proof.signatures.iter_mut() {
+                        for index in sig_reg.sig.indexes.iter_mut() {
+                        *index %= clerk.params.k - 1
+                        }
                     }
+                }else{
+                    panic!("Unexpected aggregate signature type");
```
curiecrypt: Is there a better way for handling this instead of `panic`?

mithril-stm/src/aggregate_signature/mod.rs
```diff
@@ -459,28 +459,36 @@ mod tests {
...
+                        index_list[0] += 1;
+                        MerkleBatchPath {
+                            values,
+                            indices: index_list,
+                            hasher: Default::default()
+                        }
+                    };
+                    concatenation_proof.batch_proof = batch_proof;
+                }else{
+                    panic!("Unexpected aggregate signature type");
```
curiecrypt: similarly

mithril-stm/README.md
```diff
@@ -140,6 +140,9 @@ match msig {
     Err(AggregationError::UsizeConversionInvalid) => {
         println!("Invalid usize conversion");
     }
+    Err(AggregationError::UnsupportedProofSystem(aggregate_signature_type)) => {
```
curiecrypt, line 146: In line 129, we still call the deprecated `aggregate_signatures` function. Not a big deal, but it's better to update.

curiecrypt (commented):
Left some minor comments, questions 😊

mithril-stm/src/aggregate_signature/signature.rs
jpraynaud: Fixed

mithril-stm/src/aggregate_signature/clerk.rs
jpraynaud: I don't think that it's necessary as `aggregate_signatures` already provides a deprecation message which was introduced after version `0.5.0`.

mithril-stm/src/aggregate_signature/mod.rs
jpraynaud: It's imported because it is used in the test below:
`let aggr_sig_type = AggregateSignatureType::Concatenation;`

I have updated the imports as you proposed 👍

mithril-stm/src/aggregate_signature/mod.rs
jpraynaud: It's OK to use panics in tests (e.g. `unwrap` are panics, same for `unreachable!`).

However, I have used the `to_concatenation_proof` function instead to make it more readable 👍

mithril-stm/src/aggregate_signature/signature.rs
jpraynaud: I guess that the only information that could change is the `AggregateVerificationKey` (and this would not be the case for all the proof systems). I'd rather have that modification done when we implement the SNARK-friendly version of STM (where we know this will be the case), but using a generic type should be enough to support that evolution, and it will not introduce a breaking change.

mithril-stm/README.md
jpraynaud: Nice catch, fixed

mithril-stm/src/aggregate_signature/mod.rs
jpraynaud: Same here, with the usage of `to_concatenation_proof` 👍

curiecrypt (approved):
LGTM 🚀

PR #2705, opened 2025-09-26
Fixing typo in bls PoP.

damrobi:
Fixing a typo in verify_proof_of_possession in mithril-stm/src/bls_multi_signature/verfication_key.rs.

### DIFF
--- a/mithril-stm/src/bls_multi_signature/verification_key.rs
+++ b/mithril-stm/src/bls_multi_signature/verification_key.rs
@@ -136,7 +136,7 @@ impl BlsVerificationKeyProofOfPossession {
     /// manually.
     // If we are really looking for performance improvements, we can combine the
     // two final exponentiations (for verifying k1 and k2) into a single one.
-    pub(crate) fn verify_proof_of_possesion(&self) -> Result<(), MultiSignatureError> {
+    pub(crate) fn verify_proof_of_possession(&self) -> Result<(), MultiSignatureError> {
         match self.vk.to_blst_verification_key().validate() {
             Ok(_) => {
                 let result = verify_pairing(&self.vk, &self.pop);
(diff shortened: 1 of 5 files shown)


mithril-stm/src/bls_multi_signature/verification_key.rs
```diff
@@ -164,9 +164,9 @@ impl BlsVerificationKeyProofOfPossession {
     /// manually.
     // If we are really looking for performance improvements, we can combine the
     // two final exponentiations (for verifying k1 and k2) into a single one.
-    #[deprecated(since = "0.5.0", note = "Use `verify_proof_of_possesion` instead")]
+    #[deprecated(since = "0.5.0", note = "Use `verify_proof_of_possession` instead")]
```
jpraynaud: This is probably not a good practice to update a deprecation with a breaking change after the version has been released, and regularly we should keep the `verify_proof_of_possesion` function with the typo and deprecate it. 

On the other hand, the `verify_proof_of_possession` is not public any more (when `check` was) so we can simply update the deprecation note and explain that this function is not available any more.

jpraynaud: @damrobi here is a suggestion for the replacement text:
```suggestion
    #[deprecated(since = "0.5.0", note = "The verification of the proof of possession is not part of the public API any more")]
```

damrobi: Alright I'll take care of that.
Concerning you're first comment, an appropriate way to do this change would be to let the `verify_proof_of_possesion` function there while marking it deprecated and to call the new `verify_proof_of_possession` inside to make sure there's no breaking change, right?

jpraynaud: It would have been needed only in the case where the `verify_proof_of_possession` is public. Now it is `pub(crate)` so it is not needed. You can just update the deprecation notice :+1:

Issue #2704, opened 2025-09-24
Provide Cardano database incremental snapshots needed for Amaru bootstrap


jpraynaud:
## Why
The Amaru team is using Mithril snapshots of the Cardano database for the bootstrap of the node. In particular, they need to use snapshots from `3` epochs in the past in order to reconstruct the stake distribution snapshots. The aggregator is not exposing enough entries in the routes that list the available snapshots.

## What
Provide access to entries from `3` previous epochs in the aggregator routes.

## How

- [x] Add a route `/artifact/cardano-database/epoch/{epoch}` that allows the filtering of the snapshots given an epoch number:
  - [x] with a maximum number of epochs in the past (5 epochs)
  - [x] support the expansion of `latest` value for epoch
  - [x] support the expansion of `latest-{offset}` value for epoch (`offset >= 0`)
- [x] Update the OpenAPI specs
- [x] Add a function to list artifacts by epoch in the client library
- [x] Extend the `cardano-db snapshot list --epoch $EPOCH` command in the client CLI (only for backend v2)
- [x] Test in the e2e test?
- [x] Update the client manual workflow in GitHub actions
- [x] Update the documentation
- [x] Extend the retention of the ledger state archives (ancillary) to the maximum window of epochs in the past in the storage infrastructure (update default value of `google_storage_bucket_max_age` in infra)
- [x] Assess the performance of the route:
  - [ ] ~~Do we need to add an index on the `beacon` field of the signed entity type?~~
  - [x] Or add an `epoch` column in the store?

## Other option (discarded)

- [ ] Extend the number of entries in the route that list the incremental Cardano database snapshots 
- [ ] Assess the performance impact on the aggregator
- [ ] Extend the retention of the ledger state archives (ancillary) to at least `3` epochs in the past in the storage infrastructure

jpraynaud:
Hi @jeluard can you confirm that this issue fits with the need for Amaru bootstrap?

jeluard:
It looks on spot! Thanks @jpraynaud

PR #2734, opened 2025-10-10
feat: add `/artifact/cardano-database/epoch/{epoch}` route + support epoch expansion from `latest` in some routes

Alenar:
## Content

This PR add two main features:

- add new `/artifact/cardano-database/epoch/{epoch}` route
- Allow usage of `latest` or `latest-{offset}` in routes that take an `epoch` parameters in their path

### Aggregator

- database: 
  - Schema: add `epoch` virtual column to the `signed_entity` table, computed from the json value in the `beacon` column. This column is indexed to allow fast filtering.
  - Queries: generalize `GetSignedEntityRecordQuery::cardano_stake_distribution_by_epoch` into `by_signed_entity_type_and_epoch`
  - `SignedEntityStorer`: add `get_last_signed_entities_by_type_and_epoch`
- message service: add `get_cardano_database_list_message_by_epoch` 
- http server:
  - add `/artifact/cardano-database/epoch/{epoch}` route
  - add `middleware::parameters::expand_epoch` helper. It parse the epoch if it's an integer or try to fetch its value from the epoch service if it's `latest{-offset}`.
  - make use `parameters::expand_epoch` in routes that allow epoch filtering (new route above, get `/cardano-stake-distribution/epoch/{epoch}`, `signers/registered/{epoch}`
  - Router: add `max_artifact_epoch_offset` to the `RouterConfig`, value hard-coded to `5` in the dependency injection layer.
  - Use `max_artifact_epoch_offset` as the upper limit when expanding epoch in artifacts routes

### Common

- Fake data: add methods that build only one value for each artifacts type (complement the existing methods that builds list of artifacts). This allow to avoid having to extract the value from the vector when you only need one.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation
  - [x] No new TODOs introduced

## Issue(s)

Relates to #2704

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3979,7 +3979,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-aggregator"
-version = "0.7.87"
+version = "0.7.88"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -4252,7 +4252,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-common"
-version = "0.6.21"
+version = "0.6.22"
 dependencies = [
  "anyhow",
  "async-trait",
--- a/mithril-aggregator/Cargo.toml
+++ b/mithril-aggregator/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mithril-aggregator"
-version = "0.7.87"
+version = "0.7.88"
 description = "A Mithril Aggregator server"
 authors = { workspace = true }
 edition = { workspace = true }
--- a/mithril-common/Cargo.toml
+++ b/mithril-common/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mithril-common"
-version = "0.6.21"
+version = "0.6.22"
 description = "Common types, interfaces, and utilities for Mithril nodes."
 authors = { workspace = true }
 edition = { workspace = true }
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -4,7 +4,7 @@ info:
   # `mithril-common/src/lib.rs` file. If you plan to update it
   # here to reflect changes in the API, please also update the constant in the
   # Rust file.
-  version: 0.1.53
+  version: 0.1.54
   title: Mithril Aggregator Server
   description: |
     The REST API provided by a Mithril Aggregator Node in a Mithril network.


mithril-aggregator/src/database/repository/signed_entity_store.rs
```diff
@@ -44,6 +44,14 @@ pub trait SignedEntityStorer: Sync + Send {
         total: usize,
     ) -> StdResult<Vec<SignedEntityRecord>>;
 
+    /// Get Cardano database signed entities by epoch
```
jpraynaud:
```suggestion
    /// Get signed entities by signed entity type and epoch
```

mithril-aggregator/src/dependency_injection/builder/mod.rs
```diff
@@ -449,6 +449,7 @@ impl DependenciesBuilder {
                     .configuration
                     .cardano_transactions_prover_max_hashes_allowed_by_request(),
                 cardano_db_artifacts_directory: self.get_cardano_db_artifacts_dir()?,
+                max_artifact_epoch_offset: 5,
```
jpraynaud: Maybe this could be a constant?

mithril-aggregator/src/http_server/routes/artifact_routes/cardano_database.rs
```diff
@@ -22,6 +23,21 @@ fn artifact_cardano_database_list(
         .and_then(handlers::list_artifacts)
 }
 
+/// GET /artifact/cardano-database/epoch/:epoch{-offset}
```
jpraynaud: It looks like there is a typo:
```suggestion
/// GET /artifact/cardano-database/epoch/:epoch
```

mithril-aggregator/src/http_server/routes/artifact_routes/cardano_database.rs
```diff
@@ -90,6 +107,46 @@ mod handlers {
...
+                    err.to_string(),
+                ));
+            }
+        };
+
+        if expanded_epoch.has_offset_greater_than(max_artifact_epoch_offset as u64) {
+            return Ok(reply::bad_request(
+                "invalid_epoch".to_string(),
+                format!(
+                    "offset greater than max configured: epoch:`{epoch}`, max offset:`{max_artifact_epoch_offset}`"
```
jpraynaud:
```suggestion
                    "offset greater than maximum allowed value: epoch:`{epoch}`, max offset:`{max_artifact_epoch_offset}`"
```

mithril-aggregator/src/http_server/routes/middlewares.rs
```diff
@@ -207,6 +208,91 @@ pub mod validators {
     }
 }
 
+pub mod parameters {
```
jpraynaud: This would probably deserve its own separate module.

mithril-aggregator/src/http_server/routes/middlewares.rs
```diff
@@ -207,6 +208,91 @@ pub mod validators {
...
+    ) -> StdResult<ExpandedEpoch> {
+        let epoch_str = epoch_str.to_lowercase();
+        if epoch_str == "latest" {
+            epoch_service
+                .read()
+                .await
+                .epoch_of_current_data()
+                .map(ExpandedEpoch::Latest)
+        } else if epoch_str.starts_with("latest-") {
+            let (_, offset_str) = epoch_str.split_at(7);
```
jpraynaud: This could be implemented with a regexp.

turmelclem: does something like `"latest-".len()` instead of `7` make it more understandable ?

mithril-aggregator/src/http_server/routes/middlewares.rs
```diff
@@ -360,4 +446,113 @@ mod tests {
...
+        #[tokio::test]
+        async fn error_if_given_epoch_is_not_a_number_nor_latest_nor_latest_minus_a_number() {
+            let epoch_service = fake_epoch_service(Epoch(78));
+            for invalid_epoch in [
+                "invalid",
+                "latest-",
+                "latest+",
+                "latest+0",
+                "latest+293",
+                "latest-invalid",
```
turmelclem: what about decimal ? "latest-1.0" will it succeed to cast it as number ?

mithril-aggregator/src/http_server/routes/middlewares.rs
Alenar: Yes, but as performance matter for an http endpoint and since the requirement is simple enough, I implemented this way.

> [!NOTE]
> To get the best performance while using the `regex` crate here, we would probably have [to cache the regex expression](https://docs.rs/regex/latest/regex/#avoid-re-compiling-regexes-especially-in-a-loop), this would add some complexity here which is avoided by the current design.

mithril-aggregator/src/http_server/routes/middlewares.rs
Alenar: Good catch, I will add a test case, but this should already be handled as the number is parsed as a `u64` which have no decimal.

turmelclem: Yes, I was just thinking, what will append if a "valid" decimal in a string is cast to a number. that why my example is 1.0 and not 1.2 😄

mithril-aggregator/src/dependency_injection/builder/mod.rs
Alenar: You mean that it should not be configurable at all even on the router or add a constant on this file only ?
If it's the former, it doesn't cost much to add this configuration on the router level, plus this add some flexibility for the tests and make the design more resilient to changes.

mithril-aggregator/src/http_server/routes/middlewares.rs
Alenar: I was worried that it would be inducing additional computations but [`str.len()`](https://doc.rust-lang.org/std/primitive.str.html#method.len) is `const` so it's computed at compile time :sunglasses: .

mithril-aggregator/src/dependency_injection/builder/mod.rs
jpraynaud: No, I mean that the value `5` should be defined as a `const` somewhere.

mithril-aggregator/src/http_server/routes/middlewares.rs
jpraynaud: I guess that the regexp compilation should not take more than few micro-seconds. A benchmark would have been valuable to clearly assess the impact on performances. 

However, the split on the `"latest-".len()` as suggested by @turmelclem does the job and is simpler, so let's skip the regexp :+1:

mithril-aggregator/src/dependency_injection/builder/mod.rs
Alenar: :ok_hand: , I added a `MAX_ARTIFACT_EPOCH_OFFSET` constant in the file.

Issue #2692, opened 2025-09-12
Decentralization of configuration parameters - Phase 1


jpraynaud:
## Why

We need to decentralize the configuration parameters of the Mithril network (protocol parameters and certification parameters - signed entity types and their specific configuration parameters e.g. `cardano_transactions_signing_config`).  In a first phase , we want to implement a full synchronization of the network based on the leader aggregator, and in a second phase have the configuration stored on the Cardano chain.

## What

Implement the **Phase 1** of the decentralization plan of the configuration parameters for Mithril networks.

## How

- [x] Create a new internal `internal/mithril-protocol-config` crate for the Mithril network configuration
- [x] Check if we need to publish the crate: not needed
- [x] Add the new trait `MithrilNetworkConfigurationProvider` 
- [x] Implement the **Local parameters** for the `MithrilNetworkConfigurationProvider` trait
- [x] Implement the **REST client** for the `MithrilNetworkConfigurationProvider` trait
- [x] Adapt the signer (including dependency injection)
- [x] Adapt the aggregator (including dependency injection):
  - [x] Leader
  - [x] Follower 
- [x] Integration tests
  - [x] Signer
  - [x] Leader aggregator
  - [x] Follower aggregator
- [x] End to end tests
- [x] Update documentation

## Fix for aggregator implementation

To gracefully support the new design, some modifications need to be done:

### Refactoring work
- [x] Rename `available_signed_entity_types` with `enabled_signed_entity_types`
- [x] Add `/protocol-configuration` route in aggregator to expose protocol configurations with epochs
- [x] Handle the new route in mithril-relay
- [x] Deprecate the protocol parameters and cardano transaction signing config in `EpochSettingsMessage`?
- [x] Update OpenAPI specifications
- [x] Make sure that information advertised by this route are synchronized with the `/epoch-settings` route
- [x] Update `get_network_configuration` to take an epoch as input
- [x] Update the HTTP client implementation to use the new route
- [x] Move handling discrepancies at the end of the dependency builder for the serve command

PR #2736, opened 2025-10-13
Decentralization of configuration parameters phase 1 - Aggregator local parameters

turmelclem:
## Content

This PR includes a rework of the retrieval of the network configuration parameters in the `mithril-aggregator` in order to prepare their decentralization.

#### Mithril-aggregator

Changes on both leader and follower:
- Rework `EpochService` to use the `MithrilNetworkConfigurationProvider` system from `internal/mithril-protocol-config`
  - `inform_epoch` now fetch its current/next/registration AggregatorEpochSettings from the configuration provider
  - rework how future epoch settings are inserted in db : instead of doing it at an offset of +2 in a `update_epoch_settings` that is run by the state machine after running `inform_new_epoch`, it's now done within `inform_epoch` right after retrieving the parameters from the configuration provider and at an offset of +1. This means that the parameters are now stored at the last moment right before their usage instead of ahead of time.
- Epoch setting store: use an `insert or ignore` instead of an `insert or replace` when storing the epoch settings. Now stored epoch settings are considered final, that's why we moved the time of their insertion to right before their usage.
- Rework `handle_discrepancies_at_startup` with the aim of making it ready for decentralization:
  - Source its data from the `MithrilNetworkConfigurationProvider` instead of the aggregator configuration
  - Last change + store change implies that data are now registered for the work epoch window (-1, 0, +1) instead of from -1 to +2
  - Run it later in the dependency injection: after building the `ServeCommandDependenciesContainer` instead of after building the epoch setting store. This changes limits its call to the `serve` command, previously other commands could call it even if they did not need it at all.
- Configuration:
  -  `protocol_parameters` and `cardano_transactions_signing_config` are now options, but still mandatory for a leader aggregator (the missing configuration error is now handled manually instead of automatically by the `config` crate)
  - Add optional `preload_security_parameter` with a default value of `2160`. Used to configure the transactions preloader instead of fetching the security parameter in the `cardano_transactions_signing_config` configuration.
- Single signature authenticator: log the inner error when the authentication fails, before no context were available in we could not know why an authentication failed
- Test:
  - strengthen `create_certificate_follower` integration test by making it check that the follower aggregator works without configured `protocol_parameters`
  - rework and simplify test tooling attach to the `ServeCommandDependenciesContainer`:
    - now longer stores data in the `epoch_settings` table, but only in the `signer` and `signer_registration` table. This means that the `epoch_settings` must have been filled beforehand (either by running handle discrepancies or manually).
    - update usages, notably in the certifier service tests 
    - remove now unused methods 

Leader aggregator specific: 
- Add `LocalMithrilNetworkConfigurationProvider`: a `MithrilNetworkConfigurationProvider` that fetch its data first from the `epoch_settings` table in the sqlite database, and if an entry is missing for an epoch, it fallback to the usual configuration parameters (`protocol_parameters` and `cardano_transactions_signing_config`)

Follower aggregator specific:
- Use `mithril-protocol-config::http::HttpMithrilNetworkConfigurationProvider` as its network configuration provider, fetching data from its configured leader aggregator

### Mithril-common & openapi

- `EpochSettingsMessage`:
  - make `signer_registration_protocol` field optional so future node won't fails when we remove it
  - deprecate `signer_registration_protocol` and `cardano_transactions_signing_config` fields

### Mithril-end-to-end

- Make `update_protocol_parameters` step leader only. Now it doesn't matter for the follower since it retrieve its configuration from the leader and no longer read its configuration, and keeping it introduced a flakiness since sometimes the follower aggregator restarted before the leader could restart its http server, making the handle discrepancies of the follower fails (since it do a call to the configuration provider).
- Wrap tailed logs and extracted errors in a `::group::` when running in github action. Disabled for now as there's an remaining issue to tackle first: when the e2e is retry the logs in those groups for the previous iteration are missing in the action output.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation
  - [x] No new TODOs introduced

## Comments

There's a know issue for the updated `handle_discrepancies_at_startup`: it run twice.
This is because the `ServeCommandDependenciesContainer` is also built twice, once for its purpose, a second time for the http server.
This is harmless as this will record twice the same epoch settings and the second recording will be ignored, but we should probably construct a `HttpServerDependenciesContainer` instead of reusing the one of the serve command.

# Issue(s)

Relates to #2692

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,11 +9,10 @@ As a minor extension, we have adopted a slightly different versioning convention
 
 ## Mithril Distribution [XXXX] - UNRELEASED
 
-- **UNSTABLE**:
+- Decentralization of the configuration parameters of Mithril networks:
   - Added the `/protocol-configuration/{epoch}` route to fetch aggregator configuration for a given epoch, `{epoch}` must be a number.
-
-  - Enhanced `MithrilNetworkConfigurationProvider` to return configuration with a window of three epoch.
-  - Adapt Signer to read configurations from HttpMithrilNetworkConfigurationProvider
+  - Refactor Signer and Aggregator (leader, follower) to read network configurations from a `MithrilNetworkConfigurationProvider`
+  - Support for reading network configurations from its leader aggregator for the follower aggregators.
 
 - Crates versions:
 


mithril-aggregator/src/configuration.rs
```diff
@@ -556,7 +569,11 @@ pub struct ServeCommandConfiguration {
 
     /// Cardano transactions signing configuration
     #[example = "`{ security_parameter: 3000, step: 120 }`"]
-    pub cardano_transactions_signing_config: CardanoTransactionsSigningConfig,
+    pub cardano_transactions_signing_config: Option<CardanoTransactionsSigningConfig>,
+
+    /// Blocks offset, from the tip of the chain, to exclude during the cardano transactions preload
+    /// `[default: 2160]`.
```
jpraynaud:
```suggestion
    /// Blocks offset, from the tip of the chain, to exclude during the Cardano transactions preload, defaults to 2160
    #[example = 2160]`.
```

mithril-aggregator/src/configuration.rs
```diff
@@ -981,6 +1003,9 @@ pub struct DefaultConfiguration {
     /// Cardano transactions signing configuration
     pub cardano_transactions_signing_config: CardanoTransactionsSigningConfig,
 
+    /// Blocks offset, from the tip of the chain, to exclude during the cardano transactions preload
```
jpraynaud:
```suggestion
    /// Blocks offset, from the tip of the chain, to exclude during the Cardano transactions preload
```

mithril-common/src/messages/epoch_settings.rs
```diff
@@ -9,8 +9,15 @@ pub struct EpochSettingsMessage {
     pub epoch: Epoch,
 
     /// Signer Registration Protocol parameters
-    #[serde(rename = "signer_registration_protocol")]
-    pub signer_registration_protocol_parameters: ProtocolParameters,
+    #[deprecated(
+        since = "0.6.27",
+        note = "Will be removed soon. use `mithril_protocol_config` crate instead."
```
jpraynaud:
```suggestion
        note = "Will be removed soon. use `mithril-protocol-config` crate instead."
```

mithril-common/src/messages/epoch_settings.rs
```diff
@@ -19,13 +26,16 @@ pub struct EpochSettingsMessage {
     pub next_signers: Vec<SignerMessagePart>,
 
     /// Cardano transactions signing configuration for the current epoch
+    #[deprecated(
+        since = "0.6.27",
+        note = "Will be removed soon. use `mithril_protocol_config` crate instead."
```
jpraynaud:
```suggestion
        note = "Will be removed soon. use `mithril-protocol-config` crate instead."
```

PR #2744, opened 2025-10-17
feat: add fetching cardano database snapshots by epoch in client and cli

Alenar:
## Content

This PR add new capabilities to the client library and the client-cli

- Client library:
  - `cardano_database_v2`: add new methods to fetch snapshots for a specific epoch, for the latest epoch, or for the latest epoch with an offset
  - `cardano_stake_distribution`: add new methods to fetch snapshots for the latest epoch, or for the latest epoch with an offset
- Client CLI - `cardano-database` command: add a new `--epoch` parameter to fetch snapshots for a specific epoch, for the latest epoch, or for the latest epoch with an offset (i.e. `--epoch 4`, `--epoch latest`, `--epoch latest-4`)

## Details

- doc website: update documentation for the client cli `cardano-database list` parameters table.
- end to end: add a new step that fetch the list of cardano db v2 snapshots for the `latest-5`

### `mithril-common`

- `Epoch`:
  - add new type `EpochSpecifier` to store user inputs in the form of `<number>`, `latest`, `latest-<number>` coming for either http routes path or the cli.
  - add `Epoch::parse_specifier`
  - implement `FromStr` to `Epoch` to allow child crates to do `"87".parse::<Epoch>()` instead of `"87".parse::<u64>().map(Epoch)`

### `mithril-client` library

- `cardano_stake_distribution` client:
  - add `get_for_latest_epoch` and `get_for_latest_epoch_with_offset`
  - isolate the error logic for list retrieval in order to simplify the tests 
- `cardano_database_v2` client:
  - add `list_by_epoch`, `list_for_latest_epoch`, and `list_for_latest_epoch_with_offset`
- aggregator client: add supports for the latest aggregator routes that fetch cardano database snapshots and cardano stake distribution for an epoch or `latest` or `latest-<offset>`

### `mithril-client-cli`

- `cardano-database` commands:
  - generalize to the module the `warn_unused_parameter_with_v1_backend` logic so it can be reused in other commands than `download`
  - `list` command:  add `--epoch` parameter which is parsed into a `EpochSpecifier` and is only honored if the backend is set to `v2` (else a warning is print)

### `mithril-client-wasm`

- add `get_cardano_stake_distribution_for_latest_epoch`
- add unstable `list_cardano_database_v2_per_epoch`, `list_cardano_database_v2_for_latest_epoch`
- both `***_for_latest_epoch` functions allow to pass a json object which contains an optional offset, i.e in js:
```javascript
const cardano_database_snapshots_for_latest_epoch_with_offset =
      await client.list_cardano_database_v2_for_latest_epoch({ offset: 3 });
``` 
- update `ci-test` node project to check the new functions

### `mithril-aggregator-fake`

- add `/artifact/cardano-database/epoch/{epoch}` route with support of `latest` and `latest-{offset}`
-  `/artifact/cardano-stake-distribution/epoch/{epoch}` route: add support of `latest` and `latest-{offset}`

### `mithril-build-script`

- new generated hashmap (indexed by epoch number): `cardano_database_snapshot_list_per_epoch`, `cardano_stake_distributions_per_epoch`
- new generated id list: `cardano_database_snapshot_epochs` (indexed by epoch number)
- changed `cardano_stake_distribution_epochs` key from `&str` to an epoch number

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation
  - [x] No new TODOs introduced

## Issue(s)

Relates to #2704

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,17 @@ As a minor extension, we have adopted a slightly different versioning convention
 
 - Support for multiple aggregate signature proof systems.
 
+- Mithril-Aggregator:
+  - Added a `/artifact/cardano-database/epoch/{epoch}` route to fetch the list of cardano database snapshots for a given epoch, `{epoch}` can be a number, `latest`, or `latest-{offset}`.
+  - Enhanced `/artifact/cardano-stake-distribution/epoch/{epoch}` route to support `latest` and `latest-{offset}` as `{epoch}` values.
+
+- Client library and WASM:
+  - Added listing of `cardano_database_v2` snapshots for a given epoch, the latest epoch, or the latest epoch with offset
+  - Added listing of `cardano_staked_distribution` snapshots for the latest epoch, or the latest epoch with offset
+
+- Client CLI:
+  - Added `--epoch` optional parameter to `cardano-database snapshot list` commands snapshots, the given value can be a number, `latest`, or `latest-{offset}`.
+
 - Crates versions:
 
 | Crate | Version |
@@ -21,7 +32,7 @@ As a minor extension, we have adopted a slightly different versioning convention
 
 ## Mithril Distribution [2537] - 2025-09-17
 
-- Client library, CI and WASM:
+- Client library, CLI and WASM:
   - Support for stable `cardano_database_v2` backend in the `mithril-client` library.
 
   - Support for stable `v2` backend of `cardano-db` command and decommission of the `cardano-db-v2` command in client CLI.


mithril-common/src/entities/epoch.rs
```diff
@@ -100,6 +106,32 @@ impl Epoch {
...
         self.0.abs_diff(other.0) > 1
     }
+
+    /// Parses the given epoch string into an `EpochSpecifier`.
+    ///
+    /// Accepted values are:
+    /// - a `u64` number
+    /// - `latest`
+    /// - `latest-{offset}` where `{offset}` is a `u64` number
+    pub fn parse_specifier(epoch_str: &str) -> StdResult<EpochSpecifier> {
```
jpraynaud: This would be natural that this responsibility belongs to the `EpochSpecifier`.

mithril-test-lab/mithril-aggregator-fake/src/shared_state.rs
```diff
@@ -169,11 +191,57 @@ impl AppState {
...
+
+        Ok(stake_distribution)
+    }
+
     /// return the list of Cardano database snapshots in the same order as they were read
     pub async fn get_cardano_database_snapshots(&self) -> StdResult<String> {
         Ok(self.cardano_database_snapshot_list.clone())
     }
 
+    /// return the list of Cardano database snapshots for a given epoch in the same order as they were read
```
jpraynaud, line 216: This explanation is a bit unclear.

jpraynaud (approved):
LGTM :+1:

mithril-common/src/entities/epoch.rs
Alenar: I did this like that in hope that callers would not have to import `EpochSpecifier` but ... they need the import anyway for pattern maching.
I moved this to a `parse` method directly on the specifier.

mithril-test-lab/mithril-aggregator-fake/src/shared_state.rs
Alenar: I guess that the unclear part is "in the same order as they were read", basically it try to explain that all data returned by the fake aggregator comes from configurable json strings that are mostly read and returned "as is" with minimal or no processing.

This explanation is on all other routes that returns data as list in this file.

We probably can make it a bit more clear but I have no idea on how to do that while keeping it brief.

.github/workflows/test-client.yml
```diff
@@ -250,6 +250,17 @@ jobs:
           ./mithril-client ${{ steps.prepare.outputs.debug_level }} --origin-tag CI cardano-db snapshot list
           echo "CARDANO_DATABASE_V2_SNAPSHOT_HASH=$(./mithril-client --origin-tag CI cardano-db snapshot list --json | jq -r '.[0].hash')" >> $GITHUB_ENV
 
+      - name: Cardano Database V2 Snapshot / list for latest epoch
+        if: steps.aggregator_capability_unix.outputs.cardano_database_v2_enabled == 'true' || steps.aggregator_capability_windows.outputs.cardano_database_v2_enabled == 'true'
+        shell: bash
+        working-directory: ./bin
+        run: |
+          ./mithril-client ${{ steps.prepare.outputs.debug_level }} --origin-tag CI cardano-db snapshot list --epoch latest
+          if [ $(./mithril-client ${{ steps.prepare.outputs.debug_level }} --origin-tag CI cardano-db snapshot list --epoch latest --json | jq 'lenght') -eq 0 ]; then
```
jpraynaud: Is it possible to run the command only once? 

Also, maybe it's safer to use `latest-1` to make sure that an artifact exists.

turmelclem (approved):
LGTM 👍

.github/workflows/test-client.yml
Alenar: :+1: I changed it to `latest-1`.

I can change it to a single run, but this won't follow the pattern that we use for all other steps where we use the command output. In those steps we run the command twice: once to show the output in the workflow run, the second one to store and/or test the value using the easier json output.

To still show the execution and capture the output I would have to use `tee` and change the output to json.

PR #2755, opened 2025-10-29
refactor: use shared aggregator client for signer and aggregrator

Alenar:
## Content

This PR includes a refactor of the `mithril-aggregator` and `mithril-signer` to make them use the shared `AggregatorHttpClient` defined in the `internal/mithril-aggregator-client` crate.

#### `mithril-aggregator-client`

- Add query needed by aggregator, signer, and `mithril-protocol-config` : `GetAggregatorFeaturesQuery`, `GetCertificatesListQuery`, `GetEpochSettingsQuery`, `GetProtocolConfigurationQuery`, `PostRegisterSignatureQuery`, and `PostRegisterSignerQuery`
- rename `CertificateDetailsQuery` to `GetCertificateQuery` so all queries names follow a `{Verb}{Subject}Query` scheme
- rename `AggregatorClient` to `AggregatorHttpClient`, same for the error and result types: this makes clears that those types are fully dedicated to http and can't be easily generalized.
- reorganize queries in `get` and `post` submodules instead of domain based submodules (like the previous `certificate`) as domain based naming would have result in a lot of submodules with two or three files maximum
- light refactor of the client buider to accommodate real usage in code: `with_api_version_provider` now take an `Arc` as parameter, `with_relay_endpoint` now take an option.
- add a `test_http_compression_is_enabled` that generate to easily add a test in child crates that check that http compression is enabled

#### `mithril-protocol-config`

- Refactor `HttpMithrilNetworkConfigurationProvider` to use a `mithril-aggregator-client::AggregatorHttpClient`

#### `mithril-aggregator`

- replace the locally defined AggregatorClient with a direct implementation of the business traits on the `mithril-aggregator-client::AggregatorHttpClient`

#### `mithril-signer`

- replace the locally defined AggregatorClient with a direct implementation of the business traits on the `mithril-aggregator-client::AggregatorHttpClient`
- simplify the `AggregatorClient` trait to use `StdResult` instead of `AggregatorClientError`

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation
  - [x] No new TODOs introduced

## Issue(s)

Relates to #2640

### DIFF
--- a/mithril-aggregator/Cargo.toml
+++ b/mithril-aggregator/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mithril-aggregator"
-version = "0.7.92"
+version = "0.7.93"
 description = "A Mithril Aggregator server"
 authors = { workspace = true }
 edition = { workspace = true }
--- a/mithril-signer/Cargo.toml
+++ b/mithril-signer/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mithril-signer"
-version = "0.2.274"
+version = "0.2.275"
 description = "A Mithril Signer"
 authors = { workspace = true }
 edition = { workspace = true }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3594,7 +3594,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-aggregator"
-version = "0.7.92"
+version = "0.7.93"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -3654,7 +3654,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-aggregator-client"
-version = "0.1.1"
+version = "0.1.2"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -4024,7 +4024,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-protocol-config"
-version = "0.1.1"
+version = "0.1.2"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -4102,7 +4102,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-signer"
-version = "0.2.274"
+version = "0.2.275"
 dependencies = [
  "anyhow",
  "async-trait",
--- a/internal/mithril-aggregator-client/Cargo.toml
+++ b/internal/mithril-aggregator-client/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mithril-aggregator-client"
-version = "0.1.1"
+version = "0.1.2"
 description = "Client to request data from a Mithril Aggregator"
 authors.workspace = true
 documentation.workspace = true
--- a/internal/mithril-protocol-config/Cargo.toml
+++ b/internal/mithril-protocol-config/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mithril-protocol-config"
-version = "0.1.1"
+version = "0.1.2"
 description = "Configuraton parameters for Mithril network"
 authors = { workspace = true }
 edition = { workspace = true }


internal/mithril-aggregator-client/src/external_trait_impls.rs
jpraynaud, line 1: Can you add some test for this implementation?

mithril-aggregator/Cargo.toml
```diff
@@ -43,13 +44,7 @@ paste = "1.0.15"
...
 rayon = { workspace = true }
 regex = "1.12.2"
-reqwest = { workspace = true, features = [
-    "default",
-    "gzip",
-    "zstd",
-    "deflate",
-    "brotli"
-] }
+reqwest = { workspace = true, features = ["default"] }
```
jpraynaud: Maybe we still need the traffic compression features which are used e.g. to retrieve the mapping of pool ids to the pool names by cexplorer.

mithril-signer/Cargo.toml
```diff
@@ -9,6 +9,10 @@ homepage = { workspace = true }
 license = { workspace = true }
 repository = { workspace = true }
 
+[package.metadata.cargo-machete]
+# reqwest is needed to enable its default features that we disable in the workspace (for http2 and tsl support)
```
jpraynaud: This comment lacks clarity and is a bit hard to understand.

mithril-signer/Cargo.toml
```diff
@@ -37,15 +42,7 @@ mithril-signed-entity-lock = { path = "../internal/signed-entity/mithril-signed-
...
-reqwest = { workspace = true, features = [
-    "default",
-    "stream",
-    "gzip",
-    "zstd",
-    "deflate",
-    "brotli"
-] }
-semver = { workspace = true }
+reqwest = { workspace = true, features = ["default"] }
```
jpraynaud: Do we still need the `reqwest` dependency in the signer?

Alenar: Yes, it's not used anymore in the signer code, but if we remove this line from the signer cargo toml then when building the node reqwest default features will be disabled (cf workspace `cargo.toml`) which would disable tls and http2 support (cf [reqwest default features](https://docs.rs/reqwest/latest/reqwest/#optional-features)).

Reqwest default features are disabled on the workspace level in order to let library users choose the feature set they need.

Btw since we build the binaries with a single build command in the CI, the default features of reqwest would still be enabled because of feature unification.

mithril-aggregator/Cargo.toml
Alenar: The compression is still enabled through the `enable-http-compression` feature of the `mithril-aggregator-client` dependency.
But I would not mind to keep those enabled directly as well in the aggregator.

Also I added a test in both signer and aggregator that check that the compression is enabled (the usual test that check that `accept-encoding` header is filled with `brotli`, `zstd`, `deflate`, and `gzip`)

mithril-signer/Cargo.toml
Alenar: Changed it to `# reqwest re-enables default features (HTTP/2, TLS) that are disabled in workspace`, is it better or should I find reformulate it again ?

mithril-signer/Cargo.toml
jpraynaud: OK :+1: 

Can we target only the features for HTTP/2 and TLS instead of `default` in that case?

Alenar: I can do that, but that would means that If reqwest change its default feature set, like to add http/3, we would have to enables them manually.

mithril-aggregator/Cargo.toml
Alenar: Following this morning discussion, I removed the `enable-http-compression` feature from the `mithril-aggregator-client` crate and instead the aggregator and signer are enabling the compression features through a dependency to `reqwest` directly (as it was before).

For `mithril-signer` this means importing a dependency that's is not used directly in its code which trigger a warning from `cargo machete`. I added an exclusion on the signer and documented the process in the `mithril-aggregator-client` readme.

mithril-signer/Cargo.toml
Alenar: see comment above: https://github.com/input-output-hk/mithril/pull/2755#discussion_r2486256573

PR #2761, opened 2025-11-03
Implementation of Schnorr signature module for mithril-stm.

### LINKED ISSUE #2756: Implement Schnorr signature scheme in STM
## Why
We need to implement a Schnorr signature scheme in the STM library to support the creation of SNARK Proofs.

## What
Create a new module in the STM library (e.g. `schnorr_signature`) and implement the necessary primitives to create the individual signatures.

## How

- [ ] Create a new `schnorr_signature` module
- [ ] Gate the module behind a `future_snark` feature
- [ ] Implement the primitives for creating a Schnorr signature:
  - [ ] TBD
- [ ] Implement unit tests
- [ ] Update documentation
- [ ] TBD

damrobi:
## Content

This PR includes a new module for `mithril-stm` containing an implementation for the Schnorr signature that will be used in the SNARK version of mithril.

## Pre-submit checklist

- Branch
- PR
  - [ ] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation
  - [x] No new TODOs introduced
- Finalization TODO:
  - [x] Change all Result into StmResult
  - [x] create new Error enum for Schnorr signatures and add specialized errors
  - [x] Provide context for errors
  - [x] Fix from_bytes() functions
  - [x] Add small doc comments to all public functions
  - [x] Shorten comment explanation for sign and verify functions
  - [x] Add tests to check all error types are correct/working 
  - [x] Refactor get_coordinates multiple function calls
  - [x] Investigate difference between Dusk and Midnight Poseidon

## Benchmark results

The preliminary benchmark results (on my machine) of the two signature schemes are the following.

For individual signature:
 - BLS: ~0.121ms
 - Schnorr: ~1.51ms
 - Comparison: ~12x

For individual verification:
 - BLS: ~0.750ms
 - Schnorr: ~1.52ms
 - Comparison: ~2x

The benchmark are using the Dusk library for running Poseidon. Dusk Poseidon hash:
 - For 1 scalar hashed: 0.0424ms
 - For 11 scalars hashed (same as in sign): 0.127ms

## Issue(s)

Closes #2756

### DIFF
--- a/mithril-stm/Cargo.toml
+++ b/mithril-stm/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mithril-stm"
-version = "0.6.0"
+version = "0.6.1"
 edition = { workspace = true }
 authors = { workspace = true }
 homepage = { workspace = true }
@@ -16,7 +16,7 @@ crate-type = ["lib", "cdylib", "staticlib"]
 [features]
 default = ["rug-backend"]
 rug-backend = ["rug/default"]
-num-integer-backend = ["num-bigint", "num-rational", "dep:num-traits"]
+num-integer-backend = ["dep:num-bigint", "dep:num-rational", "dep:num-traits"]
 benchmark-internals = [] # For benchmarking multi_sig
 future_proof_system = [] # For activating future proof systems
 future_snark = [
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4273,7 +4273,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-stm"
-version = "0.6.0"
+version = "0.6.1"
 dependencies = [
  "anyhow",
  "blake2 0.10.6",
--- a/docs/website/blog/2025-09-17-pre-built-linux-arm-binaries.md
+++ b/docs/website/blog/2025-09-17-pre-built-linux-arm-binaries.md
@@ -17,9 +17,9 @@ To provide a clear overview of supported platforms, the team has added a new 'Pl
 
 | Binary             | Linux x64 | Linux arm64 | macOS arm64 | Windows x64 |
 | ------------------ | :-------: | :---------: | :---------: | :---------: |
-| mithril-aggregator |    ✔     |   ✔ ⁽\*⁾   |     ⛔      |     ⛔      |
-| mithril-signer     |    ✔     |   ✔ ⁽\*⁾   |     ⛔      |     ⛔      |
-| mithril-client     |    ✔     |   ✔ ⁽\*⁾   |     ✔      |     ✔      |
+| mithril-aggregator |     ✔     |   ✔ ⁽\*⁾    |     ⛔      |     ⛔      |
+| mithril-signer     |     ✔     |   ✔ ⁽\*⁾    |     ⛔      |     ⛔      |
+| mithril-client     |     ✔     |   ✔ ⁽\*⁾    |      ✔      |      ✔      |
 
 ⁽\*⁾⚠️ Linux arm64 builds are provided on a best-effort basis and are not officially supported.
 
--- a/mithril-stm/CHANGELOG.md
+++ b/mithril-stm/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## 0.6.1 (11-27-2025)
+
+### Added
+
+- Added Schnorr signature modules.
+
 ## 0.6.0 (11-19-2025)
 
 ### Changed


mithril-stm/Cargo.toml
```diff
@@ -19,15 +19,23 @@ rug-backend = ["rug/default"]
...
 
 [dependencies]
+anyhow.workspace = true
 blake2 = "0.10.6"
 # Enforce blst portable feature for runtime detection of Intel ADX instruction set.
 blst = { version = "0.3.16", features = ["portable"] }
 digest = { workspace = true }
+ff = "0.13.1"
+group = "0.13.0"
+midnight-circuits = { git = "https://github.com/midnightntwrk/midnight-zk", rev = "c88a50c2169f060120a52ad0980de90f08bc9535" }
```
curiecrypt: I see that @kitounliu also used this revision. Don't we get the latest version of the midnight repo? 
If that is not the case, can you write a comment for these revisions, giving the date of the rev and the reason for using this specific rev.

mithril-stm/src/schnorr_signature/mod.rs
```diff
@@ -0,0 +1,168 @@
+#![allow(dead_code)]
```
curiecrypt:
```suggestion
/// TODO: Remove
#![allow(dead_code)]
```

mithril-stm/src/schnorr_signature/mod.rs
```diff
@@ -0,0 +1,168 @@
...
+};
+use sha2::{Digest, Sha256};
+
+use anyhow::{Result, anyhow};
+
+mod signature;
+mod signing_key;
+mod verification_key;
+
+/// A DST to distinguish between use of Poseidon hash
```
curiecrypt:
```suggestion

use signature::*;
use signing_key::*;
use verification_key::*;

/// A DST to distinguish between use of Poseidon hash
```
By doing so, we can simplify the imports in the files. In the future, we might make these uses public if needed.

mithril-stm/src/schnorr_signature/signing_key.rs
```diff
@@ -0,0 +1,154 @@
...
+
+use group::Group;
+
+use crate::schnorr_signature::{
+    DST_SIGNATURE, JubjubHashToCurve, get_coordinates, hash_msg_to_jubjubbase,
+    jubjub_base_to_scalar,
+};
+use crate::schnorr_signature::{
+    signature::SchnorrSignature, verification_key::SchnorrVerificationKey,
+};
```
curiecrypt:
```suggestion
use crate::schnorr_signature::{SchnorrSignature, SchnorrVerificationKey};
```

mithril-stm/src/schnorr_signature/signing_key.rs
```diff
@@ -0,0 +1,154 @@
...
+/// Schnorr Signing key, it is essentially a random scalar of the Jubjub scalar field
+#[derive(Debug, Clone)]
+pub(crate) struct SchnorrSigningKey(pub(crate) JubjubScalar);
+
+impl SchnorrSigningKey {
+    pub(crate) fn generate(rng: &mut (impl RngCore + CryptoRng)) -> Self {
+        SchnorrSigningKey(JubjubScalar::random(rng))
+    }
+
+    // TODO: Check if we want the sign function to handle the randomness by itself
```
curiecrypt: Let's explain the signature scheme in the function doc.

mithril-stm/src/schnorr_signature/signing_key.rs
```diff
@@ -0,0 +1,154 @@
...
+        // First hashing the message to a scalar then hashing it to a curve point
+        let hash = JubjubHashToCurve::hash_to_curve(&[hash_msg_to_jubjubbase(msg)?]);
+
+        // sigma = H(msg) * sk
+        let sigma = hash * self.0;
+
+        // Compute the random part of the signature with
+        // r1 = H(msg) * r
+        // r2 = g * r
+        let r = JubjubScalar::random(rng);
```
curiecrypt: We should think of more descriptive variable names within the code. WDYT @jpraynaud ?

mithril-stm/src/schnorr_signature/verification_key.rs
```diff
@@ -0,0 +1,33 @@
...
+/// Schnorr verification key, it consists of a point on the Jubjub curve
+/// vk = g * sk, where g is a generator
+#[derive(Debug, Clone, Copy, Default)]
+pub struct SchnorrVerificationKey(pub(crate) JubjubSubgroup);
+
+impl From<&SchnorrSigningKey> for SchnorrVerificationKey {
+    /// Convert a Shnorr secret key into a verification key
+    /// This is done by computing `vk = g * sk` where g is the generator
+    /// of the subgroup and sk is the schnorr secret key
+    fn from(sk: &SchnorrSigningKey) -> Self {
```
curiecrypt: Similarly, we should use more descriptive names

mithril-stm/src/schnorr_signature/signature.rs
```diff
@@ -0,0 +1,88 @@
...
+
+use group::Group;
+
+use crate::{
+    Index,
+    schnorr_signature::{
+        DST_LOTTERY, DST_SIGNATURE, JubjubHashToCurve, get_coordinates, hash_msg_to_jubjubbase,
+        jubjub_base_to_scalar, verification_key::SchnorrVerificationKey,
+    },
+};
```
curiecrypt, line 15:
```suggestion
use crate::{
    Index,
    schnorr_signature::{
        DST_LOTTERY, DST_SIGNATURE, JubjubHashToCurve, SchnorrVerificationKey, get_coordinates,
        hash_msg_to_jubjubbase, jubjub_base_to_scalar,
    },
};
```

mithril-stm/src/schnorr_signature/signature.rs
```diff
@@ -0,0 +1,88 @@
...
+
+/// Structure of the Schnorr signature to use with the SNARK
+/// This signature includes a value `sigma` which depends only on
+/// the message and the signing key.
+/// This value is used in the lottery process to determine the correct indices.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) struct SchnorrSignature {
+    pub(crate) sigma: JubjubSubgroup,
+    pub(crate) s: JubjubScalar,
```
curiecrypt: Descriptive names : )

mithril-stm/Cargo.toml
```diff
@@ -19,15 +19,23 @@ rug-backend = ["rug/default"]
 num-integer-backend = ["num-bigint", "num-rational", "num-traits"]
 benchmark-internals = [] # For benchmarking multi_sig
 future_proof_system = [] # For activating future proof systems
+future_snark = [] # For activating snark features
 
 [dependencies]
+anyhow.workspace = true
```
jpraynaud:
```suggestion
anyhow = { workspace = true }
```

mithril-stm/Cargo.toml
jpraynaud: These dependencies should be used only if the `future_snark` feature is activated.

Also a very important point: we can't merge the PR until the midnight dependencies are published on crates.io.

mithril-stm/Cargo.toml
```diff
@@ -19,15 +19,23 @@ rug-backend = ["rug/default"]
...
 digest = { workspace = true }
+ff = "0.13.1"
+group = "0.13.0"
+midnight-circuits = { git = "https://github.com/midnightntwrk/midnight-zk", rev = "c88a50c2169f060120a52ad0980de90f08bc9535" }
+midnight-curves = { git = "https://github.com/midnightntwrk/midnight-zk", rev = "c88a50c2169f060120a52ad0980de90f08bc9535" }
+num-traits = "0.2.19"
 rand_core = { workspace = true }
 rayon = { workspace = true }
 serde = { workspace = true }
+sha2 = "0.10.9"
```
jpraynaud: Same comment here

mithril-stm/src/schnorr_signature/mod.rs
```diff
@@ -0,0 +1,173 @@
+// TODO: Remove
+#![allow(dead_code)]
```
jpraynaud: We want the code to be as TODO free as possible. 

You can use public visibility to avoid the dead code warnings that we don't want to keep when merged.

mithril-stm/src/schnorr_signature/mod.rs
```diff
@@ -0,0 +1,173 @@
...
+    ecc::{hash_to_curve::HashToCurveGadget, native::EccChip},
+    hash::poseidon::PoseidonChip,
+    types::AssignedNative,
+};
+use midnight_curves::{
+    Fq as JubjubBase, Fr as JubjubScalar, JubjubAffine, JubjubExtended, JubjubSubgroup,
+};
+use sha2::{Digest, Sha256};
+
+use anyhow::{Result, anyhow};
```
jpraynaud: You should separate the module declaration from the imports and put the imports after. It is usually a good practice to create a new module to do that.

mithril-stm/src/schnorr_signature/mod.rs
```diff
@@ -0,0 +1,173 @@
...
+
+mod signature;
+mod signing_key;
+mod verification_key;
+
+use signature::*;
+// use signing_key::*;
+use verification_key::*;
```
jpraynaud: You should separate the module declaration from the imports and put the imports after. It is usually a good practice to create a new module inline to do that(you can refer to other files in the repository). A better option would be to move all this code in a new separate module like `utils`.

mithril-stm/src/schnorr_signature/mod.rs
```diff
@@ -0,0 +1,173 @@
...
+
+/// A DST to distinguish between use of Poseidon hash
```
jpraynaud:
```suggestion
/// A DST (Domain Separation Tag) to distinguish between use of Poseidon hash
```

mithril-stm/src/schnorr_signature/mod.rs
```diff
@@ -0,0 +1,173 @@
...
+    JubjubBase,
+    JubjubExtended,
+    AssignedNative<JubjubBase>,
+    PoseidonChip<JubjubBase>,
+    EccChip<JubjubExtended>,
+>;
+
+/// Convert an arbitrary array of bytes into a Jubjub scalar field element
+/// First hash the message to 256 bits use Sha256 then perform the conversion
+/// TODO: Handle the unwrap properly
```
jpraynaud:
```suggestion
/// Convert an arbitrary array of bytes into a Jubjub scalar field element
///
/// First hash the message to 256 bits use Sha256 then perform the conversion
```

mithril-stm/src/schnorr_signature/mod.rs
```diff
@@ -0,0 +1,173 @@
...
+        return Err(anyhow!(
+            "Hash of the message does not have the correct lenght."
+        ));
+    }
+    Ok(JubjubBase::from_raw([
+        u64::from_le_bytes(output[0..8].try_into()?),
+        u64::from_le_bytes(output[8..16].try_into()?),
+        u64::from_le_bytes(output[16..24].try_into()?),
+        u64::from_le_bytes(output[24..32].try_into()?),
+    ]))
```
jpraynaud: Can you add a line break before returning in this function and others? This makes the code easier to read.

mithril-stm/src/schnorr_signature/mod.rs
```diff
@@ -0,0 +1,173 @@
...
+
+/// Convert an arbitrary array of bytes into a Jubjub scalar field element
+/// First hash the message to 256 bits use Sha256 then perform the conversion
+/// TODO: Handle the unwrap properly
+pub(crate) fn hash_msg_to_jubjubbase(msg: &[u8]) -> Result<JubjubBase> {
+    let mut hash = Sha256::new();
+    hash.update(msg);
+    let hmsg = hash.finalize();
+    let mut output = [0u8; 32];
+    // Adding a check here but this
```
jpraynaud:
```suggestion
```

mithril-stm/src/schnorr_signature/mod.rs
```diff
@@ -0,0 +1,173 @@
...
+    Ok(JubjubBase::from_raw([
+        u64::from_le_bytes(output[0..8].try_into()?),
+        u64::from_le_bytes(output[8..16].try_into()?),
+        u64::from_le_bytes(output[16..24].try_into()?),
+        u64::from_le_bytes(output[24..32].try_into()?),
+    ]))
+}
+
+/// Extract the coordinates of a given point
+/// This is mainly use to feed the Poseidon hash function
```
jpraynaud:
```suggestion
/// Extract the coordinates of a given point
///
/// This is mainly used to feed the Poseidon hash function
```

mithril-stm/src/schnorr_signature/mod.rs
```diff
@@ -0,0 +1,173 @@
...
+pub(crate) fn get_coordinates(point: JubjubSubgroup) -> (JubjubBase, JubjubBase) {
+    let extended = JubjubExtended::from(point); // Convert to JubjubExtended
+    let affine = JubjubAffine::from(extended); // Convert to JubjubAffine (affine coordinates)
+    let x = affine.get_u(); // Get x-coordinate
+    let y = affine.get_v(); // Get y-coordinate
+    (x, y)
```
jpraynaud: All of these comments are not needed and could be removed with variable names which are more explicit:
```suggestion
    let extended = JubjubExtended::from(point); 
    let affine = JubjubAffine::from(extended);
    let x = affine.get_u(); 
    let y = affine.get_v(); 
    (x, y)
```

mithril-stm/src/schnorr_signature/mod.rs
```diff
@@ -0,0 +1,173 @@
...
+    use rand_chacha::ChaCha20Rng;
+    use rand_core::SeedableRng;
+
+    use crate::schnorr_signature::{
+        signing_key::SchnorrSigningKey, verification_key::SchnorrVerificationKey,
+    };
+
+    // Testing conversion from arbitrary message to scalar field element
+    #[test]
+    fn test_hash_msg_to_jubjubbase() {
```
jpraynaud: There is no need to add comments here, the name of the test should be enough to understand what it does.

mithril-stm/src/schnorr_signature/mod.rs
```diff
@@ -0,0 +1,173 @@
...
+        let msg = vec![0, 0, 0, 1];
+        let h = hash_msg_to_jubjubbase(&msg).unwrap();
+        // Correct value corresponding to the message [0,0,0,1]
+        let bytes_le = [
+            179, 7, 17, 168, 141, 112, 57, 117, 112, 92, 169, 56, 36, 70, 1, 217, 9, 13, 255, 42,
+            100, 207, 166, 110, 188, 47, 35, 211, 35, 168, 100, 25,
+        ];
+        let field_elem = JubjubBase::from_bytes_le(&bytes_le).unwrap();
+        assert_eq!(h, field_elem)
```
jpraynaud: It's a good practice to follow the Arrange, Act, and Assert (AAA) pattern to make tests more readable (can you apply that pattern to other tests?):
```suggestion
        let msg = vec![0, 0, 0, 1];
        let h = hash_msg_to_jubjubbase(&msg).unwrap();
        let bytes_le = [
            179, 7, 17, 168, 141, 112, 57, 117, 112, 92, 169, 56, 36, 70, 1, 217, 9, 13, 255, 42,
            100, 207, 166, 110, 188, 47, 35, 211, 35, 168, 100, 25,
        ];
        
        let field_elem = JubjubBase::from_bytes_le(&bytes_le).unwrap();
        
        assert_eq!(h, field_elem)
```

mithril-stm/src/schnorr_signature/signature.rs
```diff
@@ -0,0 +1,116 @@
...
+    schnorr_signature::{
+        DST_LOTTERY, DST_SIGNATURE, JubjubHashToCurve, SchnorrVerificationKey, get_coordinates,
+        hash_msg_to_jubjubbase, jubjub_base_to_scalar,
+    },
+};
+
+/// Structure of the Schnorr signature to use with the SNARK
+/// This signature includes a value `sigma` which depends only on
+/// the message and the signing key.
+/// This value is used in the lottery process to determine the correct indices.
```
jpraynaud, line 21: Can you add a line break here and anytime you create a doc comment with multiple lines (here and in the rest of the PR) as this will reflect on the cargo documentation?
```suggestion
/// Structure of the Schnorr signature to use with the SNARK
///
/// This signature includes a value `sigma` which depends only on
/// the message and the signing key.
/// This value is used in the lottery process to determine the correct indices.
```

mithril-stm/src/schnorr_signature/signature.rs
```diff
@@ -0,0 +1,116 @@
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) struct SchnorrSignature {
```
jpraynaud: All the fields should have a doc comment in all the structs for better readability and documentation.

+1 on the non descriptive names mentioned by @curiecrypt .

mithril-stm/src/schnorr_signature/signature.rs
```diff
@@ -0,0 +1,116 @@
...
+    fn evaluate_dense_mapping(&self, msg: &[u8], index: Index) -> Result<[u8; 32]> {
+        let hash = JubjubHashToCurve::hash_to_curve(&[hash_msg_to_jubjubbase(msg)?]);
+        let (hashx, hashy) = get_coordinates(hash);
+        // TODO: Check if this is the correct way to add the index
+        let idx = JubjubBase::from_raw([0, 0, 0, index]);
+        let (sigmax, sigmay) = get_coordinates(self.sigma);
+        let ev =
+            PoseidonChip::<JubjubBase>::hash(&[DST_LOTTERY, hashx, hashy, idx, sigmax, sigmay]);
+        Ok(ev.to_bytes_le())
+    }
```
jpraynaud: I'm not convinced that this function belongs here :thinking:

mithril-stm/src/schnorr_signature/signing_key.rs
jpraynaud: Yes, maybe that could help readability. It may be difficult with mathematical symbols, in that case adding some comments can help.

mithril-stm/src/schnorr_signature/signature.rs
damrobi: I replicated what was done for the BLS signature but I agree that this function probably belongs somewhere else, maybe in the SNARK aggregator or in the SingleSignature?

mithril-stm/src/schnorr_signature/signature.rs
```diff
@@ -0,0 +1,160 @@
...
+    ///     - Use the recomputed values to recompute the Poseidon hash:
+    /// c_tilde = Poseidon(DST || H(Sha256(msg)) || vk || sigma || R1 || R2), where each elliptic curve  point is converted
+    /// to their coordinates representation to feed them to the hash function.
+    ///     - Check: c == c_tilde
+    ///     
+    pub fn verify(&self, msg: &[u8], vk: &SchnorrVerificationKey) -> Result<()> {
+        let generator = JubjubSubgroup::generator();
+
+        // First hashing the message to a scalar then hashing it to a curve point
+        let hash_msg = JubjubExtended::hash_to_point(msg);
```
curiecrypt: I would use `message_hash` or `msg_hash`

mithril-stm/src/schnorr_signature/signature.rs
```diff
@@ -0,0 +1,160 @@
...
+
+        // Computing R1 = H(msg) * s + sigma * c
+        let hash_msg_times_sig = hash_msg * self.signature;
```
curiecrypt: Consequently, `msg_hash_times_sig`.
In order to avoid ambiguities, prefer the full names. I would use `signature` instead of `sig`. Because it is hard to understand at first glance whether it is sigma or signature. 
So, final recommendation: `msg_hash_times_signature`

mithril-stm/src/schnorr_signature/signature.rs
```diff
@@ -0,0 +1,160 @@
...
+    /// The step are the following:
+    ///     - Compute the Hash of the message: H(Sha256(msg))
+    ///     - Recompute the value R1 as: r1_tilde = h * s + sigma * c
+    ///     - Recompute the value R2 as: r2_tilde = g * s + vk * c
+    ///     - Use the recomputed values to recompute the Poseidon hash:
+    /// c_tilde = Poseidon(DST || H(Sha256(msg)) || vk || sigma || R1 || R2), where each elliptic curve  point is converted
+    /// to their coordinates representation to feed them to the hash function.
+    ///     - Check: c == c_tilde
+    ///     
+    pub fn verify(&self, msg: &[u8], vk: &SchnorrVerificationKey) -> Result<()> {
```
curiecrypt:
```suggestion
    pub fn verify(&self, msg: &[u8], verification_key: &SchnorrVerificationKey) -> Result<()> {
```

mithril-stm/src/schnorr_signature/signature.rs
```diff
@@ -0,0 +1,160 @@
...
+        let generator = JubjubSubgroup::generator();
+
+        // First hashing the message to a scalar then hashing it to a curve point
+        let hash_msg = JubjubExtended::hash_to_point(msg);
+
+        // Computing R1 = H(msg) * s + sigma * c
+        let hash_msg_times_sig = hash_msg * self.signature;
+        let sigma_times_challenge = self.sigma * self.challenge;
+        let random_value_1_recomputed = hash_msg_times_sig + sigma_times_challenge;
```
curiecrypt: We can use `random_scalar_1`, `random_scalar_2`. WDYT? @damrobi @jpraynaud

mithril-stm/src/schnorr_signature/mod.rs
```diff
@@ -0,0 +1,154 @@
...
+
+    use super::*;
+    use dusk_jubjub::SubgroupPoint as JubjubSubgroup;
+    use group::Group;
+    use rand_chacha::ChaCha20Rng;
+    use rand_core::SeedableRng;
+
+    use crate::schnorr_signature::{
+        signing_key::SchnorrSigningKey, verification_key::SchnorrVerificationKey,
+    };
```
curiecrypt:
```suggestion
    use crate::schnorr_signature::{SchnorrSigningKey, SchnorrVerificationKey};
```
when you put `pub use signature::*;` then you can use the above import.

mithril-stm/src/schnorr_signature/mod.rs
curiecrypt: actually, I suggested this to be removed when we are finished with all tasks of this PR.

curiecrypt (commented):
You should use full names even in function parameters. For example `verification_key` instead of `vk`.
For simplicity and descriptiveness, we can use `random_scalar_1` instead of `random_value_1`. Consequently, we can call the `generator * random_scalar_1` as `random_point_1`.
This will apply for all cases where we obtain a curve point by multiplying the generator with a scalar: For example: `point_signature = generator * signature`. 
This kind of naming seems more descriptive and pleasant. 
Final remark, if you use an underscore for separating numbers in one place, then you should do the same for all other similar cases. It should apply for separating `x` and `y` as well.

mithril-stm/src/schnorr_signature/signature.rs
damrobi: I think scalar is fine, it's more descriptive

mithril-stm/src/schnorr_signature/signature.rs
damrobi: I'll go for msg_hash for now to keep it short

mithril-stm/src/schnorr_signature/signature.rs
damrobi: Actually those values are points, so as you said in an other comment maybe random_point_1 would be better?

mithril-stm/Cargo.toml
```diff
@@ -19,13 +19,25 @@ rug-backend = ["rug/default"]
 num-integer-backend = ["num-bigint", "num-rational", "num-traits"]
 benchmark-internals = [] # For benchmarking multi_sig
 future_proof_system = [] # For activating future proof systems
+future_snark = [
+    "ff",
+    "group",
+    "num-traits",
+    "dusk-poseidon",
+    "dusk-jubjub",
+] # For activating snark features
```
jpraynaud, line 28: This avoids creating a new feature for each dependency which is activated:
```suggestion
future_snark = [
    "dep:ff",
    "dep:group",
    "dep:num-traits",
    "dep:dusk-poseidon",
    "dep:dusk-jubjub",
] # For activating snark features
```

mithril-stm/benches/schnorr_sig.rs
```diff
@@ -0,0 +1,41 @@
+use criterion::{BenchmarkId, Criterion, criterion_group, criterion_main};
+use mithril_stm::{SchnorrSigningKey, SchnorrVerificationKey};
+use rand_chacha::ChaCha20Rng;
+use rand_core::{RngCore, SeedableRng};
```
jpraynaud, line 5: Could you organize the imports according to the guidelines we follow?

mithril-stm/benches/schnorr_sig.rs
jpraynaud, line 1: What difference compared to BLS signatures? Maybe you can provide some figures in the PR description?

mithril-stm/benches/schnorr_sig.rs
```diff
@@ -0,0 +1,41 @@
...
+    let mut sigs = Vec::new();
+    for _ in 0..nr_sigs {
+        let sk = SchnorrSigningKey::generate(&mut rng);
+        let vk = SchnorrVerificationKey::from(&sk);
+        let sig = sk.sign(&msg, &mut rng_sig).unwrap();
+        sigs.push(sig);
+        mvks.push(vk);
+    }
+
+    group.bench_function(BenchmarkId::new("Individual verif", nr_sigs), |b| {
```
jpraynaud:
```suggestion
    group.bench_function(BenchmarkId::new("Signature verification", nr_sigs), |b| {
```

mithril-stm/benches/schnorr_sig.rs
```diff
@@ -0,0 +1,41 @@
...
+        b.iter(|| {
+            for (vk, sig) in mvks.iter().zip(sigs.iter()) {
+                assert!(sig.verify(&msg, vk).is_ok());
+            }
+        })
+    });
+}
+
+fn schnorr_benches(c: &mut Criterion) {
+    sign_and_verify(c, 856);
```
jpraynaud: Is there a specific reason why you chose `856` here?

mithril-stm/benches/schnorr_sig.rs
```diff
@@ -0,0 +1,41 @@
+use criterion::{BenchmarkId, Criterion, criterion_group, criterion_main};
+use mithril_stm::{SchnorrSigningKey, SchnorrVerificationKey};
+use rand_chacha::ChaCha20Rng;
+use rand_core::{RngCore, SeedableRng};
+
+fn sign_and_verify(c: &mut Criterion, nr_sigs: usize) {
```
jpraynaud, line 37: The name of the function suggests that we benchmark the signature when we just benchmark the verification. The benchmark of the signature creation is probably missing here.

mithril-stm/src/error.rs
```diff
@@ -40,6 +42,23 @@ pub enum MultiSignatureError {
...
+    SignatureInvalid(Box<SchnorrSignature>),
+
+    /// Invalid Verification key
+    #[error("Invalid Schnorr Verification key")]
+    VerificationKeyInvalid(Box<SchnorrVerificationKey>),
+
+    /// This error occurs when the the serialization of the raw bytes failed
+    #[error("Invalid bytes")]
+    SerializationError,
+}
```
jpraynaud: These errors should be created in the `schnorr_signature` module.

mithril-stm/src/error.rs
```diff
@@ -40,6 +42,23 @@ pub enum MultiSignatureError {
...
+pub enum SchnorrSignatureError {
+    /// Invalid Single signature
+    #[error("Invalid Schnorr single signature")]
+    SignatureInvalid(Box<SchnorrSignature>),
+
+    /// Invalid Verification key
+    #[error("Invalid Schnorr Verification key")]
+    VerificationKeyInvalid(Box<SchnorrVerificationKey>),
+
+    /// This error occurs when the the serialization of the raw bytes failed
```
jpraynaud:
```suggestion
    /// This error occurs when the serialization of the raw bytes failed
```

mithril-stm/src/schnorr_signature/mod.rs
```diff
@@ -0,0 +1,173 @@
+// TODO: Remove
+#![allow(dead_code)]
+
+mod signature;
+mod signing_key;
+pub(super) mod utils;
```
jpraynaud: What's the purpose of `pub(super)` when you have `pub use utils::*;` below?
```suggestion
mod utils;
```

mithril-stm/src/schnorr_signature/mod.rs
```diff
@@ -0,0 +1,173 @@
...
+
+pub use signature::*;
+pub use signing_key::*;
+pub use utils::*;
+pub use verification_key::*;
+
+use dusk_jubjub::Fq as JubjubBase;
+
+/// A DST (Domain Separation Tag) to distinguish between use of Poseidon hash
+pub(crate) const DST_SIGNATURE: JubjubBase = JubjubBase::from_raw([0u64, 0, 0, 0]);
```
jpraynaud: I guess that this is internal stuff of the `schnorr_signature` module:
```suggestion
const DST_SIGNATURE: JubjubBase = JubjubBase::from_raw([0u64, 0, 0, 0]);
```

mithril-stm/src/schnorr_signature/mod.rs
```diff
@@ -0,0 +1,173 @@
...
+    use dusk_jubjub::SubgroupPoint as JubjubSubgroup;
+    use ff::Field;
+    use group::Group;
+    use rand_chacha::ChaCha20Rng;
+    use rand_core::SeedableRng;
+
+    use crate::schnorr_signature::{SchnorrSigningKey, SchnorrVerificationKey};
+
+    #[test]
+    fn test_get_coordinates() {
```
jpraynaud: There is no need to prefix with `test_` and you could use a more explicit name, e.g. `get_coordinates_extended_from_point` (This comment applies to most of the tests below).

Also, it looks like nothing is tested in this test :thinking: Is it because the underlying function could panic?

mithril-stm/src/schnorr_signature/mod.rs
```diff
@@ -0,0 +1,173 @@
...
+pub use utils::*;
+pub use verification_key::*;
+
+use dusk_jubjub::Fq as JubjubBase;
+
+/// A DST (Domain Separation Tag) to distinguish between use of Poseidon hash
+pub(crate) const DST_SIGNATURE: JubjubBase = JubjubBase::from_raw([0u64, 0, 0, 0]);
+
+#[cfg(test)]
+mod tests {
```
jpraynaud: You could split the unit tests and put them in their respective sub-modules. This would enhance the readability.

mithril-stm/src/schnorr_signature/signature.rs
```diff
@@ -0,0 +1,152 @@
+use anyhow::{Context, anyhow};
+
+use dusk_jubjub::{
```
jpraynaud, line 2:
```suggestion
use anyhow::{Context, anyhow};
use dusk_jubjub::{
```

mithril-stm/src/schnorr_signature/signing_key.rs
```diff
@@ -0,0 +1,171 @@
+use anyhow::{Context, anyhow};
+use dusk_jubjub::{
+    ExtendedPoint as JubjubExtended, Fr as JubjubScalar, SubgroupPoint as JubjubSubgroup,
+};
+use dusk_poseidon::{Domain, Hash};
+use ff::Field;
+use rand_core::{CryptoRng, RngCore};
+
+use group::Group;
```
jpraynaud:
```suggestion
use rand_core::{CryptoRng, RngCore};
use group::Group;
```

mithril-stm/src/schnorr_signature/signing_key.rs
```diff
@@ -0,0 +1,171 @@
...
+    },
+};
+
+/// Schnorr Signing key, it is essentially a random scalar of the Jubjub scalar field
+#[derive(Debug, Clone)]
+pub struct SchnorrSigningKey(pub JubjubScalar);
+
+impl SchnorrSigningKey {
+    /// Generate a random scalar value to use as signing key
+    pub fn generate(rng: &mut (impl RngCore + CryptoRng)) -> Self {
```
jpraynaud: It'd be a better option to use static dispatch for the RNG with a generic.

mithril-stm/src/schnorr_signature/signing_key.rs
```diff
@@ -0,0 +1,171 @@
...
+        loop {
```
jpraynaud: It'd be a good idea to bound the number of iterations of the loop.

mithril-stm/src/schnorr_signature/signing_key.rs
```diff
@@ -0,0 +1,171 @@
...
+            let signing_key = JubjubScalar::random(&mut *rng);
+            if signing_key != JubjubScalar::ZERO {
+                return SchnorrSigningKey(signing_key);
+            }
+        }
+    }
+
+    /// This function is an adapted version of the Schnorr signature scheme
```
jpraynaud: What do you mean by "adapted" here?

mithril-stm/src/schnorr_signature/signing_key.rs
```diff
@@ -0,0 +1,171 @@
...
+    /// and works with the Jubjub elliptic curve and the Poseidon hash function.
+    ///
+    /// Input:
+    ///     - a message: some bytes
+    ///     - a secret key: a element of the scalar field of the Jubjub curve
+    /// Output:
+    ///     - a signature of the form (sigma, signature, challenge) where sigma is deterministic based
+    /// on the message and the secret key and the signature and challenge are computed using randomness
```
jpraynaud: This sentence is a bit confusing, which parts are deterministic vs random?

mithril-stm/src/schnorr_signature/signing_key.rs
```diff
@@ -0,0 +1,171 @@
...
+            random_point_1,
+            random_point_2.into(),
+        ]);
+
+        let mut poseidon_input = vec![DST_SIGNATURE];
+        poseidon_input.extend(points_coordinates);
+
+        let challenge = Hash::digest_truncated(Domain::Other, &poseidon_input)[0];
+
+        let signature = random_scalar - challenge * self.0;
```
jpraynaud, line 95: You can probably remove few line returns here:
```suggestion
        let mut poseidon_input = vec![DST_SIGNATURE];
        poseidon_input.extend(points_coordinates);
        let challenge = Hash::digest_truncated(Domain::Other, &poseidon_input)[0];
        let signature = random_scalar - challenge * self.0;
```

mithril-stm/src/schnorr_signature/signing_key.rs
```diff
@@ -0,0 +1,171 @@
...
+        self.0.to_bytes()
+    }
+
+    /// Convert a string of bytes into a `SchnorrSigningKey`.
+    ///
+    /// The bytes must represent a Jubjub scalar or the conversion will fail
+    pub(crate) fn from_bytes(bytes: &[u8]) -> StmResult<Self> {
+        if bytes.len() < 32 {
+            return Err(anyhow!(SchnorrSignatureError::SerializationError))
+                .with_context(|| "Not enough bytes provided to create a signing key.");
```
jpraynaud:
```suggestion
                .with_context(|| "Not enough bytes provided to create a Schnorr signing key.");
```

mithril-stm/src/schnorr_signature/signing_key.rs
```diff
@@ -0,0 +1,171 @@
...
+        }
+
+        let signing_key_bytes = bytes[0..32]
+            .try_into()
+            .map_err(|_| anyhow!(SchnorrSignatureError::SerializationError))
+            .with_context(|| "Failed to obtain signing key's bytes.")?;
```
jpraynaud:
```suggestion
            .with_context(|| "Failed to obtain Schnorr signing key's bytes.")?;
```

mithril-stm/src/schnorr_signature/signing_key.rs
```diff
@@ -0,0 +1,171 @@
...
+
+        // Jubjub returns a CtChoice so I convert it to an option that looses the const time property
```
jpraynaud: This comment is not mandatory IMO:
```suggestion
```

mithril-stm/src/schnorr_signature/signing_key.rs
```diff
@@ -0,0 +1,171 @@
...
+mod tests {
+    pub(crate) use super::*;
+    use rand_chacha::ChaCha20Rng;
+    use rand_core::SeedableRng;
+
+    impl PartialEq for SchnorrSigningKey {
+        fn eq(&self, other: &Self) -> bool {
+            self.0 == other.0
+        }
+    }
```
jpraynaud: This could probably be implemented with a `derive` directive?
Also, it's a bit weird to implement `PartialEq` in tests.

mithril-stm/src/schnorr_signature/signing_key.rs
```diff
@@ -0,0 +1,171 @@
...
+        let sk = SchnorrSigningKey::generate(&mut rng);
+
+        let bytes = sk.to_bytes();
+        let recovered_sk = SchnorrSigningKey::from_bytes(&bytes).unwrap();
+
+        assert_eq!(sk, recovered_sk);
+    }
+
+    // For now failing test, maybe change it later depending on what
+    // we want for from_bytes
```
jpraynaud: Is it still the case?

mithril-stm/src/schnorr_signature/utils.rs
```diff
@@ -0,0 +1,90 @@
+use dusk_jubjub::{
+    AffinePoint as JubjubAffine, EDWARDS_D, ExtendedPoint as JubjubExtended, Fq as JubjubBase,
+    Fr as JubjubScalar,
+};
+
+use ff::Field;
+
+use anyhow::{Context, Ok, anyhow};
```
jpraynaud:
```suggestion
use dusk_jubjub::{
    AffinePoint as JubjubAffine, EDWARDS_D, ExtendedPoint as JubjubExtended, Fq as JubjubBase,
    Fr as JubjubScalar,
};
use ff::Field;
use anyhow::{Context, Ok, anyhow};
```

mithril-stm/src/schnorr_signature/utils.rs
```diff
@@ -0,0 +1,90 @@
...
+
+    let lhs = y_square - x_square;
+    let rhs = JubjubBase::ONE + EDWARDS_D * x_square * y_square;
+
+    lhs == rhs
+}
+
+/// Extract the coordinates of a given point in an Extended form
+///
+/// This is mainly use to feed the Poseidon hash function
```
jpraynaud:
```suggestion
/// This is mainly used to feed the Poseidon hash function
```

mithril-stm/src/schnorr_signature/utils.rs
```diff
@@ -0,0 +1,90 @@
...
+    let point_affine_representation = JubjubAffine::from(point);
+    let x_coordinate = point_affine_representation.get_u();
+    let y_coordinate = point_affine_representation.get_v();
+
+    (x_coordinate, y_coordinate)
+}
+
+/// Extract the coordinates of given points in an Extended form
+///
+/// This is mainly use to feed the Poseidon hash function, the order is maintained
```
jpraynaud:
```suggestion
/// This is mainly used to feed the Poseidon hash function, the order is maintained
```

mithril-stm/src/schnorr_signature/utils.rs
```diff
@@ -0,0 +1,90 @@
...
+pub fn get_coordinates_several_points(points: &[JubjubExtended]) -> Vec<JubjubBase> {
+    let mut points_coordinates = vec![];
+    for p in points {
+        let point_affine_representation = JubjubAffine::from(p);
+        let x_coordinate = point_affine_representation.get_u();
+        let y_coordinate = point_affine_representation.get_v();
+        points_coordinates.push(x_coordinate);
+        points_coordinates.push(y_coordinate);
+    }
+    points_coordinates
```
jpraynaud: You could create a more functional implementation without the need for mutating the array.

mithril-stm/src/schnorr_signature/signature.rs
```diff
@@ -0,0 +1,152 @@
...
+        out[32..64].copy_from_slice(&self.signature.to_bytes());
+        out[64..96].copy_from_slice(&self.challenge.to_bytes());
+
+        out
+    }
+
+    /// Convert a string of bytes into a `SchnorrSignature`.
+    ///
+    /// Not sure the sigma, s and c creation can fail if the 96 bytes are correctly extracted.
+    pub fn from_bytes(bytes: &[u8]) -> StmResult<Self> {
```
jpraynaud, line 115: Can you add golden tests for the signature codec?

mithril-stm/src/schnorr_signature/verification_key.rs
```diff
@@ -0,0 +1,47 @@
...
+impl SchnorrVerificationKey {
+    /// Convert a `SchnorrVerificationKey` into a string of bytes.
+    pub(crate) fn to_bytes(self) -> [u8; 32] {
+        self.0.to_bytes()
+    }
+
+    /// Convert a string of bytes into a `SchnorrVerificationKey`.
+    ///
+    /// The bytes must represent a Jubjub Subgroup point or the conversion will fail
+    pub(crate) fn from_bytes(bytes: &[u8]) -> StmResult<Self> {
```
jpraynaud: Can you add golden tests for the verification key codec?

mithril-stm/src/schnorr_signature/verification_key.rs
```diff
@@ -0,0 +1,47 @@
...
+pub(crate) use crate::schnorr_signature::signing_key::SchnorrSigningKey;
+use crate::{StmResult, error::SchnorrSignatureError};
+
+/// Schnorr verification key, it consists of a point on the Jubjub curve
+/// vk = g * sk, where g is a generator
+#[derive(Debug, Clone, Copy, Default, PartialEq, Eq)]
+pub struct SchnorrVerificationKey(pub(crate) JubjubSubgroup);
+
+impl SchnorrVerificationKey {
+    /// Convert a `SchnorrVerificationKey` into a string of bytes.
```
jpraynaud:
```suggestion
    /// Convert a `SchnorrVerificationKey` into bytes.
```

mithril-stm/src/schnorr_signature/verification_key.rs
```diff
@@ -0,0 +1,47 @@
...
+    pub(crate) fn to_bytes(self) -> [u8; 32] {
+        self.0.to_bytes()
+    }
+
+    /// Convert a string of bytes into a `SchnorrVerificationKey`.
```
jpraynaud:
```suggestion
    /// Convert bytes into a `SchnorrVerificationKey`.
```

mithril-stm/src/schnorr_signature/verification_key.rs
```diff
@@ -0,0 +1,47 @@
...
+        let point = JubjubSubgroup::from_bytes(&verification_key_bytes)
+            .into_option()
+            .ok_or(anyhow!(SchnorrSignatureError::SerializationError))?;
+
+        Ok(SchnorrVerificationKey(point))
+    }
+}
+
+impl From<&SchnorrSigningKey> for SchnorrVerificationKey {
+    /// Convert a Shnorr secret key into a verification key
```
jpraynaud:
```suggestion
    /// Convert a Schnorr secret key into a verification key
```

mithril-stm/benches/schnorr_sig.rs
damrobi: You mean the difference between the timings, right? I will add them

jpraynaud: Yes, you can add results from running the benchmarks on your machine to get an idea on the performance impact of the Schnorr signature.

damrobi: Done, I update the benchmark functions too. For now it's a x12 overhead for the signature and a x2 for the verification. I'll check those figures to understand if it's normal or not.

mithril-stm/benches/schnorr_sig.rs
damrobi: To be the same as in the bls benchmark. I changed it to 1000 but it takes around four minutes to run signature and verification so maybe we want to lower it to have a faster benchmark.

mithril-stm/src/schnorr_signature/signing_key.rs
damrobi: I mean that it's not exactly the Schnorr signature scheme since we computed the value sigma and it's part of the signature while usually the signature is only two values. Should I give a precision on that point in the comment?

mithril-stm/src/schnorr_signature/signing_key.rs
damrobi: I did it because I don't know if we need/want to have this trait available for the signing key in general. For now, we only need the equality check in the tests.

mithril-stm/benches/schnorr_sig.rs
jpraynaud: That's fine if the benchmarks take few minutes.

mithril-stm/benches/schnorr_sig.rs
damrobi: Can you remind me the guidelines?

mithril-stm/src/schnorr_signature/signing_key.rs
jpraynaud: You can implement it with a `derive` directive for free instead of in the tests.
Other option, you can implement a `Deref` to the inner type and compare them directly.

mithril-stm/Cargo.toml
```diff
@@ -16,16 +16,28 @@ crate-type = ["lib", "cdylib", "staticlib"]
 [features]
 default = ["rug-backend"]
 rug-backend = ["rug/default"]
-num-integer-backend = ["num-bigint", "num-rational", "num-traits"]
+num-integer-backend = ["num-bigint", "num-rational", "dep:num-traits"]
```
jpraynaud: You can apply the `dep:` to all the other dependencies I guess.

mithril-stm/benches/multi_sig.rs
```diff
@@ -47,15 +47,25 @@ fn aggregate_and_verify(c: &mut Criterion, nr_sigs: usize) {
...
 
+    group.bench_function(BenchmarkId::new("Signature", nr_sigs), |b| {
+        b.iter(|| {
+            for sk in msks.iter() {
+                let _ = sk.sign(&msg);
+            }
+        })
+    });
+
     group.bench_function(BenchmarkId::new("Individual verif", nr_sigs), |b| {
```
jpraynaud:
```suggestion
    group.bench_function(BenchmarkId::new("Verification", nr_sigs), |b| {
```

mithril-stm/src/schnorr_signature/signing_key.rs
damrobi: True, I remove the `impl` and compared directly the inner values. I don't think we need equality checks outside of test but if we end up needing it we can add it to `derive`

mithril-stm/src/schnorr_signature/signing_key.rs
```diff
@@ -0,0 +1,201 @@
...
+
+        Ok(SchnorrSignature {
+            sigma,
+            signature,
+            challenge,
+        })
+    }
+
+    /// Convert a `SchnorrSigningKey` into bytes.
+    pub(crate) fn to_bytes(&self) -> [u8; 32] {
```
jpraynaud: This should probably be public:
```suggestion
    pub fn to_bytes(&self) -> [u8; 32] {
```

mithril-stm/src/schnorr_signature/signing_key.rs
```diff
@@ -0,0 +1,201 @@
...
+        self.0.to_bytes()
+    }
+
+    /// Convert bytes into a `SchnorrSigningKey`.
+    ///
+    /// The bytes must represent a Jubjub scalar or the conversion will fail
+    pub(crate) fn from_bytes(bytes: &[u8]) -> StmResult<Self> {
```
jpraynaud: Same here:
```suggestion
    pub fn from_bytes(bytes: &[u8]) -> StmResult<Self> {
```

mithril-stm/src/schnorr_signature/verification_key.rs
```diff
@@ -0,0 +1,104 @@
...
+impl SchnorrVerificationKey {
+    /// Convert a `SchnorrVerificationKey` into bytes.
+    pub(crate) fn to_bytes(self) -> [u8; 32] {
+        self.0.to_bytes()
+    }
+
+    /// Convert bytes into a `SchnorrVerificationKey`.
+    ///
+    /// The bytes must represent a Jubjub Subgroup point or the conversion will fail
+    pub(crate) fn from_bytes(bytes: &[u8]) -> StmResult<Self> {
```
jpraynaud:
```suggestion
    pub fn from_bytes(bytes: &[u8]) -> StmResult<Self> {
```

mithril-stm/src/schnorr_signature/verification_key.rs
```diff
@@ -0,0 +1,104 @@
...
+use crate::{StmResult, schnorr_signature::SchnorrSignatureError};
+
+/// Schnorr verification key, it consists of a point on the Jubjub curve
+/// vk = g * sk, where g is a generator
+#[derive(Debug, Clone, Copy, Default, PartialEq, Eq)]
+pub struct SchnorrVerificationKey(pub(crate) JubjubSubgroup);
+
+impl SchnorrVerificationKey {
+    /// Convert a `SchnorrVerificationKey` into bytes.
+    pub(crate) fn to_bytes(self) -> [u8; 32] {
```
jpraynaud:
```suggestion
    pub fn to_bytes(self) -> [u8; 32] {
```

mithril-stm/src/schnorr_signature/utils.rs
```diff
@@ -0,0 +1,157 @@
...
+        if random_scalar != JubjubScalar::ZERO {
+            return Ok(random_scalar);
+        }
+    }
+    Err(anyhow!(SchnorrSignatureError::RandomScalarGenerationError))
+        .with_context(|| "Failed to generate a non zero signing key after 100 attempts.")
+}
+
+/// Convert an element of the BLS12-381 base field to one of the Jubjub scalar field
+pub fn jubjub_base_to_scalar(x: &JubjubBase) -> StmResult<JubjubScalar> {
```
jpraynaud: It looks like this function is never used: in that case it can be removed.

mithril-stm/src/schnorr_signature/utils.rs
```diff
@@ -0,0 +1,157 @@
...
+    let lhs = y_square - x_square;
+    let rhs = JubjubBase::ONE + EDWARDS_D * x_square * y_square;
+
+    lhs == rhs
+}
+
+/// Extract the coordinates of a given point in an Extended form
+///
+/// This is mainly used to feed the Poseidon hash function
+pub fn get_coordinates_extended(point: JubjubExtended) -> (JubjubBase, JubjubBase) {
```
jpraynaud: It looks like this function is a sub-case of the `get_coordinates_several_points`, can we avoid the double implementation?

mithril-stm/src/schnorr_signature/utils.rs
```diff
@@ -0,0 +1,157 @@
...
+    let y_coordinate = point_affine_representation.get_v();
+
+    (x_coordinate, y_coordinate)
+}
+
+/// Extract the coordinates of given points in an Extended form
+///
+/// This is mainly used to feed the Poseidon hash function, the order is maintained
+/// from input to output which is important for the hash function
+pub fn get_coordinates_several_points(points: &[JubjubExtended]) -> Vec<JubjubBase> {
```
jpraynaud: I find it a bit confusing to mix x and y coordinates in the result of the function. 
Wouldn't it be less error prone with `Vec<(JubjubBase,JubjubBase)>`?

mithril-stm/src/schnorr_signature/mod.rs
jpraynaud: This is probably a good time to remove this `#[allow(dead_code)]` it will point out the remaining problems with visibility and unused functions.

mithril-stm/src/schnorr_signature/signature.rs
```diff
@@ -0,0 +1,206 @@
...
+    #[test]
+    fn from_bytes_signature_not_enough_bytes() {
+        let msg = vec![0u8; 95];
+
+        let result = SchnorrSignature::from_bytes(&msg);
+
+        assert!(result.is_err());
+    }
+
+    const GOLDEN_BYTES: &[u8; 96] = &[
```
jpraynaud: Could you put the golden tests in a `golden` sub-module for readability and maintainability?

mithril-stm/src/schnorr_signature/signing_key.rs
```diff
@@ -0,0 +1,201 @@
...
+    #[test]
+    fn from_bytes_signing_key_not_enough_bytes() {
+        let msg = vec![0u8; 31];
+
+        let result = SchnorrSigningKey::from_bytes(&msg);
+
+        assert!(result.is_err());
+    }
+
+    const GOLDEN_BYTES: &[u8; 32] = &[
```
jpraynaud: Same here, can you put golden tests in a `golden` sub-module?

mithril-stm/src/schnorr_signature/mod.rs
```diff
@@ -0,0 +1,119 @@
...
+    SerializationError,
+
+    /// This error occurs when the signing key fails to generate
+    #[error("Invalid generation of the signing key")]
+    SigningKeyGenerationError,
+
+    /// This error occurs when the random scalar fails to generate during the signature
+    #[error("Invalid generation of the signature's random scalar")]
+    RandomScalarGenerationError,
+}
```
jpraynaud: This could live in its own `error.rs` sub-module.

mithril-stm/src/schnorr_signature/mod.rs
```diff
@@ -0,0 +1,119 @@
...
+
+#[cfg(test)]
+mod tests {
```
jpraynaud: Maybe all these tests could be moved somewhere else?

mithril-stm/benches/schnorr_sig.rs
jpraynaud: Sure:
- first the `std` imports
- then line return
- then external dependencies imports
- then line return
- then internal dependencies imports
- then line return
- then crate imports

mithril-stm/src/schnorr_signature/utils.rs
damrobi: I did that to use the result directly in the poseidon hash but I can move the iter map outside of the function. I'll do that for now and I can reverse if needed.

mithril-stm/src/schnorr_signature/error.rs
```diff
@@ -0,0 +1,27 @@
...
+    /// Invalid Verification key
+    #[error("Invalid Schnorr Verification key")]
+    VerificationKeyInvalid(Box<SchnorrVerificationKey>),
+
+    /// This error occurs when the serialization of the raw bytes failed
+    #[error("Invalid bytes")]
+    SerializationError,
+
+    /// This error occurs when the signing key fails to generate
+    #[error("Invalid generation of the signing key")]
```
jpraynaud:
```suggestion
    #[error("Failed generation of the signing key")]
```

mithril-stm/src/schnorr_signature/error.rs
```diff
@@ -0,0 +1,27 @@
...
+    SigningKeyGenerationError,
+
+    /// This error occurs when the random scalar fails to generate during the signature
+    #[error("Invalid generation of the signature's random scalar")]
```
jpraynaud:
```suggestion
    #[error("Failed generation of the signature's random scalar")]
```

mithril-stm/src/schnorr_signature/signing_key.rs
```diff
@@ -0,0 +1,223 @@
...
+    },
+};
+
+/// Schnorr Signing key, it is essentially a random scalar of the Jubjub scalar field
+#[derive(Debug, Clone)]
+pub struct SchnorrSigningKey(pub(crate) JubjubScalar);
+
+impl SchnorrSigningKey {
+    /// Generate a random scalar value to use as signing key
+    pub fn generate<R: RngCore + CryptoRng>(rng: &mut R) -> StmResult<Self> {
```
jpraynaud: This is a better naming as this can fail:
```suggestion
    pub fn try_generate<R: RngCore + CryptoRng>(rng: &mut R) -> StmResult<Self> {
```

mithril-stm/src/schnorr_signature/verification_key.rs
```diff
@@ -0,0 +1,140 @@
...
+            JubjubBase::ONE,
+        ));
+
+        let result1 = sig.verify(&msg, &vk1);
+        let result2 = sig.verify(&msg, &vk2);
+
+        assert!(
+            result1.is_ok(),
+            "Correct verification key used, test should pass."
+        );
```
jpraynaud: You can replace this `assert` by an `expect` (here and everywhere else this appears in the PR):
```suggestion
        result1.expect("Correct verification key used, test should pass.");
):
```

mithril-stm/src/schnorr_signature/verification_key.rs
```diff
@@ -0,0 +1,140 @@
...
+        assert!(
+            result2.is_err(),
+            "Invalid verification key used, test should fail."
+        );
```
jpraynaud: Same here with an `expect_err` (here and everywhere else this appears in the PR):
```suggestion
        result2.expect_err("Invalid verification key used, test should fail.");
```

mithril-stm/Cargo.toml
jpraynaud:
```suggestion
num-integer-backend = ["dep:num-bigint", "dep:num-rational", "dep:num-traits"]
```

PR #2779, opened 2025-11-14
feat: simple aggregator discovery

### LINKED ISSUE #2726: Implement a simple aggregator discovery mechanism
## Why
We want to allow the discovery of available aggregators in a Mithril network by the client. This will allow a more decentralized, reliable and scalable the network, and ease of use for the end users

## What
Create an interface for a aggregator discovery mechanism and use a first simple implementation in the client.

### Create the aggregator discoverer
- [ ] Create a new `internal/mithril-aggregator-discovery` crate
- [ ] Create a `AggregatorDiscoverer` trait which gives access to a list of aggregator endpoints (given Cardano network and  expected capabilities of the aggregator -needs to be specified, can be a type of signed entity or the type of proof in the certificate chain, or is the aggregator a leader-) and returns an error otherwise
- [ ] Implement a centralized aggregator discoverer based on the `networks.json` file at the root of the repository:
  - [ ] Do we need to add a weight to the aggregators in the list for a network?
  - [ ] Do we support the configuration in a local file?
- [ ] Do we need to support a customizable URL for the `networks.json` file
- [ ] Implement the test doubles  

### Use the aggregator discoverer in the client

- [ ] In the client library:
  - [ ] Add a `automatic_aggregator_discovery` function to specify the usage of the aggregator discoverer. 
  - [ ] If no aggregator is found, the build fails. The endpoint aggregator of the builder is filled with the one given by the inner discoverer.
- [ ] In the client CLI:
  - [ ] Add a `aggregator discovery` command in `tools` (list the available aggregators)
  - [ ] Add an expansion of `AGGREGATOR_ENDPOINT=auto` to use the automatic aggregator discovery  (no guarantee to use the same aggregator for 2 distinct command calls) 
  - [ ] No breaking change in the commands API

### Documentation 
- [ ] Document how to add a new aggregator in the discovery for a network
- [ ] Publish manually the first version of the crate on crates.io + setup crate owner (@jpraynaud)

## Later
- In a follower aggregator, use the auto discovery for the chain synchronization to be able to match the capabilities of the aggregator (e.g. synchonize on a certificate chain with the same proof type)
- Can we make an auto-configuration also for the verification keys and the other parameters of the client?

jpraynaud:
## Content

This PR introduces a simple **Aggregator Discovery** functionality to the Mithril client library and CLI, allowing clients to automatically discover and connect to suitable aggregators based on their capabilities rather than requiring a hardcoded aggregator endpoint:

### New `mithril-aggregator-discovery` crate
A standalone crate providing aggregator discovery functionality:
- `AggregatorDiscoverer` trait for extensible discovery mechanisms
- `HttpConfigAggregatorDiscoverer` for fetching aggregator lists from HTTP endpoints
- `ShuffleAggregatorDiscoverer` decorator for randomizing aggregator selection
- `CapableAggregatorDiscoverer` decorator for filtering aggregators by required capabilities
- Added `MithrilNetwork` type to `mithril-common` for shared usage
- Introduced `RequiredAggregatorCapabilities` type for cleaner capability filtering

### Client integration
- New `aggregator_discoverer()` method in client builder as the recommended way to configure aggregators
- Default aggregator discoverer with shuffling support
- Deprecated `aggregator()` constructor in favor of `new` constructor

### CLI enhancements
- New `tools aggregator-discovery` command (unstable) to list available aggregators
- JSON logs support via `--json` flag

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation
  - [x] No new TODOs introduced

## Issue(s)
Closes #2726

### DIFF
--- a/internal/mithril-aggregator-discovery/Cargo.toml
+++ b/internal/mithril-aggregator-discovery/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "mithril-aggregator-discovery"
 description = "Mechanisms to discover aggregators available in a Mithril network."
-version = "0.1.0"
+version = "0.1.1"
 authors.workspace = true
 documentation.workspace = true
 edition.workspace = true
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -911,7 +911,7 @@ dependencies = [
 
 [[package]]
 name = "client-cardano-database"
-version = "0.1.37"
+version = "0.1.38"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -924,7 +924,7 @@ dependencies = [
 
 [[package]]
 name = "client-cardano-database-v2"
-version = "0.0.14"
+version = "0.0.15"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -937,7 +937,7 @@ dependencies = [
 
 [[package]]
 name = "client-cardano-stake-distribution"
-version = "0.1.13"
+version = "0.1.14"
 dependencies = [
  "anyhow",
  "clap",
@@ -950,7 +950,7 @@ dependencies = [
 
 [[package]]
 name = "client-cardano-transaction"
-version = "0.1.23"
+version = "0.1.24"
 dependencies = [
  "anyhow",
  "clap",
@@ -963,7 +963,7 @@ dependencies = [
 
 [[package]]
 name = "client-mithril-stake-distribution"
-version = "0.2.11"
+version = "0.2.12"
 dependencies = [
  "anyhow",
  "clap",
@@ -3803,7 +3803,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-aggregator-discovery"
-version = "0.1.0"
+version = "0.1.1"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -3925,7 +3925,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-client"
-version = "0.12.36"
+version = "0.12.37"
 dependencies = [
  "anyhow",
  "async-recursion",
@@ -3965,7 +3965,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-client-cli"
-version = "0.12.34"
+version = "0.12.35"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -3999,7 +3999,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-client-wasm"
-version = "0.9.8"
+version = "0.9.9"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -4017,7 +4017,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-common"
-version = "0.6.32"
+version = "0.6.33"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -4295,7 +4295,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-stm"
-version = "0.6.2"
+version = "0.6.3"
 dependencies = [
  "anyhow",
  "blake2 0.10.6",
--- a/examples/client-cardano-database-v2/Cargo.toml
+++ b/examples/client-cardano-database-v2/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "client-cardano-database-v2"
 description = "Mithril client Cardano database example"
-version = "0.0.14"
+version = "0.0.15"
 authors = ["dev@iohk.io", "mithril-dev@iohk.io"]
 documentation = "https://mithril.network/doc"
 edition = "2021"
--- a/examples/client-cardano-database/Cargo.toml
+++ b/examples/client-cardano-database/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "client-cardano-database"
 description = "Mithril client Cardano database example"
-version = "0.1.37"
+version = "0.1.38"
 authors = ["dev@iohk.io", "mithril-dev@iohk.io"]
 documentation = "https://mithril.network/doc"
 edition = "2021"
--- a/examples/client-cardano-stake-distribution/Cargo.toml
+++ b/examples/client-cardano-stake-distribution/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "client-cardano-stake-distribution"
 description = "Mithril client cardano stake distribution example"
-version = "0.1.13"
+version = "0.1.14"
 authors = ["dev@iohk.io", "mithril-dev@iohk.io"]
 documentation = "https://mithril.network/doc"
 edition = "2021"
--- a/examples/client-cardano-transaction/Cargo.toml
+++ b/examples/client-cardano-transaction/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "client-cardano-transaction"
 description = "Mithril client cardano-transaction example"
-version = "0.1.23"
+version = "0.1.24"
 authors = ["dev@iohk.io", "mithril-dev@iohk.io"]
 documentation = "https://mithril.network/doc"
 edition = "2021"
--- a/examples/client-mithril-stake-distribution/Cargo.toml
+++ b/examples/client-mithril-stake-distribution/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "client-mithril-stake-distribution"
 description = "Mithril client stake distribution example"
-version = "0.2.11"
+version = "0.2.12"
 authors = ["dev@iohk.io", "mithril-dev@iohk.io"]
 documentation = "https://mithril.network/doc"
 edition = "2021"
--- a/examples/client-wasm-nodejs/package-lock.json
+++ b/examples/client-wasm-nodejs/package-lock.json
@@ -16,7 +16,7 @@
     },
     "../../mithril-client-wasm": {
       "name": "@mithril-dev/mithril-client-wasm",
-      "version": "0.9.8",
+      "version": "0.9.9",
       "license": "Apache-2.0"
     },
     "node_modules/@mithril-dev/mithril-client-wasm": {
--- a/examples/client-wasm-web/package-lock.json
+++ b/examples/client-wasm-web/package-lock.json
@@ -20,7 +20,7 @@
     },
     "../../mithril-client-wasm": {
       "name": "@mithril-dev/mithril-client-wasm",
-      "version": "0.9.8",
+      "version": "0.9.9",
       "license": "Apache-2.0"
     },
     "node_modules/@discoveryjs/json-ext": {
--- a/mithril-client-cli/Cargo.toml
+++ b/mithril-client-cli/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mithril-client-cli"
-version = "0.12.34"
+version = "0.12.35"
 description = "A Mithril Client"
 authors = { workspace = true }
 edition = { workspace = true }
--- a/mithril-client-wasm/Cargo.toml
+++ b/mithril-client-wasm/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mithril-client-wasm"
-version = "0.9.8"
+version = "0.9.9"
 description = "Mithril client WASM"
 authors = { workspace = true }
 edition = { workspace = true }
--- a/mithril-client-wasm/ci-test/package-lock.json
+++ b/mithril-client-wasm/ci-test/package-lock.json
@@ -21,7 +21,7 @@
     },
     "..": {
       "name": "@mithril-dev/mithril-client-wasm",
-      "version": "0.9.8",
+      "version": "0.9.9",
       "license": "Apache-2.0"
     },
     "node_modules/@discoveryjs/json-ext": {
--- a/mithril-client-wasm/package.json
+++ b/mithril-client-wasm/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@mithril-dev/mithril-client-wasm",
-  "version": "0.9.8",
+  "version": "0.9.9",
   "description": "Mithril client WASM",
   "license": "Apache-2.0",
   "collaborators": [
--- a/mithril-client/Cargo.toml
+++ b/mithril-client/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mithril-client"
-version = "0.12.36"
+version = "0.12.37"
 description = "Mithril client library"
 authors = { workspace = true }
 edition = { workspace = true }
--- a/mithril-common/Cargo.toml
+++ b/mithril-common/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mithril-common"
-version = "0.6.32"
+version = "0.6.33"
 description = "Common types, interfaces, and utilities for Mithril nodes."
 authors = { workspace = true }
 edition = { workspace = true }
--- a/mithril-explorer/package-lock.json
+++ b/mithril-explorer/package-lock.json
@@ -36,7 +36,7 @@
     },
     "../mithril-client-wasm": {
       "name": "@mithril-dev/mithril-client-wasm",
-      "version": "0.9.8",
+      "version": "0.9.9",
       "license": "Apache-2.0"
     },
     "node_modules/@adobe/css-tools": {
--- a/mithril-stm/Cargo.toml
+++ b/mithril-stm/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mithril-stm"
-version = "0.6.2"
+version = "0.6.3"
 edition = { workspace = true }
 authors = { workspace = true }
 homepage = { workspace = true }


docs/website/root/manual/develop/nodes/mithril-client.md
Alenar, line 1: The parameters table  is missing for the new `tool aggregator-discovery` command

docs/website/root/manual/develop/nodes/mithril-client.md
```diff
@@ -227,27 +227,45 @@ Commands:
...
   -v, --verbose...
           Verbosity level (-v=warning, -vv=info, -vvv=debug, -vvvv=trace)
+
       --config-directory <CONFIG_DIRECTORY>
-          Directory where configuration file is located [default: ./config]
+          Directory where configuration file is located
+
+          [default: ./config]
+
       --aggregator-endpoint <AGGREGATOR_ENDPOINT>
```
Alenar, line 243: ` [env: AGGREGATOR_ENDPOINT=]` disappeared, is that normal ?
I found that strange since env source is still available in the cli argument and it's still here for `--run-mode`.

docs/website/root/manual/getting-started/bootstrap-cardano-node.md
```diff
@@ -112,25 +112,50 @@ Commands:
...
-          Verbosity level (-v=warning, -vv=info, -vvv=debug)
+          Verbosity level (-v=warning, -vv=info, -vvv=debug, -vvvv=trace)
+
       --config-directory <CONFIG_DIRECTORY>
-          Directory where the configuration file is located [default: ./config]
+          Directory where configuration file is located
+
+          [default: ./config]
+
       --aggregator-endpoint <AGGREGATOR_ENDPOINT>
```
Alenar, line 133: ` [env: AGGREGATOR_ENDPOINT=]` disappeared, is that normal ?
I found that strange since env source is still available in the cli argument and it's still here for `--run-mode`.

internal/mithril-aggregator-discovery/Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+mithril-aggregator-client = { path = "../mithril-aggregator-client" }
+mithril-common = { path = "../../mithril-common" }
+rand = { version = "0.9.2" }
+reqwest = { workspace = true, features = [
+    "default",
+    "gzip",
+    "zstd",
+    "deflate",
+    "brotli"
+] }
```
Alenar: Should we left those features activation to child crates ? Else this force their activation on child crates.

mithril-client/src/aggregator_discovery.rs
Alenar, line 1: Why is this file empty ?

mithril-client/src/type_alias.rs
```diff
@@ -85,3 +86,10 @@ pub mod common {
         pub use mithril_common::test::double::Dummy;
     }
 }
+
+/// Required capabilities for an aggregator.
+#[cfg(not(target_family = "wasm"))]
+pub use mithril_aggregator_discovery::RequiredAggregatorCapabilities;
+
+/// Mithril network
+pub use mithril_common::entities::MithrilNetwork;
```
Alenar: All other common entities are re-exported in the `pub mod common` higher, what's the rational for making an exception here ?

mithril-client/src/client.rs
```diff
@@ -1,12 +1,24 @@
 use anyhow::{Context, anyhow};
-#[cfg(feature = "fs")]
+#[cfg(any(feature = "fs", not(target_family = "wasm")))]
 use chrono::Utc;
+
+#[cfg(not(target_family = "wasm"))]
+use mithril_common::entities::MithrilNetwork;
```
Alenar: Re-exported type in `type_alias` should preferably be imported by using their re-export.

mithril-client/src/client.rs
```diff
@@ -174,36 +205,52 @@ pub struct ClientBuilder {
...
+            network.to_string(),
+        )))
+        .set_genesis_verification_key(GenesisVerificationKey::JsonHex(
+            genesis_verification_key.to_string(),
+        ))
+    }
+
+    /// Default aggregator discoverer to use to find the aggregator endpoint when in automatic discovery.
+    #[cfg(not(target_family = "wasm"))]
+    fn default_aggregator_discoverer() -> Arc<dyn AggregatorDiscoverer> {
```
Alenar: Moving this method below `pub fn discover_aggregator` would allow to reduce confusion when reading the code, positioned here I first thought that it was a constructor.

mithril-client/src/client.rs
```diff
@@ -231,6 +310,15 @@ impl ClientBuilder {
             .clone()
             .unwrap_or_else(|| Logger::root(slog::Discard, o!()));
 
+        let genesis_verification_key = match self.genesis_verification_key {
+            Some(GenesisVerificationKey::JsonHex(ref key)) => key,
+            None => {
+                return Err(anyhow!(
+                    "The genesis verification key must be provided to build the client with the 'set_genesis_verification_key' function"
```
Alenar, line 319: This is unclear, isn't the genesis vkey always mandatory ?

mithril-client/src/client.rs
```diff
@@ -337,16 +425,50 @@ impl ClientBuilder {
...
+                    .await
+                    .with_context(|| "Discovering aggregator endpoint failed")
+            })
+        })
+    }
+
     fn build_aggregator_client(
         &self,
         logger: Logger,
     ) -> Result<AggregatorHTTPClient, anyhow::Error> {
```
Alenar:
```suggestion
    ) -> MithrilResult<AggregatorHTTPClient> {
```

mithril-client-cli/src/commands/mod.rs
```diff
@@ -38,17 +40,28 @@ pub(crate) fn client_builder_with_fallback_genesis_key(
...
             "genesis_verification_key",
             fallback_genesis_verification_key,
-        ),
-    );
+        )));
 
     Ok(finalize_builder_config(builder, params))
 }
 
+fn compute_aggregator_discovery_type(
```
Alenar: I think this should be reworked as a implementation of `FromStr` over `AggregatorDiscoveryType`.

This would allow standardization and easy re-use across crates such as in the examples. 

Usage would be:
```rust
    let builder = ClientBuilder::new(AggregatorDiscoveryType::from_str(params)?)
        .set_genesis_verification_key(GenesisVerificationKey::JsonHex(params.get_or(
            "genesis_verification_key",
            fallback_genesis_verification_key,
        )));
```

mithril-client-cli/src/commands/cardano_db/list.rs
```diff
@@ -32,6 +32,14 @@ impl CardanoDbListCommand {
     /// Main command execution
     pub async fn execute(&self, context: CommandContext) -> MithrilResult<()> {
         let client = client_builder_with_fallback_genesis_key(context.config_parameters())?
+            .with_capabilities(RequiredAggregatorCapabilities::And(vec![
+                RequiredAggregatorCapabilities::SignedEntityType(
```
Alenar: The backend type is available, only the required capability should be asked, else this would fails if either one is missing.

mithril-client-cli/src/commands/cardano_db/show.rs
```diff
@@ -31,6 +34,14 @@ impl CardanoDbShowCommand {
     /// Cardano DB Show command
     pub async fn execute(&self, context: CommandContext) -> MithrilResult<()> {
         let client = client_builder_with_fallback_genesis_key(context.config_parameters())?
+            .with_capabilities(RequiredAggregatorCapabilities::And(vec![
```
Alenar: The backend type is available, only the required capability should be asked, else this would fails if either one is missing.

mithril-client-cli/src/commands/cardano_db/verify.rs
```diff
@@ -89,6 +90,14 @@ impl CardanoDbVerifyCommand {
...
         let progress_printer = ProgressPrinter::new(progress_output_type, 5);
         let client = client_builder(context.config_parameters())?
+            .with_capabilities(RequiredAggregatorCapabilities::And(vec![
+                RequiredAggregatorCapabilities::SignedEntityType(
+                    SignedEntityTypeDiscriminants::CardanoImmutableFilesFull,
+                ),
+                RequiredAggregatorCapabilities::SignedEntityType(
+                    SignedEntityTypeDiscriminants::CardanoDatabase,
+                ),
+            ]))
```
Alenar, line 97: This command is v2 only
```suggestion
            .with_capabilities(RequiredAggregatorCapabilities::And(vec![
                RequiredAggregatorCapabilities::SignedEntityType(
                    SignedEntityTypeDiscriminants::CardanoDatabase,
                ),
            ]))
```

mithril-client-cli/src/commands/tools/aggregator_discovery.rs
Alenar, line 1: Would it be better to move it in a `select.rs` file located a `aggregator_discovery` module folder ? This would be consistent with other commands.

mithril-client-cli/src/commands/tools/aggregator_discovery.rs
```diff
@@ -0,0 +1,212 @@
...
+
+impl AggregatorSelectCommand {
+    /// Main command execution
+    pub async fn execute(&self, context: &CommandContext) -> MithrilResult<()> {
+        let progress_output_type = if context.is_json_output_enabled() {
+            ProgressOutputType::JsonReporter
+        } else {
+            ProgressOutputType::Tty
+        };
+        let progress_printer = ProgressPrinter::new(progress_output_type, 1);
```
Alenar: The progress printer is mostly mean to be used for slow, multiple steps, commands. Is there discovery scenario that would take more than a second to run ?

Moreover the progress printer is not meant to output results as everything that it writes are send to `stderr`.

 I think you should remove it altogether from this command and instead just output the result of `client_builder.discover_aggregator` directly (following the same way as any list command with a `serde_json` serialization in json in json and a nice `cli_table` otherwise).

internal/mithril-aggregator-discovery/src/model.rs
```diff
@@ -0,0 +1,43 @@
+use std::time::Duration;
+
+use mithril_aggregator_client::{AggregatorHttpClient, query::GetAggregatorFeaturesQuery};
+use mithril_common::{StdResult, messages::AggregatorCapabilities};
+
+/// Representation of an aggregator endpoint
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct AggregatorEndpoint {
```
Alenar, line 10: I don't think this is possible given how this crate is structured, but it would be interesting to add the capabilities of the aggregator in this struct, this would allow reporting of them in the new cli command in a table (also this mean either make those fields public or adding a getter).

internal/mithril-aggregator-discovery/src/model.rs
```diff
@@ -0,0 +1,43 @@
+use std::time::Duration;
+
+use mithril_aggregator_client::{AggregatorHttpClient, query::GetAggregatorFeaturesQuery};
+use mithril_common::{StdResult, messages::AggregatorCapabilities};
+
+/// Representation of an aggregator endpoint
+#[derive(Debug, Clone, PartialEq, Eq)]
```
Alenar: Adding `serde::Serialize` (no need for deserialize) would be interesting to output the result in the cli command when in json.

mithril-client-cli/src/commands/tools/mod.rs
```diff
@@ -3,30 +3,55 @@
...
+            Self::UTxOHD(cmd) => cmd.execute(context).await,
+            Self::AggregatorDiscovery(cmd) => {
+                if !context.is_unstable_enabled() {
+                    Err(anyhow!(Self::unstable_flag_missing_message(
+                        "aggregator discovery",
+                        "tools"
+                    )))
+                } else {
+                    cmd.execute(context).await
+                }
```
Alenar: There's already a method for that on the context :)
```suggestion
                context.require_unstable("tools aggregator-discovery select", Some("--network mainnet"))?;
                cmd.execute(context).await
```

mithril-client-cli/src/commands/tools/aggregator_discovery.rs
```diff
@@ -0,0 +1,212 @@
...
+use crate::{
+    CommandContext,
+    utils::{ProgressOutputType, ProgressPrinter},
+};
+
+/// Clap command to select an aggregator from the available ones with automatic discovery.
+#[derive(Parser, Debug, Clone)]
+pub struct AggregatorSelectCommand {
+    /// Mithril network name
+    #[clap(long)]
```
Alenar: Maybe making it a positional argument would be interesting as it's mandatory without default, this would allow usage like `cli tools aggregator-discovery select mainnet` instead of `cli tools aggregator-discovery select --network mainnet`
```suggestion
```

mithril-client-cli/src/commands/tools/aggregator_discovery.rs
```diff
@@ -0,0 +1,212 @@
...
+    signed_entity_types: Vec<SignedEntityTypeDiscriminants>,
+
+    /// Aggregate signature types to consider for the discovery
+    ///
+    /// If not provided, all aggregate signature types are considered.
+    #[clap(long, value_parser, num_args = 0.., value_delimiter = ',')]
+    aggregate_signature_types: Vec<AggregateSignatureType>,
+}
+
+impl AggregatorSelectCommand {
```
Alenar: I found is strange that this command search using "or" for its given signed entity and aggregator signature types. I think users would expect that all the types are matched.

Also this hinder search capabilities as users won't be able to make exact search. If it were "and" they could just rerun the command and reduce or change the requirements.

internal/mithril-aggregator-discovery/src/http_config_discoverer.rs
```diff
@@ -0,0 +1,199 @@
...
+}
+
+#[async_trait::async_trait]
+impl AggregatorDiscoverer for HttpConfigAggregatorDiscoverer {
+    async fn get_available_aggregators(
+        &self,
+        network: MithrilNetwork,
+    ) -> StdResult<Box<dyn Iterator<Item = AggregatorEndpoint>>> {
+        let client = self.build_client()?;
+        let networks_configuration_response = client
```
Alenar: If possible moving the type to the `let` would make reading easier.
```suggestion
        let networks_configuration_response: NetworksConfigMessage = client
```

examples/client-cardano-database-v2/src/main.rs
```diff
@@ -39,6 +42,10 @@ pub struct Args {
     ancillary_verification_key: String,
 
     /// Aggregator endpoint URL.
+    ///
+    /// Either:
```
Alenar, line 47: Currently this comment is false since the code enforce direct url discovery, if you could share the logic from the client-cli allowing selection from the string then it would be true ! I made a comment in `mithril-client-cli/src/commands/mod.rs` that suggest a way to share that logic.

mithril-client-cli/src/commands/tools/aggregator_discovery.rs
jpraynaud: Rename to `discover-aggregator` as seen together

docs/website/root/manual/develop/nodes/mithril-client.md
jpraynaud: Fixed

docs/website/root/manual/getting-started/bootstrap-cardano-node.md
jpraynaud: Fixed

internal/mithril-aggregator-discovery/Cargo.toml
jpraynaud: Fixed

mithril-client/src/aggregator_discovery.rs
jpraynaud: Fixed

mithril-client/src/type_alias.rs
jpraynaud: Fixed

mithril-client/src/client.rs
jpraynaud: Fixed

mithril-client/src/client.rs
jpraynaud: Fixed

mithril-client/src/client.rs
jpraynaud: Fixed

mithril-client-cli/src/commands/cardano_db/verify.rs
jpraynaud: Fixed

internal/mithril-aggregator-discovery/src/model.rs
jpraynaud: Done

mithril-client-cli/src/commands/tools/mod.rs
jpraynaud: I forgot it already existed :sweat_smile:

internal/mithril-aggregator-discovery/src/http_config_discoverer.rs
jpraynaud: Fixed

mithril-client-cli/src/commands/tools/aggregator_discovery.rs
jpraynaud: You're right, it's better to use it as a positional arg. I fixed it.

mithril-client-cli/src/commands/cardano_db/show.rs
jpraynaud: Fixed

mithril-client-cli/src/commands/cardano_db/list.rs
jpraynaud: Fixed

mithril-client-cli/src/commands/tools/aggregator_discovery.rs
jpraynaud: I re-organized the `tools` command to align with the nomenclature:
- `utxo-hd` promoted to folder module
- renamed the new command to `discover-aggregator`

mithril-client-cli/src/commands/tools/aggregator_discovery.rs
jpraynaud: You're right, I switched to "and" :+1:

mithril-client-cli/src/commands/mod.rs
jpraynaud: I agree :+1: I implemented the `FromStr` trait for `AggregatorDiscoveryType`

examples/client-cardano-database-v2/src/main.rs
jpraynaud: Good catch, I fixed all the examples :+1:

docs/website/root/manual/develop/nodes/mithril-client.md
jpraynaud: Fixed

mithril-client-cli/src/commands/tools/aggregator_discovery.rs
jpraynaud: You're right, I fixed it :+1:

internal/mithril-aggregator-discovery/src/model.rs
jpraynaud: Following our conversation, these modifications will be done in a different PR.

PR #2844, opened 2025-12-05
feat: support aggregator capabilities in aggregator discovery

jpraynaud:
## Content

This PR includes **support or aggregator capabilities in the aggregator discovery process**.

The capabilities of the discovered aggregators are now available in the CLI output. This command:
```bash
./mithril-client --unstable tools discover-aggregator pre-release-preview
```

Will output:
```bash
Mithril Client CLI version: 0.12.38
+-----------------------------------------------------------------------+--------------------------+-----------------------------------------------------------------------------------------------------------------+
| Aggregator Endpoint                                                   | Aggregate Signature Type | Signed Entity Types                                                                                             |
+-----------------------------------------------------------------------+--------------------------+-----------------------------------------------------------------------------------------------------------------+
| https://aggregator.pre-release-preview.api.mithril.network/aggregator | Concatenation            | MithrilStakeDistribution,CardanoStakeDistribution,CardanoImmutableFilesFull,CardanoDatabase,CardanoTransactions |
+-----------------------------------------------------------------------+--------------------------+-----------------------------------------------------------------------------------------------------------------+
```

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Issue(s)
Relates to #2726

### DIFF
--- a/internal/mithril-aggregator-discovery/src/capabilities_discoverer.rs
+++ b/internal/mithril-aggregator-discovery/src/capabilities_discoverer.rs
@@ -289,7 +289,7 @@ mod tests {
                 .await
                 .unwrap();
 
-            let next_aggregator = aggregators.next().map(|endpoint| endpoint.into());
+            let next_aggregator = aggregators.next().map(Into::into);
             aggregator_server_mock.assert();
             assert_eq!(
                 Some(AggregatorEndpoint::new(aggregator_server.url("/"))),
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4015,7 +4015,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-aggregator-discovery"
-version = "0.1.1"
+version = "0.1.2"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -4137,7 +4137,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-client"
-version = "0.13.3"
+version = "0.13.4"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -4176,7 +4176,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-client-cli"
-version = "0.12.41"
+version = "0.12.42"
 dependencies = [
  "anyhow",
  "async-trait",
--- a/internal/mithril-aggregator-discovery/Cargo.toml
+++ b/internal/mithril-aggregator-discovery/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "mithril-aggregator-discovery"
 description = "Mechanisms to discover aggregators available in a Mithril network."
-version = "0.1.1"
+version = "0.1.2"
 authors.workspace = true
 documentation.workspace = true
 edition.workspace = true
--- a/mithril-client-cli/Cargo.toml
+++ b/mithril-client-cli/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mithril-client-cli"
-version = "0.12.41"
+version = "0.12.42"
 description = "A Mithril Client"
 authors = { workspace = true }
 edition = { workspace = true }
--- a/mithril-client/Cargo.toml
+++ b/mithril-client/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mithril-client"
-version = "0.13.3"
+version = "0.13.4"
 description = "Mithril client library"
 authors = { workspace = true }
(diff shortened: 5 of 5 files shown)


internal/mithril-aggregator-discovery/src/model.rs
```diff
@@ -43,3 +49,47 @@ impl std::fmt::Display for AggregatorEndpoint {
...
+}
+
+impl TryFrom<AggregatorEndpoint> for AggregatorEndpointWithCapabilities {
+    type Error = StdError;
+
+    fn try_from(endpoint: AggregatorEndpoint) -> Result<Self, Self::Error> {
+        let endpoint_clone = endpoint.clone();
+        let aggregator_capabilities = tokio::task::block_in_place(move || {
+            tokio::runtime::Handle::current()
+                .block_on(async move { endpoint_clone.retrieve_capabilities().await })
```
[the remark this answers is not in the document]
jpraynaud: Fixed

mithril-client-cli/src/commands/tools/aggregator_discovery.rs
```diff
@@ -48,10 +48,28 @@ impl AggregatorDiscoveryCommand {
         } else {
             let lines = lines
                 .into_iter()
-                .map(|endpoint| vec![endpoint.cell()])
+                .map(|endpoint| {
+                    let endpoint_clone = endpoint.clone();
+                    let capabilities = endpoint_clone.capabilities();
```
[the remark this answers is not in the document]
jpraynaud: False positive

internal/mithril-aggregator-discovery/src/model.rs
```diff
@@ -43,3 +49,47 @@ impl std::fmt::Display for AggregatorEndpoint {
...
+        Self { url, capabilities }
+    }
+
+    /// Get the capabilities of the aggregator
+    pub fn capabilities(&self) -> &AggregatorCapabilities {
+        &self.capabilities
+    }
+}
+
+impl TryFrom<AggregatorEndpoint> for AggregatorEndpointWithCapabilities {
```
Alenar, line 71: I think this is dangerous, like Copilot said if called outside of tokio this will panic, instead of implementing `TryFrom` I think it would be better to instead add a `retrieve_capabilities_blocking` to `AggregatorEndpoint` which would be specify to not call it outside of a tokio runtime to avoid panic.

This would still be brittle as this don't change the fact that the capabilities discoverer have to run async code in a non-async method, but would be a avoid a second level of abstraction which hide this fact.

Also it could be interesting to explore using [`Handle::try_current()`](https://docs.rs/tokio/1.49.0/tokio/runtime/struct.Handle.html#method.try_current) to avoid the panic if run outside of a tokio runtime but instead raising an error.

internal/mithril-aggregator-discovery/src/capabilities_discoverer.rs
```diff
@@ -292,7 +289,7 @@ mod tests {
                 .await
                 .unwrap();
 
-            let next_aggregator = aggregators.next();
+            let next_aggregator = aggregators.next().map(|endpoint| endpoint.into());
```
Alenar: _[Nitpick]_ you should be able to also write this conversion like this:
```suggestion
            let next_aggregator = aggregators.next().map(Into::into);
```

Alenar (approved):
LGTM with one caution (see first comment).

Overall I feel that this api is fragile, when reading this PR I see two main problems:
1. usage of async code outside async context because it returns an Iterator (the main problem)
2. the addition of a generic to the `AggregatorDiscoverer` trait, making the api less easy to use

1) may be addressed by using [`futures::stream::TryStream`](https://docs.rs/futures/latest/futures/stream/trait.TryStream.html), but there may be some hiccup when using it downstream as this means not using the standard iterator api anymore.
2) is more difficult to solve and would probably mean an extensive refactoring of the crate (maybe with returning concrete type instead of returning traits).

None of this is need immediate action, this code is okay as is.

internal/mithril-aggregator-discovery/src/capabilities_discoverer.rs
jpraynaud: Fixed

internal/mithril-aggregator-discovery/src/model.rs
jpraynaud: You're right, I will create an issue for working on these enhancements :+1:

Issue #2795, opened 2025-11-20
Implement SNARK-friendly changes in STM library


jpraynaud:
## Why
We need to refactor the STM library in order to make it SNARK-friendly. Following the definition of the architecture in #2763, we have defined an implementation plan and we want to address the refactoring part of the plan as a first step.

## What
Implement SNARK-friendly changes in the STM library.

## How
- [x] Update the Merkle tree to support multiple leaves (for concatenation and following proof systems)
- [x] Update the key registration to support the new Merkle tree structure
- [x] Move responsibility of creating single signature artifacts to the concatenation proof system (in the `proof_system` module)

PR #2866, opened 2025-12-17
SNARK-friendly STM: Support multiple Merkle Tree leaves

curiecrypt:
## Content

This PR includes the changes required for supporting multiple `MerkleTreeLeaf` types. 

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Issue(s)

Relates to #2795

### DIFF
--- a/mithril-stm/src/membership_commitment/merkle_tree/commitment.rs
+++ b/mithril-stm/src/membership_commitment/merkle_tree/commitment.rs
@@ -159,6 +159,14 @@ impl<D: Digest + FixedOutput, L: MerkleTreeLeaf> MerkleTreeBatchCommitment<D, L>
         msgp
     }
 
+    /// Serializes the Merkle Tree commitment together with a message in a single vector of bytes.
+    /// Outputs `msg || self` as a vector of bytes.
+    // todo: Do we need to concat msg to whole commitment (nr_leaves and root) or just the root?
+    #[deprecated(since = "0.5.0", note = "Use `concatenate_with_message` instead")]
+    pub fn concat_with_msg(&self, msg: &[u8]) -> Vec<u8> {
+        Self::concatenate_with_message(self, msg)
+    }
+
     /// Check a proof of a batched opening. The indices must be ordered.
     ///
     /// # Error
@@ -266,6 +274,25 @@ impl<D: Digest + FixedOutput, L: MerkleTreeLeaf> MerkleTreeBatchCommitment<D, L>
         }
 
         Err(anyhow!(MerkleTreeError::BatchPathInvalid(proof.to_bytes())))
+    }
+
+    /// Check a proof of a batched opening. The indices must be ordered.
+    ///
+    /// # Error
+    /// Returns an error if the proof is invalid.
+    // todo: Update doc.
+    // todo: Simplify the algorithm.
+    // todo: Maybe we want more granular errors, rather than only `BatchPathInvalid`
+    #[deprecated(
+        since = "0.5.0",
+        note = "Use `verify_leaves_membership_from_batch_path` instead"
+    )]
+    pub fn check(&self, batch_val: &[L], proof: &MerkleBatchPath<D>) -> StmResult<()>
+    where
+        D: FixedOutput + Clone,
+        L: MerkleTreeLeaf,
+    {
+        Self::verify_leaves_membership_from_batch_path(self, batch_val, proof)
     }
 
     /// Convert to bytes


mithril-stm/src/membership_commitment/merkle_tree/tree.rs
```diff
@@ -25,11 +26,13 @@ pub struct MerkleTree<D: Digest> {
...
     hasher: PhantomData<D>,
+    /// Phantom type to link the tree with its leafe
+    leaves: PhantomData<L>,
 }
 
-impl<D: Digest + FixedOutput> MerkleTree<D> {
+impl<D: Digest + FixedOutput, L: MerkleTreeLeaf> MerkleTree<D, L> {
     /// Provided a non-empty list of leaves, `create` generates its corresponding `MerkleTree`.
-    pub(crate) fn new(leaves: &[MerkleTreeLeaf]) -> MerkleTree<D> {
+    pub(crate) fn new(leaves: &[MerkleTreeConcatenationLeaf]) -> MerkleTree<D, L> {
```
jpraynaud: The Merkle tree should not depend on any implementation of the `MerkleTreeLeaf` trait (except in the tests):
```suggestion
    pub(crate) fn new(leaves: &[L]) -> MerkleTree<D, L> {
```

mithril-stm/src/membership_commitment/merkle_tree/commitment.rs
jpraynaud, line 1: The commitments should be using a generic `MerkleTreeLeaf` and no concrete implementation (except in tests)

mithril-stm/src/protocol/key_registration.rs
jpraynaud, line 1: The goal is to have a key registration agnostic of the leaves, which means that we don't want the key registration types to depend on the `MerkleTreeLeaf`. This must be internal stuff only (where we use the `ConcatenationMerkleLeaf` for the current Concatenation proof) . Thus only files of the `mithril-stm` library should be modified in this PR.

mithril-stm/src/membership_commitment/merkle_tree/leaf.rs
```diff
@@ -6,35 +6,43 @@ use crate::{Stake, StmResult, VerificationKey, signature_scheme::BlsVerification
 
 use super::MerkleTreeError;
 
-/// The values that are committed in the Merkle Tree.
-/// Namely, a verified `VerificationKey` and its corresponding stake.
+/// Trait defining the `from/to_bytes()` functionalities of the `MerkleTreeLeaf`.
+pub trait MerkleTreeLeaf: Clone + Send + Sync + Copy {
+    fn from_bytes(bytes: &[u8]) -> StmResult<Self>
+    where
+        Self: Sized;
```
jpraynaud: This one is not needed as we only want to use the bytes representation of the leaf in the Merkle tree.
Maybe we need to rename the function soemthing like `as_bytes_for_merkle_tree` to avoid confusion.

mithril-stm/src/membership_commitment/merkle_tree/tree.rs
```diff
@@ -61,12 +63,13 @@ impl<D: Digest + FixedOutput> MerkleTree<D> {
...
             leaf_off: num_nodes - n,
             hasher: PhantomData,
+            leaves: PhantomData,
         }
     }
 
     /// Provided a non-empty list of leaves, `create` generates its corresponding `MerkleTree`.
     #[deprecated(since = "0.5.0", note = "Use `new` instead")]
-    pub fn create(leaves: &[MerkleTreeLeaf]) -> MerkleTree<D> {
+    pub fn create(leaves: &[L]) -> MerkleTree<D, L> {
```
jpraynaud, line 72: This one can be removed at this stage (and maybe other deprecated functions in `0.5.0`).

curiecrypt: I removed the deprecated ones for batch compatibility. However, I left the other ones in order not to get unused warning. I will deal with deprecations in another pr.

mithril-stm/src/membership_commitment/merkle_tree/tree.rs
```diff
@@ -12,7 +12,7 @@ use super::{
 
 /// Tree of hashes, providing a commitment of data and its ordering.
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
-pub struct MerkleTree<D: Digest> {
+pub struct MerkleTree<D: Digest, L: MerkleTreeLeaf> {
```
damrobi, line 15: Do we want to change the Digest trait of the Merkle tree to the new MembershipDigest now? Same for the Merkle tree commitment?

mithril-stm/src/membership_commitment/merkle_tree/commitment.rs
```diff
@@ -308,14 +292,15 @@ impl<D: Digest + FixedOutput> MerkleTreeBatchCommitment<D> {
             root,
             nr_leaves,
             hasher: PhantomData,
+            leaf_type: Default::default(),
```
jpraynaud, line 322:
```suggestion
            leaf_type: PhantomData,
```

mithril-stm/src/membership_commitment/merkle_tree/leaf.rs
```diff
@@ -6,35 +6,43 @@ use crate::{Stake, StmResult, VerificationKey, signature_scheme::BlsVerification
 
 use super::MerkleTreeError;
 
-/// The values that are committed in the Merkle Tree.
-/// Namely, a verified `VerificationKey` and its corresponding stake.
+/// Trait defining the `to_bytes()` functionality of the `MerkleTreeLeaf`.
+pub trait MerkleTreeLeaf: Clone + Send + Sync + Copy {
+    fn to_bytes(self) -> Vec<u8>;
+}
```
jpraynaud: The trait lacks a bit of clarity, also I think that we should not consume the leaf when converting to bytes for Merkle tree:
```suggestion
/// Trait implemented to be used as a Merkle tree leaf.
pub trait MerkleTreeLeaf: Clone + Send + Sync + Copy {
    /// Converts the Merkle tree leaf to a bytes representation used internally by the Merkle tree
    fn as_bytes_for_merkle_tree(&self) -> Vec<u8>;
}
```

mithril-stm/src/membership_commitment/merkle_tree/leaf.rs
```diff
@@ -6,35 +6,43 @@ use crate::{Stake, StmResult, VerificationKey, signature_scheme::BlsVerification
...
-pub struct MerkleTreeLeaf(pub BlsVerificationKey, pub Stake);
+pub struct MerkleTreeConcatenationLeaf(pub BlsVerificationKey, pub Stake);
+
+impl MerkleTreeLeaf for MerkleTreeConcatenationLeaf {
+    fn to_bytes(self) -> Vec<u8> {
+        let mut result = [0u8; 104];
+        result[..96].copy_from_slice(&self.0.to_bytes());
+        result[96..].copy_from_slice(&self.1.to_be_bytes());
+        result.to_vec()
+    }
```
jpraynaud, line 32: This `to_bytes` function should be implemented out of the trait, and should be called by the `as_bytes_for_merkle_tree` function of the `MerkleTreeLeaf` trait.

mithril-stm/src/membership_commitment/merkle_tree/leaf.rs
```diff
@@ -6,35 +6,43 @@ use crate::{Stake, StmResult, VerificationKey, signature_scheme::BlsVerification
...
         let stake = Stake::from_be_bytes(u64_bytes);
-        Ok(MerkleTreeLeaf(pk, stake))
-    }
-    pub(crate) fn to_bytes(self) -> [u8; 104] {
-        let mut result = [0u8; 104];
-        result[..96].copy_from_slice(&self.0.to_bytes());
-        result[96..].copy_from_slice(&self.1.to_be_bytes());
-        result
+        Ok(MerkleTreeConcatenationLeaf(pk, stake))
     }
```
jpraynaud: This function should still be called `from_bytes`.

mithril-stm/src/membership_commitment/merkle_tree/leaf.rs
```diff
@@ -6,35 +6,43 @@ use crate::{Stake, StmResult, VerificationKey, signature_scheme::BlsVerification
 
 use super::MerkleTreeError;
 
-/// The values that are committed in the Merkle Tree.
-/// Namely, a verified `VerificationKey` and its corresponding stake.
+/// Trait defining the `to_bytes()` functionality of the `MerkleTreeLeaf`.
+pub trait MerkleTreeLeaf: Clone + Send + Sync + Copy {
```
damrobi, line 10: Do we want to enforce the Digest type used at the leaf level? To make sure it matches the one at the tree level ((Blake2b, Concatenation) and (Poseidon, SNARK)). Or do we want to keep the flexibility in the types of trees we can have?

mithril-stm/src/membership_commitment/merkle_tree/commitment.rs
```diff
@@ -102,12 +107,13 @@ impl<D: Digest + FixedOutput> MerkleTreeCommitment<D> {
     }
 
     /// Extract a `MerkleTreeCommitment` from a byte slice.
-    pub fn from_bytes(bytes: &[u8]) -> StmResult<MerkleTreeCommitment<D>> {
+    pub fn from_bytes(bytes: &[u8]) -> StmResult<MerkleTreeCommitment<D, L>> {
         let root = bytes.to_vec();
```
damrobi, line 111: Check the size of the bytes before deserializing?
```suggestion
        if bytes.len() != 32 {
            return Err(anyhow!(MerkleTreeError::SerializationError)).with_context(||"Failed to deserialize bytes into MerkleTreeCommitment");
        }
        let root = bytes.to_vec();
```

hjeljeli32: As suggested by @damrobi, I think having a length check before deserialization is a safer way.
Regarding the expected length, are we committing to 32-byte Merkle roots for both concatenation and SNARK paths? If yes, hard-coding 32 is great. If no, we may want to parameterize the expected root length so this parsing stays generic.

mithril-stm/src/membership_commitment/merkle_tree/commitment.rs
hjeljeli32, line 316: I understood from this code that we ensure a minimum length (at least 8 bytes for nr_leaves), but not a strict length for the Merkle root itself. Please correct me if my understanding is not good.

If the Merkle root is meant to be a fixed-size digest, it might be worth adding an explicit length check (e.g. bytes.len() == 8 + DIGEST_SIZE) or validating root.len() after parsing.

hjeljeli32 (approved):
LGTM. 
I only left two small suggestions around length checks in from_bytes deserialization functions, aligned what @damrobi suggested.

mithril-stm/src/membership_commitment/merkle_tree/tree.rs
jpraynaud, line 1: Can you add a golden test for the root of the Merkle tree so that we make sure we don't introduce a breaking change in the computation?

mithril-stm/src/membership_commitment/merkle_tree/tree.rs
jpraynaud: This is not needed, the Merkle tree must be agnostic of the `MembershipDigest`.

mithril-stm/src/membership_commitment/merkle_tree/leaf.rs
jpraynaud: This is something that is done at the Merkle tree level and not the leaves.

mithril-stm/src/membership_commitment/merkle_tree/commitment.rs
curiecrypt: Since this might cause some confusion when we include Poseidon hash for snark proof, we decided to handle this in the future. @hjeljeli32, @damrobi, @jpraynaud

damrobi (approved):
LGTM :+1:

Issue #2833, opened 2025-12-01
DMQ testing with SPOs on `preview`


jpraynaud:
## Why
We want to test the DMQ network in the `preview` network with volunteer SPOs before we can plan for the deployment to the mainnet.

## What

Test and qualify the DMQ network in the `pre-release-network`.

### dmq-preview
- [x] Host a static build of the DMQ node
- [x] Create a local Docker image of the DMQ node built on the static build downloaded
- [x] Update the infrastructure to support DMQ node 
- [x] Deploy on `dmq-preview` network with leader aggregator
- [x] Deploy on `dmq-follower-preview` network with follower aggregator

### pre-release-preview
- [x] Create a Docker image of the DMQ node (tools support team)
- [x] Make DMQ node build available on CDN (tools support team)
- [x] Have DMQ mini-protocols published in Pallas on crates.io
- [x] Activate the `future_dmq` feature by default
- [x] Update e2e test to run with DMQ node in e2e tests in CI
- [x] Update the infrastructure to support DMQ node
  - [x] Configure DMQ image in CI
  - [x] Configure GitHub environments vars
    - [x] `testing-preview`
    - [x] `pre-release-preview` (will be done in #2830)
    - [x] `release-preprod` (will be scheduled in #2830)
    - [x] `release-mainnet` (will be scheduled in #2830)
  - [x] Implement block-producer/relay infrastructure for DMQ nodes of signer
  - [x] Activate DMQ in pre-release
  - [x] Remove all TODOs
- [x] Update documentation
  - [x] Update network configuration with bootstrap peers
  - [x] Update DMQ config with tracers in signer and aggregator setup guides
  - [x] Update DMQ download location in signer and aggregator setup guides
  - [x] Prepare deploy on `pre-release-preview` with next distribution
- [x] Prepare testing program with volunteer SPOs
- [x] Plan for DMQ activation in release
- [x] ~~Activate the DMQ node in `pre-release-preview`~~ (will be done in #2830)

PR #2867, opened 2025-12-18
fix: KES signer period

jpraynaud:
## Content

This PR includes a **fix to use the current KES period as input to the KES signer when signing instead of the the number of evolutions of the KES secret key since creation**. This avoids circular dependency where the Operational certificate (returned by the KES signer) is needed to compute the correct number of evolutions of the KES secret key. 
It also includes some clarification on the KES evolutions used to verify the KES signature.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Issue(s)
Relates to #2833

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3700,7 +3700,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-aggregator"
-version = "0.8.3"
+version = "0.8.4"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -3995,7 +3995,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-common"
-version = "0.6.38"
+version = "0.6.39"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -4079,7 +4079,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-end-to-end"
-version = "0.4.112"
+version = "0.4.113"
 dependencies = [
  "anyhow",
  "async-recursion",
@@ -4230,7 +4230,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-signer"
-version = "0.3.0"
+version = "0.3.1"
 dependencies = [
  "anyhow",
  "async-trait",
--- a/mithril-aggregator/Cargo.toml
+++ b/mithril-aggregator/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mithril-aggregator"
-version = "0.8.3"
+version = "0.8.4"
 description = "A Mithril Aggregator server"
 authors = { workspace = true }
 edition = { workspace = true }
--- a/mithril-common/Cargo.toml
+++ b/mithril-common/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mithril-common"
-version = "0.6.38"
+version = "0.6.39"
 description = "Common types, interfaces, and utilities for Mithril nodes."
 authors = { workspace = true }
 edition = { workspace = true }
--- a/mithril-signer/Cargo.toml
+++ b/mithril-signer/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mithril-signer"
-version = "0.3.0"
+version = "0.3.1"
 description = "A Mithril Signer"
 authors = { workspace = true }
 edition = { workspace = true }
--- a/mithril-test-lab/mithril-end-to-end/Cargo.toml
+++ b/mithril-test-lab/mithril-end-to-end/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mithril-end-to-end"
-version = "0.4.112"
+version = "0.4.113"
 authors = { workspace = true }
 edition = { workspace = true }
 documentation = { workspace = true }


mithril-common/src/test/crypto_helper/cardano/kes/setup.rs
```diff
@@ -31,16 +31,12 @@ pub fn create_kes_cryptographic_material(
...
     let mut dummy_buffer = [0u8; Sum6Kes::SIZE + 4];
     let mut dummy_seed = [party_idx as u8; 32];
-    let (mut kes_secret_key, kes_verification_key) =
+    let (kes_secret_key, kes_verification_key) =
         Sum6Kes::keygen(&mut dummy_buffer, &mut dummy_seed);
-    for _ in 0..kes_period {
-        kes_secret_key
-            .update()
-            .expect("KES secret key update should not fail");
-    }
```
Alenar, line 40: For my understanding, what does this change entails ?

mithril-signer/src/runtime/runner.rs
```diff
@@ -184,7 +184,7 @@ impl Runner for SignerRunner {
         let stake = stake_distribution
             .get(&self.services.single_signer.get_party_id())
             .ok_or_else(RunnerError::NoStakeForSelf)?;
-        let (operational_certificate, protocol_operational_certificate) = match &self
+        let (_operational_certificate, protocol_operational_certificate) = match &self
```
Alenar: This could be further simplified :)
```suggestion
        let protocol_operational_certificate = match &self
```

mithril-common/src/entities/signer.rs
```diff
@@ -158,8 +158,10 @@ pub struct SignerWithStake {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub operational_certificate: Option<ProtocolOpCert>,
 
-    /// The kes period used to compute the verification key signature
-    // TODO: This kes period should not be used as is and should probably be within an allowed range of kes period for the epoch
+    /// The number of evolutions of the KES key since the start KES period of the operational certificate at the time of signature.
+    ///
+    /// Note: the naming 'kes_period' lacks clarity and should be renamed to 'kes_evolutions'
+    // TODO: This 'kes_period' should be rename to 'kes_evolutions' to avoid confusion
```
Alenar: I'm confused, the place in the code were this value is assigned is in the signer runtime, the value now assigned is:
```rust
let kes_period = self.services.chain_observer.get_current_kes_period().await?;
``` 
Before it was:
```rust
let kes_period = match operational_certificate {
    Some(operational_certificate) => Some(
        self.services
            .chain_observer
            .get_current_kes_period()
            .await?
            .unwrap_or_default()
            - operational_certificate.get_start_kes_period() as KesPeriod,
        ),
        None => None,
    };
```
I may be missing something but this is no KES evolutions, it was the evolution before the change of this PR in the signer runtime but now it seems to be the KES period at the time of the signer creation.

Also this change is far from neutral and also impact messages, the impact is low because we don't really use this value (it's ignored by the aggregator verifier), it's mostly informative.

Alenar (commented):
Maybe we should introduce two new types: `KesPeriod` and `KesEvolution`, the two would be wrappers like we do for the `BlockNumber`.
There could be a twist: the implementation of the `sub` operators on `KesPeriod` could result in `KesEvolution`.

mithril-common/src/test/crypto_helper/cardano/kes/setup.rs
jpraynaud: These updates do not have any impact on the serialized secret key, which explains why they have been removed.

mithril-common/src/entities/signer.rs
jpraynaud: You're right, good catch :+1:  

I have fixed that by using the `let current_kes_period = self.services.chain_observer.get_current_kes_period().await?;` to build the protocol initializer and keep unchanged the `kes_evolutions` to create the `Signer` entity.

mithril-common/src/entities/signer.rs
```diff
@@ -158,10 +158,10 @@ pub struct SignerWithStake {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub operational_certificate: Option<ProtocolOpCert>,
 
-    /// The kes period used to compute the verification key signature
-    // TODO: This kes period should not be used as is and should probably be within an allowed range of kes period for the epoch
+    /// The number of evolutions of the KES key since the start KES period of the operational certificate at the time of signature.
     #[serde(skip_serializing_if = "Option::is_none")]
-    pub kes_period: Option<KesPeriod>,
+    #[serde(rename = "kes_period")]
```
Alenar: You should be able to combine the two attribute :)
```suggestion
    #[serde(rename = "kes_period", skip_serializing_if = "Option::is_none")]
```

mithril-common/src/messages/message_parts/signer.rs
```diff
@@ -36,11 +36,10 @@ pub struct SignerWithStakeMessagePart {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub operational_certificate: Option<HexEncodedOpCert>,
 
-    /// The KES period used to compute the verification key signature
-    // TODO: This KES period should not be used as is and should probably be
-    //       within an allowed range of KES periods for the epoch.
+    /// The number of evolutions of the KES key since the start KES period of the operational certificate at the time of signature.
     #[serde(skip_serializing_if = "Option::is_none")]
-    pub kes_period: Option<KesPeriod>,
+    #[serde(rename = "kes_period")]
```
Alenar:
```suggestion
    #[serde(rename = "kes_period", skip_serializing_if = "Option::is_none")]
```

mithril-common/src/messages/message_parts/signer.rs
```diff
@@ -166,11 +165,10 @@ pub struct SignerMessagePart {
...
```
Alenar:
```suggestion
    #[serde(rename = "kes_period", skip_serializing_if = "Option::is_none")]
```

mithril-common/src/messages/register_signer.rs
```diff
@@ -34,11 +34,10 @@ pub struct RegisterSignerMessage {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub operational_certificate: Option<HexEncodedOpCert>,
 
-    /// The KES period used to compute the verification key signature
-    // TODO: This KES period should not be used as is and should probably be
-    //       within an allowed range of KES periods for the epoch.
+    /// The number of evolutions of the KES key since the start KES period of the operational certificate at the time of signature.
     #[serde(skip_serializing_if = "Option::is_none")]
-    pub kes_period: Option<KesPeriod>,
+    #[serde(rename = "kes_period")]
```
Alenar:
```suggestion
    #[serde(rename = "kes_period", skip_serializing_if = "Option::is_none")]
```

mithril-common/src/entities/signer.rs
jpraynaud: Fixed

mithril-common/src/messages/message_parts/signer.rs
jpraynaud: Fixed

mithril-common/src/messages/message_parts/signer.rs
jpraynaud: Fixed

mithril-common/src/messages/register_signer.rs
jpraynaud: Fixed

mithril-common/src/crypto_helper/cardano/kes/verifier_standard.rs
```diff
@@ -19,19 +19,19 @@ impl KesVerifier for KesVerifierStandard {
...
     ) -> StdResult<()> {
         operational_certificate
             .validate()
             .map_err(|_| KesVerifyError::OpCertInvalid)?;
 
-        // Check if the KES period in the operational certificate matches the provided KES period +/- 1
-        let kes_period_try_min = std::cmp::max(0, kes_period.saturating_sub(1));
-        let kes_period_try_max = std::cmp::min(64, kes_period.saturating_add(1));
-        for kes_period_try in kes_period_try_min..kes_period_try_max {
+        // Check if the signature verifies for the provided KES evolutions value +/- 1
```
curiecrypt: Should we also test valid/invalid cases for +-1 evolutions in the verification?
I cannot see a test that covers these edge cases

mithril-common/src/entities/signer.rs
```diff
@@ -71,7 +71,7 @@ impl Signer {
             verification_key,
             verification_key_signature,
             operational_certificate,
-            kes_period,
+            kes_evolutions: kes_period,
```
curiecrypt: I would rather rename the function argument as `kes_evolutions`

mithril-common/src/entities/signer.rs
```diff
@@ -200,7 +199,7 @@ impl SignerWithStake {
...
```
curiecrypt: same as above

mithril-common/src/messages/message_parts/signer.rs
```diff
@@ -256,7 +252,7 @@ impl Debug for SignerMessagePart {
                     "operational_certificate",
                     &format_args!("{:?}", self.operational_certificate),
                 )
-                .field("kes_period", &format_args!("{:?}", self.kes_period))
+                .field("kes_period", &format_args!("{:?}", self.kes_evolutions))
```
curiecrypt: Is it possible to rename the field as well? 
```suggestion
                .field("kes_evolutions", &format_args!("{:?}", self.kes_evolutions))
```

mithril-aggregator/src/database/record/signer_registration.rs
```diff
@@ -25,7 +25,10 @@ pub struct SignerRegistrationRecord {
     /// Operational certificate of the stake pool operator associated to the signer
     pub operational_certificate: Option<HexEncodedOpCert>,
 
-    /// The kes period used to compute the verification key signature
+    /// The number of evolutions of the KES key since the start KES period of the operational certificate at the time of signature.
+    ///
+    /// Note: the naming 'kes_period' lacks clarity and should be renamed to 'kes_evolutions'
+    // TODO: This 'kes_period' should be renamed to 'kes_evolutions' to avoid confusion
```
curiecrypt, line 31: +1

mithril-common/src/crypto_helper/cardano/kes/signer_with_key.rs
```diff
@@ -61,85 +65,114 @@ mod tests {
...
         let KesCryptographicMaterialForTest {
             party_id: _,
             operational_certificate_file,
             kes_secret_key_file,
         } = create_kes_cryptographic_material(
             1 as KesPartyIndexForTest,
-            0 as KesPeriod,
-            "create_valid_signature_for_message",
+            start_kes_period,
+            current_function!(),
```
curiecrypt, line 85: Just out of curiosity... Why do we need this `current_function()` now and not in the previous version?

jpraynaud: This macro returns the name of the current function and we use it to create a sub-folder for the cryptographic material.
This avoids having multiple use the same sub-folder by mistake which usually creates flakiness in the CI when tests are run in parallel.

mithril-common/src/entities/signer.rs
jpraynaud: Fixed

mithril-common/src/entities/signer.rs
jpraynaud: Fixed

mithril-common/src/messages/message_parts/signer.rs
jpraynaud: Fixed

mithril-common/src/crypto_helper/cardano/kes/verifier_standard.rs
jpraynaud: Yes, I added a test for that and also fixed a problem with the range which was too narrow.

PR #2869, opened 2025-12-19
refactor: introduce `KesEvolutions` type

jpraynaud:
## Content

This PR includes the **introduction of the type `KesEvolutions` to avoid confusion with `KesPeriod` and detect mismatch at compile time**.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation:

## Issue(s)
Relates to #2833

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3700,7 +3700,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-aggregator"
-version = "0.8.5"
+version = "0.8.6"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -3845,7 +3845,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-cardano-node-chain"
-version = "0.1.12"
+version = "0.1.13"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -3995,7 +3995,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-common"
-version = "0.6.39"
+version = "0.6.40"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -4037,7 +4037,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-dmq"
-version = "0.1.16"
+version = "0.1.17"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -4230,7 +4230,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-signer"
-version = "0.3.1"
+version = "0.3.2"
 dependencies = [
  "anyhow",
  "async-trait",
--- a/internal/cardano-node/mithril-cardano-node-chain/Cargo.toml
+++ b/internal/cardano-node/mithril-cardano-node-chain/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mithril-cardano-node-chain"
-version = "0.1.12"
+version = "0.1.13"
 authors.workspace = true
 documentation.workspace = true
 edition.workspace = true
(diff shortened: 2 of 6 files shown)


mithril-common/src/protocol/signer_builder.rs
```diff
@@ -101,7 +102,9 @@ impl SignerBuilder {
         let protocol_initializer = ProtocolInitializer::setup(
             self.protocol_parameters.clone().into(),
             kes_signer,
-            signer_with_stake.kes_evolutions,
+            signer_with_stake
+                .kes_evolutions
+                .map(|kes_evolutions| KesPeriod(0) + kes_evolutions),
```
Alenar, line 107: Why does this change entails ? I've the feeling that the code could have been left at is.

mithril-common/src/crypto_helper/cardano/kes/kes_evolutions.rs
```diff
@@ -0,0 +1,165 @@
...
+    fn from(value: u32) -> Self {
+        KesEvolutions(value.into())
+    }
+}
+
+impl From<u64> for KesEvolutions {
+    fn from(value: u64) -> Self {
+        KesEvolutions(value)
+    }
+}
```
Alenar: In all other cases we willingly did not add any direct conversion from/to the wrapper type in order to avoid making it too easy to escape from the wrapper, else this would defeat the whole objective of using the compiler to check that types are not mixed incorrectly.

We made one exception for the fallible conversion to `i64` as it's required at multiple places in the persistence layer and using explicitly the wrapper here did not bring much.

This have the added benefit of  forcing usage of the `Type(number)` notation, which is often more clear when reading the code than using `.into()`.

> [!NOTE]
> Now that we have the Dev ADR it would make sense to add an entry on how to implement arithmetic wrapper types and why avoiding implementing conversions on them.

mithril-common/src/crypto_helper/cardano/kes/kes_period.rs
```diff
@@ -0,0 +1,165 @@
...
+    fn from(value: u32) -> Self {
+        KesPeriod(value.into())
+    }
+}
+
+impl From<u64> for KesPeriod {
+    fn from(value: u64) -> Self {
+        KesPeriod(value)
+    }
+}
```
Alenar: In all other cases we willingly did not add any direct conversion from/to the wrapper type in order to avoid making it too easy to escape from the wrapper, else this would defeat the whole objective of using the compiler to check that types are not mixed incorrectly.

We made one exception for the fallible conversion to `i64` as it's required at multiple places in the persistence layer and using explicitly the wrapper here did not bring much.

This have the added benefit of  forcing usage of the `Type(number)` notation, which is often more clear when reading the code than using `.into()`.

> [!NOTE]
> Now that we have the Dev ADR it would make sense to add an entry on how to implement arithmetic wrapper types and why avoiding implementing conversions on them.

Alenar (approved):
LGTM with one reservation.

I think we should not add any `From` to wrapper types except to `i64` see the comment for the rational.

mithril-common/src/protocol/signer_builder.rs
jpraynaud: It returns a `Vec<KesPeriod>, when `signer_with_stake.kes_evolutions` returns a `Vec<KesEvolutions>`.

mithril-common/src/crypto_helper/cardano/kes/kes_evolutions.rs
jpraynaud: I agree, I have removed the conversions from primitive types :+1: 
I have also added a dev ADR explaining the guidelines for implementing an arithmetic wrapper type.

mithril-common/src/crypto_helper/cardano/kes/kes_period.rs
jpraynaud: Fixed

PR #2880, opened 2026-01-05
SNARK-friendly STM: New Key Registration

### LINKED ISSUE #2791: Refactoring STM library for SNARK-friendliness
## Why
We need to refactor the STM library in order to make it SNARK-friendly. Following the definition of the architecture in #2763, we have defined an implementation plan and we want to address the refactoring part of the plan as a first step.

## What
Implement the refactoring of the STM library for the concatenation proof.

## How
- [x] **Step 1**: Code re-organization
  - Create a `commitment_scheme` module and move `merkle_tree` in it
  - Create a `signature_scheme` module and move `bls_multi_signature` and `schnorr_signature` in it
  - Create a `proof_system` module and move `concatenation_proof` in it
  - Create a `core` or `protocol` module and move all the other modules in it
- [x] **Step 2**: Code simplification
  - Move errors to their belonging modules
  - Remove the `future_proof_system` feature and replace it with the existing `future_snark` feature
  - Replace usage of `<D: Digest>` generic with a more general one that includes digest functions for concatenation and SNARK
  - Remove the `BasicVerifier` implementation
- [x] **Step 3**: Implement SNARK-friendly changes for concatenation proof
  - Update the Merkle tree to support multiple leaves (for concatenation and following proof systems)
  - Update the key registration to support the new Merkle tree structure
  - Move responsibility of creating single signature artifacts to the concatenation proof system (in the `proof_system` module)

curiecrypt:
## Content

This PR includes new key registration functionality. 
New key registration supports registering different types of keys specific to proof system. 
Changing the key registration results in some important modifications in the other (sub-)modules in STM library.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Comments

This branch provides the STM library to be fully convenient for integrating SNARK proofs. 

## Issue(s)

Closes #2791

### DIFF
--- a/mithril-stm/src/protocol/key_registration/register.rs
+++ b/mithril-stm/src/protocol/key_registration/register.rs
@@ -1,4 +1,3 @@
-use anyhow::anyhow;
 use digest::{Digest, FixedOutput};
 use std::collections::BTreeSet;
 
@@ -32,9 +31,7 @@ impl KeyRegistration {
             self.registration_entries.insert(*entry);
             return Ok(());
         }
-        Err(anyhow!(RegisterError::KeyRegistered(Box::new(
-            entry.get_bls_verification_key()
-        ))))
+        Err(RegisterError::KeyRegistered(Box::new(entry.get_bls_verification_key())).into())
     }
 
     /// Registers a new signer with the given verification key proof of possession and stake.
@@ -69,7 +66,7 @@ impl KeyRegistration {
             .iter()
             .nth(*signer_index as usize)
             .cloned()
-            .ok_or_else(|| anyhow!(RegisterError::UnregisteredIndex))
+            .ok_or_else(|| RegisterError::UnregisteredIndex.into())
     }
 
     /// Converts the KeyRegistration into a Merkle tree


curiecrypt:
@jpraynaud tests related to hex encodings for `AggregateVerificationKey` fail:
- `restorations_from_json_hex_and_bytes_hex_give_same_certificate`
- `genesis_certificate::restorations_from_bytes_hex_succeeds`
Couldn't resolve why. 
Json serializations for `AggregateVerificationKey` work. (Golden string matches with the one produced by main implementation) See #2904

mithril-common/src/crypto_helper/types/alias.rs
```diff
@@ -53,3 +53,6 @@ pub type ProtocolInitializerError = ProtocolInitializerErrorWrapper;
 
 /// Alias of [MithrilStm:AggregationError](enum@mithril_stm::AggregationError).
 pub type ProtocolAggregationError = AggregationError;
+
+/// Alias of [MithrilStm:MithrilMembershipDigest](struct@mithril_stm::MithrilMembershipDigest).
+pub type ProtocolDigest = MithrilMembershipDigest;
```
jpraynaud: Can you get rid of the `D` type in this file and replace it with `ProtocolMembershipDigest` (and move its definition above in the file)?
```suggestion
pub type ProtocolMembershipDigest = MithrilMembershipDigest;
```

mithril-stm/README.md
```diff
@@ -94,20 +94,27 @@ let parties = (0..nparties)
...
+let mut key_reg = KeyRegistration::initialize();
 
 let mut ps: Vec<Initializer> = Vec::with_capacity(nparties as usize);
 for stake in parties {
     let p = Initializer::new(params, stake, &mut rng);
-    key_reg.register(stake, p.get_verification_key_proof_of_possession()).unwrap();
+    let entry = RegistrationEntry::new(
+        p.get_verification_key_proof_of_possession(),
+        #[cfg(feature = "future_snark")]
+        None,
```
jpraynaud: This one can be omitted at the moment.

mithril-stm/examples/key_registration.rs
```diff
@@ -128,39 +129,55 @@ fn main() {
...
+) -> ClosedKeyRegistration {
+    let mut local_keyreg = KeyRegistration::initialize();
     // data, such as the public key, stake and id.
-    for (&pk, id) in pks.iter().zip(ids.iter()) {
-        local_keyreg.register(*id, pk).unwrap();
+    for (pk, _) in pks.iter().zip(ids.iter()) {
+        let entry = RegistrationEntry::new(
+            *pk,
+            #[cfg(feature = "future_snark")]
+            None,
```
jpraynaud: Same here, this can be omitted until `future_snark` is activated.

mithril-stm/src/lib.rs
```diff
@@ -58,20 +58,25 @@
 //!     // Create keys for this party
 //!     let p = Initializer::new(params, stake, &mut rng);
 //!     // Register keys with the KeyRegistration service
-//!     key_reg
-//!         .register(p.stake, p.get_verification_key_proof_of_possession())
-//!         .unwrap();
+//!     let entry = RegistrationEntry::new(
+//!         p.get_verification_key_proof_of_possession(),
+//!         #[cfg(feature = "future_snark")]
+//!         None,
```
jpraynaud: Same comment here.

mithril-stm/src/proof_system/concatenation/clerk.rs
```diff
@@ -0,0 +1,271 @@
+use anyhow::anyhow;
+use std::collections::{BTreeMap, HashMap, HashSet};
+
+use crate::{
+    AggregationError, ClosedKeyRegistration, Index, MembershipDigest, Parameters, Signer,
+    SingleSignatureWithRegisteredParty, StmResult, proof_system::ConcatenationProofKey,
+};
+
+/// The `ConcatenationClerk` is responsible for managing the proof system related to
+/// concatenation signatures.
```
jpraynaud, line 11:
```suggestion
/// The `ConcatenationClerk` is responsible for creating concatenation proofs.
```

mithril-stm/src/proof_system/concatenation/clerk.rs
```diff
@@ -0,0 +1,271 @@
...
+
+    /// Given a slice of `sig_reg_list`, this function returns a new list of `sig_reg_list` with only valid indices.
+    /// In case of conflict (having several signatures for the same index)
+    /// it selects the smallest signature (i.e. takes the signature with the smallest scalar).
+    /// The function selects at least `self.k` indexes.
+    ///  # Error
+    /// If there is no sufficient signatures, then the function fails.
+    // todo: We need to agree on a criteria to dedup (by default we use a BTreeMap that guarantees keys order)
+    // todo: not good, because it only removes index if there is a conflict (see benches)
+    pub fn select_valid_signatures_for_k_indices<D: MembershipDigest>(
```
jpraynaud, line 65: This deduplication method should probably not be specific to Concatenation proof.

mithril-stm/src/proof_system/concatenation/proof_key.rs
```diff
@@ -0,0 +1,54 @@
...
+    },
+};
+
+/// Aggregate verification key of the concatenation proof system.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(bound(
+    serialize = "MerkleBatchPath<D::ConcatenationHash>: Serialize",
+    deserialize = "MerkleBatchPath<D::ConcatenationHash>: Deserialize<'de>"
+))]
+pub struct ConcatenationProofKey<D: MembershipDigest> {
```
jpraynaud: We should precise that this is a verification key (and maybe an aggregate verification key?):
```suggestion
pub struct AggregateVerificationKeyForConcatenation<D: MembershipDigest> {
```

mithril-stm/src/proof_system/concatenation/aggregate_key.rs
```diff
@@ -0,0 +1,54 @@
...
+}
+
+impl<D: MembershipDigest> ConcatenationProofKey<D> {
+    /// Get the Merkle tree batch commitment.
+    pub(crate) fn get_merkle_tree_batch_commitment(
+        &self,
+    ) -> MerkleTreeBatchCommitment<D::ConcatenationHash, MerkleTreeConcatenationLeaf> {
+        self.mt_commitment.clone()
+    }
+    /// Get the total stake.
```
jpraynaud, line 29:
```suggestion
    }
    
    /// Get the total stake.
```

mithril-stm/src/proof_system/concatenation/signer.rs
```diff
@@ -0,0 +1,94 @@
...
+/// Concatenation Proof single signature generator.
+/// Contains the signer's stake and signing key besides the total stake of the closed registration,
+/// protocol parameters, and the Merkle Tree commitment of the key registration for concatenation
+/// proof.
+#[derive(Debug, Clone)]
+pub(crate) struct ConcatenationProofSigner<D: MembershipDigest> {
+    pub(crate) stake: Stake,
+    total_stake: Stake,
+    parameters: Parameters,
+    signing_key: BlsSigningKey,
```
jpraynaud, line 22: Would it make sense to create a specific type for signing key?
```suggestion
    signing_key: SigningKeyForConcatenation,
```

mithril-stm/src/proof_system/concatenation/aggregate_key.rs
jpraynaud, line 1: Maybe the module also needs to be renamed into `aggregate_key.rs`?

mithril-stm/src/proof_system/concatenation/proof.rs
```diff
@@ -31,24 +33,32 @@ impl<D: MembershipDigest> ConcatenationProof<D> {
...
+                .key_registration
+                .get_registration_entry_for_index(&sig.signer_index)?;
+            sig_reg_list.push(SingleSignatureWithRegisteredParty {
                 sig: sig.clone(),
-                reg_party: clerk.closed_reg.reg_parties[sig.signer_index as usize],
-            })
-            .collect::<Vec<SingleSignatureWithRegisteredParty>>();
+                reg_party,
+            });
+        }
```
jpraynaud: This is more functional and cleaner way to implement it:

```suggestion
        let sig_reg_list = sigs
            .iter()
            .map(|sig| {
                clerk
                    .closed_key_registration
                    .key_registration
                    .get_registration_entry_for_index(&sig.signer_index)
                    .map(|reg_party| SingleSignatureWithRegisteredParty {
                        sig: sig.clone(),
                        reg_party,
                    })
            })
            .collect::<Result<Vec<_>, _>>()?;
```

mithril-stm/src/protocol/aggregate_signature/aggregate_key.rs
jpraynaud, line 1: You can get inspired by the implementations of `from_bytes` and `to_bytes` of the `AggregateSignature`:
- `to_bytes`: https://github.com/input-output-hk/mithril/blob/main/mithril-stm/src/protocol/aggregate_signature/signature.rs#L155
- `from_bytes`: https://github.com/input-output-hk/mithril/blob/main/mithril-stm/src/protocol/aggregate_signature/signature.rs#L174

mithril-stm/src/protocol/aggregate_signature/clerk.rs
```diff
@@ -1,66 +1,60 @@
...
-            closed_reg,
+            concatenation_proof_clerk: ConcatenationClerk::new_clerk_from_closed_key_registration(
+                parameters, closed_reg,
+            ),
         }
     }
-
     /// Aggregate a set of signatures with a given proof type.
-    pub fn aggregate_signatures_with_type(
+    pub fn aggregate_signatures_with_type<D: MembershipDigest>(
```
jpraynaud: Maybe we should keep the whole `Clerk` depend on the `D: MemberShipDigest` trait instead of this `aggregate_signatures_with_type` method only.

mithril-stm/src/protocol/participant/initializer.rs
```diff
@@ -3,86 +3,98 @@ use rand_core::{CryptoRng, RngCore};
...
-/// Once the latter is finished, this instance is consumed into an `Signer`.
+/// Structure responsible of creating a signer
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct Initializer {
-    /// This participant's stake.
+    /// Stake of the participant
     pub stake: Stake,
-    /// Current protocol instantiation parameters.
+    /// Protocol parameters.
+    /// TODO: Rename as `parameters`
```
jpraynaud: Can you take care of these TODOs?

mithril-stm/src/protocol/participant/signer.rs
```diff
@@ -1,111 +1,62 @@
...
-                signer_index: self.signer_index,
-            })
-        } else {
-            None
-        }
+    /// Signing function that returns an Option type.
+    pub fn sign(&self, message: &[u8]) -> Option<SingleSignature> {
+        let result = self.create_single_signature(message);
+        result.ok()
     }
```
jpraynaud, line 51: The meaning of the `Option` is not very clear. It should be used if there is no single signature created and considering this is not an error IMO.

mithril-stm/src/protocol/single_signature/signature.rs
```diff
@@ -25,6 +26,9 @@ pub struct SingleSignature {
 
 impl SingleSignature {
     /// Verify a `SingleSignature` by validating the underlying single signature for proof system.
+    /// It only works for concatenation proof system.
+    /// TODO: support snark proof single signature verification and do not use `anyhow!` for error
+    /// handling.
```
jpraynaud: This TODO can probably be fixed in this PR.

mithril-common/src/crypto_helper/types/alias.rs
jpraynaud: Maybe this is a bit more clear:
```suggestion
pub type ProtocolMembershipDigest = MithrilMembershipDigest;
```

mithril-stm/src/lib.rs
```diff
@@ -141,6 +150,9 @@ pub type Stake = u64;
 /// An aggregate signature (`StmMultiSig`) must have at least `k` unique indices.
 pub type Index = u64;
```
jpraynaud: Maybe we need to clarify this index type:
```suggestion
pub type LotteryIndex = u64;
```

mithril-stm/src/protocol/key_registration/register.rs
```diff
@@ -0,0 +1,257 @@
...
+        Self {
+            registration_entries: Default::default(),
+        }
+    }
+
+    /// Check whether the given `RegistrationEntry` is already registered.
+    /// Insert the new entry if all checks pass.
+    /// # Error
+    /// The function fails when the entry is already registered.
+    pub fn register_by_entry(&mut self, entry: &RegistrationEntry) -> StmResult<()> {
```
jpraynaud, line 29: Does it need to be public?
```suggestion
    fn register_by_entry(&mut self, entry: &RegistrationEntry) -> StmResult<()> {
```

mithril-stm/src/protocol/key_registration/register.rs
```diff
@@ -0,0 +1,257 @@
...
+            *vk_pop,
+            #[cfg(feature = "future_snark")]
+            None,
+            stake,
+        )?;
+        self.register_by_entry(&entry)
+    }
+
+    /// Gets the index of a signer registration entry
+    pub fn get_signer_index_for_registration(
```
jpraynaud, line 50: Could you move this method just after/before `get_registration_entry_for_index`?

mithril-stm/src/protocol/key_registration/registration_entry.rs
```diff
@@ -0,0 +1,123 @@
...
+    }
+}
+
+impl Ord for RegistrationEntry {
+    /// Compares the registration entries by comparing the stake first, then the verification key.
+    /// The order is backward compatible with previous implementations.
+    fn cmp(&self, other: &Self) -> Ordering {
+        self.1.cmp(&other.1).then(self.0.cmp(&other.0))
+    }
+}
```
jpraynaud, line 102: It's worth adding a test for this `Ord` implementation.

mithril-stm/src/protocol/participant/initializer.rs
```diff
@@ -3,86 +3,98 @@ use rand_core::{CryptoRng, RngCore};
...
-    /// Secret key.
-    pub(crate) sk: BlsSigningKey,
-    /// Verification (public) key + proof of possession.
-    pub(crate) pk: VerificationKeyProofOfPossession,
+    /// Protocol parameters.
+    pub parameters: Parameters,
+    /// Signing key for concatenation proof system.
+    pub bls_signing_key: BlsSigningKey,
+    /// Verification key for concatenation proof system.
+    pub bls_verification_key_proof_of_possession: VerificationKeyProofOfPossessionForConcatenation,
```
jpraynaud, line 27: You will need to update the serde names of the fields for backward compatibility (a golden test should fail with the code as is).

mithril-stm/src/protocol/key_registration/register.rs
curiecrypt: It is used in tests, benches, and the examples. Also, it could be used publicly when we have snark registration. WDYT?

mithril-stm/README.md
curiecrypt: I don't understand exactly what you mean here.

jpraynaud: I mean that the `future_snark` should not be part yet of the public documentation.

mithril-stm/src/protocol/key_registration/register.rs
jpraynaud: Let's keep it public :+1:

mithril-stm/src/protocol/aggregate_signature/clerk.rs
```diff
@@ -69,227 +67,39 @@ impl<D: MembershipDigest> Clerk<D> {
...
+    /// Get the concatenation clerk.
+    pub fn to_concatenation_clerk(&self) -> &ConcatenationClerk {
+        &self.concatenation_proof_clerk
+    }
+
+    /// Compute the aggregate verification key.
+    /// It computes only the concatenation aggregate verification key for now.
     pub fn compute_aggregate_verification_key(&self) -> AggregateVerificationKey<D> {
-        AggregateVerificationKey::from(&self.closed_reg)
+        AggregateVerificationKey::Concatenation(
```
jpraynaud, line 79: Could you leave a comment explaining that this implementation needs to be revised when implementing SNARK pre-aggregation primitives?

mithril-stm/src/protocol/key_registration/registration_entry.rs
```diff
@@ -0,0 +1,147 @@
...
+    /// Expects 96 bytes for the BLS verification key and 8 bytes for the stake (u64 big-endian).
+    /// The order is backward compatible with previous implementations.
+    pub fn from_bytes(bytes: &[u8]) -> StmResult<Self> {
+        let bls_verification_key = VerificationKeyForConcatenation::from_bytes(bytes)?;
+        let mut u64_bytes = [0u8; 8];
+        u64_bytes.copy_from_slice(&bytes[96..]);
+        let stake = Stake::from_be_bytes(u64_bytes);
+        Ok(RegistrationEntry(bls_verification_key, stake))
+    }
+}
```
jpraynaud, line 63: Do we really need these bytes conversions for this type which is only internal to the crate?

mithril-stm/src/protocol/key_registration/registration_entry.rs
```diff
@@ -0,0 +1,147 @@
...
+        Self(
+            initializer.bls_verification_key_proof_of_possession.vk,
+            initializer.stake,
+        )
+    }
+}
+
+impl Hash for RegistrationEntry {
+    /// Hashes the registration entry by hashing the stake first, then the verification key.
+    /// The order is backward compatible with previous implementations.
```
jpraynaud:
```suggestion

```

mithril-stm/src/protocol/key_registration/registration_entry.rs
```diff
@@ -0,0 +1,147 @@
...
+
+impl PartialOrd for RegistrationEntry {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(std::cmp::Ord::cmp(self, other))
+    }
+}
+
+impl Ord for RegistrationEntry {
+    /// Compares the registration entries by comparing the stake first, then the verification key.
+    /// The order is backward compatible with previous implementations.
```
jpraynaud:
```suggestion

```

mithril-stm/src/protocol/participant/initializer.rs
```diff
@@ -3,86 +3,97 @@ use rand_core::{CryptoRng, RngCore};
...
-    /// * the `avk` as built from the current registered parties (according to the registration service)
-    /// * the current total stake (according to the registration service)
-    /// # Error
-    /// This function fails if the initializer is not registered.
-    pub fn create_signer<D: MembershipDigest>(
+    /// Tries to generate a new signer. First, checks whether the registration is closed by the
+    /// total stake of the registration. Fails if the registration is not closed. Then, it checks
+    /// whether the initializer is registered and gets the signer index. If not, returns an error.
+    /// Sets up the merkle tree commitment and creates the signer of each system (for now, only the
+    /// concatenation proof system signer), then it creates and returns the signer.
```
jpraynaud: This comment 
```suggestion
/// Attempts to generate a new signer from the current registration state.
///
/// # Process
/// 1. Verifies that registration is closed (determined by total stake threshold)
/// 2. Confirms the initializer is registered and retrieves its signer index
/// 3. Constructs the Merkle tree commitment
/// 4. Creates the underlying proof system signer (currently only the concatenation proof system)
///
/// # Errors
/// Returns an error if:
/// - Registration is not yet closed
/// - The initializer is not registered
```

mithril-stm/src/protocol/participant/initializer.rs
```diff
@@ -3,86 +3,97 @@ use rand_core::{CryptoRng, RngCore};
...
+            signer_index,
+            concatenation_proof_signer,
+            closed_key_registration.clone(),
+            self.parameters,
+            registration_entry.get_stake(),
         ))
     }
 
+    /// Extract the verification key with proof of possession.
+    pub fn get_verification_key_proof_of_possession(
```
jpraynaud:
```suggestion
    pub fn get_verification_key_proof_of_possession_for_concatenation(
```

mithril-stm/src/protocol/single_signature/signature.rs
```diff
@@ -20,20 +22,26 @@ pub struct SingleSignature {
     #[serde(flatten)]
     pub(crate) concatenation_signature: SingleSignatureForConcatenation,
     /// Merkle tree index of the signer.
-    pub signer_index: Index,
+    pub signer_index: LotteryIndex,
 }
 
 impl SingleSignature {
     /// Verify a `SingleSignature` by validating the underlying single signature for proof system.
+    /// It only works for concatenation proof system.
```
jpraynaud, line 31:
```suggestion
    /// Verify a `SingleSignature` by validating the underlying single signature for proof system.
    ///
    /// It only works for concatenation proof system.
```

mithril-stm/src/protocol/aggregate_signature/aggregate_key.rs
jpraynaud: I fixed that in https://github.com/input-output-hk/mithril/pull/2880/commits/2fa32278870c5ce628bd193515f2a7820bc21466 :+1:

mithril-stm/src/protocol/aggregate_signature/clerk.rs
```diff
@@ -69,227 +67,39 @@ impl<D: MembershipDigest> Clerk<D> {
         }
     }
 
-    /// Compute the `AggregateVerificationKey` related to the used registration.
+    /// Get the concatenation clerk.
+    pub fn to_concatenation_clerk(&self) -> &ConcatenationClerk {
```
damrobi: Change `to_concatenation_clerk` to `get_concatenation_clerk` if we're not doing a conversion?
```suggestion
    pub fn get_concatenation_clerk(&self) -> &ConcatenationClerk {
```

mithril-stm/src/protocol/single_signature/signature.rs
```diff
@@ -177,22 +185,43 @@ mod tests {
 
         fn golden_value() -> SingleSignature {
```
hjeljeli32, line 187: @curiecrypt 
Golden tests for Happy paths are great. I suggest we can add a simple negative test: wrong index or wrong VK should fail verification.

mithril-stm/src/proof_system/concatenation/clerk.rs
```diff
@@ -0,0 +1,274 @@
...
+                    .collect::<Vec<LotteryIndex>>();
+                deduped_sig.sig.set_concatenation_signature_indices(&indices);
+            }
+
+            let size: Result<u64, _> =
+                deduped_sig.sig.get_concatenation_signature_indices().len().try_into();
+            if let Ok(size) = size {
+                if dedup_sigs.contains(&deduped_sig) {
+                    panic!(
+                        "Invariant violation: duplicate signature encountered in deduplication set, which should not be possible."
```
hjeljeli32, line 141: @curiecrypt 
Since the function already returns StmResult, I was wondering whether it would be better to replace the panic! with a proper error.

mithril-stm/src/proof_system/concatenation/clerk.rs
```diff
@@ -0,0 +1,274 @@
...
+                    for passed_sigs in valid_sigs {
+                        let signer = &signers[passed_sigs.sig.signer_index as usize];
+                        let verify_result = passed_sigs.sig.concatenation_signature.verify(
+                            &params,
+                            &signer.get_bls_verification_key(),
+                            &passed_sigs.reg_party.get_stake(),
+                            &avk,
+                            &msg,
+                        );
+                        assert!(verify_result.is_ok(), "All returned signatures should verify: {:?}", verify_result);
```
hjeljeli32, line 260: @curiecrypt 
The test does a good job checking that returned signatures verify, but I was wondering should we also check the returned set actually covers at least k distinct lottery indices, and that indices are unique after deduplication?

mithril-stm/src/proof_system/concatenation/aggregate_key.rs
```diff
@@ -0,0 +1,199 @@
...
+        bytes
+    }
+
+    pub fn from_bytes(bytes: &[u8]) -> StmResult<Self> {
+        let mut u64_bytes = [0u8; 8];
+        let size = bytes.len();
+
+        u64_bytes.copy_from_slice(&bytes[size - 8..]);
+        let stake = u64::from_be_bytes(u64_bytes);
+        let mt_commitment = MerkleTreeBatchCommitment::from_bytes(&bytes[0..size - 8])?;
```
hjeljeli32: @curiecrypt 
Maybe if bytes.len() < 8, this will panic (slice underflow). If that's the case, since this is a StmResult, better to return an error.

mithril-stm/src/proof_system/concatenation/clerk.rs
```diff
@@ -0,0 +1,274 @@
...
+    /// Create a `ConcatenationClerk` from a signer.
+    pub fn new_clerk_from_signer<D: MembershipDigest>(signer: &Signer<D>) -> Self {
+        Self {
+            parameters: signer.parameters,
+            closed_key_registration: signer.closed_key_registration.clone(),
+        }
+    }
+
+    /// Compute the `ConcatenationProofKey` related to the used registration.
+    pub fn compute_concatenation_proof_key<D: MembershipDigest>(
```
damrobi: Should we replace the name `compute_concatenation_proof_key` with something like `compute_aggregate_verification_key_for_concatenation` for better clarity? I don't think we have a `ConcatenationProofKey` anywhere else for now.

hjeljeli32 (commented):
LGTM. I just left a couple of comments.

mithril-stm/src/proof_system/concatenation/proof.rs
```diff
@@ -77,9 +92,9 @@ impl<D: MembershipDigest> ConcatenationProof<D> {
     fn preliminary_verify(
```
damrobi, line 92: The description of the function mentions a `BasicVerifier` which I don't think we have anymore.

mithril-stm/src/proof_system/concatenation/proof.rs
```diff
@@ -58,8 +71,10 @@ impl<D: MembershipDigest> ConcatenationProof<D> {
             .collect::<Vec<usize>>();
 
         let batch_proof = clerk
-            .closed_reg
-            .merkle_tree
+            .closed_key_registration
+            .key_registration
+            .clone()
```
damrobi: I think that since `RegistrationEntry` is Copy, we can change the `into_merkle_tree` function to take a reference of `self` and remove the `.clone()` call but maybe there's a reason not to do that.

mithril-stm/src/proof_system/concatenation/single_signature.rs
```diff
@@ -17,13 +18,13 @@ pub(crate) struct SingleSignatureForConcatenation {
     /// The underlying BLS signature
     sigma: BlsSignature,
     /// The index(es) for which the signature is valid
-    indexes: Vec<Index>,
+    indexes: Vec<LotteryIndex>,
```
damrobi, line 21: Not directly related to the PR but the struct field is called `indexes` while most functions use `indices`. Do we want to harmonize that? Maybe it's better to do at another time because it might change a lot of things.

mithril-stm/src/protocol/aggregate_signature/signature.rs
```diff
@@ -107,9 +107,11 @@ impl<D: MembershipDigest> AggregateSignature<D> {
         parameters: &Parameters,
     ) -> StmResult<()> {
         match self {
-            AggregateSignature::Concatenation(concatenation_proof) => {
-                concatenation_proof.verify(msg, avk, parameters)
-            }
+            AggregateSignature::Concatenation(concatenation_proof) => concatenation_proof.verify(
+                msg,
+                avk.to_concatenation_proof_key().unwrap(),
```
damrobi, line 112: Is there a possibility of having the self and avk of different types, self for concat and avk for snark? In which case the unwrap() would give None for now or a `AggregateVerificationKeyForSnark` in the future.

mithril-stm/src/protocol/participant/initializer.rs
```diff
@@ -3,86 +3,97 @@ use rand_core::{CryptoRng, RngCore};
...
-    ///
-    /// Returns an `Signer` specialized to
-    /// * this `Signer`'s ID and current stake
-    /// * this `Signer`'s parameter valuation
-    /// * the `avk` as built from the current registered parties (according to the registration service)
-    /// * the current total stake (according to the registration service)
-    /// # Error
-    /// This function fails if the initializer is not registered.
-    pub fn create_signer<D: MembershipDigest>(
+    /// Tries to generate a new signer. First, checks whether the registration is closed by the
```
damrobi: I'm not sure where we check that the registration is closed in this function but since it's taking a `ClosedKeyRegistration` I guess we had to check it before. However, we check the proof of possession of the Initializer and fail if it's not valid.

mithril-stm/src/protocol/participant/initializer.rs
```diff
@@ -3,86 +3,97 @@ use rand_core::{CryptoRng, RngCore};
...
+            .key_registration
+            .get_signer_index_for_registration(&registration_entry)
+        {
+            Some(index) => index,
+            None => return Err(anyhow!(RegisterError::UnregisteredInitializer)),
+        };
+
+        let key_registration_commitment = closed_key_registration
+            .key_registration
+            .clone()
```
damrobi: As before, I think we can remove the `.clone()` if we make the `into_merkle_tree` function take a reference to `self`.

mithril-stm/src/protocol/key_registration/registration_entry.rs
curiecrypt: It is used in bytes conversion of `SingleSignatureWithRegisteredParty`. 
I am making the functions `pub(crate)`.

mithril-stm/src/protocol/single_signature/signature.rs
curiecrypt: We will handle this in another ticket.

mithril-stm/src/proof_system/concatenation/clerk.rs
curiecrypt: This is a panic that we want to have. The flow should be stopped if we have a duplicate in the list. This is why we do not return an error here.

mithril-stm/src/proof_system/concatenation/single_signature.rs
curiecrypt: The field names are set year ago and we avoid changing them as much as possible for not causing any breaking change. 
However, let's keep such things in mind. We'll revise them whenever possible.

mithril-stm/src/protocol/aggregate_signature/signature.rs
curiecrypt: Let's do it while implementing pre-aggregation primitives.

mithril-stm/src/protocol/participant/initializer.rs
curiecrypt: Checking the key registration does not make any sense with this version. 
We used to have basic signer functionality which provides creating a signature without registration. 
Since we do not support this feature anymore, we do not need to check it since we have closed registration as an input to the function.

mithril-stm/src/protocol/key_registration/register.rs
```diff
@@ -0,0 +1,252 @@
...
+    /// Check whether the given `RegistrationEntry` is already registered.
+    /// Insert the new entry if all checks pass.
+    /// # Error
+    /// The function fails when the entry is already registered.
+    pub fn register_by_entry(&mut self, entry: &RegistrationEntry) -> StmResult<()> {
+        if !self.registration_entries.contains(entry) {
+            self.registration_entries.insert(*entry);
+            return Ok(());
+        }
+        Err(anyhow!(RegisterError::KeyRegistered(Box::new(
```
hjeljeli32: Small question: would it be better to return RegisterError::KeyRegistered(...) directly here instead of wrapping it with anyhow!?
This might keep error handling consistent with other functions like in registration_entry.rs and makes it easier for callers to match on the error without downcasting.

mithril-stm/src/protocol/key_registration/register.rs
```diff
@@ -0,0 +1,252 @@
...
+    /// Get the registration entry for a given signer index.
+    pub fn get_registration_entry_for_index(
+        &self,
+        signer_index: &SignerIndex,
+    ) -> StmResult<RegistrationEntry> {
+        self.registration_entries
+            .iter()
+            .nth(*signer_index as usize)
+            .cloned()
+            .ok_or_else(|| anyhow!(RegisterError::UnregisteredIndex))
```
hjeljeli32: same question here about using anyhow! vs returning the typed error for consistency across the registration API.

mithril-stm/src/proof_system/concatenation/aggregate_key.rs
```diff
@@ -0,0 +1,201 @@
...
+impl<D: MembershipDigest> From<&ClosedKeyRegistration>
+    for AggregateVerificationKeyForConcatenation<D>
+{
+    fn from(reg: &ClosedKeyRegistration) -> Self {
+        Self {
+            mt_commitment: reg
+                .key_registration
+                .clone()
+                .into_merkle_tree::<D::ConcatenationHash, RegistrationEntryForConcatenation>()
+                .unwrap()
```
hjeljeli32, line 76: Minor question: this function uses an unwrap() on into_merkle_tree().
Even if this “should never fail”, would you prefer expect(...) with context, or moving the fallible construction into a TryFrom/constructor path to avoid panics in library code?

mithril-stm/src/proof_system/concatenation/clerk.rs
```diff
@@ -0,0 +1,287 @@
...
+                    .collect::<Vec<LotteryIndex>>();
+                deduped_sig.sig.set_concatenation_signature_indices(&indices);
+            }
+
+            let size: Result<u64, _> =
+                deduped_sig.sig.get_concatenation_signature_indices().len().try_into();
+            if let Ok(size) = size {
+                if dedup_sigs.contains(&deduped_sig) {
+                    panic!(
+                        "Invariant violation: duplicate signature encountered in deduplication set, which should not be possible."
```
hjeljeli32, line 141: We panic if a duplicate signature is encountered in dedup_sigs.
Even if this is meant to be unreachable, would it be safer to return an AggregationError instead of panicking (since this function can process externally-provided signatures)?

hjeljeli32 (approved):
LGTM.

PR #2899, opened 2026-01-09
fix(cli): handle file write errors in verification report

leepl37:
## Content

This PR fixes a potential panic in the mithril-client-cli during the snapshot verification process.

The Issue: Previously, if the Immutable File verification failed, the CLI attempted to write a JSON report (immutables_verification_error.json) to the Current Working Directory (CWD) using .expect(). If the CWD was read-only, the disk was full, or permissions were missing, this caused the entire CLI process to panic/crash. 

- Refactored write_json_file_error in verify.rs to return Result<PathBuf> instead of panicking.
- Updated the caller print_immutables_verification_error to handle the error.
- The CLI now prints a warning to stderr if the report cannot be saved, but continues execution or exits with the correct status code instead of crashing.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation
  - [x] No new TODOs introduced

## Comments

Manual Verification Since this issue relates to environmental permissions (filesystem), I verified the fix manually in a local environment:

- Before Fix: The client crashed with: thread 'main' panicked at 'Could not write immutables verification error to file'.

- After Fix: The client handled the error gracefully: WARN: Could not save error report file: Permission denied (os error 13).## Issue(s)

N/A

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3943,7 +3943,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-client-cli"
-version = "0.12.37"
+version = "0.12.38"
 dependencies = [
  "anyhow",
  "async-trait",


mithril-client-cli/src/commands/cardano_db/verify.rs
```diff
@@ -273,24 +292,28 @@ impl CardanoDbVerifyCommand {
     }
 }
 
-fn write_json_file_error(date: DateTime<Utc>, lists: &ImmutableVerificationResult) -> PathBuf {
+fn write_json_file_error(
+    date: DateTime<Utc>,
+    lists: &ImmutableVerificationResult,
+) -> anyhow::Result<PathBuf> {
```
Alenar: We have a dedicated alias (which is incorrectly not used by the following method, so I understand why you missed that).
```suggestion
) -> MithrilResult<PathBuf> {
```

mithril-client-cli/src/commands/cardano_db/verify.rs
```diff
@@ -232,14 +232,29 @@ impl CardanoDbVerifyCommand {
...
+
+        // Try to write the file once. Never crash if we can't write it.
+        // Also keep JSON output clean (stdout only), log warnings to stderr.
+        let report_path: Option<PathBuf> = match write_json_file_error(utc_now, lists) {
+            Ok(path) => Some(path),
+            Err(err) => {
+                eprintln!("WARN: Could not save error report file: {:#}", err);
+                None
+            }
+        };
```
Alenar: We have the `utils::print_simple_warning` method that print to stderr and format the message to json if necessary (so it is displayed in both display mode).
```suggestion
        // Try to write the file once. Never crash if we can't write it.
        let report_path: Option<PathBuf> = write_json_file_error(utc_now, lists)
            .inspect_err(|err| {
                print_simple_warning(
                    &format!("Could not save error report file: {err:?}"),
                    json_output,
                );
            })
            .ok();
```

mithril-client-cli/src/commands/cardano_db/verify.rs
```diff
@@ -232,14 +232,29 @@ impl CardanoDbVerifyCommand {
...
+
         let error_message = "Verifying immutables files has failed";
         if json_output {
+            // Use a JSON-friendly string, and allow null when no report was written.
+            let report_path_str: Option<String> = report_path
+                .as_ref()
+                .map(|p| p.as_path().to_string_lossy().to_string());
```
Alenar: Since this will be send directly to serde json I think we can just use their types directly (which makes explicit that null is used for none):
```suggestion
            let report_path_str: serde_json::Value = match report_path {
                Some(path) => serde_json::Value::String(path.to_string_lossy().to_string()),
                None => serde_json::Value::Null,
            };
```

mithril-client-cli/src/commands/cardano_db/verify.rs
```diff
@@ -14,7 +14,7 @@ use mithril_client::{
     },
     common::{ImmutableFileNumber, SignedEntityTypeDiscriminants},
 };
-
+use crate::utils::print_simple_warning;
```
Alenar: Can you merge it in the import below ?

Alenar (approved):
LGTM :fire:

mithril-client-cli/src/commands/cardano_db/verify.rs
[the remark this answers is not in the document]
Alenar: Not useful as it's difficult to unit tests print to stdout/stderr in rust.

mithril-client-cli/src/commands/cardano_db/verify.rs
```diff
@@ -232,28 +232,49 @@ impl CardanoDbVerifyCommand {
 
     fn print_immutables_verification_error(lists: &ImmutableVerificationResult, json_output: bool) {
         let utc_now = Utc::now();
-        let json_file_path = write_json_file_error(utc_now, lists);
+
+        // Try to write the file once. Never crash if we can't write it.
+        // Also keep JSON output clean (stdout only), log warnings to stderr.
```
jpraynaud: This comment is not necessary IMO
```suggestion

```

mithril-client-cli/src/commands/cardano_db/verify.rs
```diff
@@ -232,28 +232,49 @@ impl CardanoDbVerifyCommand {
...
+            print_simple_warning(
+                &format!("Could not save error report file: {err:?}"),
+                json_output,
+            );
+        })
+        .ok();
+
         let error_message = "Verifying immutables files has failed";
         if json_output {
+            // Prepare the path value explicitly for serde
```
jpraynaud:
```suggestion

```

PR #2902, opened 2026-01-09
feat: use released Haskell DMQ node in e2e tests

jpraynaud:
## Content

This PR includes **the usage of the released Haskell DMQ node in the e2e tests and in the CI**.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Issue(s)
Relates to #2833


.github/workflows/ci.yml
```diff
@@ -392,16 +392,32 @@ jobs:
         extra_args: ["--aggregate-signature-type=Concatenation"]
 
         include:
+          # Include a test for partial decentralization with leader/follower signer registration and P2P signature registration with Haskell DMQ node (without fallback signature delayer)
```
Alenar, line 395: Given how much inclusion there's now, I think we can reduce the `run_id` list to just `['#1']`, this would lower the number of e2e runs from 19 to 11 which is enough.

When we will switch to the `Lagrange` era we may up it to `['#1","#2"]` to compensate that the removal of the `Pythagoras` era will lower the number of run by another 2.

jpraynaud: I agree :+1: I reduced the number of run ids to 1 in the matrix.

PR #2922, opened 2026-01-14
feat: add new `CardanoBlocksTransactions` to `SignedEntityTypes`

### LINKED ISSUE #2905: Create new signed entity type for Cardano Blocks and Transactions
## Why
We want to bootstrap the signed entity type for the incremental Cardano Blocks and Transactions certification.

## What
Create the associated signed entity type and the associated database migrations.

## How

> [!IMPORTANT]
> Do not use `unimplemented!` or `todo!` macro with the default implementation, return error instead.

- [x] Create the new `CardanoBlocksTransactions(Epoch,BlockNumber)` signed entity type
- [x] Adjust the timeout on the `CardanoBlocksTransactions` signed entity type to `600s`
- [x] Create the associated database migrations in the:
  - [x] signer
  - [x] aggregator

Alenar:
## Content

This PR  add new `SignedEntityTypes`: `CardanoBlocksTransactions`.

Additional changes:
- `SignedEntityTypeDiscriminants::all` exclude the new discriminant as it can't be derived without error right now which makes some tests fails
- Common `MithrilSignableBuilderService` will raise an error if the new type is ask
- Aggregator `MithrilSignedEntityService` will raise an error if the new type is ask
- Add migration in the signer and aggregator to add the new type in the `signed_entity_type` table

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation
  - [x] No new TODOs introduced

## Issue(s)

Closes #2905

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3824,7 +3824,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-aggregator"
-version = "0.8.9"
+version = "0.8.10"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -4120,7 +4120,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-common"
-version = "0.6.41"
+version = "0.6.42"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -4260,7 +4260,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-persistence"
-version = "0.2.60"
+version = "0.2.61"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -4356,7 +4356,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-signer"
-version = "0.3.3"
+version = "0.3.4"
 dependencies = [
  "anyhow",
  "async-trait",
--- a/internal/mithril-persistence/Cargo.toml
+++ b/internal/mithril-persistence/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mithril-persistence"
-version = "0.2.60"
+version = "0.2.61"
 description = "Common types, interfaces, and utilities to persist data for Mithril nodes."
 authors = { workspace = true }
 edition = { workspace = true }
--- a/mithril-aggregator/Cargo.toml
+++ b/mithril-aggregator/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mithril-aggregator"
-version = "0.8.9"
+version = "0.8.10"
 description = "A Mithril Aggregator server"
 authors = { workspace = true }
 edition = { workspace = true }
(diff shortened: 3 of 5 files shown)


mithril-common/src/entities/signed_entity_type.rs
```diff
@@ -146,7 +158,8 @@ impl SignedEntityType {
                 hasher.update(&db_beacon.epoch.to_be_bytes());
                 hasher.update(&db_beacon.immutable_file_number.to_be_bytes());
             }
-            SignedEntityType::CardanoTransactions(epoch, block_number) => {
+            SignedEntityType::CardanoTransactions(epoch, block_number)
```
Alenar, line 167: Looking at this `feed_hash` method again I remark that the discriminant is not feed to the hash, only the beacon value.
This means that for signed entity types that share a same beacon the hash will be the same. Should we change that ? If yes should we change only the new type to avoid breaking previous hashes ?

mithril-aggregator/src/database/migration.rs
```diff
@@ -276,5 +276,14 @@ pragma foreign_key_check;
 pragma foreign_keys=true;
         "#,
         ),
+        // Migration 38
+        // Add the `signed_entity_type` record for 'CardanoBlocksTransactions'
+        SqlMigration::new(
+            38,
+            r#"
+insert into signed_entity_type (signed_entity_type_id, name)
+    values  (5, 'Cardano Blocks Transactions Database');
```
jpraynaud:
```suggestion
    values  (5, 'Cardano Blocks and Transactions');
```

mithril-signer/src/database/migration.rs
```diff
@@ -128,5 +128,14 @@ insert into signed_entity_type (signed_entity_type_id, name)
     values  (4, 'Cardano Database');
 "#,
         ),
+        // Migration 8
+        // Add the `signed_entity_type` record for 'CardanoBlocksTransactions'
+        SqlMigration::new(
+            8,
+            r#"
+insert into signed_entity_type (signed_entity_type_id, name)
+    values  (5, 'Cardano Blocks Transactions Database');
```
jpraynaud:
```suggestion
    values  (5, 'Cardano Blocks and Transactions');
```

mithril-common/src/entities/signed_entity_type.rs
jpraynaud: Indeed, something is missing. You can fix only the new signed entity type to avoid breaking change and leave a TODO :+1:

jpraynaud (approved):
LGTM :+1:

mithril-common/src/entities/signed_entity_type.rs
Alenar: Done

Issue #2907, opened 2026-01-12
Import Blocks and Transactions in new store for Cardano Blocks and Transactions


jpraynaud:
## Why
We need to import and store the Cardano chain Blocks and Transactions in a local store that will be used by the aggregator and the signer to sign the `CardanoBlocksTransactions` signed entity type. The blocks and transactions will be signed by block ranges (same as for `CardanoTransactions`) and the signer store will be pruned (same as for `CardanoTransactions`).

## What
Import and store the Blocks and Transactions for the `CardanoBlocksTransactions` signed entity type.

## How

> [!IMPORTANT]
> This issue will block the release of a new distribution (to avoid discrepancies in the signature of `CardanoTransactions`)

- [x] Update the `TransactionsImporter` to support import of blocks
  - [x] In aggregator
  - [x] In signer
- [x] Update the `cardano-transactions` store:
  - [x] Reset the existing `cardano-transactions` store
  - [x] Create a `cardano_block` table
  - [x] Update the structure of the `cardano_tx` table to reference the `cardano_block` table
  - [x] Rename the `block_range_root` table in `block_range_root_legacy` table
  - [x] Create a new `block_range_root` tabel for `CardanoBlocksTransactions`
  - [x] ~~Rename the database file~~
- [x] Update the `CardanoTransactions` retrieval of information from the new store:
  - [x] Signable builder
    - [x] In aggregator 
    - [x] In signer
  - [x] Proof route of the aggregator

PR #2930, opened 2026-01-19
refactor: Shared transaction importer for signer and aggregator

Alenar:
## Content

This PR replace the `CardanoTransactionsImporter` that were identical between the `mithril-signer` and `mithril-aggregator` with a shared implementation, `CardanoChainDataImporter`, hosted in `internal/mithril-cardano-node-chain`.

With the upcoming development of the `CardanoBlocksTransactions` signed entity, this will allow to reduce the amount of work and error by removing the need of copying works from one transaction importer to the other.

### Design

`internal/mithril-cardano-node-chain` host the technical part, it defines a `CardanoNodeChain` trait which is independent from the business traits (currently only `TransactionsImporter`) defined in `mithril_common::signable_builder`.

The glue between the technical and business parts is done with a `AggregatorChainDataImporter`/`SignerChainDataImporter` struct which  and implement the business trait while using internally a `ChainDataImporter`.

This is a tradeoff, we replaced the full copy of the transactions importer service by a small boilerplate struct.

For the database part, what was the `TransactionsStore` trait have been moved to `internal/mithril-cardano-node-chain` and renamed to `ChainDataStore`.
As there's no dependency between `internal/mithril-cardano-node-chain` and `internal/mithril-persistence`, and we want to keep it this way (so both are focused on their specific domain), the `ChainDataStore` trait still have a specific implementation in the signer and the aggregator (:pencil2: a boilerplate struct was needed to implement the trait here, has we can't implement a trait on a type if both the trait and type are hosted in foreign crates).

### Testing

Previously tests of the `CardanoTransactionsImporter` were based on usage of the real sqlite repository, which was possible because the struct was hosted in the aggregator and signer which had access to everything needed.

By moving the struct to `internal/mithril-cardano-node-chain` we lost the access to the sqlite repository, instead for testing purpose a `InMemoryChainDataStore` was added.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation
  - [x] No new TODOs introduced

## Issue(s)

Relates to #2907

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3865,7 +3865,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-aggregator"
-version = "0.8.12"
+version = "0.8.13"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -4010,7 +4010,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-cardano-node-chain"
-version = "0.1.14"
+version = "0.1.15"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -4396,7 +4396,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-signer"
-version = "0.3.5"
+version = "0.3.6"
 dependencies = [
  "anyhow",
  "async-trait",
--- a/internal/cardano-node/mithril-cardano-node-chain/Cargo.toml
+++ b/internal/cardano-node/mithril-cardano-node-chain/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mithril-cardano-node-chain"
-version = "0.1.14"
+version = "0.1.15"
 authors.workspace = true
 documentation.workspace = true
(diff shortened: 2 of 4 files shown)


mithril-aggregator/src/dependency_injection/builder/mod.rs
```diff
@@ -161,6 +162,9 @@ pub struct DependenciesBuilder {
     /// Cardano transactions repository.
     pub transaction_repository: Option<Arc<CardanoTransactionRepository>>,
```
jpraynaud: Is it still needed?

Alenar: No, good catch, it's still bit used in the DI but it should be supersede by the new store, I will remove it !

Alenar: Done

PR #2936, opened 2026-01-20
feat(stm): Implementation of the digest trait for the Poseidon hash from Midnight

damrobi:
## Content

This PR includes a wrapper that implements the Digest trait to be used with the Poseidon hash function from Midnight.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation
  - [x] No new TODOs introduced

## Comments

Quick summary of the important traits implemented:
 - `Update`: updates the internal buffer with new data, takes a slice of bytes as input.
 - `FixedOutput`: contains the finalize function that outputs the result of the hash, since Poseidon takes base field elements as input there is a conversion from the bytes of the buffer to JubjubBase elements before applying the hash.

I have a question regarding the design of the digest functions:
 - I made it so the update function always pad the data received to a multiple of 32 bytes to make sure we can generate an exact number of JubjubBase elements when computing the hash. This padding can be moved to the finalize function. I did it this way because I wasn't sure how we wanted to handle the case where only one byte is received by the hasher at a time.
  - Example: We feed one byte to the hasher three times: 1u8, 2u8, 3u8. Do we want this to correspond to applying poseidon on three JubjubBase elements Poseidon(F(1), F(2), F(3)) or to one element create from those bytes Poseidon(F(b"0000 0001 0000 0010 0000 0011"))?
  - The current implementation correspond to the first case but it can be changed
 
Last remark, this problem concerns the CPU implementation of Poseidon. In the circuit, we only deal with JubjubBase elements so it is clear what is done. We need to make sure both implementations give the same results.

I think that given the current use of digest in the implementation, it does not matter too much which way we use as the example I gave does not come up but I wanted to check with more people as it might change in the future.

## Issue(s)

Relates to #2942

### DIFF
--- a/mithril-stm/src/lib.rs
+++ b/mithril-stm/src/lib.rs
@@ -137,7 +137,7 @@ use digest::{Digest, FixedOutput};
 use std::fmt::Debug;
 
 #[cfg(feature = "future_snark")]
-use crate::hash::poseidon::MidnightPoseidonDigest;
+use hash::poseidon::MidnightPoseidonDigest;
 
 /// The quantity of stake held by a party, represented as a `u64`.
 pub type Stake = u64;
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4461,7 +4461,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-stm"
-version = "0.9.6"
+version = "0.9.7"
 dependencies = [
  "anyhow",
  "blake2 0.10.6",
--- a/mithril-stm/CHANGELOG.md
+++ b/mithril-stm/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## 0.9.7 (01-23-2026)
+
+### Added
+
+- Added a Digest trait implementation for the Poseidon hash in `hash`.
+
 ## 0.9.6 (01-23-2026)
 
 ### Added
--- a/mithril-stm/Cargo.toml
+++ b/mithril-stm/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mithril-stm"
-version = "0.9.6"
+version = "0.9.7"
 edition = { workspace = true }
 authors = { workspace = true }
 homepage = { workspace = true }


mithril-stm/src/signature_scheme/unique_schnorr_signature/jubjub/poseidon_digest.rs
```diff
@@ -20,8 +23,71 @@ pub(crate) fn compute_poseidon_digest(input: &[BaseFieldElement]) -> BaseFieldEl
...
+impl Update for MidnightPoseidonDigest {
+    fn update(&mut self, data: &[u8]) {
+        // Computes the next multiple of 32 as target length
+        let target_len = (data.len() + 31) & !31;
+        let mut padded_data = Vec::with_capacity(target_len);
+        padded_data.extend_from_slice(data);
+        // Pad the data with zeros on the right to match the target length
+        padded_data.resize(target_len, 0);
+        self.buffer.extend_from_slice(&padded_data);
+    }
```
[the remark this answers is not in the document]
damrobi: A comment was added to reflect that

rrtoledo:
We have a hash function that in circuit takes as input &[Fq] but in CPU takes &[u8] (to satisfy the trait's interfaces). Hence we need to convert somewhere, when updating or finalizing the hasher, the bytes to field elements. 

As we work with prime fields, we cannot directly cast the bytes to field elements. We need to either take fewer bytes than our field's prime and cast, take 32 bytes and check we are under the modulus before casting, or take more bytes and somehow hash them to the field.

p.s. I am trying to be careful saying "somehow hash to field" here and from now on because we may not choose to use  the RFC's function `hash_to_field` for efficiency. We could simply hash with sha256/blake2 and take the modulo of it with a small loss in soundness instead.

 In @kitounliu 's circuit, we always work with field element inputs:
- the Schnorr signature's message is a field element in circuit - so it has already been converted somewhen
- the Merkle tree's leaves' data 
-- `vks` which are group element of Jubjub hence scalar of Bls12-381 
-- the target which should be small (64 bit iirc) and can be directly cast as a field

In this code, we could assume that the hasher's &[u8] input already are Fq's byte representation. In that case, we should add some comments on the Poseidon code stating it (perhaps some assert?) and in registration/signing adding the extra conversion (the target must be cast to Fq, and message hashed to Fq).
---> I think that's fine as long as it is well documented but I would like your opinion.

If we don't want to assume that the hasher's &[u8] input already are Fq's byte representations., for instance for the hash function to be generic, the design become a bit delicate.
First, do we want the property that `hasher.update(a).update(b).finalize() == hasher.update(a || b).finalize()`?

- if yes, we are forced to concatenate new bytes to the state and later on cast the state to scalar elements.
1.  Taking less than 32 bytes when converting to Fq would not be compatible with @kitounliu 's code in circuit.
2. Taking 32 bytes exactly can be compatible if we do a modulo check and somehow hash to field only when needed
3. Taking more than 32 bytes at a time would not be compatible with @kitounliu 's code in circuit.
4. For the remainder (less than 32 bytes), we can (pad and) cast or hash
- if not necessarily,
1. we can pad each update's input. In that case, the padding must be consistent with @kitounliu 's code in circuit.
2. convert each update's input to field element. We have here the same question as before of how many bytes we take at at time.
3. other solutions?

Second, depending on the choice choice above, we may want to restrict the hasher so that the update function takes a "fixed" number of bytes (e.g. at most 32 bytes or a multiple of 32 bytes). We can ensure that with `asserts`.

Maybe I am overthinking, or I missed something out, but we should all agree on what design choice to take.

mithril-stm/src/signature_scheme/unique_schnorr_signature/jubjub/poseidon_digest.rs
```diff
@@ -20,8 +23,71 @@ pub(crate) fn compute_poseidon_digest(input: &[BaseFieldElement]) -> BaseFieldEl
...
+            .map(|chunk| {
+                JubjubBase::from_raw([
+                    u64::from_le_bytes(chunk[0..8].try_into().unwrap()),
+                    u64::from_le_bytes(chunk[8..16].try_into().unwrap()),
+                    u64::from_le_bytes(chunk[16..24].try_into().unwrap()),
+                    u64::from_le_bytes(chunk[24..32].try_into().unwrap()),
+                ])
+            })
+            .collect::<Vec<JubjubBase>>();
+        let result: JubjubBase = PoseidonChip::<JubjubBase>::hash(&poseidon_input);
```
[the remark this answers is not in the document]
damrobi: A test was added to check for this behavior

mithril-stm/src/signature_scheme/unique_schnorr_signature/jubjub/poseidon_digest.rs
```diff
@@ -20,8 +23,71 @@ pub(crate) fn compute_poseidon_digest(input: &[BaseFieldElement]) -> BaseFieldEl
...
+        let poseidon_input = self
+            .buffer
+            .chunks_exact(32)
+            .map(|chunk| {
+                JubjubBase::from_raw([
+                    u64::from_le_bytes(chunk[0..8].try_into().unwrap()),
+                    u64::from_le_bytes(chunk[8..16].try_into().unwrap()),
+                    u64::from_le_bytes(chunk[16..24].try_into().unwrap()),
+                    u64::from_le_bytes(chunk[24..32].try_into().unwrap()),
+                ])
```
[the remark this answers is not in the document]
damrobi: Added a comment in the code to reflect that

mithril-stm/src/hash/poseidon.rs
```diff
@@ -0,0 +1,207 @@
...
+impl FixedOutput for MidnightPoseidonDigest {
+    fn finalize_into(self, out: &mut Output<Self>) {
+        // The data is padded during the call to the update function
+        // so there should always be a multiple of 32 bytes in the buffer
+        // We are taking chunks of 32 u8 so it should be fine to unwrap
+        let poseidon_input = self
+            .buffer
+            .chunks_exact(32)
+            .map(|chunk| {
+                // The from_raw function performs a modular reduction so
```
hjeljeli32, line 62: - Here, the digest wrapper converts each 32-byte chunk into a JubjubBase using from_raw, which performs a modular reduction.
- This mapping is not injective: in general, different 32-byte values can map to the same field element after reduction.

- This is perfectly fine if the intended invariant is that all inputs passed to this digest already represent canonical JubjubBase elements, which would match how the circuit handles field elements.

- Could we confirm that this invariant always holds for the Merkle tree usage (leaf / node encodings)?

- If this invariant is guaranteed upstream, using from_bytes_le (or otherwise enforcing canonical encodings) could make this assumption explicit. 

Sorry @damrobi I remember we discussed this point previously but I still feel not very confident about it.

mithril-stm/src/hash/poseidon.rs
```diff
@@ -0,0 +1,207 @@
...
+        let mut digest_result_bytes = [0u8; 32];
+        digest_result_bytes.copy_from_slice(&digest_result);
+        let digest_result_elem = JubjubBase::from_bytes_le(&digest_result_bytes).unwrap();
+        let poseidon_result = PoseidonChip::<JubjubBase>::hash(&[elem]);
+
+        assert_eq!(digest_result_elem, poseidon_result);
+    }
+
+    #[test]
+    fn test_digest_impl_empty_byte_array() {
```
hjeljeli32, line 153: @damrobi 
I see that the digest explicitly supports empty input, and this behavior is covered by this test.

Just to double-check: is Poseidon over an empty input something we expect to rely on in practice or is this mainly for completeness?

mithril-stm/src/membership_commitment/merkle_tree/tree.rs
```diff
@@ -390,6 +392,164 @@ mod tests {
             assert!(t.to_merkle_tree_batch_commitment().verify_leaves_membership_from_batch_path(&batch_values, &deserialized).is_ok());
         }
     }
+
+    #[cfg(feature = "future_snark")]
+    mod poseidon_digest {
```
hjeljeli32, line 399: Nice coverage: this exercises the generic Merkle tree API with MidnightPoseidonDigest (paths, batch paths, serde).
One potential gap: these tests validate Merkle correctness given the digest, but they don’t validate the byte encoding contract for Poseidon inputs (i.e., that Merkle node/leaf bytes are canonical field element encodings, avoiding unintended modular reductions/collisions).
Do we want at least one small vectored test that makes this invariant explicit?

mithril-stm/src/membership_commitment/merkle_tree/tree.rs
```diff
@@ -390,6 +392,164 @@ mod tests {
...
+        use super::*;
```
hjeljeli32, line 400: @damrobi 
Sorry if I’m missing something here, but I’d like to clarify the intent of these tests.
In these Poseidon-specific tests, we build the Merkle tree with MerkleTreeConcatenationLeaf and use Blake2b to generate 32-byte chunks for paths.
Why are we still relying on concatenation leaves and Blake2b here, instead of SNARK-oriented leaves and Poseidon throughout?
Understanding the intended use case of these tests would help me.

hjeljeli32 (commented):
LGTM. I just left a few comments/suggestions.

mithril-stm/src/hash/poseidon.rs
damrobi: Hello @hjeljeli32, you are very correct with your comment. We've been discussing with @rrtoledo what is the best way to convert bytes to field element. I was using `from_raw` as a sort of security measure to make sure we never have a failure in the finalize function but if we know for sure the inputs of the digest are JubjubBase elements then we can use other conversion functions. I need to do more some benchmarking between `from_raw` and `from_bytes_le` to see which one is faster. I also want to check the function `from_raw_bytes` which could be even better.

mithril-stm/src/hash/poseidon.rs
damrobi: I added this test following Copilot recommendation and I saw that the midnight poseidon behavior for empty input is to return the zero value of JubjubBase. It can be worth discussing if we want to follow the same behavior or not.

mithril-stm/src/membership_commitment/merkle_tree/tree.rs
damrobi: Indeed it seems important to test what happens when this contract is not respected. I will add a test for that.

mithril-stm/src/membership_commitment/merkle_tree/tree.rs
damrobi: I forgot to add a comment for those tests so you're confusion is understandable. The Snark leafs are not implemented yet on this branch so I could not perform the full tests. Those tests were used mainly to see if the Digest trait implementation for Poseidon was working correctly. I will update them as soon as we have the `MerkleTreeSnarkLeaf` implemented.

mithril-stm/src/membership_commitment/merkle_tree/tree.rs
damrobi: I thought about such a test but I am not sure how to add it. For now I added a test showing the collision behavior of the poseidon digest when the input of the digest in not handled carefully

mithril-stm/src/hash/poseidon.rs
rrtoledo: In our current context, all inputs should already be Fq so it should be fine. But if we reuse this hash elsewhere potentially not and that can be dangerous. In that case, we would need to use a `hash_to_field` or something but using a map like this at all times would be incompatible with @kitounliu's circuit.

mithril-stm/src/hash/poseidon.rs
```diff
@@ -0,0 +1,264 @@
+use digest::generic_array::typenum::U32;
+use digest::{FixedOutput, HashMarker, Output, OutputSizeUser, Reset, Update};
+use midnight_circuits::{hash::poseidon::PoseidonChip, instructions::hash::HashCPU};
+use midnight_curves::Fq as JubjubBase;
+
+/// Wrapper to implement the Digest trait for the Poseidon hash function
+/// We need this implementation to keep the merkle tree implementation
+/// generic over the Digest used.
+/// This implementation differs from the usual behavior of the digest
+/// update as documented in the update implementation.
```
jpraynaud, line 11:
```suggestion
/// Wrapper to implement the Digest trait for the Poseidon hash function
/// We need this implementation to keep the Merkle tree implementation
/// generic over the Digest used.
///
/// This implementation differs from the usual behavior of the digest
/// update as documented in the update implementation.
```

mithril-stm/src/hash/poseidon.rs
```diff
@@ -0,0 +1,264 @@
...
+    /// always represent an element of JubjubBase before calling it.
+    /// A potential way to make sure we don't loose a bit of information
+    /// could be to only allow the update function to take a maximum of 32 bytes
+    /// at a time. It would be a sort of check that makes sure we're not calling
+    /// this function on "wrong" inputs somewhere in the code
+    ///
+    /// The function is also using a padding which deviates from the tradition usage
+    /// of the digest in which update([1]).update([2]) gives the same result as
+    /// update([1, 2]). We leave this functionality as is for now since this
+    /// function is only used for the merkle tree and we prefer this behavior
```
jpraynaud: Maybe this just needs to be a comment not a doc comment?

mithril-stm/src/lib.rs
```diff
@@ -128,12 +130,13 @@ pub use signature_scheme::{
...
 pub use signature_scheme::{SchnorrSigningKey, SchnorrVerificationKey, UniqueSchnorrSignature};
 
-#[cfg(feature = "future_snark")]
-use blake2::digest::consts::U64;
 use blake2::{Blake2b, digest::consts::U32};
 use digest::{Digest, FixedOutput};
 use std::fmt::Debug;
 
+#[cfg(feature = "future_snark")]
+use crate::hash::poseidon::MidnightPoseidonDigest;
```
jpraynaud:
```suggestion
#[cfg(feature = "future_snark")]
use hash::poseidon::MidnightPoseidonDigest;
```

PR #2952, opened 2026-01-24
fix(signer): handle missing/behind KES period on registration

leepl37:
## Content

This PR hardens signer registration by avoiding implicit defaults for the KES period.

Previously, `get_current_kes_period()` returning `None` would fall back to `KesPeriod(0)` (via `unwrap_or_default`), which could produce incorrect kes_evolutions.

**Changes**
- Return `RunnerError::NoValueError` when the KES period is missing (`None`), so the signer retries safely instead of defaulting to 0.
- Guards against `current < start` by logging a warning and returning an error, preventing registration with invalid data.

**Impact** 
Prevents registration with invalid KES evolution data during sync or stale chain info.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [ ] Reviewer requested
- Documentation
  - [x] No new TODOs introduced

## Issue(s)
N/A

### DIFF
--- a/mithril-signer/src/runtime/runner.rs
+++ b/mithril-signer/src/runtime/runner.rs
@@ -214,7 +214,9 @@ impl Runner for SignerRunner {
                         "current_kes_period" => u64::from(current_kes_period),
                         "start_kes_period" => u64::from(start_kes_period)
                     );
-                    Err(RunnerError::NoValueError("kes_period_underflow".to_string()))
+                    Err(RunnerError::NoValueError(
+                        "kes_period_underflow".to_string(),
+                    ))
                 } else {
                     Ok(current_kes_period - start_kes_period)
                 }
--- a/mithril-signer/src/runtime/runner.rs
+++ b/mithril-signer/src/runtime/runner.rs
@@ -205,20 +205,15 @@ impl Runner for SignerRunner {
         let (current_kes_period, kes_evolutions) = loop {
             attempt += 1;
 
-            let kes_period = self
-                .services
-                .chain_observer
-                .get_current_kes_period()
-                .await?;
+            let kes_period = self.services.chain_observer.get_current_kes_period().await?;
 
             let kes_period = match kes_period {
                 Some(kes_period) => kes_period,
                 None => {
                     if attempt >= max_retries {
-                        return Err(RunnerError::NoValueError(
-                            "current_kes_period".to_string(),
-                        )
-                        .into());
+                        return Err(
+                            RunnerError::NoValueError("current_kes_period".to_string()).into()
+                        );
                     }
                     warn!(
                         self.logger,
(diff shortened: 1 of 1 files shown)


mithril-signer/src/runtime/runner.rs
```diff
@@ -198,10 +198,72 @@ impl Runner for SignerRunner {
...
+                        "KES period mismatch. Retrying...";
+                        "current_kes_period" => u64::from(kes_period),
+                        "start_kes_period" => u64::from(start_kes_period),
+                        "attempt" => attempt,
+                        "max_retries" => max_retries
+                    );
+                    tokio::time::sleep(retry_delay).await;
+                }
+            }
+        };
```
jpraynaud, line 266: I agree that we could trigger an error if the Cardano node is not completely synced:
- `current_kes_period` is `None`
- or if `current_kes_period` is less than `operational_certificate.get_start_kes_period()`

However, this code is too complex and does not take advantage of the retry mechanism of the state machine (an error will restart the state machine which will eventually run this same `register_signer_to_aggregator` code later).

jpraynaud (changes requested):
Thanks @leepl37 for this PR.

I left a comment explaining why it needs to be modified before we can merge it.

mithril-signer/src/runtime/runner.rs
leepl37: Thanks for the review  @jpraynaud! 
I agree that the retry loop adds complexity and that relying on the State Machine is usually the cleaner default.

The main reason I tried this workaround was to solve a race condition I kept hitting in the E2E Devnet tests:

1. The Trigger: When the `cardano-node` is briefly syncing at an epoch boundary, it reports an outdated KES period. 

2. The Problem: This error causes the State Machine to trigger its standard `run_interval` sleep (default 5s).

3. The Failure: Since Devnet epochs are extremely short, even a single backoff cycle creates a risk of missing the registration window.


Evidence:

Devnet Config: The epoch duration is only 75 seconds (`0.75s` slot length * `100` slots).

The Error: Before adding this loop, I consistently encountered this failure in E2E tests:

```
Error(Unretryable): Mithril End to End test failed
Caused by:
    0: Requesting aggregator `mithril-aggregator-1`
    1: Minimum expected mithril stake distribution epoch not reached: 20 < 21
```

This confirms the Signer missed the window for epoch 21 because it was sleeping during the transition time.

**Mainnet & CI Verification:**

- Devnet: The CI build passed with this fix, confirming it stabilizes the race condition.

- Mainnet: This logic is also safe for Mainnet. While the standard 5s backoff is negligible on Mainnet (due to 5-day epochs), adding this fast retry makes the Mainnet signer more robust against transient node glitches without introducing risks.

**Question**: Is there a preferred way to trigger an "immediate" retry (skipping the standard backoff) for this specific transient error? If not, I can remove the loop, but I am worried it will re-introduce these random CI failures.

Issue #2792, opened 2025-11-20
SNARK pre-aggregation primitives in STM library - Phase 1


jpraynaud:
## Why
We need to evolve the STM library in order to implement SNARK primitives. Following the definition of the architecture in #2763, we have defined an implementation plan and we want to address the SNARK pre-aggregation primitives part of the plan for the key registration.

## What
Implement the SNARK pre-aggregation primitives for the key registration.

## How
- [x] Implement SNARK pre-aggregation primitives (i.e. not the circuit(s) themselves)
  - [x] Create a `halo2-snark` proof system sub-module in the `proof_system` module with sub-modules:
    - [x] gated behind `future_snark` 
    - [x] `single_signature`
    - [x] ~~`proof`~~ (we'll do that later)
    - [x] ~~`clerk`~~ (we'll do that later)
    - [x] `aggregate_verification_key`
  - [x] Adapt the key registration to support the new `halo2` proof system:
    - [x] Create Merkle leaf entry for SNARK `MerkleTreeSnarkLeaf` (Schnorr verification key and Lottery Target Value)
    - [x] Create SNARK aggregate verification key (root of the Merkle tree for SNARK created, and with the Poseidon hash `MembershipDigest::SnarkHash`)
    - [x] Add the Schnorr verification key in the `RegistrationEntry` (as an option for backward compatibility)
    - [x] Move the Merkle tree creation to the `ClosedKeyRegistration`
    - [x] Implement a new `ClosedRegistrationEntry` (to be able to compute the `phi` value needed for SNARK leaf)
    - [x] Adapt the conversion of `KeyRegistration` to `ClosedKeyRegistration`
    - [x] Implement conversion trait (`From`) for the `ClosedRegistrationEntry` to `MerkleTreeSnarkLeaf`
    - [x] Adapt `Initializer`
  - [x] ~~Benchmarks~~
  - [x] Golden tests
  - [x] Property-based tests
  - [x] Update the [MVP Book](https://hackmd.io/EeCC7vaPQoWdmMVAkyDQpQ?view=#)
  - [x] Run a backward compatibility test when PR is merged

Issue #2945, opened 2026-01-21
Remove `off-circuit` module in Halo2 circuit


jpraynaud:
## Why
The prototype circuit uses some types and mirror implementations of types already implemented in the STM library. We need to switch to the existing types in STM and remove the `circuit/halo2/off-circuit` module.

## What
Switch to the existing types in STM and remove the `circuit/halo2/off-circuit` module

## How

- [ ] Switch to the existing types in STM
- [ ] Remove the `circuit/halo2/off-circuit` module

hjeljeli32:
This comment needs to be taken in consideration
https://github.com/input-output-hk/mithril/pull/2968#discussion_r2752089771

<img width="832" height="274" alt="Image" src="https://github.com/user-attachments/assets/0b32d0dd-f52f-4aaf-84f8-747e64d33483" />

hjeljeli32:
@jpraynaud @curiecrypt 
Based on the investigation of the state of STM pre-aggregation primitives, I suggest splitting the “remove off-circuit” work into 2 phases.

**Rationale:**
Some elements (notably Schnorr signatures for the SNARK path) depend on STM Phase-2 work, which is not implemented yet. Other elements (Merkle + Poseidon) are already available and can be wired today.

**Phase-A: Merkle + Poseidon (can be done now)
Phase-B: Signatures (depends on #2943)**

hjeljeli32:
**Update: signature wiring (off-circuit → STM)**

After reviewing the off-circuit signature code and the STM unique_schnorr_signature implementation, I want to document the current **semantic differences** before we start wiring. This explains why the migration needs to be done incrementally.

Observed semantic differences
- **Message format**
     - Off-circuit: message is a list of field elements (JubjubBase)
     - STM: message is raw bytes (&[u8])
- **Hash-to-curve**
     - Off-circuit: hash-to-curve over field elements
     - STM: hash-to-curve over bytes
- **Poseidon challenge preimage**: Both use Poseidon, but input ordering and transcript differ
- **DST**
     - Off-circuit: DST_UNIQUE_SIGNATURE
     - STM: internal signature DST
- **Signature shape**
     - Off-circuit: (sigma, s, c)
     - STM: { commitment_point, response, challenge }

These semantics are used in:
   - the Halo2 signature gadget
   - witness construction 
   - golden tests
   - verification key / leaf encoding at the Merkle layer

So we need to align these semantics.

**Compatibility contract**
The target end state is:
[...]

PR #2957, opened 2026-01-26
feat(stm): Merkle Tree Leaf for Snark proof

curiecrypt:
## Content

This PR includes the Merkle tree leaf for snark proof system. 

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Issue(s)

Relates to #2792

### DIFF
--- a/mithril-stm/src/membership_commitment/merkle_tree/leaf.rs
+++ b/mithril-stm/src/membership_commitment/merkle_tree/leaf.rs
@@ -137,9 +137,7 @@ impl PartialOrd for MerkleTreeSnarkLeaf {
     /// Ordering of MT Values.
     ///
     /// First we order by target value, then by key. By having this ordering,
-    /// we have the players with higher target value close together,
-    /// meaning that the probability of having several signatures in the same side of the tree, is higher.
-    /// This allows us to produce a more efficient batch opening of the merkle tree.
+    /// we have the players with higher target value close together.
     fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
         Some(std::cmp::Ord::cmp(self, other))
     }
@@ -163,6 +161,7 @@ mod tests {
 
     use super::*;
 
+    #[cfg(feature = "future_snark")]
     mod golden {
         use super::*;
         const GOLDEN_BYTES: &[u8; 104] = &[
@@ -194,18 +193,17 @@ mod tests {
         }
     }
 
-    #[cfg(feature = "future_snark")]
     mod golden_json {
         use super::*;
         const GOLDEN_JSON: &str = r#"
-        {
+        [
             [143,161,255,48,78,57,204,220,25,221,164,252,248,14,56,126,186,135,228,188,145,
             181,52,200,97,99,213,46,0,199,193,89,187,88,29,135,173,244,86,36,83,54,67,164,
             6,137,94,72,6,105,128,128,93,48,176,11,4,246,138,48,180,133,90,142,192,24,193,
             111,142,31,76,111,110,234,153,90,208,192,31,124,95,102,49,158,99,52,220,165,94,
             251,68,69,121,16,224,194],
             1
-        }"#;
+        ]"#;
 
         fn golden_value() -> MerkleTreeConcatenationLeaf {
             let mut rng = ChaCha20Rng::from_seed([0u8; 32]);
@@ -217,7 +215,7 @@ mod tests {
 
         #[test]
         fn golden_conversions() {
-            let value = serde_json::from_str(GOLDEN_JSON)
+            let value: MerkleTreeConcatenationLeaf = serde_json::from_str(GOLDEN_JSON)
                 .expect("This JSON deserialization should not fail");
             assert_eq!(golden_value(), value);
 


mithril-stm/src/membership_commitment/merkle_tree/leaf.rs
```diff
@@ -75,3 +78,78 @@ impl Ord for MerkleTreeConcatenationLeaf {
         self.1.cmp(&other.1).then(self.0.cmp(&other.0))
     }
 }
+
+#[cfg(feature = "future_snark")]
+// TODO: remove this allow dead_code directive when function is called or future_snark is activated
+#[allow(dead_code)]
+/// The values that are committed in the Merkle Tree for `SnarkProof`.
+/// Namely, a verified `SchnorrVerificationKey` and its corresponding target value.
+#[derive(Debug, Clone, Copy, Default, PartialEq, Eq, Serialize, Deserialize, Hash)]
```
jpraynaud, line 87: We probably miss some golden tests for the leaves (also for concatenation) for JSON and bytes encoding.

mithril-stm/src/lib.rs
```diff
@@ -163,6 +166,12 @@ pub type StmError = anyhow::Error;
 /// Mithril-stm result type
 pub type StmResult<T> = anyhow::Result<T, StmError>;
 
+#[cfg(feature = "future_snark")]
+// TODO: remove this allow dead_code directive when function is called or future_snark is activated
+#[allow(dead_code)]
+/// Target value type used in the lottery for snark proof system
+pub type TargetValue = crate::signature_scheme::BaseFieldElement;
```
jpraynaud: The name `TargetValue` is a bit vague, maybe something like `LotteryTargetValue` would be better?

mithril-stm/src/membership_commitment/merkle_tree/leaf.rs
```diff
@@ -75,3 +78,40 @@ impl Ord for MerkleTreeConcatenationLeaf {
...
+    fn as_bytes_for_merkle_tree(&self) -> Vec<u8> {
+        self.to_bytes()
+    }
+}
+
+#[cfg(feature = "future_snark")]
+// TODO: remove this allow dead_code directive when function is called or future_snark is activated
+#[allow(dead_code)]
+impl MerkleTreeSnarkLeaf {
+    fn to_bytes(self) -> Vec<u8> {
```
damrobi, line 103: Since this function is used in the `as_bytes_for_merkle_tree()` function, it needs to decompose the SchnorrVerificationKey into its two coordinates and returns 32 bytes for each. See the comment I left in the other issue [here](https://github.com/input-output-hk/mithril/issues/2945#issuecomment-3800350405). Let me know if this is not clear and I can give more details

mithril-stm/src/membership_commitment/merkle_tree/leaf.rs
```diff
@@ -75,3 +78,40 @@ impl Ord for MerkleTreeConcatenationLeaf {
...
+        let mut result = [0u8; 64];
+        result[..32].copy_from_slice(&self.0.to_bytes());
+        result[32..].copy_from_slice(&self.1.to_bytes());
+        result.to_vec()
+    }
+
+    pub(crate) fn from_bytes(bytes: &[u8]) -> StmResult<Self> {
+        let pk = VerificationKeyForSnark::from_bytes(&bytes[..32])
```
damrobi, line 114: Related to the comments on `to_bytes()`, we might need to assume the `SchnorrVerificationKey` is 64 bytes in this use-case.

mithril-stm/src/signature_scheme/unique_schnorr_signature/verification_key.rs
```diff
@@ -62,6 +65,12 @@ impl SchnorrVerificationKey {
     }
 }
 
+impl Hash for SchnorrVerificationKey {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        Hash::hash_slice(&self.to_bytes(), state)
```
damrobi, line 70: If we plan to use this implementation, maybe its easier to modify the `to_bytes()` implementation of `SchnorrVerificationKey` directly (see my other comments).

mithril-stm/src/membership_commitment/merkle_tree/leaf.rs
```diff
@@ -75,3 +78,78 @@ impl Ord for MerkleTreeConcatenationLeaf {
...
+    }
+}
+
+#[cfg(feature = "future_snark")]
+// TODO: remove this allow dead_code directive when function is called or future_snark is activated
+#[allow(dead_code)]
+impl PartialOrd for MerkleTreeSnarkLeaf {
+    /// Ordering of MT Values.
+    ///
+    /// First we order by target value, then by key. By having this ordering,
```
damrobi, line 139: I'm not sure this comment is valid for the Snark tree.

curiecrypt: Why?

mithril-stm/src/membership_commitment/merkle_tree/leaf.rs
curiecrypt: To update the `to_bytes` and `from_bytes` implementations of `SchnorrVerificationKey`:
- Update `to_bytes` and `from_bytes` implementations of `ProjectivePoint`:
   - Call `get_coordinates` and convert each coordinate to bytes and return it. 
   - Revise `from_bytes` accordingly.
   
Does that solve the problem? Should we modify any other implementation related to verification key or projective point?

curiecrypt: I don't think that solves.

mithril-stm/src/membership_commitment/merkle_tree/leaf.rs
kitounliu: the order seems fine, but with snark, there is no batch opening for merkle tree in the circuit

curiecrypt: Ahah copy-paste left over. Thanks for noticing that @damrobi @kitounliu

mithril-stm/src/membership_commitment/merkle_tree/leaf.rs
damrobi: I think that might solve it. If we update the to/from_bytes of the `SchnorrVerificationKey` to use the two points in the correct order and base the rest on that it should be compatible with the circuit

curiecrypt: Good. Then, I will leave this branch as it is.
So, in PR 2, you can revise the to/from bytes of verification key and snark leaf. Also, add golden tests for leaf.
Does that sound good?

damrobi: That sounds good to me 👍

mithril-stm/src/membership_commitment/merkle_tree/leaf.rs
curiecrypt: I have added golden tests for concatenation. Snark ones will be added in the next pr.

PR #2958, opened 2026-01-26
feat(stm): Lottery target value computation for snark proof

### LINKED ISSUE #2974: Compute Lottery Target Value for SNARK
## Why
The key registration leaves for SNARK do not include the stake (as it is the case with the Concatenation), they need to include the Lottery Target Value which is a pre-computation needed for eligibility verification in the SNARK circuit.
We need to implement this pre-computation and also refactor the eligibility check module and split responsibilities of the computation in the Concatenation and SNARK proof systems.

## What
Implement the Lottery Target Value computation for SNARK and refactor the eligibility check module.

## How

- [x] Implement the Lottery Target Value computation for SNARK:
  - [x] Compute the Lottery Target Value in a new `eligibility` module of the `halo2_snark` proof system
  - [x] ~~Support `rug` and `num-integer` backends~~
- [x] Move the eligibility check for Concatenation to the `concatenation` proof system
- [x] [Document discussing the target value computation](https://hackmd.io/@ME2_HxZuTma6JS_01Hd_oQ/HJjlNIzwZe/edit)

curiecrypt:
## Content

This PR includes lottery target value computation for snark proof system.

- Move eligibility check logic from `protocol` into `proof_system` 
- Implement SNARK-specific lottery eligibility: target value computation `(phi = 1 - (1 - phi_f)^(stake/total_stake)` mapped to the `BaseFieldElement`), lottery prefix hashing via Poseidon, and index eligibility checking
- Extend `BaseFieldElement` with `Neg`, `From<u64>`, and `random()` to support the new eligibility computations
- Add `cfg_num_integer!` / `cfg_rug!` macros to deduplicate conditional compilation attributes across proof system modules

**There is no precision alignment in `num_integer` implementation. I could not find an equivalent functionality in the library. Suggestions are welcome!**

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Issue(s)

Closes #2974

### DIFF
--- a/mithril-stm/src/proof_system/halo2_snark/eligibility.rs
+++ b/mithril-stm/src/proof_system/halo2_snark/eligibility.rs
@@ -41,7 +41,10 @@ cfg_num_integer! {
 
 cfg_rug! {
     #[cfg(feature = "future_snark")]
-    use rug::{Float, integer::Order, ops::Pow};
+    use rug::{Float, integer::Order, ops::Pow, Integer, float::Round};
+
+    #[cfg(feature = "future_snark")]
+    use std::cmp::Ordering;
 
     #[cfg(feature = "future_snark")]
     pub fn compute_target_bytes(phi_f: f64, stake: Stake, total_stake: Stake) -> Vec<u8> {
@@ -61,7 +64,7 @@ cfg_rug! {
         let (t_int, order) = t.to_integer_round(Round::Zero).unwrap();
         assert!(t_int >= 0);
 
-        let target = match order {
+        let target: Integer = match order {
             Ordering::Less => t_int,
             Ordering::Equal => {
                 if t_int == 0 { t_int }
@@ -74,7 +77,10 @@ cfg_rug! {
 }
 
 #[cfg(feature = "future_snark")]
-use crate::{LotteryTargetValue, Stake, signature_scheme::BaseFieldElement};
+use crate::{
+    LotteryIndex, LotteryTargetValue, SignatureError, Stake, StmResult, UniqueSchnorrSignature,
+    signature_scheme::{BaseFieldElement, DST_LOTTERY, compute_poseidon_digest},
+};
 
 #[cfg(feature = "future_snark")]
 // TODO: remove this allow dead_code directive when function is called or future_snark is activated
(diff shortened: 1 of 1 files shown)


mithril-stm/src/protocol/lottery/snark_eligibility.rs
```diff
@@ -0,0 +1,196 @@
...
+    let modulus = Integer::from_str_radix(
+        "73eda753299d7d483339d80809a1d80553bda402fffe5bfeffffffff00000001",
+        16,
+    )
+    .map_err(|_| LotteryError::BaseFieldElementConversion)?;
+
+    let w = Float::with_val(117, stake) / Float::with_val(117, total_stake);
+    let phi = Float::with_val(117, 1.0) - Float::with_val(117, 1.0 - phi_f).pow(w);
+    // increase precision
+    let phi_high = Float::with_val(300, phi);
```
rrtoledo: Can `phi_high` really have a precision of 300 if `phi` has a precision of 100 (I do not know how this work)?

rrtoledo: Just checked, and floats are _not_ deterministic.

mithril-stm/src/protocol/lottery/snark_eligibility.rs
```diff
@@ -0,0 +1,196 @@
...
+/// - `phi_f` is outside the valid range `[0.0, 1.0]`
+/// - Numeric conversion fails during computation
+pub fn compute_target_value(
+    phi_f: f64,
+    stake: Stake,
+    total_stake: Stake,
+) -> StmResult<LotteryTargetValue> {
+    // modulus - 1
+    let ev_max = -BaseFieldElement::get_one();
+    if !(0.0..=1.0).contains(&phi_f) {
```
rrtoledo: Does this remove 0?

mithril-stm/src/protocol/lottery/snark_eligibility.rs
kitounliu: 300 precision is for the following computation of t_float not for phi. Basically 255-bit precision for the integer part and around 50-bit for the fraction part

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
```diff
@@ -0,0 +1,216 @@
...
+/// where `phi_f` is the base probability parameter.
+pub fn compute_target_value(phi_f: f64, stake: Stake, total_stake: Stake) -> LotteryTargetValue {
+    // If phi_f = 1, then we automatically break with true
+    if (phi_f - 1.0).abs() < f64::EPSILON {
+        return -BaseFieldElement::get_one();
+    }
+
+    let mut bytes: Vec<u8> = compute_target_bytes(phi_f, stake, total_stake);
+    bytes.resize(32, 0);
+    BaseFieldElement::from_bytes(&bytes).unwrap()
```
damrobi, line 104: Should we add a comment explaining why the `unwrap()` is safe?

mithril-stm/src/protocol/lottery/snark_eligibility.rs
rrtoledo: We should discuss on some specs for what properties we want for the SPO's stake and targets (what is the min / max stake a SPO can have, what is the precision we want / how much granularity should we define in the target...).

rrtoledo: Just saw the explanation from the concatenation `rug` part. I am not sure we have 117 bits of precision in the halo2 however.

kitounliu: yes, and the consistency of the target computation between:
- different platform, rust version, etc.
- different packages, i.e., rug and num_integer

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
```diff
@@ -0,0 +1,216 @@
...
+            16,
+        )
+        .unwrap();
+
+        let c =
+            Ratio::from_float((1.0 - phi_f).ln()).expect("Only fails if the float is infinite or NaN.");
+        let w = Ratio::new_raw(BigInt::from(stake), BigInt::from(total_stake));
+
+        let phi = (w * c).neg();
+        let t = Ratio::from(modulus) * phi;
```
hjeljeli32, line 24: As pointed out by Copilot, it seems that the expressions of phi (theoretical and implemented) are equivalent to first-order (due to Taylor expansion), are the differences acceptable in our case? I am not sure whether Copilot's proposed method leads to a better approximation or not.

hjeljeli32 (commented):
LGTM.

One point: Copilot raised a question about the num-integer implementation using the logarithmic / Taylor-based formulation for phi versus the direct probability expression. I didn’t have time to fully validate this mathematically, and it may well be fine as-is, but it could be worth a quick double-check just to make the intended approximation and its precision guarantees explicit.

PR #2966, opened 2026-01-29
feat: optional throttling when reading blocks from the chain 

### LINKED ISSUE #2909: Add throttling to Blocks and Transactions preloader
## Why
We need to limit the resource needed on the infrastructure during the Blocks and Transactions import done by the pre-loaders in aggregator and signer.

## What
Implement a throttler to the Blocks and Transactions import done pre-loaders.

## How

- [x] Add throttling to the pre-loader to use `40%` instead of `75%` CPU (e.g. with some delays between import calls)
  - [x] In aggregator
  - [x] In signer
- [x] Add a pre-loader for Blocks and Transactions  
  - [x] In aggregator
  - [x] In signer

Alenar:
## Content

This PR add an optional throttling ability to the `CardanoBlockScanner` and `BlockStreamer` used to poll blocks from the chain. The goal is to reduce the aggregator and signer load on the servers when running the import process (which take more than 60% of the CPU time today).

It works by configuring an interval in milliseconds, the throttler then ensure that each calls that polls an group of blocks from the `BlockStreamer` take at least the given time, if the calls was slower than the interval it execute immediately.

It's configured with a default value of `400ms` in the signer and aggregator, and setting the configuration value to `0` disable the throttler entirely.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation
  - [x] No new TODOs introduced

## Issue(s)

Closes #2909

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3887,7 +3887,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-aggregator"
-version = "0.8.16"
+version = "0.8.17"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -4032,7 +4032,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-cardano-node-chain"
-version = "0.1.16"
+version = "0.1.17"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -4267,7 +4267,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-end-to-end"
-version = "0.4.116"
+version = "0.4.117"
 dependencies = [
  "anyhow",
  "async-recursion",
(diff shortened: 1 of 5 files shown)


internal/cardano-node/mithril-cardano-node-chain/src/chain_scanner/block_scanner.rs
```diff
@@ -32,9 +35,23 @@ impl CardanoBlockScanner {
         Self {
             chain_reader,
             max_roll_forwards_per_poll,
+            throttling_interval: None,
             logger,
         }
     }
+
+    /// Set the throttling interval for the block scanner
```
jpraynaud:
```suggestion
    /// Set the optional throttling interval for the block scanner
```

Issue #2912, opened 2026-01-12
Add golden tests for Halo2 circuit prototype


jpraynaud:
## Why
In order to prepare the production ready version of the Halo2 non recursive circuit, we want to create golden tests that will help confidently refactor the circuit with low risk of deviating from the specifications.

## What
Implement a suite of golden tests that covers the circuit specification and edge cases.

## How

> [!IMPORTANT]
> The code of the circuit prototype must not be modified at this stage.

- [x] Define the golden test strategy (baseline, positive edge cases, negative cases)
- [x] Propose initial test matrices (positive & negative)
- [x] Review the strategy and test matrices with
    - [x] @curiecrypt 
    - [x] @damrobi 
    - [x] @jpraynaud 
    - [x] @rrtoledo 
    - [x] @kitounliu
- [x] Explore and propose tooling to keep tests readable and maintainable
- [x] Feasibility analysis: Go through each proposed test case and classify as
    - trivial to implement
    - requires helpers/refactoring
    - impractical with current APIs  
- [x] Implement the golden tests
    - [x] Baseline
    - [x] (Edge-case) positive tests
    - [x]  Negative tests 
- [x] Make sure that the tests do not take longer in the CI than without the golden tests

hjeljeli32:
Hi @jpraynaud 
An update related to **"Make sure that the tests do not take longer in the CI than without the golden tests"**
I measured the test runtime with cargo nextest (ci profile, --features future_snark).
Suite excluding halo2 golden tests: **~19.9s (154 tests)**.
Suite including golden tests: ~**32.6s (176 tests).**
So enabling golden tests adds **~12.6s wall-time** on my machine.

Btw, these numbers were obtained after implementing the suggestion of @rrtoledo to cache pk/vk. @rrtoledo also suggested to use MockProver instead of Prover which could roughly accelerate proving time by 2x (based on previous experiments by @damrobi), I still did not implement that optimization.

PR #2968, opened 2026-01-30
 test(snark): Golden tests for the STM Halo2 circuit

hjeljeli32:
## Content

This PR adds a **comprehensive suite of golden tests** for the STM Halo2 circuit. The goal is to l**ock in the circuit specification and edge-case behavior,** so we can **confidently refactor or harden the circuit later** with low risk of  regressions.

Notes: 4 length tests **currently trigger a panic in the circuit**; they are **marked as should_panic and require hardening follow-ups**.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation
  - [x] No new TODOs introduced

## Issue(s)
Closes https://github.com/input-output-hk/mithril/issues/2912

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4461,7 +4461,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-stm"
-version = "0.9.12"
+version = "0.9.13"
 dependencies = [
  "anyhow",
  "blake2 0.10.6",
--- a/mithril-stm/CHANGELOG.md
+++ b/mithril-stm/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## 0.9.13 (02-06-2026)
+
+### Added
+
+- Added golden tests for the Halo2 STM circuit.
+
 ## 0.9.12 (02-02-2026)
 
 ### Added
--- a/mithril-stm/Cargo.toml
+++ b/mithril-stm/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mithril-stm"
-version = "0.9.12"
+version = "0.9.13"
 edition = { workspace = true }
 authors = { workspace = true }
 homepage = { workspace = true }


mithril-stm/src/circuits/halo2/circuit/mod.rs
```diff
@@ -1,5 +1,5 @@
 //! In-circuit relations and constraint systems for the Halo2 prototype.
 
-pub mod certificate;
+pub mod stm;
```
damrobi: Name suggestion for the module and structure implementing the circuit: stm_snark?

mithril-stm/src/circuits/halo2/golden/support/mod.rs
```diff
@@ -0,0 +1,3 @@
+//! Shared support code for golden test cases.
+
+pub(crate) mod stm_case;
```
damrobi: Rename to `stm_case_setup`?

mithril-stm/src/circuits/halo2/golden/helpers.rs
```diff
@@ -0,0 +1,385 @@
...
+        let usk = sks[ii].clone();
+        let uvk = leaves[ii].0;
+        let sig = usk.sign(&[merkle_root, msg], &mut OsRng);
+        sig.verify(&[merkle_root, msg], &uvk).unwrap();
+
+        let merkle_path = merkle_tree.get_path(ii);
+        let computed_root = merkle_path.compute_root(leaves[ii]);
+        assert_eq!(merkle_root, computed_root);
+
+        // any index is eligible as target is set to be the maximum
```
damrobi, line 251: I think this comment is not valid all the time anymore since the leaves are given as input.

damrobi (approved):
LGTM so far 👍 
I left some naming suggestion for the module just in case.

mithril-stm/src/circuits/halo2/circuit/mod.rs
jpraynaud: We are already in the STM library, this module is defining the circuit so the name should probably include `circuit`. Maybe this module and the `gadgets` should be at the root of the `halo2` module.

mithril-stm/src/circuits/halo2/circuit/stm.rs
```diff
@@ -22,15 +22,15 @@ type F = JubjubBase;
 type C = Jubjub;
 
 #[derive(Clone, Default, Debug)]
-pub struct Certificate {
+pub struct STM {
```
jpraynaud: Maybe this is enough?
```suggestion
pub struct Circuit {
```

mithril-stm/src/circuits/halo2/golden/cases/mod.rs
jpraynaud, line 1: IMO all of these modules with only one sub-module should be replaced with single file modules (e.g. `cases.rs`).

mithril-stm/src/circuits/halo2/golden/support/stm_case.rs
```diff
@@ -0,0 +1,385 @@
...
+use ff::Field;
+use midnight_proofs::poly::kzg::params::ParamsKZG;
+use midnight_proofs::utils::SerdeFormat;
+use midnight_zk_stdlib as zk;
+use midnight_zk_stdlib::{MidnightCircuit, MidnightPK, MidnightVK};
+use rand_core::OsRng;
+use std::fs::create_dir_all;
+use std::io::Cursor;
+use std::path::PathBuf;
+use std::time::Instant;
```
jpraynaud: We tend to follow the following order for the imports (for a better readability):
1 - Std imports
2 - Line break
3 - External imports
4 - Line break
5 - Crate imports
6 - Line break
7 - Super imports

mithril-stm/src/circuits/halo2/golden/support/stm_case.rs
```diff
@@ -0,0 +1,385 @@
...
+
+type F = JubjubBase;
+
+type WitnessEntry = (MTLeaf, MerklePath, Signature, u32);
+
+/// Shared environment for STM golden cases (SRS, relation, keys, sizing).
+pub(crate) struct STMEnv {
```
jpraynaud: All the `STM` could be replaced by `Circuit`:
```suggestion
pub(crate) struct CircuitEnv {
```

mithril-stm/src/circuits/halo2/golden/helpers.rs
```diff
@@ -0,0 +1,385 @@
...
+type WitnessEntry = (MTLeaf, MerklePath, Signature, u32);
+
+/// Shared environment for STM golden cases (SRS, relation, keys, sizing).
+pub(crate) struct STMEnv {
+    srs: ParamsKZG<Bls12>,
+    relation: STM,
+    vk: MidnightVK,
+    pk: MidnightPK<STM>,
+    num_signers: usize,
+    num_lotteries: u32,
```
jpraynaud, line 71: Could add descriptions to these fields as well?

mithril-stm/src/circuits/halo2/golden/support.rs
```diff
@@ -0,0 +1,385 @@
...
+    let tree = MerkleTree::create(&leaves);
+
+    (sks, leaves, tree)
+}
+
+/// Construct the STM relation environment and setup keys for a case.
+pub(crate) fn setup_stm_env(case_name: &str, k: u32, quorum: u32) -> STMEnv {
+    let srs = load_or_generate_params(k);
+
+    // Keep num_signers fixed for baseline comparisons.
```
jpraynaud: I'm not convinced that all these comments in this function bring a lot of value

mithril-stm/src/circuits/halo2/golden/support.rs
```diff
@@ -0,0 +1,385 @@
...
+            sig.clone(),
+            *index,
+        ));
+    }
+
+    witness
+}
+
+/// Errors returned by `prove_and_verify_result`.
+#[derive(Debug)]
```
jpraynaud: It'd be a good idea to implement `thisError` for the errors.

mithril-stm/src/circuits/halo2/circuit.rs
```diff
@@ -22,15 +22,15 @@ type F = JubjubBase;
 type C = Jubjub;
 
 #[derive(Clone, Default, Debug)]
-pub struct Certificate {
+pub struct STM {
     // k in mithril: the required number of signatures for a valid certificate
     quorum: u32,
     // m in mithril: the number of lotteries that a user can participate in to sign a message
     num_lotteries: u32,
     merkle_tree_depth: u32,
```
jpraynaud: This is probably a bit too early but this comments and field names are not accurate. At least the comment of `quorum` deserves to be updated.

mithril-stm/src/circuits/halo2/circuit/stm.rs
curiecrypt: I agree

curiecrypt: In this case, we also update the file name?

mithril-stm/src/circuits/halo2/golden/cases/negative.rs
```diff
@@ -0,0 +1,564 @@
+use crate::circuits::halo2::golden::support::stm_case::{
+    STMProofError, STMScenario, build_witness, build_witness_with_fixed_signer,
+    build_witness_with_indices, create_default_merkle_tree,
+    create_merkle_tree_with_controlled_leaf, create_merkle_tree_with_leftmost_leaf,
+    create_merkle_tree_with_rightmost_leaf, prove_and_verify_result, run_stm_case,
+    run_stm_case_default, setup_stm_env,
+};
+use crate::circuits::halo2::types::{JubjubBase, JubjubScalar};
```
curiecrypt, line 10: We should change the direct calls to `JubjubBase`, `JubjubScalar`, etc. 
It can be done in off-circuit pr.

mithril-stm/src/circuits/halo2/golden/cases/stm.rs
```diff
@@ -0,0 +1,564 @@
...
+    build_witness_with_indices, create_default_merkle_tree,
+    create_merkle_tree_with_controlled_leaf, create_merkle_tree_with_leftmost_leaf,
+    create_merkle_tree_with_rightmost_leaf, prove_and_verify_result, run_stm_case,
+    run_stm_case_default, setup_stm_env,
+};
+use crate::circuits::halo2::types::{JubjubBase, JubjubScalar};
+use ff::Field;
+
+#[test]
+fn test_stm_baseline() {
```
curiecrypt: Maybe remove all `test_` prefixes?

mithril-stm/src/circuits/halo2/circuit.rs
curiecrypt: We can handle this in the off-circuit pr. However, I am a bit concerned about another point. 
The `Parameters` structure uses `u64` for `k` and `m`. I am not sure whether it causes a compatibility problem, especially for the tests.

mithril-stm/src/circuits/halo2/golden/cases.rs
```diff
@@ -0,0 +1,564 @@
...
+use ff::Field;
+
+#[test]
+fn test_stm_baseline() {
+    const K: u32 = 13;
+    const QUORUM: u32 = 3;
+    run_stm_case_default("small", K, QUORUM);
+}
+
+#[test]
```
curiecrypt: Concise explanations would be better for each case.

curiecrypt (commented):
It looks good in general. I left some minor comments.
A big plus would be adding a doc comment on `/golden/mod.rs` that roughly explains the module. Yet, it is not critical. It is good to have!

mithril-stm/src/circuits/halo2/golden/support.rs
hjeljeli32: @jpraynaud 
You are right, these comments were added in the prototype but I will remove them.

mithril-stm/src/circuits/halo2/golden/helpers.rs
hjeljeli32: @damrobi 
Good catch, you’re right. Since target is now carried by the leaf input and can be controlled or mutated in tests, that comment was not correct.
How about I replace it like this?
“eligibility (ev ≤ target) is enforced in-circuit; targets come from the leaf input.”

mithril-stm/src/circuits/halo2/circuit.rs
hjeljeli32: > This is probably a bit too early but this comments and field names are not accurate. At least the comment of `quorum` deserves to be updated.

@jpraynaud 
Good point. The comments/field names are indeed not accurate now.
To keep this PR scoped to golden tests, I suggest deferring the fix to a follow-up circuit refactor where we can fix/enhance naming and semantics.

hjeljeli32: @curiecrypt 
Thanks for flagging this. That’s a fair concern.

I agree the u64 vs u32 discrepancy is something we should align, but to keep this PR scoped to golden tests, I propose handling it in a follow-up PR alongside other circuit/parameter cleanups.

I’ll track it explicitly so we can address it properly.

mithril-stm/src/circuits/halo2/circuit.rs
```diff
@@ -18,19 +11,27 @@ use midnight_proofs::circuit::{Layouter, Value};
...
+use crate::circuits::halo2::off_circuit::merkle_tree::{MTLeaf, MerklePath};
+use crate::circuits::halo2::off_circuit::unique_signature::Signature;
+use crate::circuits::halo2::types::{Jubjub, JubjubBase, LotteryIndex, MerkleRoot, Msg};
+
 type F = JubjubBase;
 type C = Jubjub;
 
 #[derive(Clone, Default, Debug)]
-pub struct Certificate {
+pub struct Circuit {
```
jpraynaud:
```suggestion
pub struct StmCircuit {
```

mithril-stm/src/circuits/halo2/circuit.rs
```diff
@@ -18,19 +11,27 @@ use midnight_proofs::circuit::{Layouter, Value};
...
     // k in mithril: the required number of signatures for a valid certificate
```
jpraynaud:
```suggestion
    // k in mithril: the required number of distinct lottery indices slots needed to create a valid multi-signature
```

mithril-stm/src/circuits/halo2/circuit.rs
kitounliu: the comments for k and m were originally from https://mithril.network/doc/mithril/advanced/mithril-protocol/protocol

kitounliu: k and m are small values less than 2^16, so switching between  u64 and u32 should not affect the circuit size

jpraynaud: > the comments for k and m were originally from https://mithril.network/doc/mithril/advanced/mithril-protocol/protocol

Thanks for spotting that @kitounliu, I fixed the documentation in this PR: https://github.com/input-output-hk/mithril/pull/2975

mithril-stm/src/circuits/halo2/circuit.rs
rrtoledo: Wouldn't it be better to keep mention of `Certificate` since we will eventually have several circuits (recursion)?

mithril-stm/src/circuits/halo2/golden/helpers.rs
```diff
@@ -0,0 +1,398 @@
...
+pub(crate) struct CircuitEnv {
+    /// Structured reference string used by the Halo2/KZG proving system.
+    /// Generated or loaded once per test case.
+    srs: ParamsKZG<Bls12>,
+
+    /// The Circuit relation defining all constraints enforced in-circuit.
+    relation: Circuit,
+
+    /// Verification key corresponding to `relation` and `srs`.
+    vk: MidnightVK,
```
rrtoledo, line 62: You should be able to recompute `vk` and `pk` directly from the `Circuit` but you might want to reduce the testing time? In which case, you could read/write the keys given a circuit (hash).

mithril-stm/src/circuits/halo2/golden/support.rs
```diff
@@ -0,0 +1,398 @@
...
+
+/// Prove and verify a scenario, returning a typed result for negative tests.
+pub(crate) fn prove_and_verify_result(
+    env: &CircuitEnv,
+    scenario: CircuitScenario,
+) -> Result<(), CircuitProofError> {
+    let instance = (scenario.merkle_root, scenario.msg);
+
+    let start = Instant::now();
+    let proof = zk::prove::<Circuit, blake2b_simd::State>(
```
rrtoledo: I am not sure we need to call `zk::prove`, if we want only to assert that a bad witness will lead to a proof generation failure, `MockProver` should be sufficient and faster.

mithril-stm/src/circuits/halo2/circuit.rs
rrtoledo: This could be done once we have the recursion actually.

rrtoledo: Closing this comment.

mithril-stm/src/circuits/halo2/golden/support.rs
rrtoledo: Likewise, when you do need to use the prover, we could perhaps directly read a proof from file (but I think with the `k` value you set, proof generation should be very quick anyway).

mithril-stm/src/circuits/halo2/circuit.rs
hjeljeli32: @jpraynaud,
@kitounliu confirmed that k and m are always small values, so switching them from u32 to u64 should be technically safe and should not affect the circuit size or constraints.

That said, since this PR is focused on stabilizing the golden vectors, I was leaning towards not touching the circuit types yet and doing the u32 → u64 alignment in a follow-up PR.
If you prefer, I can also update them to u64 already in this PR, assuming we agree it’s safe from a constraint point of view.

Let me know what you think.

jpraynaud: @hjeljeli32 I agree this PR should stick to not modifying the code and just create the golden tests. We can align the types later during the refactoring phase (if this is necessary given the fact that `k` and `m` are small values).

mithril-stm/src/circuits/halo2/golden/support.rs
hjeljeli32: @jpraynaud @rrtoledo 
I think the suggestion to use MockProver is indeed valid to optimize our CI's running time. I lean towards implementing this in a separate ticket in order to advance on the integration/refactor of the circuit. Please let me know if you think differently.

jpraynaud: Agreed @hjeljeli32, we'll switch to MockProver when possible in another PR :+1:

mithril-stm/src/circuits/halo2/golden/mod.rs
```diff
@@ -0,0 +1,9 @@
+//! Golden tests lock in Halo2 STM circuit behavior for safe refactors and hardening.
+//! cases.rs holds the test cases; support.rs provides shared helpers and witness builders.
+//! Includes baseline plus positive and negative scenarios.
+
+#[cfg(test)]
+pub(crate) mod support;
+
+#[cfg(test)]
+pub(crate) mod cases;
```
jpraynaud: The `#[cfg(test)]` can be set only at the upper level IMO:
```suggestion
pub(crate) mod support;
pub(crate) mod cases;
```

mithril-stm/src/circuits/halo2/mod.rs
```diff
@@ -2,6 +2,8 @@
 
 pub mod circuit;
 pub mod constants;
+pub mod gadgets;
+pub(crate) mod golden;
```
jpraynaud:
```suggestion
#[cfg(test)]
pub(crate) mod golden;
```

mithril-stm/src/circuits/halo2/golden/cases.rs
```diff
@@ -0,0 +1,759 @@
...
+/// Baseline: valid witness; expected to succeed.
+#[test]
+fn baseline() {
+    const K: u32 = 13;
+    const QUORUM: u32 = 3;
+    run_stm_circuit_case_default("small", K, QUORUM);
+}
+
+/// Sanity: larger valid witness; expected to succeed.
+/// Expensive; excluded from CI and intended for manual runs.
```
jpraynaud:
```suggestion
// Sanity: larger valid witness; expected to succeed.
// Expensive; excluded from CI and intended for manual runs.
```

mithril-stm/src/circuits/halo2/golden/cases.rs
```diff
@@ -0,0 +1,759 @@
...
+#[ignore]
+#[test]
+fn medium() {
+    const K: u32 = 16;
+    const QUORUM: u32 = 32;
+    run_stm_circuit_case_default("medium", K, QUORUM);
+}
+
+/// Very large valid witness; expected to succeed.
+/// Extremely expensive; excluded from CI and intended for manual runs.
```
jpraynaud:
```suggestion
// Very large valid witness; expected to succeed.
// Extremely expensive; excluded from CI and intended for manual runs.
```

mithril-stm/src/circuits/halo2/golden/cases.rs
```diff
@@ -0,0 +1,759 @@
...
+#[test]
+fn large() {
+    const K: u32 = 21;
+    const QUORUM: u32 = 1024;
+    run_stm_circuit_case_default("large", K, QUORUM);
+}
+
+/// Public message is zero; expected to succeed.
+#[test]
+fn msg_0() {
```
jpraynaud: These comments are not needed, instead the names of the test should be explicit enough to express the same information (applies here and in all the other tests:
```suggestion
#[test]
fn succeeds_when_messsage_is_zero() {
```

mithril-stm/src/circuits/halo2/golden/cases.rs
```diff
@@ -0,0 +1,759 @@
...
+    const K: u32 = 13;
+    const QUORUM: u32 = 3;
+    run_stm_circuit_case("msg_0", K, QUORUM, JubjubBase::ZERO);
```
jpraynaud: You could use a macro to automate this:
```suggestion
    run_stm_circuit_case(current_function!(), K, QUORUM, JubjubBase::ZERO);
```

And this macro:
```rust
#[macro_export]
macro_rules! current_function {
    () => {{
        fn f() {}
        let name = $crate::test::format_current_function_module(f);
        // The index found is the beginning of the '..', this is why we add 2.
        let function_name_index = name.rfind("::").map(|index| index + 2).unwrap_or(0);

        &name[function_name_index..]
    }};
}
```

mithril-stm/src/circuits/halo2/golden/cases.rs
```diff
@@ -0,0 +1,759 @@
...
+    let env = setup_stm_circuit_env("min_strict_indices", K, QUORUM);
+    let (sks, leaves, merkle_tree) = create_default_merkle_tree(env.num_signers());
+
+    let merkle_root = merkle_tree.root();
+    let indices = vec![0, 1, 2];
+    let witness =
+        build_witness_with_indices(&sks, &leaves, &merkle_tree, merkle_root, msg, &indices);
+
+    let scenario = StmCircuitScenario::new(merkle_root, msg, witness);
+    prove_and_verify_result(&env, scenario).expect("Proof generation/verification failed");
```
jpraynaud: You could follow an "Act Arrange Assert" pattern by grouping lines of the same part (applies to all tests):
```suggestion
    const K: u32 = 13;
    const QUORUM: u32 = 3;
    let msg = JubjubBase::from(42);
    let env = setup_stm_circuit_env("min_strict_indices", K, QUORUM);
    let (sks, leaves, merkle_tree) = create_default_merkle_tree(env.num_signers());

    let merkle_root = merkle_tree.root();
    let indices = vec![0, 1, 2];
    let witness =
        build_witness_with_indices(&sks, &leaves, &merkle_tree, merkle_root, msg, &indices);

    let scenario = StmCircuitScenario::new(merkle_root, msg, witness);
    prove_and_verify_result(&env, scenario).expect("Proof generation/verification failed");
```

mithril-stm/src/circuits/halo2/golden/cases/negative.rs
jpraynaud, line 1: Maybe you could split the test cases in different sub-modules of `golden` module (each one with its own file) to make them more readable.

mithril-stm/src/circuits/halo2/golden/helpers.rs
jpraynaud, line 1: This module could be renamed `helpers.rs` IMO.

mithril-stm/src/circuits/halo2/golden/helpers.rs
```diff
@@ -0,0 +1,432 @@
...
+
+    let merkle_root = merkle_tree.root();
+
+    let witness = build_witness(&sks, &leaves, &merkle_tree, merkle_root, msg, quorum);
+    let scenario = StmCircuitScenario::new(merkle_root, msg, witness);
+
+    prove_and_verify_result(&env, scenario).expect("Proof generation/verification failed");
+}
+
+// Load cached KZG params if present; otherwise generate and persist them for reuse.
```
jpraynaud, line 391:
```suggestion
/// Load cached KZG params if present; otherwise generate and persist them for reuse.
```

mithril-stm/src/circuits/halo2/golden/support.rs
```diff
@@ -0,0 +1,432 @@
...
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+struct VkPkKey {
+    k: u32,
+    quorum: u32,
+    num_lotteries: u32,
+    merkle_tree_depth: u32,
+}
+
+static VKPK_CACHE: OnceLock<Mutex<VkPkCache>> = OnceLock::new();
```
jpraynaud: It'd be better to use a `LazyLock` which would avoid the use a `Mutex` (and to put this static var in the function that uses it).

Also names should be more explicit: `VKPK_CACHE` is not clear enough.

mithril-stm/src/circuits/halo2/golden/support.rs
```diff
@@ -0,0 +1,432 @@
...
+    Signature, SigningKey, VerificationKey,
+};
+use crate::circuits::halo2::types::{Bls12, JubjubBase};
+use crate::circuits::test_utils::setup::{generate_params, load_params};
+
+type F = JubjubBase;
+
+type WitnessEntry = (MTLeaf, MerklePath, Signature, u32);
+type VkPkPair = (MidnightVK, MidnightPK<StmCircuit>);
+type VkPkCache = HashMap<VkPkKey, Arc<VkPkPair>>;
```
jpraynaud: All these abbreviations make all the code less readable: e.g. you should prefer `CircuitVerificationAndPrivateKeys` to `VkPkKey`. Also the suffix `Key` is ambiguous (I guess it is representing the key in a map, but it can easily be confused with a cryptographic key. If it represents a key in a map you can just remove it).

mithril-stm/src/circuits/halo2/golden/support.rs
```diff
@@ -0,0 +1,432 @@
...
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+struct VkPkKey {
```
jpraynaud: Could you add a doc comment for all the structs, enums, static vars, functions?

mithril-stm/src/circuits/halo2/golden/support.rs
```diff
@@ -0,0 +1,432 @@
...
+        leaves.push(leaf);
+    }
+
+    let tree = MerkleTree::create(&leaves);
+
+    (sks, leaves, tree, leftmost_index)
+}
+
+/// Build a full tree with a controlled leaf target at a specific index.
+pub(crate) fn create_merkle_tree_with_controlled_leaf(
```
jpraynaud: All the Merkle tree functions are doing almost the same job. Also we are making assumptions on the internal representation of the Merkle tree. This part should be refactored IMO. Maybe the Merkle tree implementation lacks a function to retrieve a leaf by index or by position (e.g. an enum with variants for `LeftMost`, `RightMost`, `Index(usize)`)

mithril-stm/src/circuits/halo2/golden/support.rs
```diff
@@ -0,0 +1,432 @@
...
+    /// Proof verification failed.
+    #[error("proof verification failed")]
+    VerifyFail,
+}
+
+/// Prove and verify a scenario, returning a typed result for negative tests.
+pub(crate) fn prove_and_verify_result(
+    env: &StmCircuitEnv,
+    scenario: StmCircuitScenario,
+) -> Result<(), StmCircuitProofError> {
```
jpraynaud:
```suggestion
) -> StmResult<()> {
```

mithril-stm/src/circuits/halo2/golden/cases.rs
```diff
@@ -0,0 +1,759 @@
...
+            if witness[i].0.to_bytes() != witness[j].0.to_bytes() {
+                mismatch_idx = Some((i, j));
+                break;
+            }
+        }
+        if mismatch_idx.is_some() {
+            break;
+        }
+    }
+    let (i, j) = mismatch_idx.expect("expected at least two distinct leaves in witness");
```
jpraynaud: We control the creation of the witness so we could pick 2 of them and make sure they are different. This would avoid this code to search for them. 

Also this code is not very functional and could be at least extracted in another function to avoid code duplication.

mithril-stm/src/circuits/halo2/golden/cases.rs
```diff
@@ -0,0 +1,759 @@
...
+    let indices = vec![6, 14, 22];
+    assert!(indices.iter().all(|i| *i < m));
+
+    let mut witness =
+        build_witness_with_indices(&sks, &leaves, &merkle_tree, merkle_root, msg, &indices);
+
+    let d = 1usize;
+    assert!(d < witness[0].1.siblings.len());
+    witness[0].1.siblings[d].0 = match witness[0].1.siblings[d].0 {
+        crate::circuits::halo2::off_circuit::merkle_tree::Position::Left => {
```
jpraynaud: Could you use imports for these?

mithril-stm/src/circuits/halo2/golden/cases.rs
hjeljeli32: Thanks for the comment @jpraynaud  👍
Just to add a bit of context: I initially chose the compact test namesto keep the tables in the MVP document readable; with longer, fully descriptive names, the tables didn’t render very well. To compensate for that, I added short comments above the tests (with the expected outcome), following a suggestion from @curiecrypt .

That said, I agree with you that we can also encode the expected behavior directly in the test names themselves if we prefer to avoid comments. 
I’m fine with either approach. Happy to go with whatever best fits the project conventions.

jpraynaud: Thanks @hjeljeli32 , I'd rather go with the expressive test names even though the tables of the MVP book could be less readable :+1:

mithril-stm/src/circuits/halo2/golden/cases/cases_negative.rs
```diff
@@ -0,0 +1,482 @@
...
+    let msg = JubjubBase::from(42);
+    let env = setup_stm_circuit_env(current_function!(), K, QUORUM);
+    let (sks, leaves, merkle_tree) = create_default_merkle_tree(env.num_signers());
+    let merkle_root = merkle_tree.root();
+    let mut witness = build_witness(&sks, &leaves, &merkle_tree, merkle_root, msg, QUORUM);
+
+    witness.pop();
+
+    let scenario = StmCircuitScenario::new(merkle_root, msg, witness);
+    let _ = prove_and_verify_result(&env, scenario);
```
jpraynaud: The result should be checked otherwise the test will pass even if `preve_and_verify` returns an error (applies to few other tests):
```suggestion
    prove_and_verify_result(&env, scenario).unwrap();
```

mithril-stm/src/circuits/halo2/golden/cases/cases_negative.rs
```diff
@@ -0,0 +1,482 @@
...
+    assert!(d < witness[0].1.siblings.len());
+    witness[0].1.siblings[d].1 += JubjubBase::ONE;
+
+    let scenario = StmCircuitScenario::new(merkle_root, msg, witness);
+    let result = prove_and_verify_result(&env, scenario);
+    assert!(matches!(result, Err(StmCircuitProofError::VerifyFail)));
+}
+
+#[test]
+fn mp_flip_pos() {
```
jpraynaud: Could you avoid any abbreviation in the tests names as it prevents readability (applies to other tests):
```suggestion
fn merkle_path_flip_position_fails() {
```

mithril-stm/src/circuits/halo2/golden/cases/negative.rs
jpraynaud, line 1: Maybe naming the modules `negative.rs` and `positive.rs` is enough as they are already in the `cases` module.

mithril-stm/src/circuits/halo2/golden/cases/mod.rs
```diff
@@ -0,0 +1,8 @@
+// Positive cases are expected to succeed; negative cases are expected to fail verification or panic.
+// Add new cases in the appropriate module to keep expectations explicit.
```
jpraynaud: This comment can be removed.

mithril-stm/src/circuits/halo2/golden/helpers.rs
```diff
@@ -0,0 +1,417 @@
...
+        leaves.push(MTLeaf(vk, -F::ONE));
+        sks.push(sk);
+    }
+    let tree = MerkleTree::create(&leaves);
+
+    (sks, leaves, tree)
+}
+
+/// Build a full tree with one controlled leaf selected by `selector` and return its index.
+pub(crate) fn create_merkle_tree_with_leaf_selector(
```
jpraynaud, line 132: This function should not be needed and duplicates a lot of code. Also, this is more the responsibility of the Merkle tree to provide leaves by position. This means that we will ultimately need one function that creates the Merkle tree and then ask the Merkle tree itself to provide the leaf for a given position.

Let's implement that in the off-circuit removal PR.

mithril-stm/src/circuits/halo2/golden/helpers.rs
```diff
@@ -0,0 +1,417 @@
...
+            merkle_root,
+            msg,
+            witness,
+        }
+    }
+}
+
+/// Build a default Merkle tree with all leaves set to the max target.
+pub(crate) fn create_default_merkle_tree(n: usize) -> (Vec<SigningKey>, Vec<MTLeaf>, MerkleTree) {
+    let mut rng = OsRng;
```
curiecrypt: Here, instead of `OsRng`, wouldn't it better to use seeded rng to make it deterministic for reproducibilty?

mithril-stm/src/circuits/halo2/golden/helpers.rs
```diff
@@ -0,0 +1,417 @@
...
+    let selected_index = match selector {
+        LeafSelector::LeftMost => 0usize,
+        LeafSelector::RightMost => n - 1,
+        LeafSelector::Index(i) => {
+            assert!(i < n, "leaf_index out of bounds for tree size");
+            i
+        }
+    };
+
+    let mut rng = OsRng;
```
curiecrypt: Similarly, I'd prefer seeded rng.

mithril-stm/src/circuits/halo2/golden/helpers.rs
```diff
@@ -0,0 +1,417 @@
...
+    let tree = MerkleTree::create(&leaves);
+
+    (sks, leaves, tree, selected_index)
+}
+
+/// Construct the STM circuit relation environment and setup keys for a case.
+pub(crate) fn setup_stm_circuit_env(case_name: &str, k: u32, quorum: u32) -> StmCircuitEnv {
+    let srs = load_or_generate_params(k);
+
+    let num_signers: usize = 3000;
```
curiecrypt: Might be defined as a constant.

mithril-stm/src/circuits/halo2/golden/helpers.rs
```diff
@@ -0,0 +1,417 @@
...
+    let depth = num_signers.next_power_of_two().trailing_zeros();
+    let num_lotteries = quorum * 10;
```
curiecrypt: This might be too.

mithril-stm/src/circuits/halo2/golden/cases/macros.rs
```diff
@@ -0,0 +1,13 @@
+macro_rules! current_function {
```
curiecrypt, line 7: We might want to document this unstable string manipulation in the comment.

mithril-stm/src/circuits/halo2/golden/cases/negative.rs
```diff
@@ -0,0 +1,482 @@
...
+
+#[test]
+fn target_lt_ev() {
+    const K: u32 = 13;
+    const QUORUM: u32 = 3;
+    let msg = JubjubBase::from(42);
+    let env = setup_stm_circuit_env(current_function!(), K, QUORUM);
+    let depth = env.num_signers().next_power_of_two().trailing_zeros();
+    let signer_index = 0usize;
+    let (sks, leaves, merkle_tree, _) = create_merkle_tree_with_leaf_selector(
```
curiecrypt, line 372: I think this test and `target_too_small` are testing the exact same scenario.
Both have `target = 0` at `signer_index = 0`, quorum, k, and indices are also the same. 
I might be missing something, is it on purpose?

hjeljeli32:
Thanks @jpraynaud  for the review! I’ve addressed your comments as follows:
1. Removed the top comment in golden/mod.rs.
2. Renamed the case modules to positive.rs and negative.rs.
3. Renamed test functions to remove abbreviations and make names more explicit.
4. For the tests you pointed out where results were ignored: these are currently panicking tests. Initially the result was ignored because the panic happens before verification. After your comment, I updated them to assert a verification failure, so once the panics are fixed later, the tests should work without further changes.
5. For create_merkle_tree_with_leaf_selector: acknowledged. Let's address this as part of the off-circuit removal PR.

hjeljeli32:
@curiecrypt Thank you for the detailed and thoughtful review 🙏
I addressed your comments as follows:
1. Reproducibility: I replaced the unseeded RNG usage in the golden helpers with a deterministic seeded RNG (ChaCha20Rng), aligned with what is already used in other parts of mithril-stm.
2. Constants: I extracted several hard-coded values into named constants to improve readability. 
3. Macro documentation: I added a comment documenting the unstable string manipulation in the macro. I did not see a straightforward stable alternative at the moment.
4. Duplicate test: Good catch. The two lottery-failure tests were covering the same scenario, so I removed one.

Regarding readability, I agree that the test suite is hard to follow given the number of scenarios. I think that reading the MVP book in parallel helps to clarify the intent of the tests. How about adding a small Markdown document in the repo, mirroring the MVP tables? I would be very happy to hear any other suggestions you might have to further improve readability.

Issue #2770, opened 2025-11-06
Follower aggregator panics in e2e tests in `run-only` mode


jpraynaud:
## Why
When running the e2e test in `run-only` mode, the follower aggregators fail after few minutes with the same error:
```
Error: task 27 panicked with message "FOREIGN KEY constraint failed (code 19)"

Stack backtrace:
   0: anyhow::error::<impl core::convert::From<E> for anyhow::Error>::from
   1: mithril_aggregator::commands::serve_command::ServeCommand::execute::{{closure}}
   2: tokio::runtime::park::CachedParkThread::block_on
   3: tokio::runtime::context::runtime::enter_runtime
   4: tokio::runtime::runtime::Runtime::block_on
   5: mithril_aggregator::main
   6: std::sys::backtrace::__rust_begin_short_backtrace
   7: std::rt::lang_start::{{closure}}
   8: std::rt::lang_start_internal
   9: main
  10: __libc_start_call_main
             at ./csu/../sysdeps/nptl/libc_start_call_main.h:58:16
  11: __libc_start_main_impl
             at ./csu/../csu/libc-start.c:360:3
  12: _start
```

The e2e test is started with the following parameters:
```
--skip-signature-delayer --number-of-aggregators=3 --number-of-signers=4 --use-relays --relay-signer-registration-mode=passthrough --relay-signature-registration-mode=p2p --aggregate-signature-type=Concatenation  --use-dmq --dmq-node-flavor=haskell --run-only --cardano-slot-length 0.25 --cardano-epoch-length 30
```

### First analysis
The problem occurs when using the DMQ node (and thus the signature processor). 
The scenario could be the following:
- An individual signature is being created for an open message which gets deleted before the insertion of the signature
- This leads to the foreign key error
- This error is triggering a panic of the `select!` call in the `run` method of the `SequentialSignatureProcessor`
- Maybe adding an `else` branch to the `select!` call

Reference of cases where `select!` panics:
```
/// The `select!` macro panics if all branches are disabled **and** there is no
/// provided `else` branch. A branch is disabled when the provided `if`
/// precondition returns `false` **or** when the pattern does not match the
/// result of `<async expression>`.
```

## What
Investigate the source of the problem and create a fix.

## How

- [ ] Investigate the source of the error
- [ ] Create a fix

PR #2977, opened 2026-02-02
feat: implement a DMQ consumer deduplicator

jpraynaud:
## Content

This PR includes **a deduplicator decorator for the DMQ consumer client that filters out duplicate messages**.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Issue(s)
Relates to #2770

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3887,7 +3887,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-aggregator"
-version = "0.8.17"
+version = "0.8.18"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -4224,7 +4224,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-dmq"
-version = "0.1.21"
+version = "0.1.22"
 dependencies = [
  "anyhow",
  "async-trait",
--- a/internal/mithril-dmq/Cargo.toml
+++ b/internal/mithril-dmq/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "mithril-dmq"
 description = "Mechanisms to publish and consume messages of a 'Decentralized Message Queue network' through a DMQ node"
-version = "0.1.21"
+version = "0.1.22"
 authors.workspace = true
 documentation.workspace = true
 edition.workspace = true
--- a/mithril-aggregator/Cargo.toml
+++ b/mithril-aggregator/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mithril-aggregator"
-version = "0.8.17"
+version = "0.8.18"
 description = "A Mithril Aggregator server"
 authors = { workspace = true }
 edition = { workspace = true }


internal/mithril-dmq/src/consumer/client/deduplicator.rs
```diff
@@ -0,0 +1,297 @@
...
+
+    /// Removes expired messages from the seen messages cache.
+    async fn remove_expired_messages(&self, current_timestamp: u64) {
+        let mut seen_messages = self.seen_messages.lock().await;
+        seen_messages
+            .retain(|_, timestamp| !self.is_expired_timestamp(*timestamp, current_timestamp));
+    }
+
+    /// Checks if a message has already been seen.
+    async fn message_has_been_seen(&self, message: &M) -> StdResult<bool> {
```
turmelclem: does `has_message_been_seen` better reflect the conditional aspect of the result?

internal/mithril-dmq/src/consumer/client/deduplicator.rs
```diff
@@ -0,0 +1,297 @@
...
+        let current_timestamp = self.current_timestamp()?;
+        self.remove_expired_messages(current_timestamp).await;
+        let messages = messages?;
+
+        let mut deduplicated_messages = Vec::new();
+        for (message, party_id) in messages {
+            if !self.message_has_been_seen(&message).await? {
+                self.mark_message_as_seen(message.clone(), current_timestamp).await?;
+                deduplicated_messages.push((message, party_id));
+            }
```
turmelclem, line 142: Is there a explanation why we can have duplicated message ? would it be interesting to keep a trace/log of duplicated messages ?

internal/mithril-dmq/src/consumer/client/deduplicator.rs
```diff
@@ -0,0 +1,297 @@
...
+            vec![
+                Ok(vec![(message_1.clone(), "party-1".to_string())]),
+                Ok(vec![(message_2.clone(), "party-2".to_string())]),
+            ],
+            1000,
+            Duration::from_secs(100),
+        );
+
+        deduplicator.consume_messages().await.unwrap();
+        assert_eq!(1, deduplicator.seen_messages_count().await);
```
turmelclem, line 272: why only one message has been seen and not two ?

internal/mithril-dmq/src/consumer/client/deduplicator.rs
jpraynaud: Fixed

internal/mithril-dmq/src/consumer/client/deduplicator.rs
jpraynaud: Yes, this is explained in the issue comments: when the aggregator disconnects from the DMQ client, it will fetch all of the messages in the mempool of the DMQ server, which will make the aggregator re-import all of the messages once again. This would create a lot of extra logs in the aggregator, which would not be there any more with the deduplicator.

internal/mithril-dmq/src/consumer/client/deduplicator.rs
jpraynaud: Because the inner results feed the sequence of responses of the `DmqConsumerFake` and the first element in that test has only one message (`vec![(message_1.clone(), "party-1".to_string())]`).

PR #2979, opened 2026-02-03
feat(stm): Computation of the target value for the SNARK lottery

damrobi:
## Content

This PR includes the computation of the lottery target for SNARK with arbitrary precision. This arbitrary precision is necessary to make sure no security is lost during the computation. The implementation uses the `BigInt` types from `num-integer` to make sure the results are stable across plateforms. 

The tests included in this PR concern specifically the `num-integer` backend as it is impossible to get matching values for target computation between the `num-integer` and the `rug` backends. The tests should cover the following properties:
Given a large total stake `total_stake = 45,000,000,000` and a fixed value `phi_f = 0.05`:
- [x] golden tests to ensure the results are consistent across configurations and OS
  - [x] For large stakes up to `total_stake` (10 highest possible stakes)
  - [x] For minimum participation stakes (around 150,000) (10 values after 100,000)
  - [x] For very small stakes down to 0 (0,1 and 2)
- [x] Make sure computing the target of a given stake several times gives the same result
  - [x] For large stakes up to `total_stake`
  - [x] For very small stakes down to 0
- [x] Make sure the order is preserved between stake value and target value (`s1 < s2 <=> t1 < t2`)
  - [x] For large stakes up to `total_stake`
  - [x] For minimum participation stakes (around 150,000)
  - [x] For very small stakes down to 0
- [x] Make sure no advantage can be gain by using smaller stakes
- [x] Add property based test to check more possible values for `stake`, `total_stake` and `phi_f`
- [ ] Once the solution is fixed, run some benchmarks

Main results:
- After running the CI with golden tests, they seem to pass for Ubuntu, macOS and Windows.
- [Here is a document](https://hackmd.io/@ME2_HxZuTma6JS_01Hd_oQ/HJjlNIzwZe/edit) summarizing the computation and giving more numbers

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Issue(s)

Relates to #2974

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4461,7 +4461,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-stm"
-version = "0.9.16"
+version = "0.9.17"
 dependencies = [
  "anyhow",
  "blake2 0.10.6",
--- a/mithril-stm/CHANGELOG.md
+++ b/mithril-stm/CHANGELOG.md
@@ -5,6 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## 0.9.17 (02-19-2026)
+
+### Added
+
+- Added the computation of the lottery target value for SNARK.
+- Switched back-end for computation from rug to num-integer.
+
 ## 0.9.16 (02-18-2026)
 
 ### Changed
--- a/mithril-stm/Cargo.toml
+++ b/mithril-stm/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mithril-stm"
-version = "0.9.16"
+version = "0.9.17"
 edition = { workspace = true }
 authors = { workspace = true }
 homepage = { workspace = true }
@@ -45,7 +45,6 @@ midnight-proofs = { version = "0.7.0", optional = true }
 midnight-zk-stdlib = { version = "1.0.0", optional = true }
 num-bigint = { version = "0.4.6", optional = true }
 num-integer = { version = "0.1", optional = true }
-num-rational = { version = "0.4.2", optional = true }
 num-traits = { version = "0.2.19", optional = true }
 rand = { version = "0.9.2", optional = true }
 rand_core = { workspace = true, features = ["std"] }


mithril-stm/src/proof_system/halo2_snark/eligibility.rs
```diff
@@ -213,4 +313,182 @@ mod tests {
...
+        let total_stake = 10000000;
+
+        let mut prev_target = compute_target_value(phi_f, 99, total_stake);
+        for stake in 1000000..1000100 {
+            let target = compute_target_value(phi_f, stake, total_stake);
+            assert!(prev_target < target);
+            prev_target = target;
+        }
+    }
+
```
kitounliu, line 479: we might also need a test to compute the same target twice and check the consistency

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
```diff
@@ -1,30 +1,59 @@
...
-        let c =
-            Ratio::from_float((1.0 - phi_f).ln()).expect("Only fails if the float is infinite or NaN.");
         let w = Ratio::new_raw(BigInt::from(stake), BigInt::from(total_stake));
+        let c =
+            Ratio::from_float((-phi_f).ln_1p()).expect("Only fails if the float is infinite or NaN.");
 
-        let phi = (w * c).neg();
-        let t = Ratio::from(modulus) * phi;
+        // With Taylor series 2
+        let exp_wc = compute_exp(c.clone(), w.clone(), 50);
```
rrtoledo: You did not precise where `50` comes from.

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
```diff
@@ -213,4 +275,283 @@ mod tests {
...
+    }
+
+    #[cfg(feature = "future_snark")]
+    #[cfg(any(feature = "num-integer-backend", target_family = "wasm", windows))]
+    #[test]
+    fn test_greatest_stake_difference() {
+        let phi_f = 0.05;
+        let total_stake = 45_000_000_000;
+
+        let first_target = compute_target_value(phi_f, 1, total_stake);
```
rrtoledo: You should run this on random values for the stake.

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
```diff
@@ -213,4 +275,283 @@ mod tests {
...
+    }
+
+    #[cfg(feature = "future_snark")]
+    #[cfg(any(feature = "num-integer-backend", target_family = "wasm", windows))]
+    #[test]
+    fn test_minimal_stake_difference() {
+        let phi_f = 0.05;
+        let total_stake = 45_000_000_000;
+        for _ in 0..100 {
+            let zero_target = compute_target_value(phi_f, 0, total_stake);
```
rrtoledo: You should put a randon offset in front of 0,1, and 2

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
```diff
@@ -213,4 +275,283 @@ mod tests {
...
+            let first_target = compute_target_value(phi_f, 1, total_stake);
+            assert!(zero_target < first_target);
+            let second_target = compute_target_value(phi_f, 2, total_stake);
+            assert!(first_target < second_target);
+        }
+    }
+
+    #[cfg(any(feature = "num-integer-backend", target_family = "wasm", windows))]
+    #[test]
+    fn test_zero_stake() {
```
rrtoledo: Why not merging this test with `test_zero_stake_bytes` ?

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
```diff
@@ -213,4 +275,283 @@ mod tests {
...
+            use midnight_curves::Fq;
+
+            let target = compute_target_value(phi_f, 0, total_stake);
+            assert!(target == BaseFieldElement(Fq::ZERO));
+        }
+    }
+
+    #[cfg(any(feature = "num-integer-backend", target_family = "wasm", windows))]
+    #[test]
+    fn test_full_stake() {
```
rrtoledo: You forgot to assert?

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
```diff
@@ -213,4 +275,283 @@ mod tests {
...
+        }
+    }
+
+    proptest! {
+        #![proptest_config(ProptestConfig::with_cases(50))]
+
+        #[test]
+        #[cfg(any(feature = "num-integer-backend", target_family = "wasm", windows))]
+        #[cfg(feature = "future_snark")]
+        fn following_stake_same_order(
```
rrtoledo: Do you need to previous tests if you have prop tests testing the same thing?

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
```diff
@@ -51,18 +80,28 @@ cfg_rug! {
...
+            // "1",
             16,
         )
         .unwrap();
 
+        // let modulus = Integer::from(1u64<<16);
+
+        // println!("modulus = {:?}", modulus);
+
         let w = Float::with_val(117, stake) / Float::with_val(117, total_stake);
```
hjeljeli32: @damrobi 
I might be not understanding this function well. My understanding is that it currently contains a Taylor-based num-integer approximation and a rug-based floating reference computation. Is this something expected? needed for the experimental side of this work?

damrobi: This function is the target computation function and has two implementations, one for the `rug` backend and one for the `num-integer` backend. With `rug`, it computes the target directly using high precision floats. With `num-integer`, it uses a Taylor approximation to compute the exponential term. I used the `rug` function at the beginning of my tests as a reference to have an idea of how precise the `num-integer` backend function was. We will not use it in the future.

mithril-stm/src/proof_system/concatenation/eligibility.rs
```diff
@@ -82,6 +82,98 @@ cfg_num_integer! {
...
+        // If exact division and target_as_int > 0, subtract 1
+        if remainder.is_zero() && !target_as_int.is_zero() {
+            target_as_int - 1
+        } else {
+            target_as_int
+        }
+    }
+
+    #[allow(dead_code)]
+    pub fn compute_target_bytes(phi_f: f64, stake: Stake, total_stake: Stake) -> Vec<u8> {
```
kitounliu: after removing rug, this function can take ln(1-phi_f) (instead of phi_f)  as input so that ln(1-phi_f) only needs to be computed once when creating a merkle tree

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
```diff
@@ -1,254 +1,295 @@
...
+        // The number of truncated bits is set at 163 for now to match the 92 bits precision
+        // of the approximations (255 - 92 = 163)
+        let truncated_target: BigInt = (target_as_int >> 163) << 163;
 
-        // If exact division and t_int > 0, subtract 1
-        let target = if remainder.is_zero() && !t_int.is_zero() {
-            t_int - 1
+        // If exact division and truncated_target > 0, subtract 1
+        let target_bigint = if remainder.is_zero() && !truncated_target.is_zero() {
+            truncated_target - 1
```
kitounliu: we need to decide if we want to truncate the target values. If the target is truncated, then we don't need truncated_target - 1 any more

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
```diff
@@ -0,0 +1,719 @@
...
+        // The number of iteration of the exponential approximation is set at 40 for now to get
+        // around 92 bits of precision for a value phi_f=0.2
+        let exp_ln_one_minus_phi_f_stake_ratio = compute_exponential_taylor_expansion(ln_one_minus_phi_f, &stake_ratio, 40);
+
+        let modulus_ratio = Ratio::from(modulus);
+        let target_as_ratio = modulus_ratio.clone() - modulus_ratio * exp_ln_one_minus_phi_f_stake_ratio;
+
+        // Floor division
+        let (target_as_int, remainder) = target_as_ratio.numer().div_rem(target_as_ratio.denom());
+        assert!(target_as_int >= BigInt::zero());
```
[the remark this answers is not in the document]
damrobi: I'm removing this assert as it is impossible for `target_as_int` to be zero as the result of the exponential of a logarithm.

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
```diff
@@ -0,0 +1,715 @@
...
+    #[allow(dead_code)]
+    /// Computes the lottery target value for SNARK proof system as a base field element.
+    ///
+    /// The target value determines the probability of winning the lottery based on the
+    /// participant's stake relative to the total stake. A higher stake results in a higher
+    /// target value, increasing the probability of eligibility.
+    ///
+    /// The target value is computed using the following formula, we need to check that:
+    /// signer_lottery_hash < p * (1 - (1 - phi_f)^w)
+    /// where p is the modulus of the field, phi_f a parameter constant comprised in ]0,1], w a variable in ]0,1]
```
curiecrypt:
```suggestion
    /// where p is the modulus of the field, phi_f is the protocol security parameter constant comprised in ]0,1], w a variable in ]0,1]
```

curiecrypt: Also, we should say that `w = stake / total_stake`

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
```diff
@@ -0,0 +1,715 @@
...
+    /// In order to do that we change the expression:
+    /// 1 - (1 - phi_f)^w = 1 - exp(w * ln(1 - phi_f))
+    /// and we use Taylor expansion to approximate the exponential and natural logarithm functions to a given precision.
+    ///
+    /// Once the precise expression obtained, we can compute the target value as:
+    /// target = floor(p * (1 - exp(w * ln(1 - phi_f))))
+    /// and truncate the unnecessary bits that are below the precision threshold to ensure more stability in the result.
+    ///
+    /// To reduce computation, this function takes ln(1 - phi_f) approximation as input as it is fixed for a given lottery
+    pub fn compute_target_value(ln_one_minus_phi_f: Ratio<BigInt>, stake: Stake, total_stake: Stake) -> LotteryTargetValue {
```
curiecrypt: I think I was not able to be clear in my suggestion. We can handle all functionality in a single function. Meaning that, we can move the flow of `compute_target_bytes` here and get rid of that function.

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
```diff
@@ -0,0 +1,715 @@
...
+    use num_traits::One;
+    use num_rational::Ratio;
+
+    #[cfg(feature = "future_snark")]
+    use crate::{
+        LotteryIndex, LotteryTargetValue, SignatureError, Stake, StmResult, UniqueSchnorrSignature,
+        signature_scheme::{BaseFieldElement, DST_LOTTERY, compute_poseidon_digest},
+    };
+
+    const JUBJUB_BASE_FIELD_MODULUS: &str = "73eda753299d7d483339d80809a1d80553bda402fffe5bfeffffffff00000001";
```
curiecrypt, line 14: Can we use the function `fn modulus<F: PrimeField>() -> BigUint` given in [here](https://github.com/input-output-hk/mithril/pull/3002/changes#r2813442206)?
@hjeljeli32 , @damrobi

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
```diff
@@ -0,0 +1,715 @@
...
+            truncated_target - 1
+        } else {
+            truncated_target
+        };
+
+        let (_, bytes) = target_bigint.to_bytes_le();
+        bytes
+    }
+
+    /// Computes a Taylor expansion of the exponential exp(c*w) up to the (N-1)th term
```
curiecrypt, line 85:
```suggestion
    /// Computes a Taylor expansion of the exponential exp(c*w) up to the (N-1)th term where N corresponds to `iterations` 
```

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
```diff
@@ -0,0 +1,715 @@
...
+    pub fn compute_exponential_taylor_expansion(x: &Ratio<BigInt>, c: &Ratio<BigInt>, iterations: usize) -> Ratio<BigInt> {
+        let cw = c * x;
+        let (num, denom, _) = exponential_approximation(0, iterations, cw.numer(), cw.denom());
+
+        Ratio::new_raw(num, denom)
+    }
+
+
+    /// Function that computes an approximation of exp(a/b) using a binomial splitting
+    /// to compute the taylor expansion terms between i and j,
```
curiecrypt:
```suggestion
    /// to compute the taylor expansion terms between `first_term` and `last_term`,
```
I guess?

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
```diff
@@ -0,0 +1,715 @@
...
+        let auxiliary_value = auxiliary_value_left * auxiliary_value_right;
+
+        // returns the numerator and denominator of the computed terms
+        // as well as an additional value to help with the rest of the computation
+        (numerator, denominator, auxiliary_value)
+    }
+
+    /// Function that computes an approximation of ln(1 - a/b) using a taylor expansion
+    /// for a given number of iterations
+    /// ln(1 - a/b) = -a/b - ((a/b)^2)/2) - ((a/b)^3)/3) - ... - ((a/b)^(N-1))/N-1) + o((a/b)^N)
```
curiecrypt:
```suggestion
    /// ln(1 - a/b) = -a/b - ((a/b)^2)/2) - ((a/b)^3)/3) - ... - ((a/b)^(N-1))/N-1) + o((a/b)^N) where N refers to `iterations`.
```

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
```diff
@@ -0,0 +1,715 @@
...
+    #[allow(dead_code)]
+    fn ln_1p_taylor_expansion(iterations: usize, a: &BigInt, b: &BigInt) -> Ratio<BigInt> {
+        let mut num = a.clone();
+        let mut denom = b.clone();
+        let mut acc = Ratio::new_raw(a.clone(),b.clone());
```
curiecrypt:
```suggestion
        let mut numerator = a.clone();
        let mut denominator = b.clone();
        let mut accumulator = Ratio::new_raw(a.clone(),b.clone());
```
and fix the rest accordingly

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
```diff
@@ -0,0 +1,715 @@
...
+    /// Verifies if a lottery index is eligible based on the signature and target value.
+    ///
+    /// This function checks whether a given index wins the lottery by computing an
+    /// evaluation value from the signature's commitment point and the index, then
+    /// comparing it against the target value. An index is eligible if its
+    /// evaluation value is less than or equal to the target.
+    ///
+    /// The evaluation is computed as: `ev = Poseidon(prefix, sigma_x, sigma_y, index)`
+    /// where `(sigma_x, sigma_y)` are the coordinates of the signature's commitment point.
+    pub fn check_index(
```
curiecrypt: Is this `verify_lottery_eligibility` given in [here](https://hackmd.io/@hmtJxv1LSXqtJkgFZZrdRA/SJ5eE63lbl#Refactor-Halo2-circuit-for-production-readiness:~:text=verify_lottery_eligibility) ?
@hjeljeli32

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
```diff
@@ -0,0 +1,715 @@
...
+        index: LotteryIndex,
+        m: u64,
+        prefix: BaseFieldElement,
+        target: LotteryTargetValue,
+    ) -> StmResult<()> {
+        if index > m {
+            return Err(SignatureError::IndexBoundFailed(index, m).into());
+        }
+
+        let idx = BaseFieldElement::from(index);
```
curiecrypt: why not using `index_as_base_field_element`?

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
```diff
@@ -0,0 +1,715 @@
...
+        let (sigma_x, sigma_y) = signature.commitment_point.get_coordinates();
```
curiecrypt: Why not using
```suggestion
        let (commitment_point_x, commitment_point_y) = signature.commitment_point.get_coordinates();
```

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
```diff
@@ -0,0 +1,715 @@
...
+        let ev = compute_poseidon_digest(&[prefix, sigma_x, sigma_y, idx]);
```
curiecrypt: I think we can use `lottery_evaluation` given [here](https://hackmd.io/@hmtJxv1LSXqtJkgFZZrdRA/SJ5eE63lbl#Refactor-Halo2-circuit-for-production-readiness:~:text=Lottery%20evaluation%20value-,lottery_evaluation,-verify_lottery) instead of `ev`

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
```diff
@@ -0,0 +1,715 @@
...
+        Ok(())
+    }
+
+    #[cfg(feature = "future_snark")]
+    // TODO: remove this allow dead_code directive when function is called or future_snark is activated
+    #[allow(dead_code)]
+    /// Computes the lottery prefix hash from a message.
+    /// The prefix is computed by prepending `DST_LOTTERY`
+    /// to the message and hashing the result using `compute_poseidon_digest`.
+    pub fn lottery_prefix(msg: &[BaseFieldElement]) -> BaseFieldElement {
```
curiecrypt: We should think of a better name for this function.
I think it is not given [here](https://hackmd.io/@hmtJxv1LSXqtJkgFZZrdRA/SJ5eE63lbl#1-Naming) @hjeljeli32 , @damrobi

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
```diff
@@ -0,0 +1,715 @@
...
+#[cfg(feature = "future_snark")]
+#[cfg(test)]
+mod tests {
+    use num_bigint::{BigInt, Sign};
+    use num_rational::Ratio;
+    use proptest::prelude::*;
+    use rand_core::OsRng;
+
+    use crate::{SchnorrSigningKey, signature_scheme::BaseFieldElement};
+
```
curiecrypt, line 214: Should we also import `LotteryTargetValue` defined in `../lib.rs`?

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
```diff
@@ -0,0 +1,715 @@
...
+    use num_rational::Ratio;
+
+    #[cfg(feature = "future_snark")]
+    use crate::{
+        LotteryIndex, LotteryTargetValue, SignatureError, Stake, StmResult, UniqueSchnorrSignature,
+        signature_scheme::{BaseFieldElement, DST_LOTTERY, compute_poseidon_digest},
+    };
+
+    const JUBJUB_BASE_FIELD_MODULUS: &str = "73eda753299d7d483339d80809a1d80553bda402fffe5bfeffffffff00000001";
+    const TAYLOR_EXPANSION_ITERATIONS: usize = 40;  // Provides ~92 bits precision for phi_f=0.2
```
curiecrypt: I think we should give a comprehensive documentation about this precision. 
The valid range of phi_f (phi_f <= 0.2) or making the iteration count dependent on phi_f to guarantee the precision invariant.

Following is what I understand:

The precision of the ln Taylor expansion depends on phi_f. 40 iterations gives: 
- ~100 bits for phi_f = 0.2, 
- ~47 bits for phi_f = 0.5, and 
- ~30 bits for phi_f = 0.65. 

The truncation retains 92 significant bits (255 - 163). 

For production phi_f = 0.2, this works with ~8 bits of headroom, but for phi_f > 0.5 the Taylor approximation itself has less precision than the retained bits, meaning the non-truncated bits contain approximation noise rather than meaningful digits. 

Am I right: @rrtoledo, @kitounliu

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
```diff
@@ -0,0 +1,715 @@
...
+                    let target = BigInt::from_bytes_le(
+                        Sign::Plus,
+                        &compute_target_bytes(&ln_one_minus_phi_f, total_stake - i, total_stake),
+                    );
+                    golden_values.push(target);
+                }
+                golden_values
+            }
+
+            #[test]
```
curiecrypt, line 602: Can you add an explanation for why we use ignore for these tests

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
```diff
@@ -0,0 +1,715 @@
...
+
+    use crate::{SchnorrSigningKey, signature_scheme::BaseFieldElement};
+
+    use super::{
+        check_index, compute_target_bytes, compute_target_value, ln_1p_taylor_expansion,
+        lottery_prefix,
+    };
+
+    #[cfg(test)]
+    mod test_bytes_computation {
```
curiecrypt:
```suggestion
    mod bytes_computation {
```

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
```diff
@@ -0,0 +1,715 @@
...
+        use super::*;
+
+        #[test]
+        fn test_zero_stake_bytes() {
```
curiecrypt: we should rename all tests and omit the prefix `test_`

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
```diff
@@ -0,0 +1,715 @@
...
+            if check_index(&sig, i, m, prefix, target).is_ok() {
+                println!("Index: {}", i);
+                counter += 1;
+            }
+        }
+        println!("Total eligible indices:{:?}", counter);
+    }
+
+    #[cfg(test)]
+    mod tests_eligibility_for_snark {
```
curiecrypt: consider a better name for this mod

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
```diff
@@ -0,0 +1,715 @@
...
+                assert!(golden_target_max_2 != golden_value_target_from_stake(44_999_999_999));
+            }
+
+            #[test]
+            #[ignore]
+            fn golden_check_following_min_stake() {
+                let golden_target_vector = golden_value_following_min_stake();
+
+                let golden_target_from_file = include_str!(
+                    "../../../tests/golden_vector_target_value/golden_vector_min_stake.txt"
```
curiecrypt: I don't like the idea of storing the vectors in a very far directory.

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
```diff
@@ -0,0 +1,715 @@
...
+                    let second_target = BigInt::from_bytes_le(
+                        Sign::Plus,
+                        &compute_target_bytes(&ln_one_minus_phi_f, 2, total_stake),
+                    );
+                    assert!(first_target < second_target);
+                }
+            }
+        }
+
+        mod ordering_tests {
```
curiecrypt: rename it as `oredering` maybe?

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
```diff
@@ -0,0 +1,715 @@
...
+        use super::*;
+
+        #[test]
+        fn test_zero_stake_bytes() {
+            // phi_f = 0.05
+            let phi_f_ratio = Ratio::new_raw(BigInt::from(1), BigInt::from(20));
+            let total_stake = 45_000_000_000;
+
+            for _ in 0..10 {
+                let target = compute_target_bytes(&phi_f_ratio, 0, total_stake);
```
curiecrypt: `compute_target_bytes` expects `ln(1 - phi_f)` as its first argument, but here phi_f itself is passed without the `ln_1p_taylor_expansion` step. 
I guess, the test still passes because stake = 0 makes the first argument irrelevant (`exp(anything * 0) = 1`). 
We should consider using `ln_1p_taylor_expansion` as done in `test_zero_stake_bytes_stable` to keep the test correct by construction, not by coincidence!

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
```diff
@@ -0,0 +1,715 @@
...
+        // The number of truncated bits is set at 163 for now to match the 92 bits precision
+        // of the approximations (255 - 92 = 163)
+        let truncated_target: BigInt = (target_as_int >> TRUNCATION_BITS) << TRUNCATION_BITS;
+
+        // If exact division and truncated_target > 0, subtract 1
+        let target_bigint = if remainder.is_zero() && !truncated_target.is_zero() {
+            truncated_target - 1
+        } else {
+            truncated_target
+        };
```
curiecrypt: I think the subtraction after truncation defeats the purpose of truncation. 

When `truncated_target` has 163 trailing zeros (0b...1010_000...000), subtracting 1 flips all 163 lower bits to ones (0b...1001_111...111). It reintroduces full precision in the bits that were just zeroed for stability. 

If our goal is to keep the target strictly below the exact value while preserving truncation, we should subtract before truncating.

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
curiecrypt: Also, the doc says `exp(c*w)` where `c = ln(1-phi_f)` and `w = stake_ratio`. 

But the call passes `x = ln_one_minus_phi_f` and `c = stake_ratio`.

The names are swapped relative to the doc. 
Of course it does not affect the correctness of the multiplication but it is confusing.

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
kitounliu: The precision/iterations should also depend on the advantage the attacker can gain from splitting stakes: adv = (1-phi(a+b)) - (1-phi(a))(1-phi(b)). By design we should have adv = 0 which means the attacker cannot gain anything from splitting/combining stakes, but the rounding during the calculation could change the value of adv and we need to minimise adv by adjusting precision/number of iterations. Same applies to whether to truncate the target values: adv = (1- target_{a+b}/modulus) - (1-target_a/modulus)(1-target_b/modulus). If a pool is split into n equal pieces, then adv = (1- target_{n*a}/modulus) - (1-target_a/modulus)^n

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
damrobi: I think we can but maybe it would be better to have it in the jubjub module directly? Either way I can update the eligibity one to use it.

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
damrobi: Indeed it is not very clear, maybe something simple like `compute_lottery_prefix` or something a bit more descriptive like `append_lottery_dst_and_hash`?

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
damrobi: I moved them into a golden_vectors repo in proof_system/halo2_snark for now

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
damrobi: That's correct, I removed the - 1

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
```diff
@@ -70,40 +59,42 @@ cfg_num_integer! {
...
         // of the approximations (255 - 92 = 163)
         let truncated_target: BigInt = (target_as_int >> TRUNCATION_BITS) << TRUNCATION_BITS;
 
-        // If exact division and truncated_target > 0, subtract 1
-        let target_bigint = if remainder.is_zero() && !truncated_target.is_zero() {
-            truncated_target - 1
-        } else {
-            truncated_target
-        };
+        // // If exact division and truncated_target > 0, subtract 1
```
curiecrypt: Can you please confirm the correctness of [this](https://github.com/input-output-hk/mithril/pull/2979#discussion_r2813981162)? @kitounliu @rrtoledo 

@damrobi you can completely remove this block if confirmed.

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
```diff
@@ -70,40 +59,42 @@ cfg_num_integer! {
...
     /// Computes a Taylor expansion of the exponential exp(c*w) up to the (N-1)th term
+    /// where N corresponds to the number of iterations
     /// exp(c*x) = 1 + c*x + (c*x)^2/2! + (c*x)^3/3! + ... + (c*x)^{N-1}/(N-1)! + O((c*x)^N)
     /// We want to stop when the next term is less than our precision target, that is epsilon = 2^{-128}
     /// Hence we stop when |(c*x)^N / N!| < epsilon
     /// We can check instead (c * x)^N < epsilon
     /// which gives us the bound N < log(epsilon) / log(|c*x|)
-    pub fn compute_exponential_taylor_expansion(x: &Ratio<BigInt>, c: &Ratio<BigInt>, iterations: usize) -> Ratio<BigInt> {
-        let cw = c * x;
+    pub fn compute_exponential_taylor_expansion(c: &Ratio<BigInt>, w: &Ratio<BigInt>, iterations: usize) -> Ratio<BigInt> {
```
curiecrypt, line 95: Can you fix the doc comment as well: `x -> w`

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
curiecrypt: I think `compute_lottery_prefix` seems reasonable. WDYT @hjeljeli32 @damrobi

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
```diff
@@ -523,20 +420,16 @@ mod tests {
...
+                let target = compute_target_value(&ln_one_minus_phi_f, stake, total_stake);
+                let same_target = compute_target_value(&ln_one_minus_phi_f, stake, total_stake);
                 assert!(target == same_target);
             }
 
         }
 
         #[allow(dead_code)]
+        // Golden tests are ignored for now as the code is not stable, i.e. we need to fix
+        // the precision/number of iterations to get stable value for the golden test
```
curiecrypt: Should we add a `TODO` for this?

mithril-stm/src/proof_system/concatenation/eligibility.rs
```diff
@@ -0,0 +1,162 @@
...
+        ) {
+            let mut ev = [0u8; 64];
+            ev.copy_from_slice(&[&ev_1[..], &ev_2[..]].concat());
+
+            let quick_result = trivial_is_lottery_won(phi_f, ev, stake, total_stake);
+            let result = is_lottery_won(phi_f, ev, stake, total_stake);
+            assert_eq!(quick_result, result);
+        }
+
+        #[cfg(any(feature = "num-integer-backend", target_family = "wasm", windows))]
```
jpraynaud, line 149: You could have used the new `cfg_num_integer!` macro you have created here.

mithril-stm/src/proof_system/concatenation/mod.rs
```diff
@@ -1,11 +1,13 @@
 mod aggregate_key;
 mod clerk;
+mod eligibility;
 mod proof;
 mod signer;
 mod single_signature;
 
 pub use aggregate_key::AggregateVerificationKeyForConcatenation;
 pub use clerk::ConcatenationClerk;
+pub(crate) use eligibility::is_lottery_won;
```
jpraynaud: Maybe this is enough:
```suggestion
pub(super) use eligibility::is_lottery_won;
```

mithril-stm/src/signature_scheme/unique_schnorr_signature/jubjub/field_elements.rs
```diff
@@ -18,6 +18,14 @@ impl BaseFieldElement {
         BaseFieldElement(JubjubBase::ONE)
     }
 
+    #[cfg(feature = "future_snark")]
```
jpraynaud: It looks like this is for tests only:
```suggestion
    #[cfg(all(test, feature = "future_snark"))]
```

mithril-stm/src/signature_scheme/unique_schnorr_signature/jubjub/poseidon_digest.rs
```diff
@@ -11,6 +11,11 @@ const DST_SIGNATURE: JubjubBase = JubjubBase::from_raw([
     0,
 ]);
 
+#[cfg(feature = "future_snark")]
+// TODO: remove this allow dead_code directive when function is called or future_snark is activated
+#[allow(dead_code)]
+pub const DST_LOTTERY: BaseFieldElement = BaseFieldElement(JubjubBase::from_raw([3, 3, 0, 0]));
```
jpraynaud: This constant should have a doc comment.

```suggestion
pub const DOMAIN_SEPARATION_TAG_LOTTERY: BaseFieldElement = BaseFieldElement(JubjubBase::from_raw([3, 3, 0, 0]));
```

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
```diff
@@ -0,0 +1,643 @@
...
+
+    #[cfg(feature = "future_snark")]
+    use crate::{
+        LotteryIndex, LotteryTargetValue, SignatureError, Stake, StmResult, UniqueSchnorrSignature,
+        signature_scheme::{BaseFieldElement, compute_poseidon_digest, DST_LOTTERY},
+    };
+
+    const JUBJUB_BASE_FIELD_MODULUS: &str = "73eda753299d7d483339d80809a1d80553bda402fffe5bfeffffffff00000001";
+    const TAYLOR_EXPANSION_ITERATIONS: usize = 40;  // Provides ~92 bits precision for phi_f=0.2
+    const TRUNCATION_BITS: usize = 163;  // Matches 92 bits precision (255 - 92 = 163)
```
jpraynaud: These constants also deserve their doc comments.

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
```diff
@@ -0,0 +1,643 @@
...
+    ///
+    /// The target value determines the probability of winning the lottery based on the
+    /// participant's stake relative to the total stake. A higher stake results in a higher
+    /// target value, increasing the probability of eligibility.
+    ///
+    /// The target value is computed using the following formula, we need to check that:
+    /// signer_lottery_hash < p * (1 - (1 - phi_f)^w)
+    /// where p is the modulus of the field, phi_f is the protocol security parameter constant comprised in ]0,1],
+    /// w = stake/total_stake
+    /// and signer_lottery_hash is the hash computed using the signer's signature and lottery index.
```
jpraynaud: This is inaccurate as we use the value computed by this function to check won lottery indices.

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
```diff
@@ -0,0 +1,643 @@
...
+    ///
+    /// Since the modulus is a 255 bits number, we need to compute (1 - (1 - phi_f)^w) with enough precision
+    /// to maintain the lottery functional, i.e. have different targets for different stakes
+    /// and maintain the same order, however close they are.
```
jpraynaud, line 40: This lacks an explanation about the necessity to have a deterministic computation IMO.

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
```diff
@@ -0,0 +1,643 @@
...
+    /// 1 - (1 - phi_f)^w = 1 - exp(w * ln(1 - phi_f))
+    /// and we use Taylor expansion to approximate the exponential and natural logarithm functions to a given precision.
+    ///
+    /// Once the precise expression obtained, we can compute the target value as:
+    /// target = floor(p * (1 - exp(w * ln(1 - phi_f))))
+    /// and truncate the unnecessary bits that are below the precision threshold to ensure more stability in the result.
+    ///
+    /// To reduce computation, this function takes ln(1 - phi_f) approximation as input as it is fixed for a given lottery
+    pub fn compute_target_value(ln_one_minus_phi_f: &Ratio<BigInt>, stake: Stake, total_stake: Stake) -> LotteryTargetValue {
+        let modulus = BigInt::from_str_radix(
```
jpraynaud, line 58: Could you add a comment which explains why this is safe to use `expect` here?

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
```diff
@@ -0,0 +1,643 @@
...
+            JUBJUB_BASE_FIELD_MODULUS,
+            16,
+        )
+        .expect("Hardcoded modulus of the Jubjub base field hex string is valid");
+
+        let stake_ratio = Ratio::new_raw(BigInt::from(stake), BigInt::from(total_stake));
+
+        // The number of iteration of the exponential approximation is set at 40 for now to get
+        // around 92 bits of precision for a value phi_f=0.2
```
jpraynaud: This comment should be associated with the `TAYLOR_EXPANSION_ITERATIONS` constant.

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
kitounliu: yes, we can remove the -1 block if the target is already truncated. I originally introduced this to convert the logic of hash < target_float into an equivalent hash <= target_int. 
There shouldn't be any overflow case, i.e., target > = modulus, because phi_f < 1  and target is truncated.

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
```diff
@@ -0,0 +1,643 @@
...
+
+        let modulus_ratio = Ratio::from(modulus);
+        let target_as_ratio = modulus_ratio.clone() - modulus_ratio * exp_ln_one_minus_phi_f_stake_ratio;
+
+        // Floor division
+        let (target_as_int, _) = target_as_ratio.numer().div_rem(target_as_ratio.denom());
+
+        // Truncate the lower bits of the target value
+        // The number of truncated bits is set at 163 for now to match the 92 bits precision
+        // of the approximations (255 - 92 = 163)
```
jpraynaud: Same remark as above.

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
```diff
@@ -0,0 +1,643 @@
...
+        let truncated_target: BigInt = (target_as_int >> TRUNCATION_BITS) << TRUNCATION_BITS;
+
+        // // If exact division and truncated_target > 0, subtract 1
+        // let target_bigint = if remainder.is_zero() && !truncated_target.is_zero() {
+        //     truncated_target - 1
+        // } else {
+        //     truncated_target
+        // };
```
jpraynaud: This should be removed if not needed.

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
```diff
@@ -0,0 +1,643 @@
...
+
+        let (_, mut bytes) = truncated_target.to_bytes_le();
+        bytes.resize(32, 0);
+        BaseFieldElement::from_bytes(&bytes).expect("Input bytes are always lower than the Jubjub modulus hence canonical.")
```
jpraynaud, line 82: This safe `expect` deserves a comment.

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
```diff
@@ -0,0 +1,643 @@
...
+    }
+
+    /// Computes a Taylor expansion of the exponential exp(c*w) up to the (N-1)th term
+    /// where N corresponds to the number of iterations
+    /// exp(c*x) = 1 + c*x + (c*x)^2/2! + (c*x)^3/3! + ... + (c*x)^{N-1}/(N-1)! + O((c*x)^N)
+    /// We want to stop when the next term is less than our precision target, that is epsilon = 2^{-128}
```
jpraynaud: I'm not sure that this is still 128 bits.

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
```diff
@@ -0,0 +1,643 @@
...
+
+        // returns the numerator and denominator of the computed terms
+        // as well as an additional value to help with the rest of the computation
+        (numerator, denominator, auxiliary_value)
+    }
+
+    /// Function that computes an approximation of ln(1 - a/b) using a taylor expansion
+    /// for a given number N of iterations
+    /// ln(1 - a/b) = -a/b - ((a/b)^2)/2) - ((a/b)^3)/3) - ... - ((a/b)^(N))/N) + o((a/b)^(N+1))
+    #[allow(dead_code)]
```
jpraynaud, line 141: Is it really dead code?
```suggestion

```

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
```diff
@@ -0,0 +1,643 @@
...
+                                let ln_one_minus_phi_f =
+                    ln_1p_taylor_expansion(40, phi_f_ratio.numer(), phi_f_ratio.denom());
+                let target = compute_target_value(&ln_one_minus_phi_f, stake, total_stake);
+                let same_target = compute_target_value(&ln_one_minus_phi_f, stake, total_stake);
+                assert!(target == same_target);
+            }
+
+        }
+
+        #[allow(dead_code)]
```
jpraynaud: That's probably a good time to stabilize this:

```suggestion

```

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
```diff
@@ -0,0 +1,643 @@
...
+                // phi_f = 0.05
+                let phi_f_ratio = Ratio::new_raw(BigInt::from(1), BigInt::from(20));
+                let ln_one_minus_phi_f =
+                    ln_1p_taylor_expansion(40, phi_f_ratio.numer(), phi_f_ratio.denom());
+
+                let total_stake = 45_000_000_000;
+
+                for _ in 0..10 {
+                    let target = compute_target_value(&ln_one_minus_phi_f, 0, total_stake);
+                    assert!(target == BaseFieldElement::from(0));
```
jpraynaud: You can fix that for all the concerned tests below:
```suggestion
                    assert_eq!(target, BaseFieldElement::from(0));
```

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
```diff
@@ -0,0 +1,643 @@
...
+                for i in 0..50 {
+                    let target =
+                        compute_target_value(&ln_one_minus_phi_f, total_stake - i, total_stake);
+                    golden_values.push(target);
+                }
+                golden_values
+            }
+
+            #[test]
+            #[ignore]
```
jpraynaud: This is also good time to stabilize this one and the others below:
```suggestion

```

mithril-stm/src/proof_system/concatenation/eligibility.rs
damrobi: I'm not sure why but the call to the macro creates a conflict with proptest

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
damrobi: Is it okay if I replace it by `signer_lottery_hash = H(DST_LOTTERY || merkle_root || msg || commitment_point || index)`?

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
damrobi: It is for now as I don't use it directly in the computation of the target anymore. It should be used before the calls to the `compute_target_value` to give the result as an input.

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
jpraynaud: The inaccurate part is ` and lottery index`.

mithril-stm/src/proof_system/concatenation/eligibility.rs
```diff
@@ -0,0 +1,162 @@
...
+        if (phi_f - 1.0).abs() < f64::EPSILON {
+            return true;
+        }
+
+        let ev_max = BigInt::from(2u8).pow(512);
+        let ev = BigInt::from_bytes_le(Sign::Plus, &ev);
+        let q = Ratio::new_raw(ev_max.clone(), ev_max - ev);
+
+        let c =
+            Ratio::from_float((1.0 - phi_f).ln()).expect("Only fails if the float is infinite or NaN.");
```
pyrros, line 43: Do we have a precision / error bound on the ln evaluation?

damrobi: According to the rust type documentation for f64 it's [undeterministic](https://doc.rust-lang.org/std/primitive.f64.html#method.ln). I guess it cannot be better than the float precision so around 52 bits.

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
damrobi: I updated it to: "signer_lottery_hash is the hash computed using the signer's signature and the index tested for the lottery."
Let me know what you think.

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
damrobi: I updated the name to `verify_lottery_eligibility`, let me know if this should be reversed or changed to something else.

rrtoledo (approved):
LGTM

Issue #2946, opened 2026-01-21
Add negative tests in STM


jpraynaud:
## Why
We want to implement negative tests in the STM library to enhance its safety.

## What
Implement negative testing in the STM library.

## How

> See this comment https://github.com/input-output-hk/mithril/pull/2880/files/2fa32278870c5ce628bd193515f2a7820bc21466#r2697318227

- [ ] Implement negative testing for single signatures
- [ ] TBD

leepl37:
Hi @jpraynaud 
I would like to work on this issue.
I have already drafted the negative tests for `SingleSignature` verification (specifically checking for `invalid verification keys` and `out-of-bounds indices`), as suggested by @hjeljeli32 in the [related discussion](https://github.com/input-output-hk/mithril/pull/2880#discussion_r2697318227).

Issue #2943, opened 2026-01-21
SNARK pre-aggregation primitives in STM library - Phase 2


jpraynaud:
## Why
We need to evolve the STM library in order to implement SNARK primitives. Following the definition of the architecture in #2763, we have defined an implementation plan and we want to address the SNARK pre-aggregation primitives part of the plan for single signatures.

## What
Implement the SNARK pre-aggregation primitives for single signatures.

## How
- [x] Implement the creation of the single signature artifacts using Schnorr signatures:
  - [x] Adapt `Signer`
  - [x] Compute the message to sign (message if prefixed with the commitment of the key registration for SNARK)
  - [x] Make sure that we compute if the same way as in the circuit (or create a converter)
  - [x] Create the Schnorr signature for the message
  - [x] We create a single signature only if the Concatenation indices list is not empty, and we don't compute the indices list for SNARK). Could this rule be enhanced?
  - [x] Adapt the `SingleSignature` and `SingleSignatureWithRegisteredParty` in the `protocol` module
  - [x] Fix issues detected in #2993 
- [x] Benchmarks
- [x] Golden tests 
- [x] Property tests
- [x] Documentation

PR #2985, opened 2026-02-04
test(stm): add negative single-signature tests

leepl37:
## Content

This PR adds negative tests for `SingleSignature::verify` in `mithril-stm` to ensure invalid inputs are rejected:

- wrong verification key
- out‑of‑bounds index
- wrong message

This aligns with the reviewer suggestion from @hjeljeli32 and addresses #2946.
**Note:** dependency on #2943 is acknowledged; I will rebase/update after it merges if needed.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [ ] Reviewer requested
- Documentation
  - [x] No new TODOs introduced

## Comments

This is a **Draft PR** as discussed with @jpraynaud. I am opening it now to allow for early review of the test logic.

## Issue(s)

Relates to #2946
Blocked by #2943 (will rebase if needed)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4507,7 +4507,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-stm"
-version = "0.9.21"
+version = "0.9.22"
 dependencies = [
  "anyhow",
  "blake2 0.10.6",
--- a/mithril-stm/CHANGELOG.md
+++ b/mithril-stm/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## 0.9.22 (03-03-2026)
+
+### Added
+
+- Added negative tests for `SingleSignature::verify` to reject invalid inputs (wrong verification key, out-of-bounds index, wrong message, and mismatched registration).
+
 ## 0.9.20 (03-02-2026)
 
 ### Changed
--- a/mithril-stm/Cargo.toml
+++ b/mithril-stm/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mithril-stm"
-version = "0.9.21"
+version = "0.9.22"
 edition = { workspace = true }
 authors = { workspace = true }
 homepage = { workspace = true }


mithril-stm/src/protocol/single_signature/signature.rs
```diff
@@ -166,11 +166,84 @@ mod tests {
...
+        RegistrationEntry, Signer, SingleSignature, StmResult,
+        VerificationKeyProofOfPossessionForConcatenation, proof_system::ConcatenationProofSigner,
+        signature_scheme::BlsSigningKey,
     };
 
+    type D = MithrilMembershipDigest;
+
+    trait IntoStmResult<T> {
+        fn into_stm_result(self) -> StmResult<T>;
+    }
```
jpraynaud: This trait is not needed IMO. What is its purpose?

mithril-stm/src/protocol/single_signature/signature.rs
```diff
@@ -306,4 +379,46 @@ mod tests {
...
+
+    #[test]
+    fn verify_fails_with_wrong_verification_key() {
+        let (params, signer, _pk_1, pk_2, avk, message) = build_single_signature_context();
+        let signature = signer
+            .create_single_signature(&message)
+            .expect("signature should be created");
+
+        let result = signature.verify(&params, &pk_2.vk, &1, &avk, &message);
+        assert!(result.is_err(), "Verification should fail with wrong VK");
```
jpraynaud: This is more readable and will give more details in case of failure (applies to all other tests):
```suggestion
        signature.verify(&params, &pk_2.vk, &1, &avk, &message).expect("Verification should fail with wrong verification key");
```

Also, you should probably check the error returned downcasts to the expected variant as the test could work but for the wrong reason (applies here and in other tests).

mithril-stm/src/protocol/single_signature/signature.rs
```diff
@@ -166,11 +166,84 @@ mod tests {
...
+    }
+
+    fn build_single_signature_context() -> (
+        Parameters,
+        Signer<D>,
+        VerificationKeyProofOfPossessionForConcatenation,
+        VerificationKeyProofOfPossessionForConcatenation,
+        AggregateVerificationKey<D>,
+        [u8; 16],
+    ) {
```
jpraynaud: You could have created a struct for this return type, this would make it easier to read IMO.

mithril-stm/src/protocol/single_signature/signature.rs
```diff
@@ -306,4 +379,46 @@ mod tests {
             assert_eq!(golden_serialized, serialized);
         }
     }
+
+    #[test]
+    fn verify_fails_with_wrong_verification_key() {
+        let (params, signer, _pk_1, pk_2, avk, message) = build_single_signature_context();
```
jpraynaud:
```suggestion
        let (params, signer_1, _pk_1, pk_2, avk, message) = build_single_signature_context();
```

mithril-stm/src/protocol/single_signature/signature.rs
```diff
@@ -166,11 +166,84 @@ mod tests {
...
+    fn build_single_signature_context() -> (
+        Parameters,
+        Signer<D>,
+        VerificationKeyProofOfPossessionForConcatenation,
+        VerificationKeyProofOfPossessionForConcatenation,
+        AggregateVerificationKey<D>,
+        [u8; 16],
+    ) {
+        let mut rng = ChaCha20Rng::from_seed([0u8; 32]);
+        let message = [42u8; 16];
```
curiecrypt: Is there a specific reason to define the message like this?

mithril-stm/src/protocol/single_signature/signature.rs
curiecrypt: I agree. Also, message and params are fixed, so instead of returning those with this function, we can hardcode them.

mithril-stm/src/protocol/single_signature/signature.rs
```diff
@@ -166,11 +166,84 @@ mod tests {
...
+            Ok(self)
+        }
+    }
+
+    impl<T> IntoStmResult<T> for StmResult<T> {
+        fn into_stm_result(self) -> StmResult<T> {
+            self
+        }
+    }
+
```
curiecrypt, line 194: > I agree. Also, message and params are fixed, so instead of returning those with this function, we can hardcode them.

I think the following would be better:
- Hardcoded parameters
- A struct for the test environment including the fields:
   - Signer: will be responsible creating single signatures in the tests
   - `Clerk` or `ClosedKeyRegistration`: They both provide all registered signers' information. 
   - `AggregateVerificationKey`: This is optional since it can be built from Clerk.
   - Message: Instead of fixed message, return a specific one for each call. 

By using this structure, we can create different instances in the tests.

mithril-stm/src/protocol/single_signature/signature.rs
```diff
@@ -166,11 +166,84 @@ mod tests {
...
+    ) {
+        let mut rng = ChaCha20Rng::from_seed([0u8; 32]);
+        let message = [42u8; 16];
+        let params = Parameters {
+            m: 10,
+            k: 5,
+            phi_f: 0.8,
+        };
+
+        let sk_1 = BlsSigningKey::generate(&mut rng);
```
curiecrypt: Using a loop would be better. If we parametrize this function with the number of signers, that would provide freedom in test scenarios.

mithril-stm/src/protocol/single_signature/signature.rs
```diff
@@ -298,4 +379,46 @@ mod tests {
...
+            .create_single_signature(&message)
+            .expect("signature should be created");
+
+        let result = signature.verify(&params, &pk_2.vk, &1, &avk, &message);
+        assert!(result.is_err(), "Verification should fail with wrong VK");
+    }
+
+    #[test]
+    fn verify_fails_with_out_of_bounds_index() {
+        let (params, signer, pk_1, _pk_2, avk, message) = build_single_signature_context();
```
curiecrypt: Using a structure will help avoid underscores in variable names.

mithril-stm/src/protocol/single_signature/signature.rs
leepl37: You’re right, this trait is not needed. 
Its only purpose was to normalize a test-only conversion step before calling expect, but that abstraction made the setup harder to read.
I removed `IntoStmResult` and replaced it with direct handling in the test setup.

mithril-stm/src/protocol/single_signature/signature.rs
leepl37: You’re right, thanks for the suggestion.
I updated the negative tests to use expect_err(...) and to assert the expected downcast error variant, so they fail for the correct reason (not just any error).
I applied this to all three tests: wrong VK, out-of-bounds index, and wrong message.

mithril-stm/src/protocol/single_signature/signature.rs
leepl37: Implemented, thanks for the suggestion.
I replaced the tuple return with a `SingleSignatureTestContext` struct to make the setup more readable, and I hardcoded shared test inputs (`TEST_MESSAGE` and `test_parameters()`) instead of returning message and params from the builder.

mithril-stm/src/protocol/single_signature/signature.rs
leepl37: No specific protocol reason — it was only a deterministic test message.
I’ve now made this explicit by using a shared test constant (`TEST_MESSAGE`) instead of returning message from the context builder, which keeps the setup clearer.

mithril-stm/src/protocol/single_signature/signature.rs
```diff
@@ -165,12 +165,97 @@ mod tests {
...
+            m: 10,
+            k: 5,
+            phi_f: 0.8,
+        }
+    }
+
+    struct SingleSignatureTestContext {
+        signer_1: Signer<D>,
+        pk_1: VerificationKeyProofOfPossessionForConcatenation,
+        pk_2: VerificationKeyProofOfPossessionForConcatenation,
```
jpraynaud:
```suggestion
        vk_1: VerificationKeyProofOfPossessionForConcatenation,
        vk_2: VerificationKeyProofOfPossessionForConcatenation,
```

mithril-stm/src/protocol/single_signature/signature.rs
```diff
@@ -165,12 +165,97 @@ mod tests {
...
+            registration.register_by_entry(&entry).unwrap();
+        }
+
+        let closed_key_registration = registration.close_registration();
+        let mut signing_keys = signing_keys.into_iter();
+        let sk_1 = signing_keys.next().expect("at least one signer exists");
+        let mut verification_keys = verification_keys.into_iter();
+        let pk_1 = verification_keys
+            .next()
+            .expect("at least one verification key exists");
```
jpraynaud: This comment is not very clear (same comment for the next line). Could you update it?

jpraynaud (commented):
Thanks for the updates @leepl37 !

Once all comments from @curiecrypt are addressed and the CI is all green, we'll be able to merge :+1:

mithril-stm/src/protocol/single_signature/signature.rs
leepl37: You’re right, thanks for pointing this out.
I updated both expect(...) messages.

mithril-stm/src/protocol/single_signature/signature.rs
leepl37: @curiecrypt Thanks for the suggestion! 
I implemented a `SingleSignatureTestContext` struct and moved fixed setup values out of the tuple-based builder. For registration-related data, I currently compute the `AVK` inside the builder and store it directly in the context, since that is what these verification tests use. This keeps the cross-context test setup simpler while preserving the intended coverage.

mithril-stm/src/protocol/single_signature/signature.rs
leepl37: thanks!  `build_single_signature_context` is now parameterized with `number_of_signers`/`rng_seed` and uses a loop to generate signers.

mithril-stm/src/protocol/single_signature/signature.rs
leepl37: thanks :) Addressed in the update: the tuple-based context was replaced with `SingleSignatureTestContext`.

leepl37:
Hi @curiecrypt and @jpraynaud! I have addressed all the feedback and implemented the new test context structure. 
Let me know if you need any other tweaks.

Issue #2908, opened 2026-01-12
Implement signable builder for Cardano Blocks and Transactions


jpraynaud:
# Why
In order to certify the Cardano Blocks and Transactions we need to implement the signable builders for the `CardanoBlocksTransactions` signed entity type.

## What
Implemement the signable builder for the `CardanoBlocksTransactions` signed entity type.

## How

> [!IMPORTANT]
> This issue will block the release of a new distribution (to avoid discrepancies in the signature of `CardanoTransactions`)

- [ ] Activate optionally the `CardanoBlocksTransactions` signed entity type in the aggregator 
- [ ] Implement the signable builder:
  - [ ] Create a Merkle map with leaves either representing Blocks (`"Block"||BlockHash||BlockNumber||SlotNumber`) or Transactions (`"Tx"||TransactionHash||BlockHash||BlockNumber||SlotNumber`) (Same structure as for `CardanoTransactions`)
  - [ ] Sort all leaves with 1/sorted blocks and then 2/sorted transactions
  - [ ] Use the `Poseidon` hash function (TBC)
  - [ ] Use the Merkle root as the message to sign
- [ ] Activate the `CardanoDatabase` signed entity type in the e2e test

PR #2995, opened 2026-02-10
feat: compute and persist blocks transactions block range roots

Alenar:
## Content

This PR:
- Define `CardanoBlockTransactionMkTreeNode` in `mithril-common`, a type that represent a leaf of the merkle tree for `CardanoBlocksTransactions` block ranges roots and can be sorted.
- Make `ChainDataImporter` in `internal/mithril-cardano-node-chain` compute and send to persistence the block ranges roots for `CardanoBlocksTransactions`
- Add necessary queries in `internal/mithril-persistence` to store/retrieve `CardanoBlocksTransactions` block ranges roots and retrieve `CardanoBlockTransactionMkTreeNode`

> [!IMPORTANT]
>  The threshold for pruning data with the `ChainDataImporterWithPruner` had to be updated to take in account that there's now two tables for block ranges roots (one "legacy" for `CardanoTransactions` and one "new" for `CardanoBlocksTransactions`).
>
> The pruner prune data that have already be computed into block ranges roots, but now with two table we have to ensure that this computation was done for both so the threshold was modified to be "the minimum between the highest start block number of the 'new' and the 'legacy' stored block range roots" (before it was simply "the highest start block number stored in the legacy block range roots"). 

#### `mithril-common`

- Add `CardanoBlockTransactionMkTreeNode`:
  - Designed to be used a the `MKTreeNode` to compute block range roots for `CardanoBlocksTransactions`
  - With an manual implementation for ordering, to allow sort rust side after data retrieval in db 
- Add `MKTree::new_from_iter` to allow construction from any `IntoIterator`, as the existing `new` works only with types that can be sliced, which is not the case of `BTreeSet` (the collection used when retrieving `CardanoBlockTransactionMkTreeNode` from the database).

#### `mithril-cardano-node-chain`

- `ChainDataStore`: add `get_highest_block_range`, `get_blocks_and_transactions_in_range`, and `store_block_range_roots`
- `BlockRangeImporter`:
  - rename `run` method to `run_legacy`
  - add new `run` that compute block ranges roots based on stored `CardanoBlocksTransactions`

#### `mithril-persistence`

- New `CardanoBlockTransactionsRecord`
- Add queries to fetch `CardanoBlockTransactionsRecord` from the sqlite database
- Add queries to fetch, insert, and delete `BlockRangeRootRecord` computed from `CardanoBlocksTransactions` and stored in the `block_range_root` table (copied from the existing 'legacy' queries)
- `CardanoTransactionRepository`:
  - Define the public methods that will be used to store and retrieve  `CardanoBlockTransactionsRecord` and the `BlockRangeRootRecord` for `CardanoBlocksTransactions`
  - rename `get_highest_start_block_number_for_legacy_block_range_roots` to `get_prune_blocks_threshold` and update it in the light of the fact that there's two block range roots tables.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation
  - [ ] No new TODOs introduced

## Issue(s)

Relates to #2908

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3887,7 +3887,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-aggregator"
-version = "0.8.21"
+version = "0.8.22"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -4032,7 +4032,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-cardano-node-chain"
-version = "0.1.19"
+version = "0.1.20"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -4182,7 +4182,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-common"
-version = "0.6.48"
+version = "0.6.49"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -4322,7 +4322,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-persistence"
-version = "0.2.63"
+version = "0.2.64"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -4418,7 +4418,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-signer"
-version = "0.3.14"
+version = "0.3.15"
 dependencies = [
  "anyhow",
  "async-trait",
--- a/internal/cardano-node/mithril-cardano-node-chain/Cargo.toml
+++ b/internal/cardano-node/mithril-cardano-node-chain/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mithril-cardano-node-chain"
-version = "0.1.19"
+version = "0.1.20"
 authors.workspace = true
 documentation.workspace = true
 edition.workspace = true
--- a/internal/mithril-persistence/Cargo.toml
+++ b/internal/mithril-persistence/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mithril-persistence"
-version = "0.2.63"
+version = "0.2.64"
 description = "Common types, interfaces, and utilities to persist data for Mithril nodes."
 authors = { workspace = true }
 edition = { workspace = true }
--- a/mithril-aggregator/Cargo.toml
+++ b/mithril-aggregator/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mithril-aggregator"
-version = "0.8.21"
+version = "0.8.22"
 description = "A Mithril Aggregator server"
 authors = { workspace = true }
 edition = { workspace = true }
--- a/mithril-common/Cargo.toml
+++ b/mithril-common/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mithril-common"
-version = "0.6.48"
+version = "0.6.49"
 description = "Common types, interfaces, and utilities for Mithril nodes."
 authors = { workspace = true }
 edition = { workspace = true }
--- a/mithril-signer/Cargo.toml
+++ b/mithril-signer/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mithril-signer"
-version = "0.3.14"
+version = "0.3.15"
 description = "A Mithril Signer"
 authors = { workspace = true }
 edition = { workspace = true }


internal/mithril-persistence/src/database/repository/cardano_transaction_repository.rs
```diff
@@ -270,13 +333,27 @@ impl CardanoTransactionRepository {
...
+        block_ranges: Vec<BlockRange>,
+    ) -> StdResult<Vec<CardanoBlockTransactionsRecord>> {
+        let mut blocks_with_transactions = vec![];
+        for block_range in block_ranges {
+            let block_range_transactions: Vec<CardanoBlockTransactionsRecord> =
+                self.connection_pool.connection()?.fetch_collect(
+                    GetCardanoBlockTransactionsQuery::by_block_ranges(vec![block_range]),
+                )?;
+            blocks_with_transactions.extend(block_range_transactions);
+        }
```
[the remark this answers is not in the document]
Alenar: False positive. This is an optimization to allow more data throughput when used by the prover.

internal/cardano-node/mithril-cardano-node-chain/src/test/double/chain_data_store.rs
```diff
@@ -607,6 +721,96 @@ mod tests {
         }
     }
 
+    #[tokio::test]
+    async fn get_blocks_and_transactions_in_range_returns_empty_when_no_transactions() {
```
turmelclem: when_store_is_empty ?

internal/cardano-node/mithril-cardano-node-chain/src/chain_importer/block_ranges_importer.rs
```diff
@@ -25,7 +25,59 @@ impl BlockRangeImporter {
...
+        let mut block_ranges_with_merkle_root: Vec<(BlockRange, MKTreeNode)> = vec![];
+        for block_range in block_ranges {
+            let blocks_and_transactions_mk_node = self
+                .transaction_store
+                .get_blocks_and_transactions_in_range(block_range.start..block_range.end)
+                .await?;
+
+            if blocks_and_transactions_mk_node.is_empty() {
+                continue;
+            }
```
jpraynaud, line 60: I'm not sure that this is supposed to happen as there will always be 15 blocks in each range ultimately (when we had the case where no transactions were available for a block range with the legacy certification).

internal/cardano-node/mithril-cardano-node-chain/src/chain_importer/block_ranges_importer.rs
```diff
@@ -25,7 +25,59 @@ impl BlockRangeImporter {
...
+
+            let merkle_root =
+                MKTree::<MKTreeStoreInMemory>::new_from_iter(blocks_and_transactions_mk_node)?
+                    .compute_root()?;
+            block_ranges_with_merkle_root.push((block_range, merkle_root));
+
+            if block_ranges_with_merkle_root.len() >= 100 {
```
jpraynaud: Maybe this `100` could be set in a constant to make it more explicit how what it represents?

internal/mithril-persistence/src/database/record/cardano_block_transactions.rs
```diff
@@ -0,0 +1,166 @@
...
+            block.block_number,
+            block.slot_number,
+            block.transactions_hashes,
+        )
+    }
+}
+
+impl CardanoBlockTransactionsRecord {
+    /// Converts the block into a vector of [CardanoBlockTransactionMkTreeNode].
+    pub fn into_mk_tree_node(self) -> Vec<CardanoBlockTransactionMkTreeNode> {
```
jpraynaud: Just curious if we could implement `Into` here?

Alenar: I though we can't, as it would be equivalent to:
```rust
impl From<CardanoBlockTransactionsRecord> for Vec<CardanoBlockTransactionMkTreeNode> { ... }
``` 
So on the right side the type is `From<something>` and on the left side it's `Vec<something else>`, as `From` and `Vec` are not members of this crate and usually this violate rust directives, I tried that in other types before and could not make it works, but ... it works now (maybe there's something different here but I don't see what).

That said I'm not sure that this is a good idea, since the target type is a collection this conversion is not "obvious" as specified in the ["When to implement From"](https://doc.rust-lang.org/core/convert/trait.From.html) in rust documentation.
> The conversion is obvious: it’s the only reasonable conversion between the two types. Otherwise it’s better to have it be a named method or constructor, like how [str::as_bytes](https://doc.rust-lang.org/core/primitive.str.html#method.as_bytes) is a method and how integers have methods like [u32::from_ne_bytes](https://doc.rust-lang.org/core/primitive.u32.html#method.from_ne_bytes), [u32::from_le_bytes](https://doc.rust-lang.org/core/primitive.u32.html#method.from_le_bytes), and [u32::from_be_bytes](https://doc.rust-lang.org/core/primitive.u32.html#method.from_be_bytes), none of which are From implementations. Whereas there’s only one reasonable way to wrap an [Ipv6Addr](https://doc.rust-lang.org/core/net/struct.Ipv6Addr.html) into an [IpAddr](https://doc.rust-lang.org/core/net/enum.IpAddr.html), thus IpAddr: From<Ipv6Addr> exists.

internal/cardano-node/mithril-cardano-node-chain/src/chain_importer/block_ranges_importer.rs
Alenar: I think the question here is not "can it happens in normal condition" but "does the system allow it to happen and what are the consequences ?".
Here the consequence would be storing a block range roots for an "empty" list of block which would then be included in the global merkle tree and chance its merkle root.

Storing nothing would also change the merkle root but is more easily recoverable as block range roots are inserted using a `insert or ignore` meaning that existing data won't be automatically update if there's a change.

PR #2996, opened 2026-02-10
fix(devnet): Docker images

jpraynaud:
## Content

This PR includes **improvements and fixes for the devnet**:
* Added a `build-ci` target to the `Makefile` for `mithril-aggregator`, `mithril-signer`, and `mithril-client-cli`, enabling architecture-aware builds for CI and improving compatibility with different Linux architectures. The corresponding `docker-build-ci` targets now depend on `build-ci` instead of the regular build. 
* Updated the `.ci` Dockerfiles for all components to consolidate and streamline package installation by combining all required packages (including `adduser`) into a single `apt-get` command for `debian:13-slim`.
* Added missing `DMQ_NODE_SOCKET_PATH` environment variable to aggregator and signer Docker images
* Fixed the startup order in the devnet run script to correctly start Cardano nodes before DMQ nodes.
* Improved the error message for the `CertificateEpochGap` variant in the Rust service interface to clarify the required actions for leader and follower aggregators during an epoch gap.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested


mithril-aggregator/Makefile
```diff
@@ -31,7 +50,7 @@ doc:
 docker-build:
 	cd ../ && docker build -t mithril/mithril-aggregator -f mithril-aggregator/Dockerfile .
 
-docker-build-ci: build
+docker-build-ci: build-ci
 	cd ../ && docker build -t mithril/mithril-aggregator -f mithril-aggregator/Dockerfile.ci --build-arg DOCKER_IMAGE_FROM --build-arg CARDANO_NODE_VERSION .
```
Alenar: The `build-ci` recipe doesn't seems to be useful as it runs the same build command as the existing `build` recipe.
Also the linux and target arch check only make senses for building the docker image.

As such I think the `build-ci` recipe should be removed and its logic, except the build command, should be transferred to the `docker-build-ci` recipe.

```suggestion
docker-build-ci: build
	# Check that we are on Linux, then build
	if [ "$$(uname -s)" != "Linux" ]; then \
		echo "Error: this target only supports Linux"; \
		exit 1; \
	fi;
	RAW_ARCH=$$(uname -m); \
	if [ "$$RAW_ARCH" = "x86_64" ]; then \
		ARCH="amd64"; \
	elif [ "$$RAW_ARCH" = "aarch64" ]; then \
		ARCH="arm64"; \
	else \
		echo "Error: unsupported architecture $$RAW_ARCH"; \
		exit 1; \
	fi; \
	mkdir -p ./bin-linux-$${ARCH}; \
	cp ../target/release/mithril-aggregator ./bin-linux-$${ARCH}/mithril-aggregator
	cd ../ && docker build -t mithril/mithril-aggregator -f mithril-aggregator/Dockerfile.ci --build-arg DOCKER_IMAGE_FROM --build-arg CARDANO_NODE_VERSION .
```

mithril-client-cli/Makefile
Alenar, line 1: Same remark as `mithril-aggregator/Makefile`

mithril-signer/Makefile
Alenar, line 1: Same remark as `mithril-aggregator/Makefile`

Alenar (approved):
LGTM with a comment.

Another point: did you forgot the `mithril-relay` ?

mithril-aggregator/Makefile
jpraynaud: You're right, fixed :+1:

mithril-client-cli/Makefile
jpraynaud: Fixed

mithril-signer/Makefile
jpraynaud: Fixed

PR #3002, opened 2026-02-13
 refactor(snark): remove halo2/off-circuit and switch to STM types

hjeljeli32:
## Content

This PR removes the circuit/halo2/off-circuit module and switches the Halo2 circuit and golden helpers to use existing STM types and implementations.

Merkle, Signature, Poseidon and Curve helpers are now fully wired to STM.
CPU and circuit implementations are consistent and golden tests remain green.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation
  - [x] No new TODOs introduced

## Issue(s)

Closes https://github.com/input-output-hk/mithril/issues/2945

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4461,7 +4461,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-stm"
-version = "0.9.15"
+version = "0.9.16"
 dependencies = [
  "anyhow",
  "blake2 0.10.6",
--- a/mithril-stm/CHANGELOG.md
+++ b/mithril-stm/CHANGELOG.md
@@ -5,6 +5,17 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## 0.9.16 (02-18-2026)
+
+### Changed
+
+- Switched Halo2 circuit and golden helpers to reuse existing STM types and implementations.
+- Applied internal refactoring and cleanup in Halo2 witness and helper layers.
+
+### Removed
+
+- Removed the `circuits/halo2/off-circuit` module.
+
 ## 0.9.15 (02-13-2026)
 
 ### Added
--- a/mithril-stm/Cargo.toml
+++ b/mithril-stm/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mithril-stm"
-version = "0.9.15"
+version = "0.9.16"
 edition = { workspace = true }
 authors = { workspace = true }
 homepage = { workspace = true }


mithril-stm/src/signature_scheme/unique_schnorr_signature/mod.rs
```diff
@@ -162,10 +164,11 @@ mod tests {
         }
 
         #[test]
-        fn signature_to_from_bytes(msg in prop::collection::vec(any::<u8>(), 1..128), seed in any::<[u8;32]>()) {
+        fn signature_to_from_bytes(mut msg in prop::collection::vec(any::<u8>(), 1..128), seed in any::<[u8;32]>()) {
```
curiecrypt: Why do we change 128 as 32 in here:
`fn valid_signing_verification(
            mut msg in prop::collection::vec(any::<u8>(), 1..32),`
and do not modify this one?

mithril-stm/src/circuits/halo2/circuit.rs
```diff
@@ -94,9 +96,18 @@ impl Relation for StmCircuit {
...
+            let vk = std_lib.jubjub().assign(
+                layouter,
+                wit.clone().map(|(x, _, _, _)| {
+                    let u = BaseFieldElement::from_bytes(&x.0[0..32])
+                        .expect("Invalid VK u-coordinate bytes");
+                    let v = BaseFieldElement::from_bytes(&x.0[32..64])
+                        .expect("Invalid VK v-coordinate bytes");
+                    PrimeOrderProjectivePoint::from_coordinates(u, v)
+                        .expect("Invalid verification key: not on curve or not prime order")
+                        .0
```
jpraynaud: We should probably use the `from_bytes` function of `SchnorrVerificationKey` 

There are a lot of `expect` which would create panic and would need to be replaced by proper error handling in the circuit.

mithril-stm/src/circuits/halo2/golden/cases/negative.rs
```diff
@@ -6,19 +6,19 @@ use crate::circuits::halo2::golden::helpers::{
     create_merkle_tree_with_leaf_selector, find_two_distinct_witness_entries,
     prove_and_verify_result, setup_stm_circuit_env,
 };
-use crate::circuits::halo2::off_circuit::merkle_tree::Position;
-use crate::circuits::halo2::types::{JubjubBase, JubjubScalar};
+use crate::circuits::halo2::types::{Msg, Position};
```
jpraynaud: You could be more precise and also avoid abbreviations:
```suggestion
use crate::circuits::halo2::types::{SignedMessageWithoutPrefix, Position};
```

mithril-stm/src/circuits/halo2/golden/helpers.rs
```diff
@@ -78,6 +81,67 @@ pub(crate) struct StmCircuitScenario {
...
+}
+
+impl SignerLeaf {
+    fn stm_leaf(&self) -> MerkleTreeSnarkLeaf {
+        MerkleTreeSnarkLeaf(self.vk, self.target_value)
+    }
+
+    fn mt_leaf(&self) -> MTLeaf {
+        MTLeaf(self.vk.to_bytes(), self.target_field)
+    }
```
jpraynaud: You could have used `Into` trait trait to do the conversions. Maybe, the introduction of a new  `SignerLeaf` type could even be avoided.

This is something that should work for conversions not in the tests: this would avoid a lot of code duplication (but needs to avoid panics and use `TryInto` instead.

mithril-stm/src/circuits/halo2/golden/helpers.rs
```diff
@@ -111,29 +175,49 @@ impl StmCircuitScenario {
...
 }
 
+impl StmMerkleTreeWrapper {
+    /// Return the Merkle root as a JubjubBase field element.
+    pub(crate) fn root(&self) -> F {
+        self.root
+    }
+
+    /// Return a Halo2-style Merkle path for the given leaf index.
+    pub(crate) fn get_path(&self, i: usize) -> MerklePath {
```
jpraynaud: Same remark for the `MerklePath`: this should be implemented with the `TryInto` trait to convert the CPU Merkle path to the circuit Merkle path.

mithril-stm/src/circuits/halo2/golden/helpers.rs
```diff
@@ -111,29 +175,49 @@ impl StmCircuitScenario {
...
+        let stm_path = self.stm_tree.compute_merkle_tree_path(i);
+        stm_path_to_halo2_path(&stm_path).expect("Failed to adapt STM Merkle path")
+    }
+}
+
 /// Build a default Merkle tree with all leaves set to the max target.
-pub(crate) fn create_default_merkle_tree(n: usize) -> (Vec<SigningKey>, Vec<MTLeaf>, MerkleTree) {
+pub(crate) fn create_default_merkle_tree(n: usize) -> (Vec<SignerLeaf>, StmMerkleTreeWrapper) {
```
jpraynaud: IMO the correct approach is to generate a CPU Merkle tree, and to enhance its implementation to provide:
- The list of leaves
- A leaf given its index if exists

This would avoid a lot of code duplication and would enhance readability.

mithril-stm/src/circuits/halo2/golden/helpers.rs
```diff
@@ -258,13 +355,16 @@ pub(crate) fn build_witness_with_indices(
...
     for i in 0..witness.len() {
-        let leaf_i = witness[i].0.to_bytes();
+        let leaf_i = witness[i].0;
         for (j, wj) in witness.iter().enumerate().skip(i + 1) {
-            if leaf_i != wj.0.to_bytes() {
+            let leaf_j = wj.0;
+            if leaf_i.0 != leaf_j.0 || leaf_i.1 != leaf_j.1 {
                 return (i, j);
             }
         }
```
jpraynaud: Why not pick 2 entries and make sure that they are distinct ? This would avoid a lot of the complexity of this function.

mithril-stm/src/circuits/halo2/types.rs
```diff
@@ -1,11 +1,47 @@
...
+use ff::Field;
+
+pub use midnight_curves::{Bls12, Fq as JubjubBase, Fr as JubjubScalar, JubjubExtended as Jubjub};
 
 pub type Target = JubjubBase;
 pub type Msg = JubjubBase;
 pub type MerkleRoot = JubjubBase;
 pub type LotteryIndex = u32;
+
+// Merkle witness types
```
jpraynaud: Can you use doc comments (here and below)?

mithril-stm/src/circuits/halo2/utils/mod.rs
```diff
@@ -0,0 +1,78 @@
+use ff::PrimeField;
+use num_bigint::BigUint;
+use num_traits::{Num, One};
+
+/// Split a field element into lower/upper limbs at `num_bits` (little-endian).
+pub(crate) fn split<F: PrimeField>(value: &F, num_bits: u32) -> (F, F) {
```
jpraynaud: Maybe the name of this function is not explicit enough.

mithril-stm/src/circuits/halo2/utils/mod.rs
```diff
@@ -0,0 +1,78 @@
...
+    let value_big = BigUint::from_bytes_le(value.to_repr().as_ref());
+    let lower_mask = (BigUint::one() << num_bits) - BigUint::one();
+    let lower_big = value_big.clone() & &lower_mask;
+    let upper_big = value_big >> num_bits;
+    let lower = big_to_fe::<F>(lower_big);
+    let upper = big_to_fe::<F>(upper_big);
+    (lower, upper)
+}
+
+fn modulus<F: PrimeField>() -> BigUint {
```
jpraynaud: Could you add a doc comment?

mithril-stm/src/circuits/halo2/utils/mod.rs
```diff
@@ -0,0 +1,78 @@
...
+    BigUint::from_str_radix(&F::MODULUS[2..], 16).unwrap()
+}
+
+fn big_to_fe<F: PrimeField>(e: BigUint) -> F {
```
jpraynaud: Could you add a doc comment and make the function name more explicit:
```suggestion
fn big_unsigned_integer_to_field_element<F: PrimeField>(e: BigUint) -> F {
```

mithril-stm/src/circuits/halo2/utils/mod.rs
```diff
@@ -0,0 +1,78 @@
...
+    let modulus = modulus::<F>();
+    let e = e % modulus;
+    F::from_str_vartime(&e.to_str_radix(10)[..]).unwrap()
```
jpraynaud: We don't want any panic in production code, this should be handled with an error.

mithril-stm/src/circuits/halo2/utils/mod.rs
```diff
@@ -0,0 +1,78 @@
...
+    for (i, value) in stm_path.values.iter().enumerate() {
+        let bytes: [u8; 32] = value
+            .as_slice()
+            .try_into()
+            .map_err(|_| MerklePathAdapterError::InvalidDigestLength)?;
+        let node =
+            digest_bytes_to_base(&bytes).ok_or(MerklePathAdapterError::NonCanonicalDigest)?;
+        let bit = (stm_path.index >> i) & 1;
+        // STM uses even idx => H(h || sibling), odd idx => H(sibling || h);
+        // map even to Position::Right so Halo2 folds as H(acc || sibling).
```
jpraynaud: This comment is unclear, can you elaborate a bit further?

mithril-stm/src/signature_scheme/unique_schnorr_signature/jubjub/curve_points.rs
```diff
@@ -59,13 +59,13 @@ impl From<&PrimeOrderProjectivePoint> for AffinePoint {
 pub(crate) struct ProjectivePoint(pub(crate) JubjubExtended);
 
 impl ProjectivePoint {
-    /// Hashes input BaseFieldElements to a projective point on the Jubjub curve
+    /// Hashes input bytes to a projective point on the Jubjub curve
+    /// For now we leave the SHA call in the function since the SHA
+    /// function is not used anywhere else. This might change in the future.
```
jpraynaud: You need to rebase your branch as this comes from a PR that has already been merged.

mithril-stm/src/circuits/halo2/types.rs
```diff
@@ -1,11 +1,47 @@
...
+
+#[derive(Debug, Copy, Clone)]
+pub struct MTLeaf(pub [u8; 64], pub Target);
```
damrobi: I'm not sure we want to store the verification key as bytes for the witness. If I'm not wrong the key is stored as a verification key in all structures.

mithril-stm/src/circuits/halo2/circuit.rs
```diff
@@ -94,9 +96,18 @@ impl Relation for StmCircuit {
 
             pre_index = index.clone();
 
-            let vk = std_lib
-                .jubjub()
-                .assign(layouter, wit.clone().map(|(x, _, _, _)| x.0.0))?;
+            let vk = std_lib.jubjub().assign(
+                layouter,
+                wit.clone().map(|(x, _, _, _)| {
+                    let u = BaseFieldElement::from_bytes(&x.0[0..32])
```
damrobi: I think if we store the key directly as a point we don't have to change this part. If we really need to it would be better to have this conversion outside the circuit when forming the witness.

mithril-stm/src/circuits/halo2/circuit.rs
```diff
@@ -125,13 +136,33 @@ impl Relation for StmCircuit {
...
+            )?;
+            let c_native = std_lib.assign(
+                layouter,
+                wit.map(|(_, _, sig, _)| {
+                    let bytes = sig.challenge.to_bytes();
+                    JubjubBase::from_bytes_le(&bytes)
+                        .into_option()
+                        .expect("Invalid challenge base field bytes")
+                }),
+            )?;
```
damrobi: I'm not sure I understand why we need those conversion, it feels like we do two conversions in a row. If it's because the STM types do not match the midnight ones, I think accessing the value inside our wrappers might be enough since JubjubScalar and JubjubBase are just aliases

mithril-stm/src/circuits/halo2/utils/mod.rs
curiecrypt: Also maybe more descriptive function name?

mithril-stm/src/circuits/halo2/off_circuit/merkle_tree.rs
```diff
@@ -1,323 +0,0 @@
...
-use crate::circuits::halo2::off_circuit::unique_signature::VerificationKey;
-use crate::circuits::halo2::types::{JubjubBase, Target};
-
-type F = JubjubBase;
-
-#[derive(Debug, Copy, Clone)]
-pub struct MTLeaf(pub VerificationKey, pub Target);
-
-impl MTLeaf {
-    pub fn to_field(&self) -> [F; 3] {
```
curiecrypt, line 14: Are we going to need this function?

mithril-stm/src/circuits/halo2/utils/mod.rs
curiecrypt: I agree. 
The terminology `even idx` is a bit ambiguous. I guess it refers to the index bit at level `i`. 
We might rephrase to: `STM bit=0 means current node is the left child, so sibling is on the Right`?

mithril-stm/src/circuits/halo2/golden/helpers.rs
```diff
@@ -258,13 +355,16 @@ pub(crate) fn build_witness_with_indices(
 /// Find two witness entries with distinct leaves, for negative test construction.
 pub(crate) fn find_two_distinct_witness_entries(witness: &[WitnessEntry]) -> (usize, usize) {
     assert!(witness.len() >= 2, "expected at least two witness entries");
-    if witness[0].0.to_bytes() != witness[1].0.to_bytes() {
+    let leaf0 = witness[0].0;
+    let leaf1 = witness[1].0;
+    if leaf0.0 != leaf1.0 || leaf0.1 != leaf1.1 {
```
curiecrypt: This comparison between `[u8; 64]` and `JubjubBase` seems to have no problems. 
But, we can derive `PartialEq` on `MTLeaf` so that we can simply write `leaf0 != leaf1`?

mithril-stm/src/circuits/halo2/golden/helpers.rs
```diff
@@ -211,45 +301,52 @@ pub(crate) fn setup_stm_circuit_env(case_name: &str, k: u32, quorum: u32) -> Stm
...
+            .expect("STM signature generation failed");
+        assert_challenge_endianness(&stm_sig);
+        stm_sig
+            .verify(
+                &[BaseFieldElement(merkle_root), BaseFieldElement(msg)],
+                &material.vk,
+            )
+            .expect("STM signature verification failed");
 
         let merkle_path = merkle_tree.get_path(ii);
```
curiecrypt: The old code had a `compute_root` + `assert_eq!` here to verify that the adapted path creates the expected root before handing the witness to the circuit. 
This was catching bugs in the tests (especially the `stm_path_to_halo2_path` adapter) with a clear assertion message, rather than Halo2 error "constraint not satisfied".
Since we don't have `compute_root` anymore, I think, you can replace it with STM's own path verification.

mithril-stm/src/circuits/halo2/golden/helpers.rs
```diff
@@ -276,41 +376,40 @@ pub(crate) fn find_two_distinct_witness_entries(witness: &[WitnessEntry]) -> (us
...
-        signer_index < leaves.len(),
-        "signer_index out of bounds for leaves"
+        signer_index < signer_leaves.len(),
+        "signer_index out of bounds for signer_leaves"
     );
     let mut witness = Vec::new();
-    let usk = sks[signer_index].clone();
-    let uvk = leaves[signer_index].0;
+    let material = &signer_leaves[signer_index];
     let merkle_path = merkle_tree.get_path(signer_index);
```
curiecrypt: See my comment on line 346.

mithril-stm/src/circuits/halo2/circuit.rs
```diff
@@ -125,13 +136,33 @@ impl Relation for StmCircuit {
...
+                    PrimeOrderProjectivePoint::from_coordinates(u, v)
+                        .expect("Invalid commitment point: not on curve or not prime order")
+                        .0
+                }),
+            )?;
+            let s: AssignedScalarOfNativeCurve<C> = std_lib.jubjub().assign(
+                layouter,
+                wit.clone().map(|(_, _, sig, _)| {
+                    let bytes = sig.response.to_bytes();
+                    JubjubScalar::from_bytes(&bytes)
```
curiecrypt: `sig.response` is already a `ScalarFieldElement` holding a valid `JubjubScalar` inside. The `to_bytes` -> `from_bytes` round-trip seems to reconstruct the same inner value with an unnecessary serialization step. And also a panic that can never occur.

You might consider extracting the inner value directly (`sig.response.0`) to stay consistent with how VK and sigma are handled above: wrapper validation first, then .0 extraction.

mithril-stm/src/circuits/halo2/circuit.rs
```diff
@@ -125,13 +136,33 @@ impl Relation for StmCircuit {
...
+                }),
+            )?;
+            let c_native = std_lib.assign(
+                layouter,
+                wit.map(|(_, _, sig, _)| {
+                    let bytes = sig.challenge.to_bytes();
+                    JubjubBase::from_bytes_le(&bytes)
+                        .into_option()
+                        .expect("Invalid challenge base field bytes")
+                }),
```
curiecrypt: same as above

mithril-stm/src/circuits/halo2/circuit.rs
hjeljeli32: @jpraynaud I will switch it to use SchnorrVerificationKey::from_bytes.

Regarding the remaining expect usages: I agree they should not remain long-term. However, in this PR I intentionally avoided modifying the circuit code. I will address all panics in the upcoming circuit refactor ticket.

mithril-stm/src/circuits/halo2/golden/helpers.rs
hjeljeli32: @jpraynaud 
I updated the helpers to use standard conversions (.into()) instead of conversion methods.

Regarding SignerLeaf: it’s intentionally a small “material” container we generate once per signer (SK, VK, target field, target value). We need the same material both to build the STM Merkle tree leaves and later to derive the witness entries (leaf bytes + signature + path). That was my logic why I introduced SignerLeaf. Maybe the name is confusing, I can rename it to something like SignerMaterial / LeafMaterial.

mithril-stm/src/circuits/halo2/circuit.rs
hjeljeli32: @jpraynaud I removed all expect from helpers.rs / positive.rs / negative.rs

mithril-stm/src/signature_scheme/unique_schnorr_signature/mod.rs
```diff
@@ -14,6 +14,7 @@ mod verification_key;
 
 pub use error::*;
 pub use jubjub::BaseFieldElement;
+pub(crate) use jubjub::DST_SIGNATURE;
```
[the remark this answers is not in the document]
hjeljeli32: Copilot’s suggestion is not quite right in this case.
These two pub(crate) re-exports are serving different module boundaries

mithril-stm/src/circuits/halo2/utils/mod.rs
```diff
@@ -0,0 +1,96 @@
...
+///
+/// Conversion uses `from_str_vartime` on the reduced decimal representation.
+fn big_unsigned_integer_to_field_element<F: PrimeField>(e: BigUint) -> Result<F, Halo2UtilsError> {
+    let modulus = field_modulus_as_biguint::<F>()?;
+    let e = e % modulus;
+    F::from_str_vartime(&e.to_str_radix(10)[..]).ok_or(Halo2UtilsError::FieldElementConversion)
+}
+
+#[cfg(test)]
+pub(crate) use merkle_path_adapter::MerklePathAdapterError;
```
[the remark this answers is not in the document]
hjeljeli32: Copilot is incorrect here.

In mod.rs MerklePathAdapterError is defined inside #[cfg(test)] mod merkle_path_adapter.
The TryFromfor Halo2MerklePath is also inside that same #[cfg(test)] module.
The re-export is also #[cfg(test)

mithril-stm/src/circuits/halo2/golden/helpers.rs
```diff
@@ -40,6 +43,71 @@ type CircuitVerificationAndProvingKeyPair = (MidnightVK, MidnightPK<StmCircuit>)
...
+    /// Challenge bytes decode but do not match the native challenge field value.
+    #[error("challenge endianness mismatch")]
+    ChallengeEndiannessMismatch,
+    /// Merkle root digest has an invalid byte length.
+    #[error("invalid merkle root digest length")]
+    InvalidMerkleRootDigestLength,
+    /// Merkle root digest is not a canonical base field element encoding.
+    #[error("non-canonical merkle root digest")]
+    NonCanonicalMerkleRoot,
+    /// Merkle path adaptation from STM format to Halo2 witness format failed.
```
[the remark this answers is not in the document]
hjeljeli32: Current setup is consistent StmCircuitProofError is pub(crate), not public API.
It lives in helpers.rs which is test-only support code via circuits/halo2/golden.

mithril-stm/src/circuits/halo2/golden/cases/negative.rs
```diff
@@ -243,211 +245,229 @@ fn merkle_path_flip_position() {
...
 }
 
 #[should_panic]
 #[test]
 fn merkle_path_length_long() {
+    merkle_path_length_long_body()
+        .unwrap_or_else(|error| panic!("Unexpected build error: {error}"));
+}
+
+fn merkle_path_length_long_body() -> Result<(), StmCircuitProofError> {
```
jpraynaud: IMO adding a result type in all of the tests does not bring a lot as asserts will panic in the first place. They can be removed.

mithril-stm/src/circuits/halo2/gadgets.rs
```diff
@@ -36,7 +36,12 @@ fn decompose_unsafe(
     // Decompose 255-bit value into 127-bit and 128-bit values.
     let x_value = x.value();
     let base127 = F::from_u128(1_u128 << 127);
-    let (x_low, x_high) = x_value.map(|v| split(v, 127)).unzip();
+    let (x_low, x_high) = x_value
+        .map_with_result(|v| split_field_element_into_le_limbs(v, 127))
+        .map_err(|_| {
+            Error::Synthesis("failed to split field element into little-endian limbs".to_string())
```
jpraynaud: It would be interesting to have the source error in the synthesis error message here and in other occurences.

mithril-stm/src/circuits/halo2/golden/helpers.rs
jpraynaud: Yes, maybe you can rename into `SignerFixture`.

mithril-stm/src/circuits/halo2/golden/helpers.rs
jpraynaud: It looks like you forgot this one @hjeljeli32

mithril-stm/src/circuits/halo2/golden/helpers.rs
jpraynaud: It looks like you forgot this one @hjeljeli32

mithril-stm/src/circuits/halo2/golden/helpers.rs
hjeljeli32: Thanks @jpraynaud for double-checking.
I implemented this in the golden helper layer.
We now build the CPU Merkle tree once in a shared `build_merkle_tree_wrapper`, and the wrapper owns the leaf material with indexed access.
Both `create_default_merkle_tree` and `create_merkle_tree_with_leaf_selector` call that builder, so the duplication is removed and the “leaf by index (if exists)” is handled via typed errors.
If you meant enhancing the STM MerkleTree API itself (instead of golden helpers), we can follow up in a separate PR.

mithril-stm/src/circuits/halo2/golden/helpers.rs
hjeljeli32: @jpraynaud 
Thanks for pointing it out!
We now use a simple strategy: return the first two witness entries if their leaves are distinct.

```
/// Return the first two witness entries if their leaves are distinct.
///
/// This helper assumes callers construct witnesses so entries `0` and `1` come
/// from different signer fixtures (negative-test precondition). If not, it
/// returns an error instead of searching later entries.
pub(crate) fn find_two_distinct_witness_entries(
    witness: &[WitnessEntry],
) -> StmResult<(usize, usize)> {
    if witness.len() < 2 {
        return Err(StmCircuitProofError::WitnessTooShort);
    }
    let leaf0 = witness[0].0;
    let leaf1 = witness[1].0;
    if leaf0 != leaf1 {
        return Ok((0, 1));
    }
    Err(StmCircuitProofError::NoDistinctWitnessEntries)
}
```

mithril-stm/src/circuits/halo2/golden/cases/negative.rs
hjeljeli32: @jpraynaud 
Sure. Removed Result return types from all tests; they now fail via asserts. Kept should_panic 4 tests minimal.

mithril-stm/src/circuits/halo2/golden/helpers.rs
jpraynaud: Yes, that's what I meant. Let's do that in the refactoring that follows :+1:

PR #3009, opened 2026-02-16
feat(stm): Signer and Signature for SNARK Proof System

### LINKED ISSUE #2943: SNARK pre-aggregation primitives in STM library - Phase 2

curiecrypt:
## Content

This PR includes:
- Implement the SNARK proof signer (`SnarkProofSigner<D>`) that generates unique Schnorr signatures and checks the lottery
- Add `SingleSignatureForSnark` to `SingleSignature`, gated behind `future_snark` feature, with full serialization (bytes + JSON) support
- Integrate SNARK signing into the `Signer` flow: `create_single_signature` now produces both concatenation and SNARK signatures when `future_snark` is enabled
- Extend Initializer to generate and carry Schnorr key pairs (`SchnorrSigningKey, VerificationKeyForSnark`)
- Update `AggregateVerificationKey` to include a SNARK aggregate key built from the SNARK Merkle tree commitment
- Update `SingleSignature::verify` to validate the SNARK signature against the SNARK AVK when present
- Add deterministic golden tests for `SingleSignature` and `SingleSignatureWithRegisteredParty` covering `future_snark`

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Comments
`From<(RegistrationEntry, Stake)> for ClosedRegistrationEntry` is updated to set `modulus - 1` to the target value instead of 1. 

## Issue(s)
Closes #2943

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4507,7 +4507,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-stm"
-version = "0.9.23"
+version = "0.9.24"
 dependencies = [
  "anyhow",
  "blake2 0.10.6",
--- a/mithril-stm/CHANGELOG.md
+++ b/mithril-stm/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## 0.9.24 (03-04-2026)
+
+### Added
+
+- Added SNARK compatibility for signer and single signature.
+
 ## 0.9.23 (03-03-2026)
 
 ### Changed
--- a/mithril-stm/Cargo.toml
+++ b/mithril-stm/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mithril-stm"
-version = "0.9.23"
+version = "0.9.24"
 edition = { workspace = true }
 authors = { workspace = true }
 homepage = { workspace = true }


mithril-stm/src/proof_system/halo2_snark/signer.rs
```diff
@@ -0,0 +1,314 @@
...
+    /// `ev = Poseidon(prefix, commitment_point_x, commitment_point_y, index)` where
+    /// `(commitment_point_x, commitment_point_y)` are coordinates of signature's commitment point.
+    pub(super) fn verify_lottery_eligibility(
+        signature: &UniqueSchnorrSignature,
+        lottery_index: LotteryIndex,
+        m: u64,
+        prefix: BaseFieldElement,
+        target: LotteryTargetValue,
+    ) -> StmResult<()> {
+        if lottery_index > m {
```
hjeljeli32: @curiecrypt 
In the circuit, we require lottery_index to be strictly smaller than m. If that's the right bound, this check should be lottery_index >= m

mithril-stm/src/membership_commitment/merkle_tree/commitment.rs
```diff
@@ -69,6 +72,25 @@ impl<D: Digest + FixedOutput, L: MerkleTreeLeaf> MerkleTreeCommitment<D, L> {
...
+    /// `BaseFieldElement` via `BaseFieldElement::try_from(&[u8])`, which SHA-256 hashes the input
+    /// and then applies Montgomery reduction to obtain a field element. This means any message
+    /// length is accepted. The output is an array of two `BaseFieldElement`s, where the first
+    /// element is the root and the second element is the message.
+    ///
+    /// # Error
+    /// Returns an error if the conversion of the root or the message to `BaseFieldElement` fails.
+    pub fn build_snark_message(&self, message: &[u8]) -> StmResult<[BaseFieldElement; 2]> {
+        let root = self.root.as_slice();
+        let root_as_base_field_element = BaseFieldElement::try_from(root)
```
hjeljeli32: @curiecrypt 
I see a root-encoding mismatch risk with Halo2
BaseFieldElement::try_from hashes input with SHA-256 before field mapping, so this is SHA256(root_bytes) -> field.

Halo2 currently expects the actual Merkle root field element (canonical decode)

mithril-stm/src/signature_scheme/unique_schnorr_signature/jubjub/poseidon_digest.rs
```diff
@@ -11,6 +11,16 @@ const DST_SIGNATURE: JubjubBase = JubjubBase::from_raw([
     0,
 ]);
 
+/// Domain Separation Tag (DST) for the Poseidon hash used in lottery contexts.
+// TODO: remove this allow dead_code directive when function is called or future_snark is activated
+#[allow(dead_code)]
+pub const DST_LOTTERY: BaseFieldElement = BaseFieldElement(JubjubBase::from_raw([
+    0x4C4F_5454_4552_5953, // "LOTTERY" (ASCII), little-endian u64
```
hjeljeli32: @curiecrypt 
Today in Halo2 circuit, the Lottery DST is the following:
pub const DST_LOTTERY: JubjubBase = JubjubBase::from_raw([3, 3, 0, 0]);
So we need to align them: either we change STM side or I change circuit side in refactoring ticket.

mithril-stm/src/proof_system/halo2_snark/single_signature.rs
```diff
@@ -1,39 +1,95 @@
...
     }
 
-    /// Return `indices` of the single signature
-    pub(crate) fn get_indices(&self) -> &[LotteryIndex] {
-        &self.indices
+    /// Verify a `SingleSignatureForSnark`. First try to build the message to verify using the
+    /// provided `AggregateVerificationKeyForSnark` and the input `message`. Then verify the Schnorr
+    /// signature for the built message and the provided `VerificationKeyForSnark`. Finally, verify
+    /// that the lottery index associated with this signature actually won the lottery.
+    pub(crate) fn verify<D: MembershipDigest>(
```
hjeljeli32, line 37: @curiecrypt 
There could be a mismatch between STM and circuit
 - If I understood correctly, STM's Verification is based on `minimum_winning_lottery_index` only. So checks only one index: minimum_winning_lottery_index
 - Halo2 witness/circuit checks lottery/index constraints per witness entry across the full witness vector.

mithril-stm/src/membership_commitment/merkle_tree/commitment.rs
damrobi: Indeed, the merkle root should already be a BaseFieldElement as an output of a Poseidon hash so it doesn't need to be converted with try_from.

mithril-stm/src/proof_system/halo2_snark/signer.rs
```diff
@@ -0,0 +1,314 @@
...
+
+impl<D: MembershipDigest> SnarkProofSigner<D> {
+    /// Creates a new `SnarkProofSigner` with the given protocol parameters, keys,
+    /// lottery target value, and Merkle tree commitment of the key registration.
+    pub fn new(
+        parameters: Parameters,
+        signing_key: SchnorrSigningKey,
+        verification_key: VerificationKeyForSnark,
+        lottery_target_value: LotteryTargetValue,
+        key_registration_commitment: MerkleTree<D::SnarkHash, RegistrationEntryForSnark>,
```
damrobi: I'm not sure we need to give the Signer the full merkle tree, maybe only the root is enough, isn't it? It might also apply to the ConcatenationSigner. WDYT?

mithril-stm/src/proof_system/halo2_snark/signer.rs
```diff
@@ -0,0 +1,314 @@
...
+        prefix: BaseFieldElement,
+        target: LotteryTargetValue,
+    ) -> StmResult<()> {
+        if lottery_index > m {
+            return Err(SignatureError::IndexBoundFailed(lottery_index, m).into());
+        }
+
+        let lottery_index_as_base_field_element = BaseFieldElement::from(lottery_index);
+        let (commitment_point_x, commitment_point_y) = signature.commitment_point.get_coordinates();
+        let lottery_evaluation = compute_poseidon_digest(&[
```
damrobi: The function `compute_poseidon_digest` adds a `DST_SIGNATURE` to the input when it's called so it won't match the circuit. We can change that function to not add a DST when it's called, since we already added the `DST_LOTTERY` to the inputs here. We'll need to modify the unique schnorr signature to compensate.

mithril-stm/src/proof_system/halo2_snark/signer.rs
```diff
@@ -0,0 +1,314 @@
...
+
+    /// Computes the lottery prefix by hashing the message with the lottery DST.
+    /// The prefix is computed by prepending `DST_LOTTERY` to the message and hashing the result
+    /// using `compute_poseidon_digest`.
+    pub(super) fn compute_lottery_prefix(
+        message_as_base_field_element: &[BaseFieldElement],
+    ) -> BaseFieldElement {
+        let mut prefix = vec![DST_LOTTERY];
+        prefix.extend_from_slice(message_as_base_field_element);
+        compute_poseidon_digest(&prefix)
```
damrobi: Same problem with the `compute_poseidon_digest` adding the `DST_SIGNATURE` as in `verify_lottery_eligibility`.

mithril-stm/src/proof_system/halo2_snark/signer.rs
```diff
@@ -0,0 +1,314 @@
...
+                )
+                .is_ok()
+            })
+            .ok_or_else(|| SignatureError::LotteryLost.into())
+    }
+
+    /// Computes the lottery prefix by hashing the message with the lottery DST.
+    /// The prefix is computed by prepending `DST_LOTTERY` to the message and hashing the result
+    /// using `compute_poseidon_digest`.
+    pub(super) fn compute_lottery_prefix(
```
damrobi: This function and the one below `verify_lottery_eligibility` are also defined in the `eligibility.rs` that will soon be merged so they can be removed from there if we judge this location to be better for them.

mithril-stm/src/proof_system/halo2_snark/single_signature.rs
curiecrypt: @hjeljeli32 thanks for your comment.
However, we had an extensive discussion related to this topic in this Slack [thread](https://input-output-rnd.slack.com/archives/C09RY8KEYUF/p1771423501790559)

So, basically, the plan is to store the minimum winning index in the snark single signature. In the aggregation, winning indices will be computed for all valid signatures. This trick will be useful in several aspects. The following are some of those:
- Individual signers will spend less time on signature generation since they won't explore all possible indices; they can stop whenever they find one (meaning that that signature has at least 1 winning lottery). 
- We will be able to transfer less data from signer to aggregator. 
- The aggregator won't need to check the validity of all indices for all signatures. 

Do you still have any concerns about the inconsistency between the schemes?

mithril-stm/src/proof_system/halo2_snark/signer.rs
curiecrypt: I agree. We do not actually need to store the entire Merkle tree in any of the individual signers. Furthermore, a Merkle tree can be built from closed key registration whenever needed. 
WDYT @jpraynaud

mithril-stm/src/proof_system/halo2_snark/signer.rs
curiecrypt: I think storing those functions in signer makes more sense. I will update the eligibility accordingly.

curiecrypt: I changed my mind. I store them in eligibility, outisde `num_integer` gate

curiecrypt:
@hjeljeli32 @damrobi 
Changes regarding the comments related to circuit compatibility of `build_snark_message`, `DST_LOTTERY`, and `compute_poseidon_digest`:

- Merkle root encoding: `build_snark_message` now uses `from_raw` instead of `from_bytes`, matching the circuit's field element encoding.
- `DST_LOTTERY` mismatch: CPU's `DST_LOTTERY` has been updated to `from_raw([3, 3, 0, 0])`, aligned with the circuit's value in constants.rs. After deciding the final DST values, dependent golden conversions of any changed DST will need to be updated accordingly.
- `compute_poseidon_digest` auto-prepending DST: Already resolved; the function no longer silently prepends `DST_SIGNATURE`.
Additionally, 4 circuit-compatibility tests were added (`mod circuit_compatibility` in `halo2_snark/mod.rs`) that verify the CPU's intermediate values (message encoding, Schnorr challenge ordering, lottery prefix, lottery evaluation) match the circuit's expectations. These serve as a regression guard against future drift and can be extended as the CPU implementation evolves.

mithril-stm/src/membership_commitment/merkle_tree/commitment.rs
curiecrypt: see [my comment](https://github.com/input-output-hk/mithril/pull/3009#issuecomment-3937249491)

mithril-stm/src/proof_system/halo2_snark/signer.rs
curiecrypt: see [my comment](https://github.com/input-output-hk/mithril/pull/3009#issuecomment-3937249491)

mithril-stm/src/proof_system/halo2_snark/signer.rs
curiecrypt: see [my comment](https://github.com/input-output-hk/mithril/pull/3009#issuecomment-3937249491)

mithril-stm/src/signature_scheme/unique_schnorr_signature/jubjub/poseidon_digest.rs
curiecrypt: see [my comment](https://github.com/input-output-hk/mithril/pull/3009#issuecomment-3937249491).
Also, let's decide on the dst values and make the related revisions. You can initiate it @hjeljeli32

mithril-stm/src/proof_system/halo2_snark/signer.rs
curiecrypt: done

mithril-stm/src/membership_commitment/merkle_tree/commitment.rs
```diff
@@ -69,6 +72,32 @@ impl<D: Digest + FixedOutput, L: MerkleTreeLeaf> MerkleTreeCommitment<D, L> {
...
+    /// # Error
+    /// Returns an error if the root or the message is not exactly 32 bytes or either cannot be
+    /// converted to a `BaseFieldElement`.
+    pub fn build_snark_message(&self, message: &[u8]) -> StmResult<[BaseFieldElement; 2]> {
+        let root = self.root.as_slice();
+
+        let root_bytes: [u8; 32] = root
+            .try_into()
+            .with_context(|| "Merkle tree root must be exactly 32 bytes.")?;
+        let root_as_base_field_element = BaseFieldElement::from_raw(&root_bytes)
```
damrobi: I think we can change that to `from_bytes` as the merkle root should be a representation of a `BaseFieldElement`. `from_bytes` can fail if it's not the case which might be something we want?

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
```diff
@@ -7,7 +7,7 @@ cfg_num_integer! {
     #[cfg(feature = "future_snark")]
     use crate::{
         LotteryIndex, LotteryTargetValue, SignatureError, Stake, StmResult, UniqueSchnorrSignature,
-        signature_scheme::{BaseFieldElement, compute_poseidon_digest, DOMAIN_SEPARATION_TAG_LOTTERY},
+        signature_scheme::{BaseFieldElement, compute_poseidon_digest, DST_LOTTERY},
```
damrobi: We updated the `DST_LOTTERY` in a previous PR to `DOMAIN_SEPARATION_TAG_LOTTERY` and I think it was reverted here

mithril-stm/src/signature_scheme/unique_schnorr_signature/jubjub/poseidon_digest.rs
```diff
@@ -11,19 +11,18 @@ pub(crate) const DST_SIGNATURE: JubjubBase = JubjubBase::from_raw([
...
 
-#[cfg(feature = "future_snark")]
 /// Domain Separation Tag (DST) for the lottery check. It is used as a prefix when computing
 /// the eligibility value of a signature.
-// TODO: remove this allow dead_code directive when function is called or future_snark is activated
-#[allow(dead_code)]
-pub const DOMAIN_SEPARATION_TAG_LOTTERY: BaseFieldElement =
-    BaseFieldElement(JubjubBase::from_raw([3, 3, 0, 0]));
+#[cfg(feature = "future_snark")]
+pub const DST_LOTTERY: BaseFieldElement = BaseFieldElement(JubjubBase::from_raw([
```
damrobi: As mentioned before, this was updated to `DOMAIN_SEPARATION_TAG_LOTTERY`. If we want to keep this update, we might consider updating the others DSTs.

mithril-stm/src/membership_commitment/merkle_tree/commitment.rs
```diff
@@ -69,6 +72,32 @@ impl<D: Digest + FixedOutput, L: MerkleTreeLeaf> MerkleTreeCommitment<D, L> {
...
 
+    /// Build the message with merkle tree root and the given message to be used by the single
+    /// signature of the SNARK proof system. Both the root and the message are converted to
+    /// `BaseFieldElement` via `from_raw`, which interprets the bytes as a little-endian integer
+    /// and applies modular reduction.
+    ///
+    /// # Error
+    /// Returns an error if the root or the message is not exactly 32 bytes or either cannot be
+    /// converted to a `BaseFieldElement`.
+    pub fn build_snark_message(&self, message: &[u8]) -> StmResult<[BaseFieldElement; 2]> {
```
jpraynaud: The Merkle tree implementation should be agnostic of what is needed for signature. This function likely belongs to the `halo2_snark` proof system module and should take a merkle root as argument.

mithril-stm/src/membership_commitment/merkle_tree/commitment.rs
```diff
@@ -90,6 +119,16 @@ impl<D: Digest + FixedOutput, L: MerkleTreeLeaf> MerkleTreeCommitment<D, L> {
...
 
+#[cfg(feature = "future_snark")]
+impl<D: Digest, L: MerkleTreeLeaf> PartialEq for MerkleTreeCommitment<D, L> {
+    fn eq(&self, other: &Self) -> bool {
+        self.root == other.root
+    }
+}
+
+#[cfg(feature = "future_snark")]
+impl<D: Digest, L: MerkleTreeLeaf> Eq for MerkleTreeCommitment<D, L> {}
```
jpraynaud: Maybe derive implementation is enough here. Is there any reason not to use them?

mithril-stm/src/proof_system/halo2_snark/aggregate_key.rs
```diff
@@ -1,29 +1,41 @@
...
+    }
+}
+
+impl<D: MembershipDigest> PartialEq for AggregateVerificationKeyForSnark<D> {
+    fn eq(&self, other: &Self) -> bool {
+        self.merkle_tree_commitment == other.merkle_tree_commitment
     }
+}
+
+impl<D: MembershipDigest> Eq for AggregateVerificationKeyForSnark<D> {}
```
jpraynaud: Same remark as above for the usage of derived `PartialEq`/`Eq`.

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
```diff
@@ -152,52 +152,56 @@ cfg_num_integer! {
...
+/// `ev = Poseidon(prefix, commitment_point_x, commitment_point_y, index)` where
+/// `(commitment_point_x, commitment_point_y)` are coordinates of signature's commitment point.
+#[cfg(feature = "future_snark")]
+pub(crate) fn verify_lottery_eligibility(
+    signature: &UniqueSchnorrSignature,
+    lottery_index: LotteryIndex,
+    m: u64,
+    prefix: BaseFieldElement,
+    target: LotteryTargetValue,
+) -> StmResult<()> {
```
jpraynaud: I guess that we need to differentiate error cases from the case where a lottery is lost for an index. It is also mis-leading to name the error with the same name `LotteryLost` as a variant which has different meaning in Concatenation proof system
```suggestion
) -> StmResult<bool> {
```

mithril-stm/src/proof_system/halo2_snark/mod.rs
```diff
@@ -1,3 +1,435 @@
...
+
+                // CPU encoding via build_snark_message (uses from_raw)
+                let [root_cpu, _] = commitment.build_snark_message(&msg).unwrap();
+
+                // Circuit encoding: golden helpers use from_bytes for the root
+                let root_array: [u8; 32] = root_bytes
+                    .try_into()
+                    .expect("root should be 32 bytes");
+                let root_circuit = BaseFieldElement::from_bytes(&root_array)
+                    .expect("Poseidon root must be canonical");
```
jpraynaud, line 148: It'd be a good idea to create 
```suggestion
                // Circuit encoding: golden helpers use from_bytes for the root
                assert_eq!(32, root_bytes.len());
                let root_circuit = BaseFieldElement::from_bytes(&root_bytes)
                    .expect("Poseidon root must be canonical");
```

mithril-stm/src/proof_system/halo2_snark/mod.rs
```diff
@@ -1,3 +1,435 @@
...
+    }
+
+    /// Circuit-compatibility tests for the SNARK signature flow.
+    ///
+    /// These tests verify that CPU-side signature generation and verification
+    /// produce intermediate values compatible with the circuit in
+    /// `circuits::halo2`. They serve as a **compatibility contract**: if the
+    /// CPU code diverges from the circuit's expectations, these tests will
+    /// catch it.
+    mod circuit_compatibility {
```
jpraynaud, line 104: I'm not sure that this is the best place to put all of these tests :thinking:

mithril-stm/src/proof_system/halo2_snark/signer.rs
```diff
@@ -0,0 +1,254 @@
...
+    ///
+    /// Computes the lottery prefix from the message, then iterates over each index
+    /// to verify eligibility. Returns the first winning index, or
+    /// `SignatureError::LotteryLost` if no index is won.
+    fn check_lottery(
+        m: u64,
+        msg: &[BaseFieldElement],
+        signature: &UniqueSchnorrSignature,
+        lottery_target_value: LotteryTargetValue,
+    ) -> StmResult<LotteryIndex> {
```
jpraynaud: This should probably return the list of won lottery indices (as stated by the domment "Checks the lottery for all indices `0..m`."). We can later choose to keep the first index if it makes sense.

mithril-stm/src/protocol/aggregate_signature/aggregate_key.rs
```diff
@@ -24,12 +36,30 @@ impl<D: MembershipDigest> AggregateVerificationKey<D> {
...
+        {
+            conct_eq
+                && self.to_snark_aggregate_verification_key()
+                    == other.to_snark_aggregate_verification_key()
+        }
+
+        #[cfg(not(feature = "future_snark"))]
+        conct_eq
     }
 }
```
jpraynaud: This could probably be implemented with derive.

mithril-stm/src/protocol/aggregate_signature/clerk.rs
```diff
@@ -74,11 +74,15 @@ impl<D: MembershipDigest> Clerk<D> {
 
     /// Compute the aggregate verification key.
     /// It computes only the concatenation aggregate verification key for now.
-    /// Needs to be revised when implementing SNARK pre-aggregation primitives
+    /// Needs to be revised when implementing SNARK aggregation primitives
+    /// TODO: Replace None with the actual SNARK verification key when implementing
+    /// SNARK aggregation primitives.
```
jpraynaud:
```suggestion
    // TODO: Replace None with the actual SNARK verification key when implementing
    // SNARK aggregation primitives.
```

mithril-stm/src/protocol/key_registration/closed_registration_entry.rs
```diff
@@ -162,15 +162,16 @@ impl Serialize for ClosedRegistrationEntry {
...
 /// TODO: Compute the lottery target value based on the total stake and the entry's stake.
 impl From<(RegistrationEntry, Stake)> for ClosedRegistrationEntry {
     fn from(entry_total_stake: (RegistrationEntry, Stake)) -> Self {
         let (entry, _total_stake) = entry_total_stake;
         #[cfg(feature = "future_snark")]
         let (schnorr_verification_key, target_value) = {
             let vk = entry.get_verification_key_for_snark();
-            let target = vk.map(|_| LotteryTargetValue::get_one());
+            let target =
+                vk.map(|_| &LotteryTargetValue::default() - &LotteryTargetValue::get_one());
```
jpraynaud, line 174: This can probably be computed accurately now that https://github.com/input-output-hk/mithril/issues/2974 is completed.

mithril-stm/src/protocol/participant/initializer.rs
```diff
@@ -90,26 +95,54 @@ impl Initializer {
             )
             .ok_or_else(|| anyhow!(RegisterError::UnregisteredInitializer))?;
 
-        let key_registration_commitment = closed_key_registration
-            .to_merkle_tree::<D::ConcatenationHash, RegistrationEntryForConcatenation>();
-
-        // Create concatenation proof signer
+        // --- Concatenation proof signer creation ---
```
jpraynaud: This comment is not necessary

mithril-stm/src/protocol/participant/initializer.rs
```diff
@@ -90,26 +95,54 @@ impl Initializer {
...
             registration_entry.get_stake(),
             closed_key_registration.total_stake,
             self.parameters,
             self.bls_signing_key,
             self.bls_verification_key_proof_of_possession.vk,
-            key_registration_commitment,
+            key_registration_commitment_for_concatenation,
         );
 
+        // ------- Snark proof signer creation -------
```
jpraynaud: Same for this comment

mithril-stm/src/protocol/single_signature/signature_registered_party.rs
```diff
@@ -111,9 +120,68 @@ mod tests {
         ]
         "#;
 
+        #[cfg(feature = "future_snark")]
+        const GOLDEN_JSON: &str = r#"
```
jpraynaud, line 124: We need to be extra careful with this encoding as it embeds the problematic serde encoding with arrays. Not sure that this will not introduce a breaking change :thinking:

mithril-stm/src/signature_scheme/unique_schnorr_signature/jubjub/poseidon_digest.rs
```diff
@@ -40,8 +39,8 @@ mod test {
         use crate::signature_scheme::{BaseFieldElement, compute_poseidon_digest};
 
         const GOLDEN_BYTES: &[u8; 32] = &[
-            101, 79, 70, 206, 167, 240, 180, 227, 96, 239, 132, 120, 73, 204, 173, 158, 7, 70, 226,
-            165, 250, 3, 119, 186, 80, 22, 132, 135, 21, 128, 227, 98,
+            168, 249, 251, 86, 251, 84, 26, 25, 244, 24, 31, 5, 153, 234, 64, 53, 32, 67, 114, 26,
+            250, 90, 108, 196, 47, 25, 60, 234, 221, 176, 255, 92,
```
jpraynaud, line 48: Why do this golden value need to be modified?

mithril-stm/src/proof_system/halo2_snark/single_signature.rs
```diff
@@ -1,39 +1,91 @@
...
 
 /// Single signature for the Snark proof system.
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub(crate) struct SingleSignatureForSnark {
     /// The underlying Schnorr signature
     schnorr_signature: UniqueSchnorrSignature,
-    /// The index(es) for which the signature is valid
-    indices: Vec<LotteryIndex>,
+    /// The minimum winning lottery index for which the signature is valid
+    minimum_winning_lottery_index: LotteryIndex,
```
jpraynaud: The single signature should include a list of indices instead of the minimum lottery index. This structure can support a broader range of cases (no index as originally planned, first won index and all indices), in particular for codecs backward compatibility. This will probably also avoid complications in the witness preparation phase of the circuit.

mithril-stm/src/proof_system/halo2_snark/single_signature.rs
```diff
@@ -1,39 +1,91 @@
...
+            .verify(&message_to_verify, verification_key)
+            .with_context(|| "Schnorr signature verification failed for SNARK proof system.")?;
+
+        self.verify_winning_lottery_index(*lottery_target_value, &message_to_verify, parameters.m)?;
+
+        Ok(())
+    }
+
+    /// Verifies that the lottery index associated with this signature actually won the lottery.
+    fn verify_winning_lottery_index(
```
jpraynaud: We probably need a `compute_winning_lottery_indices` function that takes a range of lottery index and computes the winning ones, and use it here for verification (and later in the circuit witness preparation)

mithril-stm/src/proof_system/halo2_snark/single_signature.rs
jpraynaud, line 1: Maybe computing only the first winning index is a bit of an over optimization at this stage and could be addressed once we have a first working end to end prototype?

mithril-stm/src/protocol/single_signature/signature.rs
```diff
@@ -45,7 +60,7 @@ impl SingleSignature {
         )
     }
 
-    /// Verify that all indices of a signature are valid.
+    /// Verify that all indices of a concatenation signature are valid.
```
jpraynaud: This comment lacks clarity, maybe it's worth explaining a bit more what's the impact of SNARKs here.

mithril-stm/src/protocol/single_signature/signature.rs
```diff
@@ -77,6 +95,15 @@ impl SingleSignature {
...
 
         output.extend_from_slice(&self.signer_index.to_be_bytes());
+
+        #[cfg(feature = "future_snark")]
+        if let Some(snark_signature) = &self.snark_signature {
+            let unique_schnorr_signature = snark_signature.get_schnorr_signature();
+            let minimum_winning_lottery_index = snark_signature.get_minimum_winning_lottery_index();
+            output.extend_from_slice(&unique_schnorr_signature.to_bytes());
+            output.extend_from_slice(&minimum_winning_lottery_index.to_be_bytes());
+        }
```
jpraynaud, line 113: This encoding is not future proof and will likely create backward compatibility issues.

mithril-stm/src/proof_system/halo2_snark/single_signature.rs
hjeljeli32: Thanks @curiecrypt for explaining the context and direction.
I think the plan will be optimal at several aspects as you explained.
 
Just to double-confirm the flow:
- signer provides only one winning index
- aggregator verifies each single signature
- aggregator derives enough winning indices off-circuit and populate quorum witness tuples.
Is my understanding correct?

IMO, your proposed plan is **compatible with current circuit constraints as-is**.

curiecrypt: > Is my understanding correct?

Exactly! 

Thanks @hjeljeli32

mithril-stm/src/membership_commitment/merkle_tree/commitment.rs
curiecrypt: derived `Eq`, `PartialEq` instead of custom implementation

mithril-stm/src/protocol/aggregate_signature/aggregate_key.rs
curiecrypt: derived `Eq`, `PartialEq` instead of custom implementation

mithril-stm/src/signature_scheme/unique_schnorr_signature/jubjub/poseidon_digest.rs
curiecrypt: Because I modified how we compute the poseidon digest. It used to prepend DST_SIGNATURE by default. Now, the DST is included in the input array.

mithril-stm/src/proof_system/halo2_snark/aggregate_key.rs
curiecrypt: derived `Eq`, `PartialEq` instead of custom implementation

mithril-stm/src/proof_system/halo2_snark/mod.rs
curiecrypt: resolved.

mithril-stm/src/protocol/aggregate_signature/clerk.rs
curiecrypt: addressed

mithril-stm/src/protocol/participant/initializer.rs
curiecrypt: addressed

mithril-stm/src/protocol/participant/initializer.rs
curiecrypt: addressed

mithril-stm/src/protocol/single_signature/signature.rs
curiecrypt: written a clearer comment.

mithril-stm/src/membership_commitment/merkle_tree/commitment.rs
curiecrypt: moved `build_snark_message` to `halo2_snark`

mithril-stm/src/proof_system/halo2_snark/signer.rs
curiecrypt: Moved to `mod.rs`. Returns `StmResult<Vec<LotteryIndex>>`.

mithril-stm/src/proof_system/halo2_snark/single_signature.rs
curiecrypt: it is modified as `indices: Vec<LotteryIndex>,`

mithril-stm/src/proof_system/halo2_snark/single_signature.rs
curiecrypt: I modified the related implementations to stick with the initial plan.

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
curiecrypt: Modified to return `StmResult<bool>`. 
In addition, it does not throw the error `LotteryLost`. Instead, it directly returns: `Ok(lottery_evaluation <= target)`.

mithril-stm/src/proof_system/halo2_snark/mod.rs
curiecrypt: The tests located here are completely removed. We probably consider adding some tests to the circuit module for checking the compatibility between the circuit and CPU computations. Let's discuss it on grooming session.

mithril-stm/src/protocol/key_registration/closed_registration_entry.rs
curiecrypt: As agreed, this will be handled in another branch.

mithril-stm/src/proof_system/halo2_snark/single_signature.rs
curiecrypt: Added `compute_winning_lottery_indices` function to `eligibility.rs` that returns a list of all winning indices or an error if there is no winning index. 
The `verify_winning_lottery_index` function is renamed as `check_lottery_for_index`. This function is called in the loop from 0 to m to check whether the lottery is won for the current index by `compute_winning_lottery_indices`.

mithril-stm/src/signature_scheme/unique_schnorr_signature/jubjub/poseidon_digest.rs
curiecrypt: Note that the golden values must be modified when the DST PR is merged.

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
curiecrypt: We will revise them all at once when the DST PR is merged.

mithril-stm/src/signature_scheme/unique_schnorr_signature/jubjub/poseidon_digest.rs
curiecrypt: same as above.

mithril-stm/src/proof_system/halo2_snark/single_signature.rs
curiecrypt: FYI: We'll go with the initial plan in this phase: Single signature structure has the field `Indices` with type `Vec<LotteryIndex>`. The signature will store an empty vector of `LotteryIndex`s for now. We will consider further optimizations in the following phase.

mithril-stm/src/membership_commitment/merkle_tree/commitment.rs
curiecrypt: I fixed this conversion, and the conversion of message bytes in the same manner. Now, they both use `from_bytes` instead of `from_raw` since both the root and the message are already compatible with `BaseFieldElement` limitations.

mithril-stm/src/protocol/single_signature/signature_registered_party.rs
curiecrypt: Will be handled in _Prepare refactor all bytes codec in STM library_

mithril-stm/src/protocol/single_signature/signature.rs
curiecrypt: Will be handled in _Prepare refactor all bytes codec in STM library_

mithril-stm/src/protocol/participant/signer.rs
```diff
@@ -24,23 +33,42 @@ impl<D: MembershipDigest> Signer<D> {
...
+    /// * a snark proof signer is available.
     pub fn create_single_signature(&self, message: &[u8]) -> StmResult<SingleSignature> {
         let concatenation_signature =
             self.concatenation_proof_signer.create_single_signature(message)?;
+
+        // TODO: Update this rng. We might want to use a different RNG for the snark signature generation, or pass it as an argument to the function.
+        #[cfg(feature = "future_snark")]
+        let snark_signature = if let Some(snark_signer) = &self.snark_proof_signer {
+            let mut rng = rand_core::OsRng;
+            Some(snark_signer.create_single_signature(message, &mut rng)?)
```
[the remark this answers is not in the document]
curiecrypt: option c applied. instead of manual impl of `ok()`, used: `snark_signer.create_single_signature(message, &mut rng).ok()`

mithril-stm/README.md
```diff
@@ -131,7 +131,7 @@ let avk: AggregateVerificationKey<D>  = clerk.compute_aggregate_verification_key
 // Check all parties can verify every sig
 for s in sigs.iter() {
     let entry = closed_reg.get_registration_entry_for_index(&s.signer_index).unwrap();
-    assert!(s.verify::<D>(&params, &entry.get_verification_key_for_concatenation(), &entry.get_stake(), &avk, &msg).is_ok(), "Verification failed");
+    assert!(s.verify::<D>(&params, &entry.get_verification_key_for_concatenation(), &entry.get_stake(), &avk, &msg, #[cfg(feature = "future_snark")] None).is_ok(), "Verification failed");
 }
```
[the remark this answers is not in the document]
curiecrypt: Disagree - should be revisited when updating all doc tests to also cover snark proof system.

mithril-stm/src/protocol/aggregate_signature/mod.rs
```diff
@@ -297,14 +297,14 @@ mod tests {
 
         #[test]
         /// Test that when a party creates a signature it can be verified
-        fn test_sig(msg in any::<[u8;16]>()) {
+        fn test_sig(msg in any::<[u8;32]>()) {
             let params = Parameters { m: 1, k: 1, phi_f: 0.2 };
             let ps = setup_equal_parties(params, 1);
```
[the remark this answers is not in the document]
curiecrypt: In another ticket

mithril-stm/src/protocol/aggregate_signature/mod.rs
```diff
@@ -203,7 +203,7 @@ mod tests {
         fn test_aggregate_sig(nparties in 2_usize..30,
                               m in 10_u64..20,
                               k in 1_u64..5,
-                              msg in any::<[u8;16]>()) {
+                              msg in any::<[u8;32]>()) {
```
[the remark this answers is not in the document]
curiecrypt: In another ticket

mithril-stm/src/protocol/aggregate_signature/mod.rs
```diff
@@ -371,7 +371,7 @@ mod tests {
         /// Test that when the adversaries do not hold sufficient stake, they can not form a quorum
         fn test_adversary_quorum(
             (adversaries, parties) in arb_parties_adversary_stake(8, 30, 16, 4),
-            msg in any::<[u8;16]>(),
+            msg in any::<[u8;32]>(),
         ) {
```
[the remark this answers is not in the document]
curiecrypt: In another ticket

mithril-stm/src/protocol/aggregate_signature/mod.rs
```diff
@@ -341,13 +341,13 @@ mod tests {
...
+                assert!(sig_deser.verify(&params, &ps[0].get_bls_verification_key(), &ps[0].concatenation_proof_signer.stake, &avk, &msg, #[cfg(feature = "future_snark")] None).is_ok());
             }
         }
 
         #[test]
         fn test_multisig_serialize_deserialize(nparties in 2_usize..10,
-                                          msg in any::<[u8;16]>()) {
+                                          msg in any::<[u8;32]>()) {
             let params = Parameters { m: 10, k: 5, phi_f: 1.0 };
             let ps = setup_equal_parties(params, nparties);
```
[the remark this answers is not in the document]
curiecrypt: In another ticket

mithril-stm/src/protocol/aggregate_signature/mod.rs
```diff
@@ -332,7 +332,7 @@ mod tests {
         }
 
         #[test]
-        fn test_sig_serialize_deserialize(msg in any::<[u8;16]>()) {
+        fn test_sig_serialize_deserialize(msg in any::<[u8;32]>()) {
             let params = Parameters { m: 1, k: 1, phi_f: 0.2 };
             let ps = setup_equal_parties(params, 1);
```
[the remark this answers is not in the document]
curiecrypt: In another ticket

mithril-stm/src/protocol/aggregate_signature/clerk.rs
```diff
@@ -74,11 +74,14 @@ impl<D: MembershipDigest> Clerk<D> {
...
-    /// Needs to be revised when implementing SNARK pre-aggregation primitives
+    // TODO: Replace None with the actual SNARK verification key when implementing
+    // SNARK aggregation primitives.
     pub fn compute_aggregate_verification_key(&self) -> AggregateVerificationKey<D> {
         AggregateVerificationKey::new(
             self.concatenation_proof_clerk
                 .compute_aggregate_verification_key_for_concatenation(),
+            #[cfg(feature = "future_snark")]
+            None,
         )
```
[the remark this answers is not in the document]
curiecrypt: will be done in the next phase

mithril-stm/src/protocol/participant/signer.rs
```diff
@@ -24,23 +33,42 @@ impl<D: MembershipDigest> Signer<D> {
...
+        } else {
```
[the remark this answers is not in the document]
curiecrypt: In another ticket

mithril-stm/src/proof_system/halo2_snark/single_signature.rs
```diff
@@ -26,14 +29,40 @@ impl SingleSignatureForSnark {
...
 
-    /// Set `indices` of single signature to given `indices`
-    pub(crate) fn set_indices(&mut self, indices: &[LotteryIndex]) {
-        self.indices = indices.to_vec()
+    /// Set `indices` of single signature to given value
+    // TODO: remove this allow dead_code directive when function is called or future_snark is activated
+    #[allow(dead_code)]
+    pub(crate) fn set_indices(&mut self, indices: Vec<LotteryIndex>) {
+        self.indices = indices;
     }
```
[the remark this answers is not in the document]
curiecrypt: fixed

mithril-stm/src/proof_system/halo2_snark/signer.rs
```diff
@@ -0,0 +1,228 @@
...
+        let signature = self.signing_key.sign(&message_to_sign, rng)?;
+
+        compute_winning_lottery_indices(
+            self.parameters.m,
+            &message_to_sign,
+            &signature,
+            self.lottery_target_value,
+        )?;
+
+        Ok(SingleSignatureForSnark::new(signature, vec![]))
```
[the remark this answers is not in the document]
curiecrypt: fixed - adjust the field + docs to reflect that indices are intentionally not carried at this stage

mithril-stm/benches/size_benches.rs
```diff
@@ -78,7 +68,7 @@ fn main() {
 
     let params: [(u64, u64, usize); 2] = [(445, 2728, 3000), (554, 3597, 3000)];
     for (k, m, nparties) in params {
-        size::<MembershipDigestU64>(k, m, nparties, "Blake2b 512");
+        size::<MithrilMembershipDigest>(k, m, nparties, "Blake2b 512");
         size::<MithrilMembershipDigest>(k, m, nparties, "Blake2b 256");
```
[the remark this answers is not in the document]
curiecrypt: fixed - revert back the benchmark only membership structure (with a more descriptive name).
concatenation hash is overwritten with `Blake2b<U64>`, snark hash is kept as MithrilPoseidonDigest.
To import MithrilPoseidonDigest in this bench file, `lib.rs` is updated to publicly export it for benchmark internals and future snark features are enabled.

mithril-stm/src/protocol/aggregate_signature/clerk.rs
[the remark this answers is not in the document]
curiecrypt: will be handled in the next phase

mithril-stm/src/protocol/participant/signer.rs
```diff
@@ -24,23 +33,42 @@ impl<D: MembershipDigest> Signer<D> {
...
+    /// * a snark proof signer is available.
     pub fn create_single_signature(&self, message: &[u8]) -> StmResult<SingleSignature> {
         let concatenation_signature =
             self.concatenation_proof_signer.create_single_signature(message)?;
+
+        // TODO: Update this rng. We might want to use a different RNG for the snark signature generation, or pass it as an argument to the function.
+        #[cfg(feature = "future_snark")]
+        let snark_signature = if let Some(snark_signer) = &self.snark_proof_signer {
+            let mut rng = rand_core::OsRng;
+            snark_signer.create_single_signature(message, &mut rng).ok()
```
hjeljeli32, line 62: If I understood correctly, SNARK signature creation uses .ok() and drops all errors. Therfore, caller gets “success” with snark_signature=None without knowing SNARK path failed.

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
```diff
@@ -142,52 +142,81 @@ cfg_num_integer! {
...
+/// `ev = Poseidon(prefix, commitment_point_x, commitment_point_y, index)` where
+/// `(commitment_point_x, commitment_point_y)` are coordinates of signature's commitment point.
+#[cfg(feature = "future_snark")]
+pub(crate) fn check_lottery_for_index(
+    signature: &UniqueSchnorrSignature,
+    lottery_index: LotteryIndex,
+    m: u64,
+    prefix: BaseFieldElement,
+    target: LotteryTargetValue,
+) -> StmResult<bool> {
```
hjeljeli32, line 193: This function returns StmResult<bool>, but the tests in mod.rs still use .is_ok(). I think the tests need to be updated.

hjeljeli32 (approved):
LGTM. I just left 2 minor comments about error handling.

mithril-stm/src/protocol/participant/signer.rs
jpraynaud: I agree with @hjeljeli32 this is not ideal. The problem is that we don't want a failing SNARK signature failing because no indices was won to prevent the signature for Concatenation.

Ideally, the `create_single_signature` of the SNARK signer would return a `StdResult<Option<_>>`
- An Error is returned when something is wrong with the cryptographic signature 
- `Ok(None)` is returned when no lottery is won
- `OK(Some(_))` otherwise

We'll do that in a separate PR :+1:

mithril-stm/src/proof_system/halo2_snark/eligibility.rs
curiecrypt: You are right. I'll fix them as the other suggestion in a separate PR. 👍

Issue #3035, opened 2026-02-23
Refactor SNARK circuit - Error handling


jpraynaud:
## Why

As described in the [SNARK MVP Book](https://hackmd.io/EeCC7vaPQoWdmMVAkyDQpQ?view=#Refactor-Halo2-circuit-for-production-readiness) (Workstream 3 - Error Handling), the Halo2 SNARK circuit prototype contains `panic!` and `assert!` paths that cause unrecoverable crashes instead of returning structured errors. Additionally, under some circumstances, the Midnight library panics. This makes the circuit fragile and not suitable for production.

## What

Replace all panic paths in the circuit with structured `Result` based error handling.

## How

- [x] Search for all `panic!`, `assert!`, `unwrap()`, and `expect()` occurrences in `circuit.rs` and `gadgets.rs`
- [x] Define a dedicated `CircuitError` enum covering all identified cases
- [x] Replace the `assert!(self.quorum < self.num_lotteries)` in `circuit.rs` with a `Result` 
- [x] Replace any remaining panic paths with `StmResult<_>` returns
- [x] For panics originating from the Midnight stdlib (witness size mismatches), add validation guards before calling Midnight functions to catch invalid inputs early and return circuit-level errors
- [x] Add negative tests for each new error
- [x] Measure circuit constraint count before and after to ensure no power-of-two boundary (`2^k`) increase
- [x] Update the [SNARK MVP Book](https://hackmd.io/EeCC7vaPQoWdmMVAkyDQpQ?view=#Refactor-Halo2-circuit-for-production-readiness) to reflect the changes made
- [x] Create an issue in Midnight repository to enhance their errors

hjeljeli32:
Dear @jpraynaud 
I created the following issue https://github.com/midnightntwrk/midnight-zk/issues/248 on midnight-zk repo.
Can you please review it?

PR #3055, opened 2026-02-27
refactor(stm): error handling

hjeljeli32:
## Content

This PR removes all remaining panic-based failure paths from the SNARK circuit prototype and replaces them with structured Result-based error handling.

Previously, the circuit relied on panic!, assert!, and implicit failures (including from the Midnight stdlib) for invalid inputs such as malformed witnesses or inconsistent parameters. This made the circuit fragile and unsuitable for production environments.

Changes
 - Introduced a dedicated CircuitError enum to represent circuit-level failures.
- Replaced `assert!(self.quorum < self.num_lotteries)` with explicit validation returning a typed error.
- Added validation guards to prevent Midnight stdlib panics (e.g., witness and Merkle path length mismatches).
- Replaced remaining panic paths in circuit.rs and gadgets.rs with Result-based returns.
- Updated negative tests to assert structured errors instead of relying on `#[should_panic]`.
- Verified that circuit constraint count emain unchanged.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation
  - [x] No new TODOs introduced

## Issue(s)
Closes https://github.com/input-output-hk/mithril/issues/3035

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4507,7 +4507,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-stm"
-version = "0.9.24"
+version = "0.9.25"
 dependencies = [
  "anyhow",
  "blake2 0.10.6",
--- a/mithril-stm/CHANGELOG.md
+++ b/mithril-stm/CHANGELOG.md
@@ -5,6 +5,16 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## 0.9.25 (03-06-2026)
+
+### Changed
+
+- Removed panic-based failure paths from the Halo2 SNARK circuit prototype and switched to structured `Result`-based error handling.
+- Replaced `assert!(self.quorum < self.num_lotteries)` with explicit validation returning a typed circuit error.
+- Added pre-validation guards to reject malformed witnesses and inconsistent Merkle path shapes before Midnight stdlib panics.
+- Reworked remaining failure paths in `circuits/halo2/circuit.rs` and `circuits/halo2/gadgets.rs` to return structured errors.
+- Updated Halo2 negative tests to assert structured errors rather than using `#[should_panic]`.
+
 ## 0.9.24 (03-04-2026)
 
 ### Added
--- a/mithril-stm/Cargo.toml
+++ b/mithril-stm/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mithril-stm"
-version = "0.9.24"
+version = "0.9.25"
 edition = { workspace = true }
 authors = { workspace = true }
 homepage = { workspace = true }


mithril-stm/src/circuits/halo2/circuit.rs
```diff
@@ -40,6 +45,53 @@ impl StmCircuit {
             merkle_tree_depth,
         }
     }
+
+    pub(crate) fn validate_parameters(&self) -> CircuitResult<()> {
```
curiecrypt: Should give doc comments to functions in this implementation

mithril-stm/src/circuits/halo2/errors.rs
```diff
@@ -0,0 +1,116 @@
...
+pub enum CircuitError {
+    /// Invalid relation parameters: quorum must be strictly lower than number of lotteries.
+    #[error(
+        "invalid circuit parameters: quorum ({quorum}) must be lower than num_lotteries ({num_lotteries})"
+    )]
+    InvalidCircuitParameters { quorum: u32, num_lotteries: u32 },
+
+    /// Witness vector length does not match the configured quorum.
+    #[error("witness length mismatch: expected quorum {expected_quorum}, got {actual}")]
+    WitnessLengthMismatch {
```
curiecrypt: I see why you used `usize` type for some enums. However, I think the error enum should store `u32` (for now, later we'll switch to u64) uniformly and cast to usize only at comparison time inside the validators.

mithril-stm/src/circuits/halo2/errors.rs
```diff
@@ -0,0 +1,116 @@
...
+        actual: usize,
+    },
+
+    /// The proving backend returned an error while executing this circuit.
+    #[error("circuit execution failed in proving backend")]
+    CircuitExecutionFailed,
+}
+
+impl CircuitError {
+    /// Reconstruct typed circuit errors from the raw synthesis payload emitted
```
curiecrypt: This deserves a better doc. Without it, a future reader will think this is a design choice rather than a forced workaround and may try to fix it.

mithril-stm/src/circuits/halo2/errors.rs
```diff
@@ -0,0 +1,116 @@
...
+    /// Merkle position path length does not match the configured Merkle depth.
+    #[error("merkle position length mismatch: expected depth {expected_depth}, got {actual}")]
+    MerklePositionLengthMismatch {
+        expected_depth: usize,
+        actual: usize,
+    },
+
+    /// The proving backend returned an error while executing this circuit.
+    #[error("circuit execution failed in proving backend")]
+    CircuitExecutionFailed,
```
curiecrypt: In `map_proving_backend_error`, the original error information is discarded when: 
- the synthesis message doesn't match any known pattern, or 
- the error isn't `PlonkError::Synthesis` at all.

We might consider `CircuitExecutionFailed(String)` to preserve the original message for diagnostics.

mithril-stm/src/circuits/halo2/circuit.rs
```diff
@@ -40,6 +45,53 @@ impl StmCircuit {
...
+        let expected_depth = self.merkle_tree_depth as usize;
+        if actual != expected_depth {
+            return Err(CircuitError::MerklePositionLengthMismatch {
+                expected_depth,
+                actual,
+            });
+        }
+
+        Ok(())
+    }
```
curiecrypt: Both functions validate `x.siblings.len()` against `merkle_tree_depth`. The siblings and positions are extracted from the same `x.siblings` vector. If one passes, the other always will. The position validation (line 178) in the circuit code can never fail independently. It's correct but worth a comment acknowledging this.

mithril-stm/src/circuits/halo2/mod.rs
```diff
@@ -2,6 +2,7 @@
 
 pub mod circuit;
 pub mod constants;
+pub mod errors;
```
curiecrypt: Do we really need to publicly expose the errors? Maybe `pub(crate)`? 
We should consider the visibility of other mods as well, probably in actual revision of circuit code.

mithril-stm/src/circuits/halo2/errors.rs
```diff
@@ -0,0 +1,116 @@
...
+}
+
+impl CircuitError {
+    /// Reconstruct typed circuit errors from the raw synthesis payload emitted
+    /// by `StmCircuit::circuit(...)` before Midnight wraps it in `plonk::Error`.
+    pub fn from_synthesis_message(message: &str) -> Option<Self> {
```
curiecrypt: As far as I understand, if someone edits an error message, this parser silently falls through to `CircuitExecutionFailed` with no compile or test run signal. 
We might consider adding a round-trip unit test that passes each variant through `to_string()`  - `from_synthesis_message()`.

mithril-stm/src/circuits/halo2/golden/helpers.rs
```diff
@@ -100,17 +101,21 @@ pub(crate) enum StmCircuitProofError {
...
     SignatureVerification,
-    /// Proof generation failed.
-    #[error("proof generation failed")]
-    ProveFail,
-    /// Proof verification failed.
-    #[error("proof verification failed")]
-    VerifyFail,
+    /// Proving/verification failures shared with Halo2 error model.
+    #[error(transparent)]
+    Proof(#[from] StmProofError),
```
curiecrypt: _nit:_
As far as I assume, the structure of the wrapping content here: `StmProofError` -> `ProovingError` -> `CircuitError`. 
Meaning that in the tests, we will have assertions with 4 levels. 
It does not cause a problem for sure, but it could make the test readability harder.

mithril-stm/src/circuits/halo2/golden/cases/negative.rs
```diff
@@ -1,40 +1,76 @@
...
-    LeafSelector, StmCircuitProofError, StmCircuitScenario, build_witness,
+    LOTTERIES_PER_QUORUM, LeafSelector, StmCircuitProofError, StmCircuitScenario, build_witness,
     build_witness_with_fixed_signer, build_witness_with_indices, create_default_merkle_tree,
     create_merkle_tree_with_leaf_selector, find_two_distinct_witness_entries,
     prove_and_verify_result, setup_stm_circuit_env,
 };
 use crate::circuits::halo2::types::{Position, SignedMessageWithoutPrefix};
 use crate::signature_scheme::{BaseFieldElement, ScalarFieldElement};
 
+fn assert_verification_failed(result: Result<(), StmCircuitProofError>) {
```
curiecrypt: _nit:_
The name sounds like it asserts that verification specifically failed, but it actually asserts the full prove-and-verify flow returned a verification phase error (for example: proof was generated successfully, then verification rejected it). 
Something like `assert_proof_rejected_by_verifier` would make the distinction clearer. Especially since other tests in this file assert proving phase failures.

curiecrypt (changes requested):
Good overall! Thank you @hjeljeli32 
- The move from panics to Result-based error handling is well done.
- The error hierarchy is clean. 

Can be merged after addressing the suggestions. The following are the more prior ones IMO:
- Use uniform data type in `CircuitError` fields. (`usize` vs. `u32` for now, `u64` later)
- Better document needed.  Especially for the Midnight error type workaround.
- Add a round-trip test for `from_synthesis_message` to catch silent fall to default case.

There are also some minor nits, nothing blocking.

mithril-stm/src/circuits/halo2/circuit.rs
jpraynaud: There is already a `StmResult` which should be used everywhere a `Result` is needed:
```suggestion
    pub(crate) fn validate_parameters(&self) -> StmResult<()> {
```

mithril-stm/src/circuits/halo2/circuit.rs
```diff
@@ -23,6 +24,10 @@ use crate::signature_scheme::{PrimeOrderProjectivePoint, UniqueSchnorrSignature}
 type F = JubjubBase;
 type C = Jubjub;
 
+fn synthesis_error(error: CircuitError) -> Error {
```
jpraynaud: This function should probably not be necessary as we follow these rules when implementing errors: https://mithril.network/doc/adr/6

mithril-stm/src/circuits/halo2/circuit.rs
```diff
@@ -57,7 +109,11 @@ impl Relation for StmCircuit {
         instance: Value<Self::Instance>,
         witness: Value<Self::Witness>,
     ) -> Result<(), Error> {
-        assert!(self.quorum < self.num_lotteries);
+        self.validate_parameters().map_err(synthesis_error)?;
+        let witness = witness.map_with_result(|witness| -> Result<_, Error> {
```
jpraynaud:
```suggestion
        let witness = witness.map_with_result(|witness| -> StmResult<_> {
```

mithril-stm/src/circuits/halo2/errors.rs
```diff
@@ -0,0 +1,116 @@
+use thiserror::Error;
+
+/// Circuit-scoped errors for Halo2 STM validation and execution.
+#[derive(Debug, Error, Clone, PartialEq, Eq)]
+pub enum CircuitError {
+    /// Invalid relation parameters: quorum must be strictly lower than number of lotteries.
+    #[error(
+        "invalid circuit parameters: quorum ({quorum}) must be lower than num_lotteries ({num_lotteries})"
```
jpraynaud: Could you capitalize all the error descriptions?
```suggestion
        "Invalid circuit parameters: quorum ({quorum}) must be lower than num_lotteries ({num_lotteries})"
```

mithril-stm/src/circuits/halo2/errors.rs
jpraynaud: It looks like this error is shadowing why the circuit failed. IMO we can probably simply skip it.

mithril-stm/src/circuits/halo2/errors.rs
jpraynaud: We really want to avoid relying on textual analysis of the underlying error messages (except when we have no other choice in the tests when we want to test that a specific error occured). We should just forward it to the caller with some anyhow context. 

If we consider that something could be enhanced in the error system of Midnight, I'd suggest to create an issue in their repository.

mithril-stm/src/circuits/halo2/errors.rs
```diff
@@ -0,0 +1,116 @@
...
+                    actual,
+                }
+            })
+        })
+    }
+}
+
+/// Proving-side error categories for Halo2 STM.
+#[derive(Debug, Error, Clone, PartialEq, Eq)]
+pub enum ProvingError {
```
jpraynaud: I'd suggest to keep a single `CircuitError` and to replace the `#[from]` errors with wrapping into anyhow error with context.

mithril-stm/src/circuits/halo2/circuit.rs
hjeljeli32: Thanks @curiecrypt for the suggestion. I added doc comments to the validation functions.

mithril-stm/src/circuits/halo2/circuit.rs
hjeljeli32: @curiecrypt 
You’re right. Both checks rely on x.siblings.len(), so the position validation cannot fail independently under the current witness structure. That's why I could not have a dedicated negative test that returns this error.
I added a short comment to clarify that this is a defensive guard and future-proofing.

mithril-stm/src/circuits/halo2/errors.rs
hjeljeli32: Agreed. I’ll align the error enum to use u32 consistently.

mithril-stm/src/circuits/halo2/errors.rs
hjeljeli32: Agreed. I will expand the doc to clarify the synthesis-message parsing is a workaround because of Midnight error surface.

mithril-stm/src/circuits/halo2/errors.rs
hjeljeli32: Agreed. I will store the original message in CircuitExecutionFailed(String) when we can’t map to a known typed error.

mithril-stm/src/circuits/halo2/mod.rs
hjeljeli32: Agreed. I will change it to pub(crate).

mithril-stm/src/circuits/halo2/golden/cases/negative.rs
hjeljeli32: Makes sense. I will update it accordingly.

mithril-stm/src/circuits/halo2/errors.rs
hjeljeli32: Makes sense. I added round-trip unit tests for for the different errors.

mithril-stm/src/circuits/halo2/golden/helpers.rs
hjeljeli32: Agree @curiecrypt I implemented a small readability cleanup for golden tests without changing behavior.
I kept the production error model as-is, but added helper assertions in golden helpers to unwrap proving-phase circuit errors and verifier rejections.
So tests no longer match the full nested chain (StmCircuitProofError -> StmProofError -> ProvingError -> CircuitError) each time.
Would you be okay with this direction?

mithril-stm/src/circuits/halo2/circuit.rs
hjeljeli32: @jpraynaud 
Sure I aligned with the guidelines defined in the adr and removed this function. Please let me know if there is still any gap.

mithril-stm/src/circuits/halo2/circuit.rs
hjeljeli32: @jpraynaud 
I aligned to ADR as far as possible (typed thiserror internally, StdResult aliases, context on subsystem boundaries, explicit Midnight adapter).
For circuit validation guards so far I kept Result<T, CircuitError>, because the Halo2 relation API requires plonk::Error at the boundary and we need typed circuit errors until that conversion point.
If you prefer naming consistency, I can rename CircuitResult to StmCircuitResult to reflect project style while preserving typed behavior.

I am worried that if we return StmResult in the circuit code, we will lose the typed circuit errors early as StmResult (anyhow) hides CircuitError structure and it would be harder to pattern-match the specific guard failures.

Please let me know your thoughts.

mithril-stm/src/circuits/halo2/circuit.rs
hjeljeli32: @jpraynaud Discussed in my previous answer.

mithril-stm/src/circuits/halo2/errors.rs
hjeljeli32: @jpraynaud Done.

mithril-stm/src/circuits/halo2/errors.rs
hjeljeli32: @jpraynaud 
Good point; I addressed it by preserving original details in `CircuitExecutionFailed(String)` instead of a reason-less error variant.
Is this okay?

mithril-stm/src/circuits/halo2/errors.rs
hjeljeli32: @jpraynaud 
I aligned the runtime path with your guidance: we no longer parse synthesis strings in production error mapping.
We now forward backend errors with explicit context (CircuitExecutionFailed(String)).
Textual analysis is kept only in tests where we need to assert specific guard failures.

mithril-stm/src/circuits/halo2/errors.rs
hjeljeli32: Sure @jpraynaud 
Runtime flow no longer depends on synthesis-string parsing; backend errors are forwarded with context

mithril-stm/src/circuits/halo2/circuit.rs
```diff
@@ -57,7 +121,11 @@ impl Relation for StmCircuit {
         instance: Value<Self::Instance>,
         witness: Value<Self::Witness>,
     ) -> Result<(), Error> {
-        assert!(self.quorum < self.num_lotteries);
+        self.validate_parameters().map_err(Error::from)?;
+        let witness = witness.map_with_result(|witness| -> Result<_, Error> {
+            self.validate_witness_length(witness.len()).map_err(Error::from)?;
```
damrobi: Is it possible to fuse this line with line 150 (`let witness = witness.transpose_vec(self.quorum as usize);`)? From what I understand this check on line 126 makes it impossible for the `transpose_vec` to fail but in this case it is handled separatly while the others (l.173 and l.186) perform both in one variable assignement.

damrobi (approved):
LGTM 👍 
I left one small comment about merge two lines.

mithril-stm/src/circuits/halo2/golden/helpers.rs
curiecrypt: I think this solves the readability problem in a smart way. Thanks @hjeljeli32

mithril-stm/src/circuits/halo2/circuit.rs
curiecrypt: Actually, in my first review I thought the same thing with @jpraynaud. On the other hand, I also had the concern of losing the typed circuit errors. This is why I did not write a suggestion related to this topic. 
However, it will be much better if we can use `StmResult` and still keep the circuit error content.

mithril-stm/src/circuits/halo2/circuit.rs
```diff
@@ -33,13 +34,76 @@ pub struct StmCircuit {
...
         Self {
             quorum,
             num_lotteries,
             merkle_tree_depth,
         }
     }
+
+    /// Validates global circuit parameters before synthesis.
+    /// Enforces `quorum < num_lotteries` returning
+    /// `StmCircuitError::InvalidCircuitParameters` when violated.
```
jpraynaud, line 74: This will give a line break in the cargo doc (you can apply it in other occurences):
```suggestion
    /// Validates global circuit parameters before synthesis.
    ///
    /// Enforces `quorum < num_lotteries` returning
    /// `StmCircuitError::InvalidCircuitParameters` when violated.
```

mithril-stm/src/circuits/halo2/errors.rs
```diff
@@ -0,0 +1,84 @@
...
+    )]
+    MerklePositionLengthMismatch { expected_depth: u32, actual: u32 },
+
+    /// The proving backend returned an error while executing this circuit.
+    #[error("Circuit execution failed in proving backend: {0}")]
+    CircuitExecutionFailed(String),
+}
+
+/// Result alias for Halo2 circuit-local operations.
+pub type StmCircuitResult<T> = Result<T, StmCircuitError>;
```
jpraynaud: There is already a `StmResult` in the STM library which relies on `anyhow`. Using it would avoid creating anther `Result` alias.

mithril-stm/src/circuits/halo2/errors.rs
```diff
@@ -0,0 +1,84 @@
...
+
+/// Boundary adapter required by Midnight's circuit API.
+///
+/// The circuit relation must return `plonk::Error`, so typed `StmCircuitError` values are
+/// converted only at that boundary.
+impl From<StmCircuitError> for PlonkError {
```
jpraynaud: This is a bit weird to convert the STM circuit error back to a Midnight proof system error. Is it really needed?

mithril-stm/src/circuits/halo2/circuit.rs
```diff
@@ -57,7 +121,11 @@ impl Relation for StmCircuit {
         instance: Value<Self::Instance>,
         witness: Value<Self::Witness>,
     ) -> Result<(), Error> {
```
jpraynaud, line 149: It'd be much more convenient if this function returned a `StmResult`.

mithril-stm/src/circuits/halo2/circuit.rs
jpraynaud: @hjeljeli32 is there any specific reason not to use `StmResult` here?

mithril-stm/src/circuits/halo2/golden/cases/negative.rs
```diff
@@ -293,20 +360,30 @@ fn merkle_path_length_short() {
...
     let scenario = StmCircuitScenario::new(merkle_root, msg, witness);
-    let result = prove_and_verify_result(&env, scenario);
-    assert!(matches!(result, Err(StmCircuitProofError::VerifyFail)));
+    assert_proving_backend_message_contains(
+        prove_and_verify_result(&env, scenario),
+        &format!(
+            "Circuit::validate_merkle_sibling_length failed: expected depth {expected_depth}, got {}",
+            expected_depth - 1
+        ),
+    );
```
jpraynaud, line 370: When the error downcasts to a typed error, it's more robust to check that instead of the text message of the error (which is what was done previously). This comment also applies to all other tests.

mithril-stm/src/circuits/halo2/golden/helpers.rs
```diff
@@ -47,69 +49,127 @@ type CircuitKeysCache = HashMap<StmCircuitConfig, Arc<CircuitVerificationAndProv
...
+    #[error("Proof flow failed: {source}")]
+    Proof {
+        #[source]
+        source: crate::StdError,
+    },
 }
 
 /// Result type for STM circuit golden helpers.
-type StmResult<T> = Result<T, StmCircuitProofError>;
+type GoldenResult<T> = Result<T, StmCircuitProofError>;
```
jpraynaud: Same comment as above for this `Result` alias.

mithril-stm/src/lib.rs
```diff
@@ -164,11 +164,17 @@ pub type LotteryIndex = u64;
...
-/// Mithril-stm error type
-pub type StmError = anyhow::Error;
+/// Standard crate-wide error type.
+pub type StdError = anyhow::Error;
+
+/// Standard crate-wide result type.
+pub type StdResult<T> = anyhow::Result<T, StdError>;
+
+/// Backward-compatible Mithril-stm error alias.
+pub type StmError = StdError;
```
jpraynaud: IMO this is not needed: it creates more aliases on `anyhow::Result<T, StmError>` which is not needed and prevents readability and maintainability.

mithril-stm/src/circuits/halo2/circuit.rs
hjeljeli32: Thanks @curiecrypt for your comment!
@jpraynaud 
I agree that it is not an easy decision. How about this: We can expose `StmResult` at module boundaries and keep `StmCircuitError` as source in the error chain, so we align with ADR and still preserve circuit-error granularity?

mithril-stm/src/circuits/halo2/circuit.rs
hjeljeli32: @jpraynaud 
This is related to your previous comment on `StmResult` usage.

In the current code, the validation guards still return typed `StmCircuitResult<()> (StmCircuitError)` so we keep precise circuit-level errors as long as possible.
At the Relation::circuit boundary, we convert to `plonk::Error (map_err(Error::from))` because the Midnight API requires that return type.

For the `map_with_result`, we currently use `Result<_, Error>` for the same reason: returning `StmResult<_>` there would require an extra conversion layer back to `plonk::Error`, without improving error fidelity.

If this interpretation matches your expectation, we can keep this approach (typed internally, explicit conversion only at Halo2 boundary).
If you prefer strict `StmResult` signatures even inside synthesis closures, I can adapt it, but it would introduce earlier anyhow wrapping in this path.

mithril-stm/src/circuits/halo2/circuit.rs
hjeljeli32: Thanks @damrobi for the suggestion. Done.

mithril-stm/src/circuits/halo2/circuit.rs
hjeljeli32: Done.

mithril-stm/src/circuits/halo2/errors.rs
hjeljeli32: @jpraynaud 
These are part of the same earlier discussion about typed circuit errors vs `StmResult<anyhow>`. Current code keeps typed guard errors until the mandatory Halo2 boundary (`plonk::Error`). If you prefer strict `StmResult `usage, I can refactor accordingly.

mithril-stm/src/circuits/halo2/errors.rs
hjeljeli32: @jpraynaud 
These are part of the same earlier discussion about typed circuit errors vs `StmResult<anyhow>`. Current code keeps typed guard errors until the mandatory Halo2 boundary (`plonk::Error`). If you prefer strict `StmResult `usage, I can refactor accordingly.

mithril-stm/src/circuits/halo2/circuit.rs
hjeljeli32: @jpraynaud 
These are part of the same earlier discussion about typed circuit errors vs `StmResult<anyhow>`. Current code keeps typed guard errors until the mandatory Halo2 boundary (`plonk::Error`). If you prefer strict `StmResult `usage, I can refactor accordingly.

mithril-stm/src/circuits/halo2/golden/cases/negative.rs
hjeljeli32: @jpraynaud Thanks for the suggestion.
For proving-time failures coming from Midnight, we currently receive `PlonkError::Synthesis(String)` only, and the structured variant is already erased at that boundary.
That is why these specific negative tests assert text message content today.
If you’d like, I can add a short comment in the helper/tests to make this limitation explicit until Midnight improves its error handling.

mithril-stm/src/circuits/halo2/golden/helpers.rs
hjeljeli32: @jpraynaud 
This is related to the same earlier discussion on StmResult vs typed local errors.
Here  I kept a typed alias because tests/helpers pattern-match specific failure variants.
Using `StmResult<anyhow>` here would remove that structure at the function signature level and push us to runtime downcasting/string checks earlier.
If you prefer strict `StmResult` usage for consistency, I can refactor accordingly.

mithril-stm/src/lib.rs
hjeljeli32: Done.

mithril-stm/src/circuits/halo2/circuit.rs
```diff
@@ -35,13 +37,98 @@ pub struct StmCircuit {
...
             num_lotteries,
             merkle_tree_depth,
         }
     }
+
+    /// Validates global circuit parameters before synthesis.
+    ///
+    /// Enforces `quorum < num_lotteries` returning
+    /// `StmCircuitError::InvalidCircuitParameters` when violated.
+    pub(crate) fn validate_parameters(&self) -> crate::StmResult<()> {
```
jpraynaud: Can you use imports for `StmResult` and `StmError`?

mithril-stm/src/circuits/halo2/golden/helpers.rs
```diff
@@ -43,73 +43,71 @@ type CircuitVerificationAndProvingKeyPair = (MidnightVK, MidnightPK<StmCircuit>)
 /// Cache map for verification/proving keys keyed by STM circuit configuration.
 type CircuitKeysCache = HashMap<StmCircuitConfig, Arc<CircuitVerificationAndProvingKeyPair>>;
 
-/// Errors returned by `prove_and_verify_result`.
-#[derive(Debug, Error)]
-pub(crate) enum StmCircuitProofError {
```
jpraynaud, line 48: All of these variants are valuable and you could keep them (maybe merged in the `StmCircuitErrror`).

mithril-stm/src/circuits/halo2/golden/helpers.rs
```diff
@@ -170,12 +169,12 @@ impl From<&SignerFixture> for MTLeaf {
     }
 }
 
-fn target_value_from_field(target: F) -> StmResult<LotteryTargetValue> {
+fn target_value_from_field(target: F) -> crate::StmResult<LotteryTargetValue> {
     LotteryTargetValue::from_bytes(&target.to_bytes_le())
-        .map_err(|_| StmCircuitProofError::InvalidLotteryTargetBytes)
+        .map_err(|e| anyhow!("Invalid lottery target bytes: {e}"))
```
jpraynaud: This should work, like that we keep the strongly typed error which is wrapped in a `StmResult` (also applies to other tests below):
```suggestion
        .map_err(|_| StmCircuitError::InvalidLotteryTargetBytes.into())
```

Issue #3050, opened 2026-02-25
Cardano blocks and transaction import stalled in `dev` aggregators


jpraynaud:
## Why

On `dev-mainnet`, the aggregator state machine is healthy but the blocks and transactions import is stuck, while the signer import succeeds normally (running Cardano `10.6.2`):

```bash
sqlite3 -table data/mainnet/mithril-aggregator/mithril/stores/cardano-transaction.sqlite3 
SQLite version 3.37.2 2022-01-06 13:25:41
Enter ".help" for usage hints.
sqlite> select * from cardano_block order by rowid desc limit 5;
+------------------------------------------------------------------+--------------+-------------+
|                            block_hash                            | block_number | slot_number |
+------------------------------------------------------------------+--------------+-------------+
| 6ef18b443f592f18cd31059311b8bb20c823fb28c70d8cc7c0fee81a9769a605 | 13079034     | 180319545   |
| 67d200df2a6fda59091397153f43d4559696f259d583daef3961ae205bec66b4 | 13079033     | 180319508   |
| 05d4327bd97f01827bd13bb10a95f22771c84be733d0c5e461247cae4338f3ec | 13079032     | 180319498   |
| f184f519351444b0b203bb765c14f736a23a033d3e3061ac9ea8947a57dc8094 | 13079031     | 180319487   |
| cd0f7ba4a6c12a8204d6a0dd7d980773cd0deedcf590b69bc51962550448cfd1 | 13079030     | 180319473   |
+------------------------------------------------------------------+--------------+-------------+

```

```bash
sqlite3 -table data/mainnet/mithril-signer-1/mithril/stores/cardano-transaction.sqlite3 
SQLite version 3.37.2 2022-01-06 13:25:41
Enter ".help" for usage hints.
sqlite> select * from cardano_block order by rowid desc limit 5;
+------------------------------------------------------------------+--------------+-------------+
|                            block_hash                            | block_number | slot_number |
+------------------------------------------------------------------+--------------+-------------+
| 81d46cca959c5f59069021e9baef8422e23fbc096ca416b62ca4b8f2d160ab3d | 13084829     | 180437367   |
| 157d1beb1bad6b002802174d99521a3a026e000422523d6ec0e2ad2409f21c03 | 13084828     | 180437362   |
| ec095dfb2744d399e4e595fdc225ac9ed0c263a4ea2b085cb4a6c17f9ae0c36f | 13084827     | 180437360   |
[...]

PR #1723, opened 2024-05-31
Make Cardano transaction prover able to certify more than 5 transactions

jpraynaud:
## Content
This PR includes a fix to make the **Cardano transaction prover** able to certify more than `5` transactions.

### Benchmarks

Screenshot from 2024-05-31 16-36-16

| Total Requests | Concurrency | Transactions/Request | Pool (x50): Request/s | Pool (x50) + Optimization 'get_all_transactions_for_block_ranges': Request/s |
| -------------- | ----------- | -------------------- | --------------------- | ---------------------------------------------------------------------------- |
| 1000           | 100         | 1                    | **233.7**                 | **256.29**                                                                       |
| 1000           | 100         | 2                    | **170.75**                | **181.31**                                                                       |
| 1000           | 100         | 3                    | **127.82**                | **135.56**                                                                       |
| 1000           | 100         | 4                    | **99.07**                 | **112.84**                                                                       |
| 1000           | 100         | 5                    | **83.36**                 | **94.28**                                                                        |
| 1000           | 100         | 6                    | **0**                     | **78.23**                                                                        |
| 1000           | 100         | 7                    | **0**                     | **69.29**                                                                        |
| 1000           | 100         | 8                    | **0**                     | **60.99**                                                                        |

> Running on **Linux / 8 cores / 64 GB RAM / 2 TB SSD**

## Pre-submit checklist

- Branch
  - [x] Tests are provided (if possible)
  - [x] Crates versions are updated (if relevant)
  - [x] Commit sequence broadly makes sense
  - [x] Key commits have useful messages
- PR
  - [x] No clippy warnings in the CI
  - [x] Self-reviewed the diff
  - [x] Useful pull request description
  - [x] Reviewer requested

## Issue(s)
Relates to #1722

PR #3057, opened 2026-02-27
perf: optimize blocks and transactons retrieval from sqlite

Alenar:
## Content

This PR includes optimizations and fixes related to the cardano blocks and transactions.

- feature(persistence): add `Optimize` task to `SqliteCleaner` (than run either `pragma optimize;` or `pragma optimize=0x10002;` depending of the provided `OptimizationMode`)
- chore(cardano-node-chain): make the block ranges importers (new and legacy) logs progress when working (only when storing a batch to avoid flooding the logs)
- perf(persistence): run `SqliteCleaner::optimize(OptimizationMode::Default)` after opening a connection if migrations were applied
- perf(cardano-node-chain): execute a `ChainDataStore::optimize` after storing the blocks and transactions and after computing the block ranges (the implementation run `SqliteCleaner::optimize(OptimizationMode::Default)`)
- perf(persistence): replace `GetCardanoBlockTransactionsQuery::by_block_ranges` by `between_block` as the simple fact that a `1 = 0 or (...)` was added made the query plan far worse and we do not retrieve blocks and transactions with more than one block range at a time since #1723 anyway.
- fix(persistence): storing more than 11 000 blocks at once crashed the app (wrong chunking that was set to 100 000, making 300 000 bindings at once maximum, far above the sqlite limit of `32 766`).

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation
  - [x] No new TODOs introduced

## Issue(s)

Relates to #3050

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3932,7 +3932,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-aggregator"
-version = "0.8.28"
+version = "0.8.29"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -4078,7 +4078,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-cardano-node-chain"
-version = "0.1.21"
+version = "0.1.22"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -4229,7 +4229,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-common"
-version = "0.6.52"
+version = "0.6.53"
 dependencies = [
  "anyhow",
  "async-trait",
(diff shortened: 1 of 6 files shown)


internal/mithril-persistence/src/sqlite/cleaner.rs
```diff
@@ -38,7 +64,7 @@ impl SqliteCleaningTask {
 pub struct SqliteCleaner<'a> {
     connection: &'a SqliteConnection,
     logger: Logger,
-    tasks: Vec<SqliteCleaningTask>,
+    tasks: BTreeSet<SqliteCleaningTask>,
```
jpraynaud, line 67: The order we use when creating the tasks with `with_tasks` will not be followed any more. Maybe it's worth mentioning that the order of the variants `SqliteCleaningTask` will impact the order of execution of the tasks.

PR #3060, opened 2026-03-03
refactor: rework sqlite connection pool

Alenar:
## Content

~~This PR reworks the SQLite connection pool system, switching its logic from "creating all connections at build to distribute them afterwards" to "opening a new connection each time one is asked, with a token systems to ensure not more than 'x' connections are opened at the same time".~~

[updated] This PR reworks the SQLite connection pool system to add a scheduled reset ability, when called it makes a pooled connection reopening itself the next time it's returned to the pull.

The rational being this change is the fact that when running an `analyze` or `pragma optimize;`, only the connection on which the command was run and the new connections opened afterwards benefit of the optimizations, the others connections live at the same times does not benefits of the updated query plan.
This explain why even if we did not see performance improvement despite adding `pragma optimize` between the blocks and transactions storage and the blocks ranges computations (in #3057).

> [!WARNING]
> There's currently no consistent ways of using shared [in-memory connections](https://sqlite.org/inmemorydb.html) using SQLite, the only way to do that needs the [shared cache](https://sqlite.org/sharedcache.html), but since it's discouraged SQLite binaries are often built without it (this is the case for the sqlite crate `bundled` feature) making it not an option.
>
> This forced me to update **all** tests that used a `SqliteConnectionPool` to use a file in place of a in-memory database.

### global

- update all tests that use a `SqliteConnectionPool` to use a file database when they were using one in-memory

#### mithril-cardano-node-chain

- `ChainReaderBlockStreamer`: updated `poll_next` log to add the `last_polled_point` + add a new log when running `try_new`

#### mithril-resource-pool

- add `ResourcePool::reset_available_resources` to immediately run `reset` on all resources that are still available in the pool.

#### mithril-persistence

- refactor `SqliteConnectionPool`:
  - ~~create connections on the fly when asked, use a tokens system to ensure no more than the set limit connections can exist at the same time~~
  - add `schedule_reset`: this makes the `SqlitePooledConnection` re-open themselves the next time they are returned to the pool
  - change: use the `ConnectionBuilder` to build connections instead of a arbitrary function
- `ConnectionBuilder`:
  - change: don't consume self when running `build` so it can build more than one connection
  - new: `build_without_migrations` to be used when a database have already be initialized and we only want a connection
  - new: `build_pool` run the migrations and return a `SqliteConnectionPool` that use the builder

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation
  - [x] No new TODOs introduced

## Issue(s)

Relates to #3050

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3932,7 +3932,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-aggregator"
-version = "0.8.31"
+version = "0.8.32"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -4078,7 +4078,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-cardano-node-chain"
-version = "0.1.22"
+version = "0.1.23"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -4369,7 +4369,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-persistence"
-version = "0.2.66"
+version = "0.2.67"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -4430,7 +4430,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-resource-pool"
-version = "0.0.9"
+version = "0.0.10"
 dependencies = [
  "anyhow",
  "mithril-common",
@@ -4464,7 +4464,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-signer"
-version = "0.3.21"
+version = "0.3.22"
 dependencies = [
  "anyhow",
  "async-trait",
--- a/internal/cardano-node/mithril-cardano-node-chain/Cargo.toml
+++ b/internal/cardano-node/mithril-cardano-node-chain/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mithril-cardano-node-chain"
-version = "0.1.22"
+version = "0.1.23"
 authors.workspace = true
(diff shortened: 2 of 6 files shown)


turmelclem (approved):
LGTM 👍

internal/mithril-persistence/src/sqlite/connection_pool.rs
```diff
@@ -1,76 +1,228 @@
...
-        let connection = self.connection_pool.acquire_resource(timeout)?;
+        let connection = self.connections.acquire_resource(timeout)?;
 
         Ok(connection)
     }
 
-    /// Returns a single resource pool connection
-    pub fn build_from_connection(connection: SqliteConnection) -> Self {
-        let connection_pool = ResourcePool::new(1, vec![SqlitePooledConnection::new(connection)]);
+    /// Schedule a reset of the connections in this pool
```
jpraynaud: Maybe `schedule` is a bit misleading as the reset is done right away?

Alenar: I agree, but I also don't want to name this function `reset_available_connections_and_schedule_reset_for_those_not_available`. I may rename it simply to `reset_connections` or `renew_connections`, WDYT ?

jpraynaud: `renew_connections` sounds good :+1:

Issue #3037, opened 2026-02-23
Refactor SNARK circuit - Midnight library decoupling


jpraynaud:
## Why

As described in the [SNARK MVP Book](https://hackmd.io/EeCC7vaPQoWdmMVAkyDQpQ?view=#Refactor-Halo2-circuit-for-production-readiness) (Workstream 4 - Type Consistency, circuit-local wrapper types), the Halo2 circuit layer is tightly coupled to the Midnight library through direct imports of four Midnight crates (`midnight_curves`, `midnight_circuits`, `midnight_proofs`, `midnight_zk_stdlib`) across `circuit.rs`, `gadgets.rs`, `types.rs`, and test helpers. This coupling means that any change in the Midnight API directly impacts the circuit logic, making the circuit harder to maintain and test independently.

## What

Introduce wrapper types and abstraction boundaries to decouple the Halo2 circuit from direct Midnight library dependencies, ensuring the circuit layer depends only on Mithril-internal abstractions.

## How

- [ ] Audit all Midnight imports in `circuit.rs`, `gadgets.rs`, `types.rs`, and `utils/`
- [ ] Introduce circuit-local wrapper types for externally imported Midnight domain types (curve types, assigned types, instruction traits). These types should have private internal fields and provide `Deref` trait implementation.
- [ ] Centralize all Midnight-to-Halo2 conversions inside dedicated `adapter` module
- [ ] Consolidate the duplicated `type F = JubjubBase` and `type C = Jubjub` aliases (currently defined identically in `circuit.rs`, `gadgets.rs`, and `golden/helpers.rs`) into a single shared location
- [ ] Ensure `circuit.rs` and `gadgets.rs` no longer import Midnight crates directly but use only Mithril-internal type aliases and wrappers
- [ ] Re-export only the minimal necessary Midnight types from `types.rs` (currently re-exports `Bls12`, `JubjubBase`, `JubjubScalar`, `Jubjub`)
- [ ] Measure circuit constraint count before and after to ensure no power-of-two boundary (`2^k`) increase
- [ ] Update the [SNARK MVP Book](https://hackmd.io/EeCC7vaPQoWdmMVAkyDQpQ?view=#Refactor-Halo2-circuit-for-production-readiness) to reflect the changes made

PR #3075, opened 2026-03-07
feat(stm): Prepare Prover Witness

curiecrypt:
## Content

This PR implements the witness preparation logic for the SNARK prover. It introduces `SnarkProverInput::prepare_prover_input`, which takes a set of single signatures and transforms them into the witness and public instance format expected by the circuit. 

- Introduce the witness submodule with `WitnessEntry` and `SnarkProverInput` structs
- Implement `SnarkProverInput::prepare_prover_input`: filter, verify, compute lottery indices, deduplicate, and expand into witness entries
- Add `SnarkClerk` with signature deduplication (`select_valid_signatures_for_k_indices`)
- Convert Merkle paths to circuit-facing `JubjubBase` field representation
- Compute the public instance (`merkle_root`, `message`) as `BaseFieldElement` pair

## Pre-submit checklist

- Branch
  - [x] Tests are provided (if possible)
  - [x] Crates versions are updated (if relevant)
  - [x] CHANGELOG file is updated (if relevant)
  - [x] Commit sequence broadly makes sense
  - [x] Key commits have useful messages
- PR
  - [x] All check jobs of the CI have succeeded
  - [x] Self-reviewed the diff
  - [x] Useful pull request description
  - [x] Reviewer requested

### Tests that should be added/modified?
- Mix of valid + stripped SNARK sigs still succeeds: some sigs have `snark_signature = None`, but enough valid ones remain for `k`
- Succeeds when some sigs fail Schnorr verification: tamper with a few sigs so verify fails, but enough pass
- Signer with out-of-range index is skipped: check the `Err(_) => return None` / `continue` path in registration lookup
- Witness has exactly `k` entries with unique lottery indices: structural invariant check on the output
- Assert `NotEnoughSignatures` variant for empty input: strengthen `empty_signatures_fails` to check the specific error variant
- Assert `NotEnoughSignatures` variant for all-stripped input: strengthen `skips_signatures_without_snark_component` similarly
- Witness entries are sorted by lottery index: verify the ordering guarantee from the doc comment

## Issue(s)

Closes #2887

PR #3076, opened 2026-03-09
refactor(halo2): Decouple circuit from Midnight types

hjeljeli32:
## Content
This PR improves the Halo2 circuit code structure for by introducing decoupled circuit-local types.

This PR includes:
- Added a circuit-local field wrapper and shared aliases in `types.rs`:
   - `CircuitBaseField`, `CircuitBase`, `CircuitCurve`
- Replaced duplicated local `F/C` aliases with shared types from `types.rs` across:
   - `circuit.rs`, `gadgets.rs`, `golden/helpers.rs`
- Implemented type conversions using `From/Into`
- Introduced a test-only adapter module (`adapters.rs`) for STM Merkle path → Halo2 witness conversion.
- Removed `utils/mod.rs` by inlining field-limb split logic into `gadgets.rs`
- Unified synthesis error mapping with `to_synthesis_error` in `errors.rs`

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Issue(s)

Closes https://github.com/input-output-hk/mithril/issues/3037

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4507,7 +4507,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-stm"
-version = "0.9.26"
+version = "0.9.27"
 dependencies = [
  "anyhow",
  "blake2 0.10.6",
--- a/mithril-stm/CHANGELOG.md
+++ b/mithril-stm/CHANGELOG.md
@@ -5,6 +5,17 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## 0.9.27 (03-11-2026)
+
+### Changed
+
+- Introduced circuit-local Halo2 types in `circuits/halo2/types.rs` with shared `CircuitBaseField`, `CircuitBase`, and `CircuitCurve`.
+- Replaced duplicated local `F`/`C` aliases with shared circuit types across Halo2 circuit, gadgets, and golden helper code.
+- Standardized Halo2 conversion paths using `From`/`Into` implementations for circuit/domain field wrappers.
+- Added `circuits/halo2/adapters.rs` to convert STM Merkle paths into Halo2 witness paths for circuit consumption.
+- Removed `circuits/halo2/utils/mod.rs` and inlined field-limb split logic into `circuits/halo2/gadgets.rs`.
+- Unified synthesis error mapping through `to_synthesis_error` in `circuits/halo2/errors.rs`.
+
 ## 0.9.26 (03-09-2026)
 
 ### Changed
--- a/mithril-stm/Cargo.toml
+++ b/mithril-stm/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mithril-stm"
-version = "0.9.26"
+version = "0.9.27"
 edition = { workspace = true }
 authors = { workspace = true }
 homepage = { workspace = true }


mithril-stm/src/circuits/halo2/golden/helpers.rs
```diff
@@ -17,7 +17,9 @@ use rand_core::SeedableRng;
 use crate::LotteryTargetValue;
 use crate::circuits::halo2::circuit::StmCircuit;
 use crate::circuits::halo2::errors::StmCircuitError;
-use crate::circuits::halo2::types::{Bls12, JubjubBase, MTLeaf, MerklePath};
+use crate::circuits::halo2::types::{
+    CircuitBase as CF, MTLeaf, MerklePath, SignedMessageWithoutPrefix as F,
```
jpraynaud: You can use directly the `CircuitBase` and `SignedMessageWithoutPrefix` types to remove the `CF` and `F` types which prevent readbility.

mithril-stm/src/circuits/halo2/adapters.rs
jpraynaud, line 1: It'd be a good idea to move all test related modules in a dedicated `test` module as mixing them with other modules tends to be confusing. 

Also, could you avoid creating modules with `mod` inside other modules and prefer working with folder modules for more clarity of the structure of the code?

mithril-stm/src/circuits/halo2/adapters.rs
```diff
@@ -0,0 +1,50 @@
...
+    /// Errors returned when adapting STM Merkle paths to Halo2 witness paths.
+    #[derive(Debug, Error)]
+    pub enum MerklePathAdapterError {
+        #[error("invalid merkle digest length")]
+        InvalidDigestLength,
+        #[error("non-canonical merkle digest")]
+        NonCanonicalDigest,
+    }
+
+    impl<D: Digest> TryFrom<&StmMerklePath<D>> for Halo2MerklePath {
```
jpraynaud: Is this really something limited to testing? IMO this is also something that needs to be implemented for production code.

mithril-stm/src/circuits/halo2/circuit.rs
```diff
@@ -147,26 +127,27 @@ impl Relation for StmCircuit {
...
         let witness = witness
             .map_with_result(|witness| -> StmResult<_> {
                 self.validate_witness_length(witness.len())?;
                 Ok(witness)
             })
-            .map_err(Self::synthesis_error)?
+            .map_err(to_synthesis_error)?
             .transpose_vec(self.quorum as usize);
 
+        // Wrapper-to-circuit-base conversions use `From`/`Into` implementations.
```
jpraynaud: This comment is not needed.

mithril-stm/src/circuits/halo2/gadgets.rs
```diff
@@ -9,12 +10,35 @@ use midnight_circuits::types::{
...
+use num_traits::{Num, One};
+
+use crate::StmResult;
+use crate::circuits::halo2::errors::{StmCircuitError, to_synthesis_error};
+use crate::circuits::halo2::types::{CircuitBase as F, CircuitCurve as C};
+
+/// Splits a field element into `(lower, upper)` limbs at `num_bits` using LE encoding.
+fn split_field_element_into_le_limbs<F: PrimeField>(value: &F, num_bits: u32) -> StmResult<(F, F)> {
+    let value_big = BigUint::from_bytes_le(value.to_repr().as_ref());
+    let lower_mask = (BigUint::one() << num_bits) - BigUint::one();
```
jpraynaud, line 34: `num_bits` probably needs to be checked to be within a valid range as this operation could easily overflow.

mithril-stm/src/circuits/halo2/types.rs
```diff
@@ -2,18 +2,102 @@
...
+pub(crate) type CircuitCurve = MidnightJubjub;
+
+/// Field type boundaries:
+/// - `BaseFieldElement`: STM/domain field wrapper.
+/// - `CircuitBaseField`: circuit statement/input wrapper.
+/// - `CircuitBase` (`MidnightBaseField`): raw proving-library field.
+///
+/// Circuit-local wrapper for base-field values present in public inputs/witness data.
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Default)]
+pub struct CircuitBaseField(MidnightBaseField);
```
jpraynaud: Here and in all other occurrences of `MidnightBaseField`:
```suggestion
pub struct CircuitBaseField(CircuitBase);
```

mithril-stm/src/circuits/halo2/circuit.rs
```diff
@@ -12,21 +12,19 @@ use midnight_proofs::circuit::{Layouter, Value};
...
 
-use crate::circuits::halo2::errors::StmCircuitError;
+use crate::StmResult;
+use crate::circuits::halo2::errors::{StmCircuitError, to_synthesis_error};
 use crate::circuits::halo2::gadgets::{
     verify_lottery, verify_merkle_path, verify_unique_signature,
 };
 use crate::circuits::halo2::types::{
-    Jubjub, JubjubBase, LotteryIndex, MTLeaf, MerklePath, MerkleRoot, SignedMessageWithoutPrefix,
+    CircuitBase as F, CircuitCurve as C, LotteryIndex, MTLeaf, MerklePath, MerkleRoot,
```
damrobi: I think I would prefer if we used directly `CircuitBase` and `CircuitCurve` instead of `F` and `C` for readability. Let me know what you think.

mithril-stm/src/circuits/halo2/types.rs
```diff
@@ -2,18 +2,102 @@
...
-pub type Target = JubjubBase;
+pub type Target = CircuitBaseField;
 /// Signed message value used by the circuit transcript, without any domain prefix.
-pub type SignedMessageWithoutPrefix = JubjubBase;
+pub type SignedMessageWithoutPrefix = CircuitBaseField;
 /// Merkle root public input committed by the STM membership commitment tree.
-pub type MerkleRoot = JubjubBase;
+pub type MerkleRoot = CircuitBaseField;
 /// Lottery index (`i`) used for per-lottery checks in witness entries.
 pub type LotteryIndex = u32;
```
damrobi: Just a warning, this type is removed in the type consistency PR to used the general STM `LotteryIndex` which is a `u64` so it might create conflicts.

mithril-stm/src/circuits/halo2/golden/helpers.rs
```diff
@@ -318,15 +317,15 @@ pub(crate) fn create_merkle_tree_with_leaf_selector(
...
 }
 
 fn assert_challenge_endianness(sig: &UniqueSchnorrSignature) -> StmResult<()> {
     let challenge_bytes = sig.challenge.to_bytes();
-    let challenge_native = JubjubBase::from_bytes_le(&challenge_bytes)
+    let challenge_native = CF::from_bytes_le(&challenge_bytes)
         .into_option()
         .ok_or_else(|| anyhow!(StmCircuitError::InvalidChallengeBytes))?;
-    if challenge_native != sig.challenge.0 {
+    if F::from(challenge_native) != F::from(sig.challenge.0) {
```
damrobi: Is it possible to implement a `From<CircuitBase> for BaseFieldElement` to avoid using the `.0`?

mithril-stm/src/circuits/halo2/golden/helpers.rs
```diff
@@ -172,7 +171,7 @@ impl From<&SignerFixture> for MTLeaf {
 }
 
 fn target_value_from_field(target: F) -> StmResult<LotteryTargetValue> {
-    LotteryTargetValue::from_bytes(&target.to_bytes_le())
+    LotteryTargetValue::from_bytes(&CF::from(target).to_bytes_le())
```
damrobi: I think you can do a simpler conversion, for example something like:
```suggestion
    let lottery_target_value: LotteryTargetValue = target.into();
```

mithril-stm/src/circuits/halo2/adapters.rs
```diff
@@ -0,0 +1,50 @@
...
+        type Error = MerklePathAdapterError;
```
curiecrypt: Expanding on @jpraynaud 's  question: this isn't just hypothetically needed in production. The `SnarkProof::aggregate_signatures` flow (in `proof.rs` of this [pr](https://github.com/input-output-hk/mithril/pull/3075)) already calls `convert_snark_single_signature_to_witness_entries`, which needs exactly this `TryFrom<&StmMerklePath<D>> for Halo2MerklePath` conversion to build real prover witnesses from aggregated signatures. Gating it #[cfg(test)] means either duplicating the logic or moving it later. 

We should decide which strategy to keep on main. The conversion proposed here or the one proposed in the mentioned pr.

mithril-stm/src/circuits/halo2/types.rs
```diff
@@ -25,17 +109,17 @@ pub type LotteryIndex = u32;
...
     Left,
     Right,
 }
 
-impl From<Position> for JubjubBase {
-    fn from(value: Position) -> Self {
-        match value {
-            Position::Left => JubjubBase::ZERO,
-            Position::Right => JubjubBase::ONE,
+impl From<Position> for MidnightBaseField {
```
curiecrypt: `Position` only converts to `MidnightBaseField` currently, but callers working with `CircuitBaseField` (the wrapper) must go through an extra `.into()` hop. Adding a direct conversion keeps the wrapper layer self-contained:

```rust
impl From<Position> for CircuitBaseField {
    fn from(position: Position) -> Self {
        Self(match position {
            Position::Left => CircuitBase::ZERO,
            Position::Right => CircuitBase::ONE,
        })
    }
}
```

Keep the existing `MidnightBaseField impl` too. It's still needed by the Halo2 assign path in `circuit.rs:245`.

mithril-stm/src/circuits/halo2/gadgets.rs
```diff
@@ -9,12 +10,35 @@ use midnight_circuits::types::{
...
 use midnight_zk_stdlib::ZkStdLib;
+use num_bigint::BigUint;
+use num_traits::{Num, One};
+
+use crate::StmResult;
+use crate::circuits::halo2::errors::{StmCircuitError, to_synthesis_error};
+use crate::circuits::halo2::types::{CircuitBase as F, CircuitCurve as C};
+
+/// Splits a field element into `(lower, upper)` limbs at `num_bits` using LE encoding.
+fn split_field_element_into_le_limbs<F: PrimeField>(value: &F, num_bits: u32) -> StmResult<(F, F)> {
```
curiecrypt: _nit:_ 
`split_field_element_into_le_limbs<F: PrimeField>` introduces a generic `F` that shadows the module-level use `CircuitBase` as `F` (at line 18). This compiles fine but is confusing when reading: the `F` inside the function body is the generic, not `CircuitBase`. Consider renaming the generic:
```rust
fn split_field_element_into_le_limbs<Fp: PrimeField>(value: &Fp, num_bits: u32) -> StmResult<(Fp, Fp)> {
```

Same for `field_modulus_as_biguint<F>` and `big_unsigned_integer_to_field_element<F>` (at lines 31 and 36).

mithril-stm/src/circuits/halo2/circuit.rs
```diff
@@ -1,7 +1,7 @@
 use anyhow::anyhow;
 use ff::Field;
```
curiecrypt, line 2: We still directly import `ff::Field` and `group::Group` from Midnight's deps: used for `F::ZERO`, `F::from_u128`, and `Group::generator()`. 

This doesn't break the decoupling (they're standard traits), but worth noting that full decoupling would eventually push these behind wrapper methods on `CircuitBaseField`. 

Could be tracked as a _follow-up_ if needed.

mithril-stm/src/circuits/halo2/gadgets.rs
hjeljeli32: @jpraynaud 
Thanks for the suggestion, I applied it with a typed error.
I added `StmCircuitError::InvalidBitDecompositionRange { num_bits, field_bits } `and now `split_field_element_into_le_limbs` validates` num_bits`before doing the operation.

mithril-stm/src/circuits/halo2/types.rs
hjeljeli32: Thanks @damrobi for the warning 

Let me confirm my understanding from the discussion you had with @kitounliu / @jpraynaud  on slack:
- keep u64 on STM,
- keep circuit constraints with lottery index as u32,
- add explicit range validation before proving (index fits in u32 and within circuit bounds).

If this seems reasonable, I’ll apply it in this PR.

mithril-stm/src/circuits/halo2/adapters.rs
hjeljeli32: @curiecrypt 
I agree this conversion is needed in production now since witness preparation relies on  it.
I’ll remove the test-only gating.

I also noticed your PR already has an equivalent conversion function; I agree we should unify to a single implementation on main to avoid duplication. Let me know if this is convenient approach.

mithril-stm/src/circuits/halo2/types.rs
damrobi: What you describe was indeed applied to the parameters of the circuit. The `LotteryIndex` was not a parameter and was only used in the circuit computation and casted as a u64 to create a field element. In the type consistency PR, I removed the circuit version of `LotteryIndex` and imported the one from STM to remove this extra casting so now there is only the u64 version of it.

mithril-stm/src/circuits/halo2/mod.rs
```diff
@@ -1,10 +1,15 @@
 //! Halo2 prototype integration (feature-gated by `future_snark`).
 
-pub mod circuit;
-pub(crate) mod errors;
-pub mod gadgets;
-pub mod types;
-pub(crate) mod utils;
+pub mod adapters;
+// Temporary while Halo2 production wiring is in progress.
+// Remove once these modules are wired into STM.
```
jpraynaud: Maybe you can convert this to a `TODO` comment

rust-clippy-results.sarif
jpraynaud, line 1: No need to commit this file which is local test results.

mithril-stm/src/circuits/halo2/types.rs
hjeljeli32: Thanks for the clarification, @damrobi.
I did the following:
- removed the circuit-local LotteryIndex alias (u32)
- switched Halo2 to the STM LotteryIndex (u64)
- kept circuit comparisons at 32 bits (lower_than(..., 32))
- added pre-synthesis guards to ensure each index fits u32 and is < num_lotteries
- I also updated negative tests to cover these guard paths.

Issue #3015, opened 2026-02-18
Certification stopped in `testing-preview` network


jpraynaud:
## Why

The certification has stopped in the `testing-preview` network since 2/17/2026, 2:59:55 PM.

The error received in the aggregator is:
```bash
[...]

PR #3095, opened 2026-03-19
feat: avoid storage and diffusion of unconsistent network configuration and epoch settings

### LINKED ISSUE #3030: Enhance the support for signed entity types with configuration
## Why
In #3015, we have noticed that there is a problem with the storage of the signed entity config for Cardano blocks and transactions which is done too late: this means that when we activate the new signed entity type on the leader aggregator, it fails to create the associated open message for the next 2 epochs.

## What
Make sure that the signed entity type config is stored earlier to avoid problems when activating the signed entity type.

## How

- [x] Save configurations in leader aggregator even if signed entity type not activated
- [x] Deactivate signed entity types if configuration is incoherent (and add a warning log)

## Later
- Activate the Cardano blocks and transactions signed entity type on `testing-preview`

Alenar:
## Content

This PR add a consistency check capability for `SignedEntityConfig`, which is used by the `mithril-aggregator` to filter out unusable signed entities when retrieving them at the start of each epoch.

This is a permanent fix for the issue found in #3015.

#### `mithril-common`

- `SignedEntityConfig` - add `check_consistency`: check that all discriminant in its allowed list that have their required configuration if they need some, if it's missing returns a dedicated error which contains the split between the usable and not usable discriminant.

#### `mithril-aggregator`

- `LocalMithrilNetworkConfigurationProvider`: run `SignedEntityConfig::check_consistency` before returning the network configuration for an epoch, filtering out signed entities that can't be enabled without raising an error.
   - this have cascading effect: the corrected parameters are transmitted to the `EpochService` which save them in the `epoch_setting` table after crossing epoch boundaries for later use and diffusion to mithril-signers
- Configuration: Change how `ConfigurationSource::get_leader_aggregator_epoch_settings_configuration` behave to always returns cardano transactions and cardano blocks transactions signings, regardless of if there associated signed entity is enabled or not
- new integration test - `leader_signed_entity_config_propagation`: reproduce the bug found in #3015 by trying to put the runtime in a state were an epoch settings exists with signed entities that don't have their required configuration.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation
  - [x] No new TODOs introduced

## Issue(s)

Closes #3030

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3951,7 +3951,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-aggregator"
-version = "0.8.36"
+version = "0.8.37"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -4251,7 +4251,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-common"
-version = "0.6.58"
+version = "0.6.59"
 dependencies = [
  "anyhow",
  "async-trait",
--- a/mithril-aggregator/Cargo.toml
+++ b/mithril-aggregator/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mithril-aggregator"
-version = "0.8.36"
+version = "0.8.37"
 description = "A Mithril Aggregator server"
 authors = { workspace = true }
 edition = { workspace = true }
--- a/mithril-common/Cargo.toml
+++ b/mithril-common/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mithril-common"
-version = "0.6.58"
+version = "0.6.59"
 description = "Common types, interfaces, and utilities for Mithril nodes."
 authors = { workspace = true }
 edition = { workspace = true }


mithril-aggregator/src/services/network_configuration_provider.rs
```diff
@@ -57,8 +63,29 @@ impl LocalMithrilNetworkConfigurationProvider {
...
+            cardano_blocks_transactions_signing_config: epoch_settings
+                .cardano_blocks_transactions_signing_config
+                .clone(),
+        })
+        .check_consistency()
+        {
+            Ok(usable_discriminants) => usable_discriminants,
+            Err(err) => {
+                warn!(
+                    self.logger, "Some allowed signed entity could be enabled for epoch {epoch}; using only the usable subset";
```
jpraynaud:
```suggestion
                    self.logger, "Some allowed signed entity could not be enabled for epoch {epoch}; using only the usable subset";
```

mithril-aggregator/tests/leader_signed_entity_config_propagation.rs
```diff
@@ -0,0 +1,183 @@
...
+    assert_metrics_eq!(
+        tester.metrics_verifier,
+        ExpectedMetrics::new()
+            .certificate_total(4)
+            .artifact_mithril_stake_distribution_total(4)
+    );
+}
+
+#[macro_export]
+macro_rules! assert_signable_discriminants_contains {
```
jpraynaud: This would be a better name as you use `assert_eq!`:
```suggestion
macro_rules! assert_signable_discriminants_eq {
```

Issue #3066, opened 2026-03-04
Refactoring and enhancements for SNARK-friendly STM


jpraynaud:
## Why

The STM library has accumulated structural debt as SNARK support was introduced alongside the existing concatenation proof system.

## What

Perform targeted refactorings and enhancements across the STM library to improve separation of concerns, test efficiency, API readiness for SNARK signing, and overall code hygiene.

## Current iteration
- [ ] Use blake2 transcript hash function for non recursive circuit
- [ ] Better RNG support for RNG in STM protocol signer (https://github.com/input-output-hk/mithril/blob/main/mithril-stm/src/protocol/participant/signer.rs#L61-L62)

## Following iteration
- [ ] Enhance error handling with Midnight zk library (need new version of Midnight zk released)
- [ ] Refactor `RegistrationEntry` into struct with fields
- [ ] Refactoring `ClosedRegistrationEntry` (To avoid having double options for vk and stake/lottery target value, use a `ClosedRegistrationEntryForConcatenation` and a `Option<ClosedRegistrationEntryForSnark>`)
- [ ] Refactor tests for CPU/circuit compatibility (Create a dedicated module for CPU/non-recursive circuit compatibility tests)

## Later
- [ ] Get rid of Options for SNARK verification keys and signatures
- [ ] Optimize computation of Lottery Target Value with [Binomial Series](https://en.wikipedia.org/wiki/Binomial_series)?
> (1-phi_f)^w = 1 - w * phi_f + w(w-1) * (phi_f^2) / 2! - w(w-1)(w-2) * (phi_f^3) / 3! + ...

jpraynaud:
> Hi [@jpraynaud](https://github.com/jpraynaud)! :)
> 
> I would like to work on this issue ([#3066](https://github.com/input-output-hk/mithril/issues/3066)), if that is okay.
> 
> My plan is to split the work into small, reviewable PRs, starting with:
> 
> 1. RegistrationEntry / ClosedRegistrationEntry refactor
> 2. test structure improvements (CPU/circuit compatibility + mock prover usage)
> 3. RNG support and cleanup of remaining TODOs
> 4. performance follow-up for SNARK Merkle leaf computation (rayon), if still needed after the refactor
> 
> If this plan looks good, could you assign the issue to me?
> 
> Thank you!

[...]

PR #3103, opened 2026-03-23
feat: wire SNARK proof in aggregate signature

### LINKED ISSUE #3042: SNARK aggregation primitives: Wire SNARK proof in aggregate signature
## Why

This phase integrates it into the existing protocol layer by replacing the `Future` placeholder with the real `Snark` variant, connecting proof creation and verification through the `Clerk` and `AggregateSignature` dispatch.

## What

Replace the `Future` aggregate signature placeholder with the new `Snark` variant and wire the SNARK proof system through the protocol layer.

## How

- [x] Replace the `Future` variant with `Snark` in [`AggregateSignatureType`](https://github.com/input-output-hk/mithril/blob/main/mithril-stm/src/protocol/aggregate_signature/signature.rs) (use byte encoding prefix `1` instead of `255`)
- [x] Replace the `Future` variant with `Snark` in [`AggregateSignature<D>`](https://github.com/input-output-hk/mithril/blob/main/mithril-stm/src/protocol/aggregate_signature/signature.rs)
- [x] Wire `SnarkProof::verify` into [`AggregateSignature::verify`](https://github.com/input-output-hk/mithril/blob/main/mithril-stm/src/protocol/aggregate_signature/signature.rs#L103)
- [x] Wire `SnarkProof::aggregate_signatures` into [`Clerk::aggregate_signatures_with_type`](https://github.com/input-output-hk/mithril/blob/main/mithril-stm/src/protocol/aggregate_signature/clerk.rs#L47)

jpraynaud:
## Content

This PR includes the **wiring of the SNARK proof in the aggregate signature**.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Issue(s)
Closes #3042

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4544,7 +4544,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-stm"
-version = "0.9.35"
+version = "0.9.36"
 dependencies = [
  "anyhow",
  "blake2 0.10.6",
--- a/mithril-stm/Cargo.toml
+++ b/mithril-stm/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mithril-stm"
-version = "0.9.35"
+version = "0.9.36"
 edition = { workspace = true }
 authors = { workspace = true }
 homepage = { workspace = true }


mithril-stm/src/protocol/aggregate_signature/signature.rs
```diff
@@ -150,9 +156,18 @@ impl<D: MembershipDigest> AggregateSignature<D> {
                     ConcatenationProof::batch_verify(&concatenation_proofs, msgs, &avks, parameters)
                 }
                 #[cfg(feature = "future_snark")]
-                AggregateSignatureType::Future => Err(anyhow!(
-                    AggregateSignatureError::UnsupportedProofSystem(aggregate_signature_type)
-                )),
+                AggregateSignatureType::Snark => {
+                    for (((aggregate_signature, msg), avk), parameters) in aggregate_signatures
```
damrobi: I might be wrong but I don't think we are checking that the lenght of the arrays match. Could the number of signatures be greater than the number of messages for example in which case the last signatures are not verified? I see that we perform this check for the concatenation type and I don't think there can be more snark proofs than concatenation proofs but I might be wrong there.

jpraynaud: You're right, I fixed it 👍

mithril-stm/src/protocol/aggregate_signature/signature.rs
```diff
@@ -169,7 +202,7 @@ impl<D: MembershipDigest> AggregateSignature<D> {
                 concatenation_proof.to_bytes()
             }
             #[cfg(feature = "future_snark")]
-            AggregateSignature::Future => vec![],
+            AggregateSignature::Snark(snark_proof) => snark_proof.to_bytes().unwrap(),
```
curiecrypt: I think we should find a better way to handle this unwrap until the deterministic proof generation is offered by midnight. 
I suggest adding an item related to this in #3066.

mithril-stm/src/protocol/aggregate_signature/signature.rs
```diff
@@ -197,7 +232,16 @@ impl<D: MembershipDigest> AggregateSignature<D> {
...
             AggregateSignature::Concatenation(proof) => Some(proof),
             #[cfg(feature = "future_snark")]
-            AggregateSignature::Future => None,
+            AggregateSignature::Snark(_) => None,
+        }
+    }
+
+    /// If the aggregate signature is a SNARK proof, return it.
+    #[cfg(feature = "future_snark")]
+    pub fn to_snark_proof(&self) -> Option<&SnarkProof<D>> {
```
curiecrypt:
```suggestion
    pub fn get_snark_proof(&self) -> Option<&SnarkProof<D>> {
```

I think this name is more accurate. If you agree we should rename `to_concatenation_proof` as `get_concatenation_proof`.

mithril-stm/src/protocol/aggregate_signature/signature.rs
```diff
@@ -8,6 +8,9 @@ use crate::{
     proof_system::ConcatenationProof,
 };
 
+#[cfg(feature = "future_snark")]
+use crate::proof_system::SnarkProof;
```
curiecrypt, line 12: We have a `#[allow(dead_code)]` on `SnarkProof` impl. I think we can remove it and its TODO comment now.

mithril-stm/src/protocol/aggregate_signature/signature.rs
```diff
@@ -63,7 +66,7 @@ impl FromStr for AggregateSignatureType {
         match s {
             "Concatenation" => Ok(AggregateSignatureType::Concatenation),
             #[cfg(feature = "future_snark")]
-            "Future" => Ok(AggregateSignatureType::Future),
+            "Snark" => Ok(AggregateSignatureType::Snark),
             _ => Err(anyhow!("Unknown aggregate signature type: {}", s)),
```
curiecrypt: AFAIS `UnsupportedProofSystem` is no longer referenced. So, instead of `anyhow!("Unknown aggregate signature type: {}", s)` we might want to return this error type. 
If we go with the `anyhow`, we can remove the `UnsupportedProofSystem` error.

mithril-stm/src/protocol/aggregate_signature/signature.rs
```diff
@@ -113,19 +116,40 @@ impl<D: MembershipDigest> AggregateSignature<D> {
...
+                let snark_avk = avk.to_snark_aggregate_verification_key().ok_or_else(|| {
+                    anyhow!(AggregateSignatureError::MissingSnarkAggregateVerificationKey)
+                })?;
+                snark_proof.verify(msg, snark_avk)
+            }
         }
     }
 
-    /// Batch verify a set of aggregate signatures
+    /// Batch verify a set of aggregate signatures.
```
curiecrypt, line 130: I guess I don't completely understand the batching logic for SNARK. 
What I understand is when the input contains a mix of Concatenation and Snark signatures, the grouping by HashMap discards positional information. The Snark batch then zips `aggregate_signatures` (only the Snark ones) with full arrays of msgs, avks, parameters. That seems to be resulting in wrong message/key pairings. 
Is my understanding accurate?

I know that batching is not the real concern now. But, at least we can add an explanation about it in the doc comment.

mithril-stm/src/protocol/aggregate_signature/signature.rs
jpraynaud: Fixed

mithril-stm/src/protocol/aggregate_signature/signature.rs
jpraynaud: I created a new error variant `UnknownProofSystem(String)` which applies when the string representation is not a know `AggregatrSignatureType`, when the `UnsupportedProofSystem(AggregateSignatureType)` is for a proof system not (yet) supported.

mithril-stm/src/protocol/aggregate_signature/signature.rs
jpraynaud: Fixed

mithril-stm/src/protocol/aggregate_signature/signature.rs
jpraynaud: You'right, I fixed it. Now the first step consists in creating a `HashMap` with `AggregateSignatureType` as key and `Vec<(AggregateSignature<D>, AggregateVerificationKey<D>, Vec<u8>, Parameters)>>` as value which guarantees consistent association of message, proofs, keys and parameters.

mithril-stm/src/protocol/aggregate_signature/signature.rs
jpraynaud: ~~Fixed, the `to_bytes` function does not fail so I removed the `Result` return type which is not needed.~~

Fixed, I have updated the function `to_bytes` to return a `StmResult` :+1:

Issue #3107, opened 2026-03-25
Implement SNARK proof generation and verification in end to end tests


jpraynaud:
## Why

We want to run a successful end to end test with SNARK proof generation and verification. We also need to assess how the test can be run at a fast pace, as what we do for the concatenation proof, where the execution of the e2e test is below 2 minutes.

## What

Implement the SNARK proof system in the end to end test and make it run as fast as for Concatenation proof system.

## How

- [ ] Allow the end to end test to run at a slow pace
- [ ] Run a successful end to end test without time limit
- [ ] Identify bottlenecks due to the SNARK proof system
- [ ] Fix bottlenecks due to SNARK proof system

hjeljeli32:
@damrobi @jpraynaud 
Small follow-up from my side after rerunning the SNARK e2e experiments on `jpraynaud/3107-snark-e2e-test` (`4a43490ac`).

Setup used:
- startup params: `(5,9,0.95)`
- epoch length: `100`
- slot length: `0.5`
- run interval: `2000`
- aggregate signature type: `Snark`

Method:
- metrics were extracted from the aggregator certificate store (`aggregator.sqlite3`)
- measured certificates = all rows whose `protocol_parameters` match the measured tuple
- certification duration = `sealed_at - initiated_at`

Results:

| Measured params `(k,m,phi_f)` | Count | Avg cert duration | Sum cert duration | Approx non-SNARK overhead | Total runtime | Status |
|---|---:|---:|---:|---:|---:|---|
| `(3,9,0.94)` | `37` | `~2.429s` | `~89.858s` | `~10m36.077s` | `12m05.935s` | Success |
| `(4,9,0.94)` | `37` | `~2.490s` | `~92.142s` | `~10m26.201s` | `11m58.343s` | Success |
| `(5,9,0.94)` | `36` | `~2.705s` | `~97.369s` | `~9m45.554s` | `11m22.923s` | Success |
| `(7,9,0.94)` | `27` | `~2.946s` | `~79.530s` | `~10m07.413s` | `11m26.943s` | Success |
| `(9,12,0.94)` | `40` | `~3.235s` | `~129.390s` | `~10m08.986s` | `12m18.376s` | Success |
| `(11,12,0.94)` | `33` | `~3.611s` | `~119.174s` | `~10m14.478s` | `12m13.652s` | Success |
| `(13,18,0.94)` | `44` | `~3.590s` | `~157.944s` | `~9m49.574s` | `12m27.518s` | Success |

PR #3115, opened 2026-03-31
feat: support SNARK proof generation

damrobi:
## Content

This PR includes a fix to the way the hex encoded message is processed in mithril-stm and some modifications to allow for caching of the circuit verification key and saving the generated srs in some temporary directory.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Issue(s)

Relates to #3107

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4543,7 +4543,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-stm"
-version = "0.9.36"
+version = "0.9.37"
 dependencies = [
  "anyhow",
  "blake2 0.10.6",
--- a/mithril-stm/CHANGELOG.md
+++ b/mithril-stm/CHANGELOG.md
@@ -5,6 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## 0.9.37 (04-01-2026)
+
+### Added
+
+- Added a fix for the encoded message received not matching the expected length
+- Added a way to save and load the srs and to cache the circuit verification key (only for testing purposes)
+
 ## 0.9.36 (03-27-2026)
 
 ### Added
--- a/mithril-stm/Cargo.toml
+++ b/mithril-stm/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mithril-stm"
-version = "0.9.36"
+version = "0.9.37"
 edition = { workspace = true }
 authors = { workspace = true }
 homepage = { workspace = true }


mithril-stm/src/proof_system/halo2_snark/unsafe_helpers.rs
jpraynaud, line 1: We probably miss some tests here.

mithril-test-lab/mithril-end-to-end/src/assertions/exec.rs
```diff
@@ -79,9 +79,9 @@ pub async fn update_protocol_parameters(aggregator: &Aggregator) -> StdResult<()
     info!("> stopping aggregator");
     aggregator.stop().await?;
     let protocol_parameters_new = ProtocolParameters {
-        k: 145,
-        m: 210,
-        phi_f: 0.80,
+        k: 14,
+        m: 21,
+        phi_f: 0.94,
     };
```
jpraynaud, line 85: Can you rollback these modifications as we don't want to use these parameters for Concatenation at the moment?

mithril-stm/src/proof_system/halo2_snark/unsafe_helpers.rs
```diff
@@ -54,37 +64,41 @@ impl SnarkSetup {
...
-            .join("circuits")
-            .join("halo2")
-            .join("assets")
+        // Uses a temporary directory to store the srs generated
+        // and to access it again during execution
+        let srs_path = env::temp_dir()
+            .join("mithril-srs")
             .join(format!("params_kzg_unsafe_{}", circuit_degree));
 
         let srs = load_or_generate_srs(
```
hjeljeli32, line 78: I am wondering whether we can have a race here around the SRS file.
My understanding from the Mithril protocol that only 1 aggregator builds/verifies the Snark setup, but if we are in a situation where there are multiple aggregators, another process may touch the same SRS path.

mithril-stm/src/proof_system/halo2_snark/unsafe_helpers.rs
```diff
@@ -54,37 +64,41 @@ impl SnarkSetup {
...
+        let circuit_degree = MidnightCircuit::from_relation(&circuit).min_k();
 
-        let srs_path = PathBuf::from(env!("CARGO_MANIFEST_DIR"))
-            .join("src")
-            .join("circuits")
-            .join("halo2")
-            .join("assets")
+        // Uses a temporary directory to store the srs generated
+        // and to access it again during execution
+        let srs_path = env::temp_dir()
```
hjeljeli32, line 74: I saw that in `mithril-test-lab/mithril-end-to-end/src/mithril/aggregator.rs` there is `MITHRIL_STM_SRS_DIR`, can we use it when it is present and only fallback to `env::temp_dir()`  when it is not?

damrobi: I believe I removed the mentions of `MITHRIL_STM_SRS_DIR` as commented by @jpraynaud in the other PR. I will check again that they are removed correctly but you are right that they should not be used anymore as we switch to `env::temp_dir()`

jpraynaud: This is just a temporary fix until we implement a production ready version for SRS and circuit keys. Also I'm not convinced that using an env var is the good way to provide such a configuration.

mithril-stm/src/proof_system/halo2_snark/unsafe_helpers.rs
damrobi: I don't think this should be a problem for the current tests as only the aggregator needs to access this file so unless we run two aggregators on the same device there should not be any access problems.

mithril-test-lab/mithril-end-to-end/src/assertions/exec.rs
damrobi: I rolled back the parameters

mithril-stm/src/proof_system/halo2_snark/unsafe_helpers.rs
damrobi: I added some tests in `unsafe_helpers.rs`, let me know if I missed some functionalities

mithril-stm/src/proof_system/halo2_snark/unsafe_helpers.rs
curiecrypt: Even though we should add a TODO for that, just to be on the safe side. Maybe the file doc could be updated with the TODO.

mithril-stm/src/proof_system/halo2_snark/unsafe_helpers.rs
```diff
@@ -95,20 +109,220 @@ fn load_or_generate_srs(circuit_degree: u32, path: &str) -> StmResult<ParamsKZG<
...
         Err(e) => Err(anyhow!("Failed to open SRS file at path '{path}': {e}")),
     }
 }
 
-/// Compute the circuit degree from the protocol parameter `k`.
+/// Persist a KZG SRS to disk so subsequent calls to `load_or_generate_srs` can load it.
+fn persist_srs(srs: &ParamsKZG<Bls12>, path: &str) -> StmResult<()> {
+    if let Some(parent) = std::path::Path::new(path).parent() {
+        std::fs::create_dir_all(parent)
+            .with_context(|| format!("Failed to create SRS directory at '{}'", parent.display()))?;
```
curiecrypt, line 130: We used a lot of `with_context` in this file. Most of them could be unnecessary or need to be modified when we have the actual SRS fix. However, it still makes me consider defining specific errors.
WDYT, @damrobi @jpraynaud ?

jpraynaud: It's better to use `with_context` when forwarding an error (especially when it comes from file system), otherwise we would have a `file does not exist`-type of error which is difficult to locate in the code.

I think that this code is just temporary and used during the build . It needs to be rewritten for production use in a future PR.

mithril-stm/src/proof_system/halo2_snark/proof.rs
```diff
@@ -243,11 +243,19 @@ mod tests {
...
             },
         },
     };
 
+    const BASE_CIRCUIT_DEGREE: u32 = 11;
+
+    /// Compute the circuit degree from the protocol parameter `k`.
+    ///
+    /// Returns an error if `k` is zero, since a quorum threshold of zero is invalid.
+    fn compute_circuit_degree(k: u64) -> u32 {
```
jpraynaud: This does not return an error actually.

damrobi: Fixed, it was a copy and paste leftover from when I imported it!

mithril-stm/src/proof_system/halo2_snark/unsafe_helpers.rs
damrobi: I will leave it like that for now then. I agree that the code needs to be adapted for production and rewritten in a cleaner way.

mithril-stm/src/proof_system/halo2_snark/unsafe_helpers.rs
damrobi: I added a TODO comment to warn about the use of the temporary directory

Issue #3099, opened 2026-03-23
Support partial block range in Cardano blocks and transactions


jpraynaud:
## Why
We want to certify partial block ranges (which do not have yet all their blocks and transactions). This will allow a better precision and regularity in the blocks and transactions which are signed.

## What
Support the certification of partial block range in Cardano Blocks and Transactions.

## How

- [x] Update the computation of `CardanoBlocksTransactionsSigningConfig::compute_block_number_to_be_signed` and remove the truncation to the step size
- [x] Update the computation of the `MithrilProverService::compute_proof` to support a partial last block range (to be used to retrieve all the nodes)
- [x] Update the `MithrilProverService::compute_cache` to compute on 
- [x] Make sure that the block range roots are not stored if their block range is not complete
- [x] Recompute on the fly the block range root for a last and incomplete block range in the `BlockRangeRootRetriever`

PR #3116, opened 2026-04-01
feat: support partial block ranges for blocks transactions certifications

Alenar:
## Content

This PR includes partial block ranges support for `CardanoBlocksTransactions` signatures, allowing signatures closer to the tip of the chain.

- relax `CardanoBlocksTransactionsSigningConfig` step configuration to allow step lower than a block range length and remove step adjustment to be a multiple of a block range length
- update merkle map computation to account for partial block range
- update prover to account for partial block range

> [!IMPORTANT]
> If a block range is partial it won't be stored in database, instead it will be computed in memory.

#### `mithril-common`

- signed entity config:
    - `CardanoBlocksTransactionsSigningConfig`: allow step smaller than a block range length and remove step adjustment to a multiple of block range length
    - `CardanoTransactionsSigningConfig` is unchanged and keep existing behavior
- add partial range support for `BlockRangeRootRetriever::compute_merkle_map_from_block_range_roots` blanket
- block range: new `is_complete_up_to` method
- crypto helpers:
  - new shortcut methods `compute_root_from_iter` for `MKMap` and `MKTree`
  - new `MKmap::keys(&self)` method: used to retrieve the last range included in the map (used by the 
  - fix `MKMap::from_blocks_with_transactions` test helpers: now order the given blocks and transactions before passing them to the map

#### `mithril-persistence`

- change behavior of `CardanoTransactionRepository::get_blocks_with_transactions_by_block_ranges`:
  - add `up_to_included: BlockNumber`  parameter to limit data retrieved for the given ranges list
  - ranges that intersect with `up_to_included` with it will be partially returned
  - ranges that start after `up_to_included` will be skipped
  - ranges that end below `up_to_included` are unchanged

#### `mithril-cardano-node-chain`

- Rewrite rustdoc for `BlockRangeImporter` runs methods and `InMemoryChainDataStoreBuilder::compute_block_ranges` to make more explicit how the end bound is computed

#### `mithril-aggregator`

- prover - add partial range support:
  - the code is mostly unchanged as the heavy lifting is done in the `BlockRangeRootRetriever`
  - add test coverage for the partial block ranges cases
- integration tests:
  - change `create_certificate` blocks transactions signing config so it produces partial ranges
  - new `prove_blocks_transactions` test:
    - with a small step of 5, producing a partial range two out of three times
    - compute two certificates: one for a complete range and one for a partial range
    - check that for the partial range no block range roots were computed and stored in database
    - produce a proof for block and one for transaction and validate them

#### `mithril-signer`

- new `test_create_cardano_blocks_transactions_single_signature` integration tests: check signature creation for blocks and transactions with complete and partial block range case 
- Refactor: move all `mock!` definition to a `test::double::mocks` module (same as the aggregator)

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation
  - [x] No new TODOs introduced

## Issue(s)

Relates to #3099

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@ As a minor extension, we have adopted a slightly different versioning convention
     - Support for computing and sending signatures for the `CardanoBlocksTransactions` signed entity type in the signer.
     - Support for verifying and aggregating signatures for the `CardanoBlocksTransactions` signed entity type in the aggregator.
     - Support for computing proofs of inclusions for a list of Cardano blocks or transactions hashes in the aggregator; the proof includes extra metadata about the included blocks or transactions.
+    - Support for signing with an interval lower than 15 blocks.
 
 - **UNSTABLE**:
   - Support for **DMQ consumer deduplication** to prevent processing the same message multiple times upon reconnection to the DMQ server.


mithril-aggregator/src/services/prover.rs
```diff
@@ -158,7 +161,7 @@ impl<S: MKTreeStorer> MithrilProverService<S> {
 
         // 2 - Fetch all nodes contained in the block ranges
         let nodes_per_block_ranges = self
-            .get_all_nodes_for_block_ranges(block_ranges.into_iter().collect())
+            .get_all_nodes_for_block_ranges(block_ranges.into_iter().collect(), up_to)
```
jpraynaud: IMO it would have been easier to compute adjusted block ranges in step 1, and rely on the block ranges to provide the upper bound. It would have avoided the introduction of a somewhat repetitive block number to represent this upper bound.

mithril-aggregator/src/services/prover.rs
```diff
@@ -406,11 +613,11 @@ mod tests {
                 .build_blocks_for_block_ranges(5);
             let blocks = into_blocks(&blocks_with_txs);
             // Only certify blocks for the first three ranges
-            let beacon = blocks_with_txs[3 * 3].block_number;
+            let beacon = blocks_with_txs[3 * 3].block_number - 1;
```
jpraynaud, line 738: Can you elaborate why this change is needed?

mithril-common/src/crypto_helper/merkle_map.rs
```diff
@@ -64,6 +64,21 @@ impl<K: MKMapKey, V: MKMapValue<K>, S: MKTreeStorer> MKMap<K, V, S> {
...
 
+    /// Get the keys of the merkelized map
+    pub fn keys(&self) -> impl DoubleEndedIterator<Item = &K> {
+        self.inner_map_values.keys()
+    }
+
+    /// Get the root of the merkle tree of a merkelized map built from an iterator
+    ///
+    /// Shortcut for `MKMap::new_from_iter(entries)?.compute_root()`
+    pub fn compute_root_from_iter<T: IntoIterator<Item = (K, V)>>(
```
jpraynaud: This function is useful only in the tests to avoid double unwraps. Can it be moved to the test extension?

mithril-common/src/crypto_helper/merkle_tree.rs
```diff
@@ -371,6 +371,16 @@ impl<S: MKTreeStorer> MKTree<S> {
         Ok(Self { inner_tree })
     }
 
+    /// Generate root of the Merkle tree built from the given leaves
+    ///
+    /// Shortcut for `MKTree::new_from_iter(leaves)?.compute_root()`
+    pub fn compute_root_from_iter<T: IntoIterator<Item = U>, U: Into<MKTreeNode>>(
```
jpraynaud: Same remark as above here.

mithril-common/src/entities/block_range.rs
```diff
@@ -37,6 +37,26 @@ impl BlockRange {
...
+    ///
+    /// ```rust
+    /// # use mithril_common::entities::{BlockNumber, BlockRange};
+    /// let range = BlockRange::from_block_number(BlockNumber(15));
+    ///
+    /// assert!(range.is_complete_up_to(BlockNumber(29)));
+    /// assert!(range.is_complete_up_to(BlockNumber(30)));
+    /// assert!(!range.is_complete_up_to(BlockNumber(28)));
+    /// ```
+    pub fn is_complete_up_to(&self, block_number: BlockNumber) -> bool {
```
jpraynaud: I think the "is complete up to" is not very clear. It would probably be better to refer to a block number being included or not in a range.

mithril-common/src/entities/signed_entity_config.rs
```diff
@@ -315,88 +313,273 @@ mod tests {
...
+            };
+            // ((150 - 0).div_euclid(15 - 1) * (15 - 1)) - 1 = (150.div_euclid(14) * 14) - 1 = 10 * 14 - 1 = 139
+            assert_eq!(
+                signing_config.compute_block_number_to_be_signed(block_number),
+                139,
+            );
+        }
+
+        #[test]
+        fn can_use_step_right_after_than_a_multiple_of_a_block_range_length() {
```
jpraynaud:
```suggestion
        fn can_use_step_right_after_a_multiple_of_a_block_range_length() {
```

mithril-common/src/entities/signed_entity_config.rs
```diff
@@ -315,88 +313,273 @@ mod tests {
...
+            };
+            // ((150 - 0).div_euclid(15 + 1) * (15 + 1)) - 1 = (150.div_euclid(16) * 16) - 1 = 9 * 16 - 1 = 143
+            assert_eq!(
+                signing_config.compute_block_number_to_be_signed(block_number),
+                143,
+            );
+        }
+
+        #[test]
+        fn can_use_step_higher_than_a_block_range_length_and_between_two_multiples_of_a_block_range_length()
```
jpraynaud: This name is not clear, in particular the `between_two_multiples`. Do you mean `between_two_consecutive_multiples`?

mithril-common/src/entities/signed_entity_config.rs
```diff
@@ -315,88 +313,273 @@ mod tests {
...
+            };
+            // ((150 - 0).div_euclid(15 + 6) * (15 + 6)) - 1 = (150.div_euclid(21) * 21) - 1 = 7 * 21 - 1 = 146
+            assert_eq!(
+                signing_config.compute_block_number_to_be_signed(block_number),
+                146,
+            );
+        }
+
+        #[test]
+        fn can_use_step_lower_than_a_block_range_length_and_between_two_multiples_of_a_block_range_length()
```
jpraynaud: Same remark as above here.

mithril-common/src/signable_builder/cardano_blocks_transactions.rs
```diff
@@ -37,9 +50,42 @@ pub trait BlockRangeRootRetriever<S: MKTreeStorer>: Send + Sync {
...
+        //
+        // Several cases:
+        // - `up_to_beacon` is not partial: no partial block range to compute and insert
+        // - `up_to_beacon` is partial but contained in last already stored range: no partial block range to compute and insert
+        // - `up_to_beacon` is partial and not contained in last already computed range: compute and insert the partial block range
+        let is_beacon_contained_in_last_computed_range = mk_hash_map
+            .keys()
+            .next_back()
+            .map(|range| range.contains(&up_to_beacon))
+            .unwrap_or_default();
```
jpraynaud, line 67:
```suggestion
        let is_beacon_contained_in_last_computed_range = mk_hash_map
            .keys()
            .nth_back(0)
            .map(|range| range.contains(&up_to_beacon))
            .unwrap_or_default();
```

mithril-common/src/signable_builder/cardano_blocks_transactions.rs
```diff
@@ -165,4 +211,250 @@ mod tests {
...
+                .concat()
+                .into_iter()
+                .map(|(k, v)| (k, v.into())),
+            )
+            .unwrap();
+            assert_eq!(expected_mk_map_root, retrieved_mk_map_root);
+        }
+
+        #[tokio::test]
+        async fn compute_when_given_block_is_partial_but_a_range_were_already_computed() {
```
jpraynaud:
```suggestion
        async fn compute_when_given_block_is_partial_but_a_range_was_already_computed() {
```

mithril-common/src/crypto_helper/merkle_tree.rs
Alenar: This one is used in production code by the `BlockRangeImporter`.

mithril-aggregator/src/services/prover.rs
Alenar: I transferred the range to retrieve computation to the prover and changed the retrieval of blocks and txs to be filtered by `Vec<Range<BlockNumber>>` instead of `Vec<BlockRange>` (as the former is not limited to steps of 15 blocks).

mithril-common/src/entities/signed_entity_config.rs
Alenar: I changed it to `can_use_step_higher_than_a_block_range_length_and_out_not_equal_to_a_range_boundaries`

mithril-common/src/entities/signed_entity_config.rs
Alenar: I changed it to `can_use_step_lower_than_a_block_range_length_and_out_not_equal_to_a_range_boundaries`

mithril-common/src/crypto_helper/merkle_tree.rs
Alenar: Done

mithril-common/src/entities/block_range.rs
Alenar: I renamed it to `is_fully_covered_at`. WDYT ?

mithril-aggregator/src/services/prover.rs
Alenar: This is because the 9th blocks computed here have a block number of `45`, meaning that leaving it without the `-1`would make this test fall into the "partial" case, which at this point of the development were not supported.
Also I wanted to characterize the "partial" case in other tests.

If you want I can remove it now, as without it the test is still green.

mithril-common/src/entities/block_range.rs
jpraynaud: It's better :+1:

PR #3185, opened 2026-04-13
fix: unstable signed entity hash computation

Alenar:
## Content

This PR includes a fix for signed entity hash computation which was unstable across target.

For `CardanoBlocksTransactions` only it feed the signed entity `index` to the hash, which was based on a `usize`. But `usize` representation change based on the target, e.g for a 32 bit target it would be an `u32` and for a 64 bit target a `u64`.

> [!TIP]
> This made the `wasm` based explorer certificate validation fails since we target a 32 bit target and the hashes were computed on linux which use a 64 target.

This PR fix this index type to an `u16`, ensuring stability across targets and giving us far enough available numbers for any future signed entity type (`65535`).

### Details

- add `SignedEntityTypeId` type alias for `u16`
- use `SignedEntityTypeId` instead of usize everywhere a signed entity type id/index was used
- `[additional]` fix a clippy issue in the aggregator when the `future_snark` feature was not used

> [!NOTE]
> In database conversion error message I replaced `usize` mention with `u16` instead of  `SignedEntityTypeId` but I wonder if the message should be reword instead to drop the concrete target type.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation
  - [x] No new TODOs introduced

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3962,7 +3962,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-aggregator"
-version = "0.8.43"
+version = "0.8.44"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -4265,7 +4265,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-common"
-version = "0.6.64"
+version = "0.6.65"
 dependencies = [
  "anyhow",
  "async-trait",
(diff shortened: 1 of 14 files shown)


jpraynaud (approved):
LGTM

turmelclem (approved):
LGTM 👍

PR #3204, opened 2026-04-16
Add golden tests for IVC

### LINKED ISSUE #3125: Add golden tests for recursive SNARK circuit prototype
(maybe add unit tests before for logic, ...)

hjeljeli32:
## Content

This PR includes the first Layer D golden-test safety net for the `halo2_ivc` prototype.

It adds:
- committed asset readers for the recursive verification context, stored chain checkpoint, and stored recursive step output
- local generators to rebuild those assets with deterministic setup
- a small positive golden suite covering:
  - validation of the stored recursive chain state asset
  - validation of the stored recursive step output asset
  - genesis/base-case coverage with `MockProver`
  - a positive next-epoch recursive-step check with `MockProver`
  - a positive same-epoch recursive-step check with `MockProver `
  - a representative chained-flow check that replays the stored step exactly
- documentation for the golden scenarios, the stored assets and regeneration commands

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation
  - [x] Add README file

## Issue(s)
Closes https://github.com/input-output-hk/mithril/issues/3125


mithril-stm/src/circuits/halo2_ivc/tests/golden/asset_readers.rs
```diff
@@ -0,0 +1,216 @@
...
+    let mut verifier_param_bytes = Vec::new();
+    reader.read_to_end(&mut verifier_param_bytes)?;
+
+    let mut verifier_params_reader = Cursor::new(&verifier_param_bytes);
+    let verifier_params =
+        ParamsVerifierKZG::<E>::read(&mut verifier_params_reader, SerdeFormat::RawBytesUnchecked)?;
+
+    let mut verifier_tau_reader = Cursor::new(&verifier_param_bytes);
+    let verifier_tau_in_g2 =
+        <E as Engine>::G2::read(&mut verifier_tau_reader, SerdeFormat::RawBytesUnchecked)?.into();
```
curiecrypt, line 255: This reads the first `G2` point from the raw `ParamsVerifierKZG`, assuming `s_g2` is serialized first. 
If the upstream serialization format changes, this will silently extract the wrong value. 
You can add a comment explaining the layout dependency.

mithril-stm/src/circuits/halo2_ivc/tests/golden/helpers.rs
```diff
@@ -0,0 +1,286 @@
...
+};
+use crate::circuits::halo2_ivc::{
+    Accumulator, AssignedAccumulator, C, CERT_VK_NAME, E, F, IVC_ONE_NAME, K, VerifyingKey,
+    circuit::IvcCircuit,
+    state::{Global, fixed_bases_and_names, trivial_acc},
+};
+
+const ASSET_SEED: u64 = 42;
+const CERTIFICATE_CIRCUIT_DEGREE: u32 = 13;
+const RECURSIVE_CIRCUIT_DEGREE: u32 = 19;
```
curiecrypt: We have the same constants defined in generators.rs as well. 
It would be better to define them in `mod.rs` with the minimum enough visibility.

mithril-stm/src/circuits/halo2_ivc/tests/golden/helpers.rs
```diff
@@ -0,0 +1,286 @@
...
+    pub(crate) certificate_fixed_bases: BTreeMap<String, C>,
+    pub(crate) recursive_fixed_bases: BTreeMap<String, C>,
+    pub(crate) combined_fixed_bases: BTreeMap<String, C>,
+    pub(crate) universal_verifier_params: ParamsVerifierKZG<E>,
+    pub(crate) verifier_tau_in_g2: <E as Engine>::G2Affine,
+    pub(crate) trivial_accumulator:
+        crate::circuits::halo2_ivc::Accumulator<crate::circuits::halo2_ivc::S>,
+}
+
+fn build_deterministic_params(circuit_degree: u32) -> ParamsKZG<Bls12> {
```
curiecrypt: This function is a duplicate of the one in generators.rs

mithril-stm/src/circuits/halo2_ivc/tests/golden/generators.rs
```diff
@@ -0,0 +1,1155 @@
...
+/// Wraps the protocol-message API so the call sites stay explicit about which
+/// fields define a generated asset.
+fn insert_protocol_message_part(
+    protocol_message: &mut ProtocolMessage,
+    part_key: ProtocolMessagePartKey,
+    encoded_value: Vec<u8>,
+) {
+    // Keep the protocol-message boundary explicit at the call site.
+    protocol_message.set_message_part(part_key, encoded_value);
+}
```
curiecrypt: This function does nothing beyond forwarding its arguments to `set_message_part`. The comment says it's about keeping the boundary "explicit," but the wrapper itself is what adds the indirection. 
Is there a specific reason for having it that I miss?

mithril-stm/src/circuits/halo2_ivc/tests/golden/generators.rs
```diff
@@ -0,0 +1,1155 @@
...
+        certificate_commitment_parameters
+    };
+
+    let recursive_commitment_parameters = {
+        let mut recursive_commitment_parameters = universal_kzg_parameters.clone();
+        if RECURSIVE_CIRCUIT_DEGREE < shared_srs_degree {
+            recursive_commitment_parameters.downsize(RECURSIVE_CIRCUIT_DEGREE);
+        }
+        recursive_commitment_parameters
+    };
```
curiecrypt: We can have a little helper for this selected functionality since it's been used in several places.

mithril-stm/src/circuits/halo2_ivc/tests/golden/asset_readers.rs
```diff
@@ -0,0 +1,216 @@
...
+            .map_err(|_| io::Error::new(io::ErrorKind::InvalidData, "invalid UTF-8 key"))?;
+
+        let point = C::read(reader, SerdeFormat::RawBytesUnchecked)?;
+        map.insert(name, point);
+    }
+
+    Ok(map)
+}
+
+pub(crate) fn load_recursive_chain_state_asset(
```
curiecrypt, line 191: Doc comment should be added.

mithril-stm/src/circuits/halo2_ivc/tests/golden/cases/mod.rs
```diff
@@ -0,0 +1 @@
+pub(crate) mod positive;
```
curiecrypt: does it have to be `pub(crate)`?

curiecrypt (approved):
The architecture is clean, the cryptographic flow is correct, and the documentation is thorough. I left a few comments, the rest LGTM 👍

mithril-stm/src/circuits/halo2_ivc/tests/golden/cases/positive.rs
```diff
@@ -0,0 +1,191 @@
+use midnight_circuits::types::Instantiable;
+
```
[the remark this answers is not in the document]
hjeljeli32: AssignedAccumulator::as_public_input(...) is a trait method from Instantiable. Given the code in positive.rs, the import is justified.

mithril-stm/src/circuits/halo2_ivc/tests/golden/helpers.rs
```diff
@@ -0,0 +1,286 @@
+use std::collections::BTreeMap;
+
+use group::Group;
+use midnight_circuits::hash::poseidon::PoseidonState;
+use midnight_circuits::types::Instantiable;
```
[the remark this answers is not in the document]
hjeljeli32: Same answer as above.

mithril-stm/src/circuits/halo2_ivc/tests/golden/generators.rs
```diff
@@ -0,0 +1,1155 @@
+use std::{
+    fs::{self, File},
+    io::{BufWriter, Write as IoWrite},
+    path::PathBuf,
+    time::Instant,
+};
+
+use ff::Field;
+use group::Group;
+use midnight_circuits::{hash::poseidon::PoseidonState, types::Instantiable};
```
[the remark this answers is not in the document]
hjeljeli32: Same answer as above

mithril-stm/src/circuits/halo2_ivc/tests/golden/asset_readers.rs
```diff
@@ -0,0 +1,216 @@
...
+    reader.read_to_end(&mut verifier_param_bytes)?;
+
+    let mut verifier_params_reader = Cursor::new(&verifier_param_bytes);
+    let verifier_params =
+        ParamsVerifierKZG::<E>::read(&mut verifier_params_reader, SerdeFormat::RawBytesUnchecked)?;
+
+    let mut verifier_tau_reader = Cursor::new(&verifier_param_bytes);
+    let verifier_tau_in_g2 =
+        <E as Engine>::G2::read(&mut verifier_tau_reader, SerdeFormat::RawBytesUnchecked)?.into();
+
```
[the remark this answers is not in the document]
hjeljeli32: The comment assumes ParamsVerifierKZG exposes something like s_g2(). In this codebase/API, it does not.
That is exactly why we kept the explicit read from the serialized verifier-param bytes

mithril-stm/src/circuits/halo2_ivc/tests/golden/generators.rs
hjeljeli32: @curiecrypt 
You are right. The wrapper was originally meant to keep the protocol-message boundary visible at the call sites. There is no strong reason anymore.
I’ll remove it and call `set_message_part(...)` directly.

mithril-stm/src/circuits/halo2_ivc/tests/golden/cases/mod.rs
hjeljeli32: You are right.
This module is only used inside the golden test subtree, so plain `mod positive;` is sufficient.

hjeljeli32:
> The architecture is clean, the cryptographic flow is correct, and the documentation is thorough. I left a few comments, the rest LGTM 👍

Thank you @curiecrypt for the review. I applied your suggestions 👍 .

mithril-stm/src/circuits/halo2_ivc/tests/golden/asset_readers.rs
kitounliu: when ParamsVerifierKZG is serialised, it only serialises s_g2: 
https://github.com/midnightntwrk/midnight-zk/blob/next/proofs/src/poly/kzg/params.rs#L247
Other elements in ParamsVerifierKZG are precomputed for pairing optimisation.

mithril-stm/src/circuits/halo2_ivc/tests/golden/GOLDEN_TEST_ASSETS.md
jpraynaud, line 1: Could you merge the 2 markdown file in a single `README.md` file?

mithril-stm/src/circuits/halo2_ivc/tests/fixtures.rs
```diff
@@ -0,0 +1,185 @@
...
+/// - `global_fes`: 5 field elements
+/// - `state`: 7 field elements
+/// - `proof`: length-prefixed proof bytes
+/// - `accumulator`: serialized accumulator
+#[derive(Debug)]
+pub(crate) struct ChainStateFixture {
+    pub(crate) global_fes: Vec<F>,
+    pub(crate) state: State,
+    pub(crate) proof: Vec<u8>,
+    pub(crate) accumulator: Accumulator<crate::circuits::halo2_ivc::S>,
```
jpraynaud, line 30: Could you move the import with the others ate the top of the file?

mithril-stm/src/circuits/halo2_ivc/tests/fixtures.rs
jpraynaud, line 1: Can you add doc comments to each field instead of adding them to the doc comment of the structs (here and in all other occurrences)? Also, doc comments are missing to all the functions.

mithril-stm/src/circuits/halo2_ivc/tests/fixtures.rs
```diff
@@ -0,0 +1,185 @@
...
+/// - `next_state`: 7 field elements
+#[derive(Debug)]
+pub(crate) struct AggregatorResultFixture {
+    pub(crate) proof: Vec<u8>,
+    pub(crate) next_accumulator: Accumulator<crate::circuits::halo2_ivc::S>,
+    pub(crate) next_state: State,
+}
+
+fn assets_dir() -> PathBuf {
+    PathBuf::from("mithril-stm/src/circuits/halo2_ivc/tests/assets")
```
jpraynaud, line 60: It'd be better to include all these files at compile time with `include_str!`

mithril-stm/src/circuits/halo2_ivc/tests/fixtures.rs
```diff
@@ -0,0 +1,185 @@
...
+
+pub(super) fn aggr_result_fixture_path() -> PathBuf {
+    assets_dir().join("aggr_result.bin")
+}
+
+fn open_fixture(path: &Path) -> io::Result<BufReader<File>> {
+    File::open(path).map(BufReader::new)
+}
+
+fn read_fe<R: Read>(reader: &mut R) -> io::Result<F> {
```
jpraynaud, line 79: This name `read_fe` is not very readable.

mithril-stm/src/circuits/halo2_ivc/tests/fixtures.rs
```diff
@@ -0,0 +1,185 @@
...
+        read_fe(reader)?,
+        read_fe(reader)?,
+        read_fe(reader)?,
+        read_fe(reader)?,
+        read_fe(reader)?,
+        read_fe(reader)?,
+    ))
+}
+
+fn read_proof<R: Read>(reader: &mut R) -> io::Result<Vec<u8>> {
```
jpraynaud, line 95: Why not use the bytes codec which leverages `Serialize` and `Deserialize` here?
Also, it'd be better to use the standard `StmResult` as return type.

mithril-stm/src/circuits/halo2_ivc/tests/golden/generators.rs
```diff
@@ -0,0 +1,1080 @@
...
+    let certificate_commitment_parameters = derive_commitment_params(
+        &universal_kzg_parameters,
+        shared_srs_degree,
+        CERTIFICATE_CIRCUIT_DEGREE,
+    );
+    let recursive_commitment_parameters = derive_commitment_params(
+        &universal_kzg_parameters,
+        shared_srs_degree,
+        RECURSIVE_CIRCUIT_DEGREE,
+    );
```
damrobi, line 243: I might be wrong but I'm not sure you need this split in the parameters generated. The `zk_lib::setup_vk` function mentions downsizing to the correct degree if necessary. It doesn't make much difference anyway since the operation needs to be performed at some point.

mithril-stm/src/circuits/halo2_ivc/tests/golden/generators.rs
```diff
@@ -0,0 +1,1080 @@
...
+        &[&[&[], public_inputs]],
+        random_generator,
+        &mut transcript,
+    )
+    .expect("IVC proof generation should not fail");
+    transcript.finalize()
+}
+
+/// Verifies a recursive proof using the Poseidon transcript and returns the prepared MSM.
+fn verify_and_prepare_poseidon_ivc(
```
damrobi, line 324: I think this function is the same as `verify_and_prepare_poseidon_recursive_proof` from the `helpers.rs` file. Same thing below with the blake2b version.

damrobi (approved):
LGTM 👍 
I left a couple of non-blocking comments. One thing I noticed but I'm not sure of is that every time we take a step in the chain we change epoch in the test. Let me know if I got this wrong! Otherwise do we have plan to have recursive steps within the same epoch as the circuit check for that are slightly different? I guess that can come at a later point.

mithril-stm/src/circuits/halo2_ivc/tests/golden/helpers.rs
jpraynaud, line 1: Could you re-order imports following our conventions? 

Also I see that many imports are inlined in the code, could you harmonize that and gather all the imports from `halo2_ivc` at the beginning of the file.

mithril-stm/src/circuits/halo2_ivc/tests/golden/helpers.rs
```diff
@@ -0,0 +1,242 @@
...
+}
+
+/// Runs `MockProver` on the recursive circuit and asserts that all constraints hold.
+pub(crate) fn assert_recursive_mock_prover_accepts(circuit: IvcCircuit, public_inputs: Vec<F>) {
+    let prover = MockProver::run(K, &circuit, vec![vec![], public_inputs])
+        .expect("recursive MockProver setup should succeed");
+    assert!(
+        prover.verify().is_ok(),
+        "recursive MockProver should accept the provided circuit and public inputs"
+    );
```
jpraynaud, line 91:
```suggestion
    prover.verify().expect("recursive MockProver should accept the provided circuit and public inputs");
```

mithril-stm/src/circuits/halo2_ivc/tests/golden/generators.rs
jpraynaud, line 324: These functions could be generic with the transcript hash to avoid code duplication.

mithril-stm/src/circuits/halo2_ivc/tests/golden/generators.rs
jpraynaud, line 1: This module is much too dense and hard to understand. Could you create refactor it into sub-modules with clear responsibilities?

mithril-stm/src/circuits/halo2_ivc/tests/golden/cases/positive.rs
```diff
@@ -0,0 +1,191 @@
...
+    helpers::{
+        assert_recursive_mock_prover_accepts, build_recursive_mock_prover_setup,
+        compute_exact_next_accumulator_from_assets, compute_expected_next_accumulator,
+        verify_and_prepare_blake2b_recursive_proof, verify_and_prepare_poseidon_recursive_proof,
+    },
+};
+use crate::circuits::halo2_ivc::{AssignedAccumulator, circuit::IvcCircuit, state::State};
+
+#[test]
+fn slow_genesis_base_case_mock_prover() {
```
jpraynaud, line 23: The names of the test functions lack clarity, could you update them?

mithril-stm/src/circuits/halo2_ivc/tests/test_certificate.rs
jpraynaud, line 1: Could you add a TODO and a comment stating that this is duplicate code used during the build phase and that it will be removed after and replaced with the existing non-recursive circuit implementation?

mithril-stm/src/circuits/halo2_ivc/tests/golden/GOLDEN_TEST_SCENARIOS.md
jpraynaud, line 1: IMO this these information should be accessible directly in the code and not in a markdown file.

mithril-stm/src/circuits/halo2_ivc/tests/golden/GOLDEN_TEST_ASSETS.md
jpraynaud, line 1: Could you rename this file into `README.md`?

mithril-stm/src/circuits/halo2_ivc/tests/golden/generators.rs
hjeljeli32: @damrobi You are right that `setup_vk(...)` can downsize internally, but here the circuit-specific params are also reused later on 2 different flows:
- certificate params for certificate proving / verification
- recursive params for recursive VK generation and recursive proving

So the main benefit is clarity: each functionality gets the params for its own circuit degree.

mithril-stm/src/circuits/halo2_ivc/tests/fixtures.rs
hjeljeli32: @jpraynaud 

I think this one is outdated.

The code has moved out of `fixtures.rs` into the golden test asset modules, and in the current files the imports are already grouped at the top.

mithril-stm/src/circuits/halo2_ivc/tests/fixtures.rs
hjeljeli32: Thanks @jpraynaud 

For the result type, I updated the current golden asset helper layer to use `StmResult` consistently, including the raw asset parsing/writing helpers, so it now matches the rest of the Halo2/STM test style more closely.

For the serialization part, I did not switch these helpers to the STM bytes codec yet. The reason is that this file is parsing a test-only binary asset container that bundles several heterogeneous proof-system values, some of which already rely on their own custom `read` / `write` APIs. Is it acceptable?

hjeljeli32:
> One thing I noticed but I'm not sure of is that every time we take a step in the chain we change epoch in the test. Let me know if I got this wrong! Otherwise do we have plan to have recursive steps within the same epoch as the circuit check for that are slightly different? I guess that can come at a later point.

Thanks so much @damrobi for pointing this out.
You are right, there is currently a gap: we don't have today a positive test for the recursive steps within the same epoch. I will add 1 or 2 positive tests that cover this.

mithril-stm/src/circuits/halo2_ivc/tests/fixtures.rs
hjeljeli32: @jpraynaud 

I think the original comment is on the outdated `fixtures.rs`, but I did revisit your idea in the current code.

I first tried the compile-time approach using `include_bytes!` since these assets are binary proof/state artifacts. That made the positive golden tests self-contained, but it also introduced an awkward bootstrap problem for the manual regeneration workflow: the generator entrypoints could no longer compile cleanly when the asset files were absent or had to be recreated from scratch.

I ended up reverting that part and going back to file-based loading while keeping the shared reader-based parsing/writing helpers. 

So I did explore the compile-time inclusion route, but in this case I think the simpler file-based workflow is the better tradeoff.
WDYT?

mithril-stm/src/circuits/halo2_ivc/tests/golden/generators/asset_generation.rs
```diff
@@ -0,0 +1,374 @@
...
+    build_recursive_global, build_recursive_proving_key, build_shared_recursive_context,
+};
+use super::transitions::{
+    build_genesis_base_case_next_state, build_genesis_base_case_witness,
+    build_next_certificate_asset_data,
+};
+use crate::circuits::halo2_ivc::state::trivial_acc;
+
+/// Generates and writes the stored recursive chain snapshot asset.
+pub(crate) fn generate_recursive_chain_state_asset(
```
jpraynaud: Maybe you can split this function which is a bit too long?

mithril-stm/src/circuits/halo2_ivc/tests/golden/generators/setup.rs
```diff
@@ -0,0 +1,287 @@
...
+        Self {
+            recursive_chain_state: base_dir.join("recursive_chain_state.bin"),
+            verification_context: base_dir.join("verification_context.bin"),
+            recursive_step_output: base_dir.join("recursive_step_output.bin"),
+        }
+    }
+}
+
+/// Returns the default committed asset paths under `tests/golden/assets`.
+pub(crate) fn default_asset_paths() -> AssetPaths {
```
jpraynaud: This could be replaced by a more idiomatic `Default` trait implementation

jpraynaud (approved):
LGTM 👍

damrobi (approved):
LGTM 👍 
The tests for the certificate chain within the same epoch look good!

curiecrypt (approved):
LGTM 👍

Issue #3206, opened 2026-04-17
Support multiple tiers for tests in CI


jpraynaud:
## Why
Some tests running in the CI are slow and this prevents a good developer experience. We need to split the tests in multiple tiers (e.g. `fast` and `slow`), always run the `fast` tier in the CI, and run the `slow` tier in the CI under some specific conditions.

## What
Adapt the CI to support multiple tiers for tests.

## How

- [ ] Update the cargo tests run in the GitHub CI workflow:
  - [ ] Create a whitelist of modules for which we detect that a modification has been made in the PR (start with `mithril-stm/circuits`)
  - [ ] Always run the `fast` tier
  - [ ] Run the `slow` tier for the modified modules
  - [ ] Run the `slow` tier for the modified modules when a label is added to the PR  (`with-slow-tests-run`)
- [ ] Run the `fast` tier in the Hydra CI 
- [ ] Run the `slow` tier in the nightly tests
- [ ] Create a DEV-ADR to describe the rules for test tiers

Alenar:
### Current slow tests inventory

| Slow tests (more than 30s run time)                           | when to run                                                                                       |
|---------------------------------------------------------------|---------------------------------------------------------------------------------------------------|
| **mithril-common test::builder::certificate_chain_builder::** |                                                                                                   |
| ~~test::builds_certificate_chain_with_correct_length~~        | reworked to be fast                                                                               |
| **mithril-stm circuits::halo2::tests::golden::cases::**       |                                                                                                   |
| negative::leaf_wrong_verification_key                         | when the `circuit::halo2` module change                                                           |
| negative::merkle_path_flip_position                           | when the `circuit::halo2` module change                                                           |
| negative::merkle_tree_commitment_mismatch                     | when the `circuit::halo2` module change                                                           |
| negative::signature_bad_challenge                             | when the `circuit::halo2` module change                                                           |
| negative::signature_bad_commitment                            | when the `circuit::halo2` module change                                                           |
| negative::signature_other_message                             | when the `circuit::halo2` module change                                                           |
| negative::signature_bad_response                              | when the `circuit::halo2` module change                                                           |
| negative::signature_verification_key_mismatch                 | when the `circuit::halo2` module change                                                           |
| negative::target_less_than_evaluation                         | when the `circuit::halo2` module change                                                           |
| negative::witness_duplicate_entry                             | when the `circuit::halo2` module change                                                           |
[...]

PR #3221, opened 2026-04-22
ci/test: multi speed tiers for tests

Alenar:
## Content

This PR introduce a two tiers mechanisms for tests:
- most tests have no changes, and run as before
- some tests are marked as "slow" (more on that below), by default they don't run in the CI except if their tested codes have changes

### Slow tests

The base rule is: any tests that **consistently** takes more than 30 seconds to run.

`cargo nextest` already reports by himself when a tests takes more than `60s`, the `ci-profile` have been modified to lower that threshold to the previously said `30s`.

But this is not enough:
- it depends on the running computer: tests will probably be faster on the developer computers, but sometimes its the reverse (it was for me with most `mithril_stm::circuits::halo2::tests::golden`)
- and execution time can very on a same machine depending on its load

This means that the `30s` must not considered as a strict threshold, we may want to include tests that are a little faster than that (in the 20~30s range) or exclude some that are slower (e.g. I choose to always run integration tests in `mithril-stm/tests/stm_protocol.rs`).

When choosing to mark a test as slow, the following change is made:
The tests is moved to a `slow` submodule inside the current file (most likely after all "fast" tests).

Moving it to a submodule allow to easily identify them using `cargo nextest` filter capabilities or even with `cargo test` using `-- --skip slow::` (to exclude all of them for this example).

### Mechanisms

A new devbook script is introduced, `filter-slow-tests.sh`, this script looks for the changed files on the current branch compared on `origin/main` (customizable) and compare it to a list of paths to watch, if any of those paths are in the changed files then it enable the related tests to runs.

This works by constructing a `cargo nextest` filterset ([reference](https://nexte.st/docs/filtersets/reference/)), which by default removes all tests located in a `slow` submodule but add an exception for all detected paths.

For example, if the paths lists contains only the following entries:

```shell
readonly -a SLOW_MITHRIL_STM_TESTS=(
  "mithril-stm/src/#protocol::aggregate_signature::"
  "mithril-stm/src/circuits/halo2#circuits::halo2::"
 )
```

If the files and folders container in `mithril-stm/src/circuits/halo2` does not contains any change then the produced default filterset will exclude all slow tests without exception:
```
not test(#*slow::*)
``` 

But, if there's a change to any file or folders in `mithril-stm/src/circuits/halo2`, then the an exclusion is added to the returned filterset:
```
not test(#*slow::*) or (package(mithril-stm) and (test(#protocol::aggregate_signature::*slow::*) or test(#circuits::halo2::*slow::*)))
```

Said filterset can be used directly with `cargo nextest` commands, e.g:
```shell
cargo nextest list --workspace --profile ci --filterset "$(./docs/devbook/filter-slow-tests/filter-slow-tests.sh)
```

> [!NOTE]
> - `cargo nextest` won't crash nor complain if a filterset does not apply to any tests of a targeted crate, meaning it can be applied freely on all environments even if they don't run `mithril-stm` tests.
> - The script is located in the devbook instead of the github workflows script so it can be more easily be discovered and used by Mithril developers
> - the script have two options: 
>   - `-c`/`--commit-ref` to change the comparison points for changed files discovery
>   - `-a`/`--all` to change the returned filterset to `all()`, useful if the triggering condition is not related to changed files (e.g. a PR label)

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation
  - [x] No new TODOs introduced

## Issue(s)

Relates to #3206

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -245,6 +245,9 @@ jobs:
           if [[ "${{ contains(github.event.pull_request.labels.*.name, 'run-slow-tests') }}" == "true" ]]; then
             FILTERSET="$(.github/workflows/scripts/filter-slow-tests.sh --all)"
           else
+            # For pull-request HEAD^1 targets the "simulated merge commit" created by GitHub Actions.
+            # For push (i.e. merging on main branch), HEAD^1 refers be an actual, not simulated, merge commit.
+            # Note: this approach does not correctly detect changes when multiple commits are pushed directly to the main branch at once.
             FILTERSET="$(.github/workflows/scripts/filter-slow-tests.sh --commit-ref "HEAD^1")"
           fi
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -227,6 +227,8 @@ jobs:
     steps:
       - name: Checkout sources
         uses: actions/checkout@v5
+        with:
+          fetch-depth: 2 # Needed to get the diff of the last commit, which is used to filter out slow tests
 
       - name: Install stable toolchain, tools, and restore cache
         uses: ./.github/workflows/actions/toolchain-and-cache
@@ -235,11 +237,20 @@ jobs:
           cargo-tools: cargo-nextest
           github-token: ${{ secrets.GITHUB_TOKEN }}
 
+      - id: prepare
+        name: Prepare
+        shell: bash
+        run: |
+          FILTERSET="$(./docs/devbook/filter-slow-tests/filter-slow-tests.sh --commit-ref "HEAD^1")"
+
+          echo "Tests filterset: $FILTERSET"
+          echo "filterset=$FILTERSET" >> $GITHUB_OUTPUT
+
       - name: Build tests
         run: cargo nextest run --cargo-profile ci-tests --no-run ${{ matrix.test-args }}
 
       - name: Run tests
-        run: cargo nextest run --profile ci --cargo-profile ci-tests ${{ matrix.test-args }}
+        run: cargo nextest run --profile ci --cargo-profile ci-tests --filterset "${{ steps.prepare.outputs.filterset }}" ${{ matrix.test-args }}
 
       - name: Run doc tests
         run: cargo test --profile ci-tests --doc ${{ matrix.test-args }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -243,9 +243,9 @@ jobs:
         run: |
           # Run all tests, including slow tests, if "run-slow-tests" label is in the PR labels
           if [[ "${{ contains(github.event.pull_request.labels.*.name, 'run-slow-tests') }}" == "true" ]]; then
-            FILTERSET="$(./docs/devbook/filter-slow-tests/filter-slow-tests.sh --all)"
+            FILTERSET="$(.github/workflows/scripts/filter-slow-tests.sh --all)"
           else
-            FILTERSET="$(./docs/devbook/filter-slow-tests/filter-slow-tests.sh --commit-ref "HEAD^1")"
+            FILTERSET="$(.github/workflows/scripts/filter-slow-tests.sh --commit-ref "HEAD^1")"
           fi
 
           echo "Tests filterset: $FILTERSET"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -241,7 +241,12 @@ jobs:
         name: Prepare
         shell: bash
         run: |
-          FILTERSET="$(./docs/devbook/filter-slow-tests/filter-slow-tests.sh --commit-ref "HEAD^1")"
+          # Run all tests, including slow tests, if "run-slow-tests" label is in the PR labels
+          if [[ "${{ contains(github.event.pull_request.labels.*.name, 'run-slow-tests') }}" == "true" ]]; then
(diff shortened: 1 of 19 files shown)


mithril-stm/src/protocol/aggregate_signature/mod.rs
```diff
@@ -484,4 +360,137 @@ mod tests {
...
+
+        proptest! {
+            #![proptest_config(ProptestConfig::with_cases(5))]
+
+            #[test]
+            /// Test that when the adversaries do not hold sufficient stake, they can not form a quorum
+            fn test_adversary_quorum(
+                (adversaries, parties) in arb_parties_adversary_stake(4, 15, 16, 4),
+                msg in any::<[u8;16]>(),
+            ) {
```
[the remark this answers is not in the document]
Alenar: You are right, but this is an existing offender and this PR won't modify stm tests, only move them if they are slow.

.github/workflows/ci.yml
```diff
@@ -235,11 +237,25 @@ jobs:
...
+      - id: prepare
+        name: Prepare
+        shell: bash
+        run: |
+          # Filter on PR but not when merging or if "with-slow-tests-run" label is in the PR labels
+          if [[ "${{ github.event_name }}" != "pull_request" ]] || [[ "${{ contains(github.event.pull_request.labels.*.name, 'run-slow-tests') }}" == "true" ]]; then
+            FILTERSET="$(./docs/devbook/filter-slow-tests/filter-slow-tests.sh --all)"
+          else
+            FILTERSET="$(./docs/devbook/filter-slow-tests/filter-slow-tests.sh --commit-ref "HEAD^1")"
+          fi
```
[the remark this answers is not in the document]
Alenar: This is not true here, but I understand why the mistake is made.

For PR what is actually fetch by `actions/checkout@v5` is a `refs/pull/<pr_number>/merge` which is "a simulated merge commit that represents what the repository would look like if the pull request were merged right now", see : https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/about-pull-requests#pull-request-refs-and-merge-branches

So `HEAD^1` does contains all the modified files of the whole PR.

Also, when merging, which trigger the `push` event, it's the actual merge commit that's used by `actions/checkout@v5`.

.github/workflows/scripts/filter-slow-tests.sh
jpraynaud, line 1: I'd rather have this script in the `.ci/scripts` folder now, and then initiate a discussion (and a refactoring) about where these scripts should live, how they should be documented, ..., with a dev ADR.

docs/devbook/README.md
```diff
@@ -4,7 +4,8 @@ This page gathers the available guides and tools for Mithril nodes developers.
...
-| **Upgrade the repository dependencies** | [upgrade-repository-dependencies](./upgrade-repository-dependencies/README.md) | Upgrade the project's dependencies in the repository                                          |
+| Operation                               | Location                                                                       | Description                                                                                            |
+| --------------------------------------- | ------------------------------------------------------------------------------ | ------------------------------------------------------------------------------------------------------ |
+| **Bump versions**                       | [bump-versions](bump-versions/README.md)                                       | Script to easily bump crates, js packages, and openapi versions before merging a pull request          |
+| **Filter slow tests**                   | [filter-slow-tests](filter-slow-tests/README.md)                               | Script that generates a `cargo nextest` filterset that excludes slow tests with unchanged dependencies |
```
jpraynaud:
```suggestion
| **Filter slow tests**                   | [filter-slow-tests](filter-slow-tests/README.md)                               | Script to generate a `cargo nextest` filterset that excludes slow tests with unchanged dependencies |
```

mithril-common/src/test/builder/certificate_chain_builder.rs
```diff
@@ -773,20 +761,48 @@ mod test {
     }
 
     #[test]
-    fn builds_certificate_chain_with_correct_length() {
```
jpraynaud, line 776: It's be worth keeping a test which builds the full chain maybe for a limited subset of the previous cases.

jpraynaud (approved):
LGTM 👍

mithril-stm/src/signature_scheme/bls_multi_signature/mod.rs
```diff
@@ -258,50 +258,59 @@ mod tests {
...
-        fn batch_verify(num_batches in 2..10usize,
-                              seed in any::<[u8;32]>(),
-        ) {
-            let mut rng = ChaCha20Rng::from_seed(seed);
-            let num_sigs = 10;
-            let mut batch_msgs = Vec::new();
-            let mut batch_vk = Vec::new();
-            let mut batch_sig = Vec::new();
-            for _ in 0..num_batches {
+    mod test {
```
damrobi: I'm not sure but I think this should be a `slow` module, shouldn't it?

damrobi (approved):
LGTM 👍 
I mainly had a look at the tests of the mithril-stm library and everything looks good there!

turmelclem (approved):
LGTM 👍

.github/workflows/scripts/filter-slow-tests.sh
Alenar: done

mithril-stm/src/signature_scheme/bls_multi_signature/mod.rs
Alenar: Good catch ! I fixed it.

PR #3232, opened 2026-04-27
ci: publish crates all at once

### LINKED ISSUE #3230: Refactor the crates publication
## Why
We need to refactor the crates publication to support introducing a new crate without having the CI failing on the dry-run publication.

## What
Publish all crates at once in the CI instead of publishing them sequentially.

## How 
 
- [x] Replace the matrix of publication of the crates with a single step to publish all crates:
  - [x] In CI workflow (dry run)
  - [x] In Pre-release workflow (dry-run)
  - [x] In Release workflow
- [x] Use the `CRATES_IO_API_TOKEN` secret for publication
- [x] ~~Delete all other crates publication tokens in the GitHub settings~~ (will be done in #2967)
- [x] Remove all constraints for importing Mithril crates in `Cargo.toml` files

Alenar:
## Content

This PR includes reworks the publication of rust crates to leverage the new ability of cargo to works on multiple crates at once.

Currently our publish have huge limitations which hurt our ability to deliver crates with correct dependencies to crates.io :
- we can't use exact versions for published crates depending on another published crates. We set those versions using `>=x.y.z` as a stopgap, but this means that a published `mithril-client` can be seen as compatible with an older `mithril-common` even if they are not.
- we are forced to do a first manual publication when we want to publish a new package

Those limitations are mostly due to the fact that we are publishing the crates one at a time, meaning that if crates `B` depends on a updated version of `A`, versions which is not yet published (likely because running publish in dry-run for the CI), the

The idea behind the introduction of the publish test in the CI was to protect ourselves against unpublished dependency graphs when releasing a new version. E.G., if crates `A` was already published but we want to publish a new version which introduce a dependency against a crate `B` we want this CI job to fail if `B` was not taken into account for the publication.

This PR removes those limitations by running `cargo publish`, dry run or not, on all crates that needs it at once. This means that now `cargo publish` is able to verify that the whole sets of crates to publish is coherent, solving our issues.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation
  - [x] No new TODOs introduced

## Issue(s)

Closes #3230

### DIFF
--- a/.github/workflows/actions/publish-rust-crates/action.yml
+++ b/.github/workflows/actions/publish-rust-crates/action.yml
@@ -1,12 +1,12 @@
-name: publish-crate-package
+name: publish-rust-crates
 description: |
-  Deploy the crate package to crates.io
+  Deploy a list of rust crates to crates.io
 inputs:
   dry_run:
     description: Dry run will not publish to crates.io, just test it.
     required: true
-  package:
-    description: crate package name.
+  packages:
+    description: comma separated list of rust crates to publish.
     required: true
   api_token:
     description: crates.io API token.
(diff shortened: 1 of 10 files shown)


jpraynaud (approved):
LGTM 👍

turmelclem (approved):
LGTM 👍

PR #3235, opened 2026-04-28
Test the ledger snapshot conversion in the e2e test

### LINKED ISSUE #3213: Test the ledger snapshot conversion in the e2e test
## Why
We need to detect early breaking changes in the `snapshot-converter` Cardano tool which is used under the hood by the `mithril-client tools utxo-hd snapshot-converter` command. Currently, we need to wait until the new version of the Cardano node is deployed on `testing-preview` before we can make sure that there is (or not) a breaking change.

## What
Integrate the snapshot converter as a step in the end to end test.

## How

- [ ] Update the `mithril-client tools utxo-hd snapshot-converter` to use an optional `--binary-path` location (when specified, no need to download the `snapshot-converter` from GitHub)
- [ ] In the end to end test, add an option `--with-snapshot-converter` to check the ledger snapshot conversion step which leverages the `--binary-path` option in the client CLI

turmelclem:
## Content

In order to detect early breaking changes in the `snapshot-converter,` the `mithril-client tools utxo-hd snapshot-converter `command has been improved to support to new parameters to specify the path of the binary of the snapshot converter to use and its configuration.  With so, no need to download Cardano-node when running the snapshot conversion.
-  `--binary-path` (for example `.../cardano-node-10.7.1-linux-amd64/bin/snapshot-converter`)
-  `--config-path` (for example `.../cardano-node-10.7.1-linux-amd64/share/preview/config.json`)

Those parameters are optional but if one of them is filled to other one must be filled to. 

on the E2E tests side, a new option has been introduced, `--check-client-cli-snapshot-converter`,  to run test scenario with client CLI snapshot conversion verification. 
- It required at least mithril-client-cli to be in `0.13.10` or upper
- the `utxo_hd_flavor` will be `LSM` for Cardano Node version `10.7.0 `and upper otherwise `LMDB`

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [ ] Reviewer requested
- Documentation
  - [x] No new TODOs introduced

Closes #3213


mithril-client-cli/src/commands/tools/utxo_hd/snapshot_converter.rs
```diff
@@ -303,7 +316,21 @@ impl SnapshotConverterCommand {
...
+            let mut steps = minimal_steps;
+            if !is_local_bin_mode {
+                steps += 2
+            };
+            if self.commit {
+                steps += 1
+            };
+            steps
+        };
+        let mut current_step = 0;
```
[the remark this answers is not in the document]
turmelclem: tests has been added on steps calculation, and command parsing

mithril-test-lab/mithril-end-to-end/src/assertions/check.rs
```diff
@@ -889,3 +890,42 @@ pub async fn assert_client_can_verify_cardano_stake_distribution(
...
+
+    let utxo_hd_flavor = if cardano_node_version.is_below("10.7.0") {
+        "LMDB"
+    } else {
+        "LSM"
+    };
+
+    let binary_path = artifacts_dir.join("bin").join("snapshot-converter");
+    let config_path = artifacts_dir
+        .join("node-full1")
```
Alenar: A magic string for this should be avoided, this should be a parameter instead

mithril-test-lab/mithril-end-to-end/src/assertions/check.rs
```diff
@@ -889,3 +890,42 @@ pub async fn assert_client_can_verify_cardano_stake_distribution(
 
     Ok(())
 }
+
+pub async fn assert_client_can_convert_the_ledger_snapshot(
+    client: &mut Client,
+    aggregator: &Aggregator,
```
Alenar: Imo you should pass a `FullNode` instead of the aggregator, the fact that it's the aggregator node that's used should not be known by this method.

mithril-test-lab/mithril-end-to-end/src/assertions/check.rs
```diff
@@ -889,3 +890,42 @@ pub async fn assert_client_can_verify_cardano_stake_distribution(
...
+        "LSM"
+    };
+
+    let binary_path = artifacts_dir.join("bin").join("snapshot-converter");
+    let config_path = artifacts_dir
+        .join("node-full1")
+        .join("snapshot-converter-config.json");
+
+    client
+        .run(ClientCommand::Tools(ToolsCommand::UtxoHd(
```
Alenar, line 925: Isn't this is running against a currently still running node ? This should be raising some warnings / errors from at least the node (on which you are replacing its ledger live) and maybe on the aggregator (because its node may stop working).

Instead you should copy the minimal amount needed (only the ledger dir ?) and run that on the copy.

This is not perfect as the node is still running but the alternatives would be stopping everything before running this step. Which requires additional changes to be handled correctly (maybe an addition of a "post run" scenario step which is executed after stopping the infrastructure).

mithril-test-lab/mithril-end-to-end/src/end_to_end_spec.rs
```diff
@@ -373,4 +378,21 @@ impl Spec {
 
         Ok(target_epoch)
     }
+
+    async fn verify_snapshot_conversion(
```
Alenar: Is this method really needed ? I would rather put its code directly where this method is currently called.

Alenar (approved):
LGTM with some remarks

mithril-client-cli/src/commands/tools/utxo_hd/snapshot_converter.rs
```diff
@@ -273,7 +286,7 @@ pub struct SnapshotConverterCommand {
     )]
     cardano_network: Option<CardanoNetworkCliArg>,
 
-    /// UTxO-HD flavor to convert the ledger snapshot to (`Legacy` or `LMDB`).
+    /// UTxO-HD flavor to convert the ledger snapshot to (`Legacy`, `LMDB` or `LMS`).
```
jpraynaud:
```suggestion
    /// UTxO-HD flavor to convert the ledger snapshot to (`Legacy`, `LMDB` or `LSM`).
```

jpraynaud (approved):
LGTM 👍

Issue #3210, opened 2026-04-20
Nightly tests are flaky


jpraynaud:
## Why
The nightly tests are flaky, and the client tests are failing randomly with the following error:
```
 Apr 08 01:48:25.906 ERRO Failed downloading and unpacking immutable_file_23651 for location FileUri(FileUri("https://storage.googleapis.com/cdn.aggregator.testing-preview.api.mithril.network/cardano-database/immutable/23651.tar.zst")), error: Download of location FileUri(FileUri("https://storage.googleapis.com/cdn.aggregator.testing-preview.api.mithril.network/cardano-database/immutable/23651.tar.zst")) failed after 3 attempts

Caused by:
    0: Download: could not write 0 bytes to stream.
    1: sending on a closed channel
Error: Can not download and unpack cardano db snapshot for hash: '916bcb816dcb2f2e6c23dc9d1c7913d99f063b68c44c54843641d110dd6b697c'

Caused by:
    All locations failed for immutable_file_23651, tried locations: https://storage.googleapis.com/cdn.aggregator.testing-preview.api.mithril.network/cardano-database/immutable/23651.tar.zst
```

Here is a list of recent failures:
- https://github.com/input-output-hk/mithril/actions/runs/24113008881/job/70351379474#step:6:1774
- https://github.com/input-output-hk/mithril/actions/runs/24167864127/job/70533186421#step:5:23
- https://github.com/input-output-hk/mithril/actions/runs/24222035085/job/70715406997#step:6:394
- https://github.com/input-output-hk/mithril/actions/runs/24321711203/job/71009048300#step:6:65
- https://github.com/input-output-hk/mithril/actions/runs/24431883249/job/71377894371#step:5:30
- https://github.com/input-output-hk/mithril/actions/runs/24487332278/job/71565033033#step:6:1046
- https://github.com/input-output-hk/mithril/actions/runs/24487332278/job/71565033026#step:6:65
- https://github.com/input-output-hk/mithril/actions/runs/24487332278/job/71565033027#step:6:60

## What

Investigate the source of the problem, and create a fix.

## How 

- [ ] Investigate the source of the problem
- [ ] Create a fix

PR #3248, opened 2026-05-04
fix(client): rare race condition when a file download finish

Alenar:
## Content

This PR a fix for a rare race condition that can occurs in the `mithril-client` library when it download a file.

### Issue

In our nightly test lately we often have one `test-client / bin-cdb-download` out of nine which fail with the following error:

> [!NOTE]
> The failing immutable vary 

```
src: CardanoDatabaseClient
 May 04 01:54:46.485 ERRO Failed downloading and unpacking immutable_file_22084 for location FileUri(FileUri("https://storage.googleapis.com/cdn.aggregator.testing-preview.api.mithril.network/cardano-database/immutable/22084.tar.zst")), error: Download of location FileUri(FileUri("https://storage.googleapis.com/cdn.aggregator.testing-preview.api.mithril.network/cardano-database/immutable/22084.tar.zst")) failed after 3 attempts

Caused by:
    0: Download: could not write 0 bytes to stream.
    1: sending on a closed channel
Error: Can not download and unpack cardano db snapshot for hash: 'e3bd1ae37e80af363de96a021f9f8dc9b1915b50d5f14a9e1dcabc6e24afbc8a'

Caused by:
    All locations failed for immutable_file_22084, tried locations: https://storage.googleapis.com/cdn.aggregator.testing-preview.api.mithril.network/cardano-database/immutable/22084.tar.zst
``` 

The code who throw the error is the `HttpFileDownloader::download_remote_file` method, specifically the `send_async` method of flume.
Looking at flume code, the `sending on a closed channel` is only sent when the receiver channel was drop, which can happen very rarely when the file download do a last loop, who will read 0 bytes, but the receiver was closed because the unpack task finished (probably because the archive unpacker may know just by looking at the bytes that no more bytes are needed).

### Fix

If no more bytes can be read, the file downloader now stop and don't try to send more bytes to the receiver. An error will still be send if the sender is disconnected with the receiver and there is still bytes to send.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation
  - [x] No new TODOs introduced

## Issue(s)

Relates to #3210

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4217,7 +4217,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-client"
-version = "0.14.7"
+version = "0.14.8"
 dependencies = [
  "anyhow",
  "async-trait",
--- a/mithril-client/Cargo.toml
+++ b/mithril-client/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mithril-client"
-version = "0.14.7"
+version = "0.14.8"
 description = "Mithril client library"
 authors = { workspace = true }
 edition = { workspace = true }
--- a/mithril-client/src/file_downloader/http.rs
+++ b/mithril-client/src/file_downloader/http.rs
@@ -240,7 +240,7 @@ mod tests {
 
     use httpmock::MockServer;
 
-    use mithril_common::{entities::FileUri, test::TempDir};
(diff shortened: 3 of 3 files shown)


turmelclem (approved):
LGTM 👍

jpraynaud (approved):
LGTM 👍

PR #3249, opened 2026-05-04
feat(stm): Trusted setup validation

### LINKED ISSUE #2944: Trusted setup for Halo2 circuit in STM
## Why

Halo2 circuits rely on a trusted setup (aka powers of tau) which is generated with a specific multi-party ceremony.
We have decided in #2800 to use the ceremony provided by Midnight ZK and we need to make it production ready.

## What

Implement the trusted setup for Mithril Halo2 circuits of STM library.

## How

- [ ] ~~Find a place for it (3GB) or truncate it to smaller size and provide tooling to do/verify our computation~~
- [ ] ~~How to deal with it in the CI (vs space limitation)?~~
- [ ] Use the [SRS](https://github.com/midnightntwrk/midnight-trusted-setup/blob/main/MIDNIGHT_SRS_CATALOG.md) from Midnight (~800MB)
- [ ] ~~Create a build script for STM library to download the SRS~~
- [ ] ~~Gate the snark compilation behind `future_snark` feature~~
- [ ] ~~Assess the compilation time with full SRS~~
- [ ] Update the MVP book

damrobi:
## Content

This PR includes a structure `SrsManager` managing the existence and path of the locally stored SRS. This structure implements functions allowing to verify the integrity of the SRS file against hard-coded values, to download the SRS file if it doesn't exist locally and to load it from disk to recreate the SRS.

Main functions:
- `verify_file_sha256_hash`: computes the SHA256 of the local file and checks it against its expected value
- `verify_bytes_sha256_hash`: computes the SHA256 of given bytes and checks it against its expected value
- `download_srs_file`: fetches the SRS from the stored URL and writes to disk
- `store_srs_bytes_to_file`: Saves the given bytes to the stored path 
- `download_srs_file_if_not_cached`: checks if a file exists at the stored path, otherwise downloads, verifies and stores it if its hash was correct
- `get_trusted_setup_parameters`: Loads the SRS from the path and returns it, downloading it if it is not available

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation
  - [ ] ~No new TODOs introduced~

## Issue(s)

Closes #2944


mithril-stm/src/circuits/mod.rs
```diff
@@ -7,6 +7,7 @@
 
 pub mod halo2;
 pub mod halo2_ivc;
+mod trusted_setup_verification;
```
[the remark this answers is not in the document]
damrobi: I updated the visibility of the module to `pub`. It is not yet used in the rest of the code but this will be done in another PR.

mithril-stm/src/circuits/trusted_setup.rs
```diff
@@ -0,0 +1,325 @@
...
+    /// and deserializes it into memory.
+    fn load(&self) -> StmResult<ParamsKZG<Bls12>> {
+        self.ensure_available()?;
+
+        let file = File::open(&self.path)
+            .with_context(|| format!("Failed to open SRS file at {:?}", self.path))?;
+        let mut reader = BufReader::new(file);
+
+        ParamsKZG::read_custom(&mut reader, SerdeFormat::RawBytes)
+            .with_context(|| "Failed to deserialize SRS from file")
```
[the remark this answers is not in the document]
damrobi: I added a test showing that the SRS can be loaded in both bytes format but I'm not sure it is worth keeping if we know that both format work. Maybe we can replace the test with a doc comment for the load function.

mithril-stm/Cargo.toml
```diff
@@ -55,6 +55,7 @@ rand = { version = "0.10.1", optional = true }
 rand_chacha = { workspace = true, optional = true }
 rand_core = { workspace = true, features = ["std"] }
 rayon = { workspace = true }
+reqwest = { workspace = true, features = ["blocking", "default"] }
```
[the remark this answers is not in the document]
damrobi: Done

mithril-stm/src/circuits/trusted_setup_verification.rs
```diff
@@ -0,0 +1,325 @@
...
+            .parent()
+            .ok_or(anyhow!("The given path contains no parent directory!"))?;
+        println!(
+            "Creating temporary directories to store SRS file: {:?}",
+            parent
+        );
+        std::fs::create_dir_all(parent)
+            .with_context(|| "Subdirectory creation should have succeeded!")?;
+
+        std::fs::write(&self.path, srs_bytes)?;
```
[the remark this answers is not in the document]
damrobi: I update the `download` code to enforce this writing to a temporary file and atomic renaming.

mithril-stm/src/circuits/trusted_setup.rs
jpraynaud, line 1: The module should be named `trusted_setup` IMO.

mithril-stm/src/circuits/trusted_setup.rs
```diff
@@ -0,0 +1,363 @@
...
+/// TODO: remove allow(dead_code) when the constants are used or remove the constatnts
+#[allow(dead_code)]
+/// Constant storing the hash of the SRS of degree 22 used to create proof in production
+///
+/// If the degree of the SRS used were to change, this hash would need to be updated using
+/// the proper value available here: https://github.com/midnightntwrk/midnight-trusted-setup/blob/main/MIDNIGHT_SRS_CATALOG.md
+const SRS_HASH_K22: &str = "e8ad5eed936d657a0fb59d2a55ba19f81a3083bb3554ef88f464f5377e9b2c2f";
+#[allow(dead_code)]
+/// Constant storing temporary path of the SRS of degree 22 used to create proof in production
+const SRS_PATH_K22: &str = "/tmp/trusted_setup/midnight-srs-2p22";
```
jpraynaud: It'd be better to have a folder than a filename here.
Also there is no notion of temporary storage, you should probably use local storage instead.

mithril-stm/src/circuits/trusted_setup.rs
```diff
@@ -0,0 +1,363 @@
...
+#[allow(dead_code)]
+/// Constant storing URL to download the SRS of degree 22 used to create proof in production
```
jpraynaud, line 23: Can you add a mention to the fact that we use Midnight SRS

mithril-stm/src/circuits/trusted_setup.rs
jpraynaud, line 1: Can you add test to make sure the SRS url and hash are not modified inadvertently?

mithril-stm/src/circuits/trusted_setup_verification.rs
```diff
@@ -0,0 +1,363 @@
...
+/// Constant storing temporary path of the SRS of degree 22 used to create proof in production
+const SRS_PATH_K22: &str = "/tmp/trusted_setup/midnight-srs-2p22";
+#[allow(dead_code)]
+/// Constant storing URL to download the SRS of degree 22 used to create proof in production
+const SRS_URL_K22: &str = "https://srs.midnight.network/midnight-srs-2p22";
+
+/// Manages the local storage and integrity verification of an SRS file.
+/// Stores the path of the local file, its expected hash value and a
+/// URL for where to download it if the file is missing locally.
+pub struct SrsManager {
```
jpraynaud: The name `SrsManager` is a bit too generic.

mithril-stm/src/circuits/trusted_setup.rs
```diff
@@ -0,0 +1,363 @@
...
+const SRS_URL_K22: &str = "https://srs.midnight.network/midnight-srs-2p22";
+
+/// Manages the local storage and integrity verification of an SRS file.
+/// Stores the path of the local file, its expected hash value and a
+/// URL for where to download it if the file is missing locally.
+pub struct SrsManager {
+    path: PathBuf,
+    expected_hash: String,
+    url: String,
```
jpraynaud: The field name should be more explicit and have doc comments for clarity.

mithril-stm/src/circuits/trusted_setup_verification.rs
jpraynaud: You should implement the `Default` trait with the constants you declared above in the code.

mithril-stm/src/circuits/trusted_setup.rs
```diff
@@ -0,0 +1,363 @@
...
+}
+
+/// TODO: remove allow(dead_code) when used
+#[allow(dead_code)]
+impl SrsManager {
+    fn new(
+        path: impl Into<PathBuf>,
+        expected_hash: impl Into<String>,
```
jpraynaud: Could you use generic types for `Into<_>` here?

mithril-stm/src/circuits/trusted_setup.rs
```diff
@@ -0,0 +1,363 @@
...
+        hex::encode(hasher.finalize())
+    }
+
+    /// Reads the local SRS file at the stored path and checks its SHA256 hash against the expected value.
+    // This function can be removed if it is not used.
+    fn verify_hash(&self) -> StmResult<bool> {
+        println!(
+            "Verifying integrity of the download file by checking its Sha256 hash value. Expected hash: {:?}",
+            self.expected_hash
+        );
```
jpraynaud: Can you avoid `println!` in production code?

mithril-stm/src/circuits/trusted_setup.rs
```diff
@@ -0,0 +1,363 @@
...
+        let mut file = File::open(&self.path)
+            .with_context(|| "Loading of the SRS file should have succeeded!")?;
+        let mut srs_buffer = vec![];
+        file.read_to_end(&mut srs_buffer)
+            .with_context(|| "Reading the SRS file should have succeeded!")?;
```
jpraynaud: It would be better to stream the bytes to compute the hash to avoid loading the full SRS in memory.

mithril-stm/src/circuits/trusted_setup_verification.rs
```diff
@@ -0,0 +1,363 @@
...
+    path: PathBuf,
+    expected_hash: String,
+    url: String,
+}
+
+/// TODO: remove allow(dead_code) when used
+#[allow(dead_code)]
+impl SrsManager {
```
jpraynaud: The name of the methods that should be public lack clarity and should probably be more explicit.

mithril-stm/src/circuits/trusted_setup.rs
```diff
@@ -0,0 +1,363 @@
...
+#[cfg(test)]
+mod tests {
+    use httpmock::MockServer;
+    use tempfile::NamedTempFile;
+
+    use super::*;
+
+    const SRS_HASH_K1: &str = "bbe04fe3c70d0c138447cb086b4baddc30cb8bb2a004114bc02e6f739516280e";
+
+    // Can be moved to a file in an asset directory
```
jpraynaud: Can you remove this comment?

mithril-stm/src/circuits/trusted_setup.rs
```diff
@@ -0,0 +1,363 @@
...
+            .with_context(|| format!("Failed to open SRS file at {:?}", self.path))?;
+        let mut reader = BufReader::new(file);
+
+        ParamsKZG::read_custom(&mut reader, SerdeFormat::RawBytes)
+            .with_context(|| "Failed to deserialize SRS from file")
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
jpraynaud, line 181: The names of the tests functions are not clear enough and should be more explicit.

mithril-stm/src/circuits/trusted_setup.rs
```diff
@@ -0,0 +1,363 @@
...
+    fn download(&self) -> StmResult<Vec<u8>> {
+        let response = reqwest::blocking::Client::builder()
+            // TODO: For now a timeout but this should be updated depending on the behavior we want
+            .timeout(std::time::Duration::from_secs(600))
+            .build()?
+            .get(&self.url)
+            .header("User-Agent", "mithril-stm")
+            .send()?
+            .error_for_status()?;
+        let bytes = response.bytes()?;
```
hjeljeli32, line 100: Same concern as pointed by @jpraynaud: `response.bytes()` buffers the entire response in RAM before returning. It would be better to stream the response body directly to a BufWriter and hash in flight, avoiding holding the whole file in memory at once.

mithril-stm/src/circuits/trusted_setup.rs
```diff
@@ -0,0 +1,363 @@
...
+        println!(
+            "Creating temporary directories to store SRS file: {:?}",
+            parent
+        );
+        std::fs::create_dir_all(parent)
+            .with_context(|| "Subdirectory creation should have succeeded!")?;
+
+        let mut temporary_file = NamedTempFile::new_in(parent)
+            .with_context(|| "Failed to generate temporary file to store the SRS")?;
+        std::fs::write(&mut temporary_file, srs_bytes)?;
```
hjeljeli32: `fs::write` opens a fresh file descriptor internally, so it bypasses the handle we already have open. IMO it would be better to write directly through the Write impl:
```
use std::io::Write;
BufWriter::new(&mut temporary_file).write_all(srs_bytes)?;
```

hjeljeli32 (commented):
Looks good, I left a couple of suggestions that can enhance how we manipulate the files.

mithril-stm/src/circuits/trusted_setup_verification.rs
damrobi: What do you think of `TrustedSetupHandle`? Or `SrsHandle`?

mithril-stm/src/circuits/trusted_setup.rs
damrobi: I added golden tests for the constants to detect a change, one would need to change both by mistake to not realize it. I could replace the check by the hash of the string to make it more robust.

damrobi: Renamed

mithril-stm/src/circuits/trusted_setup.rs
damrobi: Mention and link added

mithril-stm/src/circuits/trusted_setup.rs
damrobi: I tried to rework this path constant to be a local directory and added a path using the `HOME` environment variable for the default implementation. Let me know what you think

mithril-stm/src/circuits/trusted_setup.rs
damrobi: I improved the naming and added doc comments

mithril-stm/src/circuits/trusted_setup_verification.rs
damrobi: I added the `Default` implementation using the constants

mithril-stm/src/circuits/trusted_setup.rs
damrobi: I replace the `Into` by generic types

mithril-stm/src/circuits/trusted_setup.rs
damrobi: Removed

mithril-stm/src/circuits/trusted_setup_verification.rs
damrobi: I renamed the functions to have more explicit names, let me know if some are still not clear enough

mithril-stm/src/circuits/trusted_setup.rs
damrobi: Removed

mithril-stm/src/circuits/trusted_setup.rs
damrobi: I renamed the test functions

mithril-stm/src/circuits/trusted_setup.rs
damrobi: As we discussed, I will leave the implementation like that for now and the streaming aspect can be added in a later PR

mithril-stm/src/circuits/trusted_setup.rs
damrobi: As mentioned in JP's comment, we will leave this step for a later PR.

mithril-stm/src/circuits/trusted_setup.rs
damrobi: I applied your suggestion

mithril-stm/src/circuits/trusted_setup.rs
```diff
@@ -0,0 +1,376 @@
...
+#[allow(dead_code)]
+/// Constant storing local path of the SRS of degree 22 used to create proof in production
+const SRS_PATH_K22: &str = ".local/local_trusted_setup/";
+#[allow(dead_code)]
+/// Constant storing URL to download the SRS of degree 22 used to create proof in production
+const SRS_URL_K22: &str = "https://srs.midnight.network/midnight-srs-2p22";
+
+/// An handle to manage the trusted setup srs. It stores the local path of the srs file
+/// and information to download the file and verify integrity if it is missing.
+pub struct TrustedSetupHandle {
```
jpraynaud: Maybe this name is better?
```suggestion
pub struct TrustedSetupProvider {
```

mithril-stm/src/circuits/trusted_setup.rs
```diff
@@ -0,0 +1,376 @@
...
+    url_to_download_srs: String,
+}
+
+/// TODO: remove allow(dead_code) when used
+#[allow(dead_code)]
+impl TrustedSetupHandle {
+    fn new<P, S>(local_srs_path: P, srs_expected_hash: S, url_to_download_srs: S) -> Self
+    where
+        P: Into<PathBuf>,
+        S: Into<String>,
```
jpraynaud:
```suggestion
    fn new<P: Into<PathBuf>, S: Into<String>>(local_srs_path: P, srs_expected_hash: S, url_to_download_srs: S) -> Self
```

mithril-stm/src/circuits/trusted_setup.rs
```diff
@@ -0,0 +1,376 @@
...
+
+        ParamsKZG::read_custom(&mut reader, SerdeFormat::RawBytes)
+            .with_context(|| "Failed to deserialize SRS from file")
+    }
+}
+
+impl Default for TrustedSetupHandle {
+    fn default() -> Self {
+        Self::new(
+            PathBuf::from(std::env::var("HOME").unwrap_or_else(|_| ".".to_string()))
```
jpraynaud: Can you use a `mithril-circuit/srs` folder in the the temp folder by default?
Ideally `std::env::temp_dir().join(MITHRIL_CIRCUIT_CACHE_FOLDER).join("srs')` with `MITHRIL_CIRCUIT_CACHE_FOLDER` in the mod.rs of `circuits` (to be used in the circuit cache later).

mithril-stm/src/circuits/trusted_setup.rs
```diff
@@ -0,0 +1,376 @@
...
+        BufWriter::new(&mut temporary_file).write_all(srs_bytes)?;
+        temporary_file
+            .persist(&self.local_srs_path)
+            .with_context(|| "Failed to atomically rename SRS temp file.")?;
+        Ok(())
+    }
+
+    /// Ensures the SRS file is present. If the file is missing,
+    /// downloads it, verifies its hash and stores it if the hash is valid.
+    fn ensure_srs_file_is_available(&self) -> StmResult<()> {
```
jpraynaud: ~~The caller will need the final path of the cached SRS file:~~
```suggestion
    pub fn get_verified_local_srs_file(&self) -> StmResult<PathBuf> {
```

~~or the KZG params:~~
```suggestion
    pub fn get_verified_local_srs_file(&self) -> StmResult<ParamsKZG<Bls12>> {
```

mithril-stm/src/circuits/trusted_setup.rs
```diff
@@ -0,0 +1,376 @@
...
+    fn verify_bytes_sha256_hash(&self, srs_bytes: &[u8]) -> bool {
+        let recomputed_hash = Self::compute_hash(srs_bytes);
+
+        self.srs_expected_hash == recomputed_hash
+    }
+
+    /// Fetches the SRS from `self.url_to_download_srs` and returns its bytes.
+    fn download_srs_file(&self) -> StmResult<Vec<u8>> {
+        let response = reqwest::blocking::Client::builder()
+            // TODO: For now a timeout but this should be updated depending on the behavior we want
```
jpraynaud: In that case, you can make the timeout a parameter of the struct and set a default value in the `Default` implementation.

mithril-stm/src/circuits/trusted_setup.rs
curiecrypt: This function only verifies the hash on the download branch, the file-exists branch returns `Ok(())` unconditionally and `check_available_file_then_load_srs` then deserializes whatever is on disk into `ParamsKZG`. 
Any of the following silently corrupts every produced proof: 
- a partial write from a previous crashed run by an older version of this code
- an SRS rotation where the constant changes but the cached file doesnt
- user error placing the wrong-degree SRS at the path
- and maybe others ?

mithril-stm/Cargo.toml
```diff
@@ -55,9 +55,11 @@ rand = { version = "0.10.1", optional = true }
 rand_chacha = { workspace = true, optional = true }
 rand_core = { workspace = true, features = ["std"] }
 rayon = { workspace = true }
+reqwest = { workspace = true, default-features = true, features = ["blocking"] }
 serde = { workspace = true }
 sha2 = { version = "0.10.9", optional = true }
 subtle = "2.6.1"
+tempfile = "3.27.0"
```
jpraynaud: You should not need this crate. You should instead download the file to a non final file name and proceed to the renaming when the file has been completely downloaded.

mithril-stm/Cargo.toml
```diff
@@ -55,9 +55,11 @@ rand = { version = "0.10.1", optional = true }
 rand_chacha = { workspace = true, optional = true }
 rand_core = { workspace = true, features = ["std"] }
 rayon = { workspace = true }
+reqwest = { workspace = true, default-features = true, features = ["blocking"] }
```
jpraynaud: This should be gated behind `future-snark`.

mithril-stm/src/circuits/trusted_setup.rs
```diff
@@ -0,0 +1,376 @@
...
+        let file = File::open(&self.local_srs_path)
+            .with_context(|| format!("Failed to open SRS file at {:?}", self.local_srs_path))?;
+        let mut reader = BufReader::new(file);
+
+        ParamsKZG::read_custom(&mut reader, SerdeFormat::RawBytes)
+            .with_context(|| "Failed to deserialize SRS from file")
+    }
+}
+
+impl Default for TrustedSetupHandle {
```
curiecrypt: This code silently falls back to "." when HOME is unset.
We could either drop Default entirely and require an explicit path at the call site, or expose a fallible named constructor (sth like `TrustedSetupHandle::production_k22() -> StmResult<Self>`) that errors when HOME is unset.

mithril-stm/src/circuits/trusted_setup.rs
```diff
@@ -0,0 +1,376 @@
...
+
+    /// Ensures the SRS file is present. If the file is missing,
+    /// downloads it, verifies its hash and stores it if the hash is valid.
+    fn ensure_srs_file_is_available(&self) -> StmResult<()> {
+        if !self.local_srs_path.exists() {
+            let srs_bytes = self
+                .download_srs_file()
+                .with_context(|| "Download of the SRS file should have succeeded!")?;
+            if !self.verify_bytes_sha256_hash(&srs_bytes) {
+                return Err(anyhow!(
```
curiecrypt: We don't have any diagnostic context here. Should we define error enums instead of only returning manual errors here? Of course with diagnostic data. WDYT?

mithril-stm/src/circuits/trusted_setup.rs
```diff
@@ -0,0 +1,376 @@
...
+    /// Fetches the SRS from `self.url_to_download_srs` and returns its bytes.
+    fn download_srs_file(&self) -> StmResult<Vec<u8>> {
+        let response = reqwest::blocking::Client::builder()
+            // TODO: For now a timeout but this should be updated depending on the behavior we want
+            .timeout(std::time::Duration::from_secs(600))
+            .build()?
+            .get(&self.url_to_download_srs)
+            .header("User-Agent", "mithril-stm")
+            .send()?
+            .error_for_status()?;
```
curiecrypt, line 99: I guess this only flags non-**2xx**. 
For example, a captive portal returning 200 OK with an HTML "login required" page passes through to `bytes()` and gives the same generic "hash does not match". 
The hash check eventually catches it but the resulting message wont be accurate.
We might need to handle this catch in a more granular wat?

mithril-stm/src/circuits/trusted_setup.rs
```diff
@@ -0,0 +1,376 @@
...
+/// Constant storing local path of the SRS of degree 22 used to create proof in production
+const SRS_PATH_K22: &str = ".local/local_trusted_setup/";
+#[allow(dead_code)]
+/// Constant storing URL to download the SRS of degree 22 used to create proof in production
+const SRS_URL_K22: &str = "https://srs.midnight.network/midnight-srs-2p22";
+
+/// An handle to manage the trusted setup srs. It stores the local path of the srs file
+/// and information to download the file and verify integrity if it is missing.
+pub struct TrustedSetupHandle {
+    /// Path of the local srs file
```
curiecrypt:
```suggestion
    /// Path of the local SRS file
```

curiecrypt: There are a few more like this one.

mithril-stm/src/circuits/trusted_setup.rs
```diff
@@ -0,0 +1,376 @@
...
+        let recomputed_hash = Self::compute_hash(srs_bytes);
+
+        self.srs_expected_hash == recomputed_hash
+    }
+
+    /// Fetches the SRS from `self.url_to_download_srs` and returns its bytes.
+    fn download_srs_file(&self) -> StmResult<Vec<u8>> {
+        let response = reqwest::blocking::Client::builder()
+            // TODO: For now a timeout but this should be updated depending on the behavior we want
+            .timeout(std::time::Duration::from_secs(600))
```
curiecrypt: Should we define a const for this duration?

mithril-stm/src/circuits/trusted_setup.rs
```diff
@@ -0,0 +1,376 @@
...
+            TrustedSetupHandle::new(srs_file.path(), "", "").ensure_srs_file_is_available();
+
+        assert!(result.is_ok());
+    }
+
+    mod mock_server_test {
+        use super::*;
+
+        #[test]
+        fn missing_srs_file_triggers_download_verifiation_and_storage() {
```
curiecrypt:
```suggestion
        fn missing_srs_file_triggers_download_verification_and_storage() {
```

mithril-stm/src/circuits/trusted_setup.rs
```diff
@@ -0,0 +1,376 @@
...
+    fn compute_hash(bytes: &[u8]) -> String {
+        let mut hasher = Sha256::new();
+        hasher.update(bytes);
+
+        hex::encode(hasher.finalize())
+    }
+
+    /// Reads the local SRS file at the stored path and checks its SHA256 hash against the expected value.
+    // This function can be removed if it is not used.
+    fn verify_file_sha256_hash(&self) -> StmResult<bool> {
```
curiecrypt: This function returns `StmResult<bool>` and `verify_bytes_sha256_hash` returns bool. 
A future `verify_file_sha256_hash()?` call would give errors but silently discards a hash mismatch with `Ok(false)`. 
In this case a tampered SRS could load into the prover with no error.

curiecrypt (commented):
The first round of review, I left some comments and happy to discuss, since I am not sure about all of them

mithril-stm/Cargo.toml
damrobi: Done

mithril-stm/src/circuits/trusted_setup.rs
damrobi: Added timeout parameter

mithril-stm/src/circuits/trusted_setup.rs
damrobi: Updated the structure name

mithril-stm/src/circuits/trusted_setup.rs
damrobi: This is true but the design we opted for, described in the issue, assumes that if a file is present and named after the SRS file, it can be considered valid. If it is corrupted, the creation of the SRS from this file will most likely fail and if the version is wrong it is possible that some proof are generated but they won't verify with the correct verification key. We can also change the behavior to verify the hash before returning the SRS in the `check_available_file_then_load_srs` function.

mithril-stm/src/circuits/trusted_setup.rs
damrobi: This function is not used yet so I am not sure we need to keep it unless we plan to have this functionality of verifying the content of a SRS file.
Regarding your comment, it could indeed return an Ok(false) so its output needs to be checked in a similar way to the `verify_bytes_sha256_hash` function. I will describe its behavior better in the doc comment.

mithril-stm/src/circuits/trusted_setup.rs
damrobi: fixed

mithril-stm/src/circuits/trusted_setup.rs
damrobi: I added the timeout value to the struct as suggested by JP

mithril-stm/src/circuits/trusted_setup.rs
damrobi: I updated the comments to fix them

mithril-stm/src/circuits/trusted_setup.rs
damrobi: Done

mithril-stm/src/circuits/trusted_setup.rs
damrobi: I am fine with adding a specific error here, maybe we will have more when implementing the next step with verification and proving keys caching

mithril-stm/src/circuits/trusted_setup.rs
damrobi: True! There should be no problem like that with midnight's server but I can add a check for a HTLM response to return a more precise error.

jpraynaud: IMO this not necessary, this will complicate the code and this case is very unlikely to happen.

mithril-stm/src/circuits/trusted_setup.rs
jpraynaud: Let's stick with the current design: if the aggregator is corrupted, it will create a proof which will not match the circuit verification key whitelisted. However, in the long run, if the hash computation is very fast compared to the proving time we could add this safety check at every proof generation.

mithril-stm/src/circuits/trusted_setup.rs
jpraynaud: The return type should be simply `StdResult<()>` to avoid confusion 👍

jpraynaud: If the function is not used, it needs to be removed :+1:

mithril-stm/src/circuits/trusted_setup.rs
damrobi: Done

mithril-stm/Cargo.toml
damrobi: Removed

mithril-stm/src/circuits/trusted_setup.rs
damrobi: I updated this part with JP's suggestion to use the temporary directory again so we don't use HOME anymore

mithril-stm/src/circuits/trusted_setup.rs
damrobi: I removed the function

mithril-stm/src/circuits/trusted_setup.rs
damrobi: I created an error enum that for now only contains the hash verification error for now. The rest of the errors from this module are mostly linked to read/write and download so I don't know if we want to define custom errors for them.

mithril-stm/src/circuits/mod.rs
```diff
@@ -7,10 +7,25 @@
...
+/// Errors which can be outputted by the trusted setup verification.
+#[cfg(feature = "future_snark")]
+#[derive(Debug, Clone, thiserror::Error, PartialEq, Eq)]
+pub enum TrustedSetupError {
+    /// The hash verification of the SRS bytes failed
+    #[error(
+        "The hash of the SRS file does not match the hard-coded value! Expected: {0}, Current hash: {1}"
+    )]
+    VerifyHashFail(String, String),
+}
```
jpraynaud: You can move these errors to the `trusted_setup` module.

mithril-stm/src/circuits/mod.rs
```diff
@@ -7,10 +7,25 @@
...
+
+/// Errors which can be outputted by the trusted setup verification.
+#[cfg(feature = "future_snark")]
+#[derive(Debug, Clone, thiserror::Error, PartialEq, Eq)]
+pub enum TrustedSetupError {
+    /// The hash verification of the SRS bytes failed
+    #[error(
+        "The hash of the SRS file does not match the hard-coded value! Expected: {0}, Current hash: {1}"
+    )]
+    VerifyHashFail(String, String),
```
jpraynaud:
```suggestion
    #[error(
        "The hash of the SRS file does not match the hard-coded value! Expected: {expected}, Current hash: {computed}"
    )]
    ```suggestion
    VerifyHashFail{
      expected: String
      computed: String
    },
```
```

File: mithril-stm/src/circuits/trusted_setup.rs [hunk 33]
```diff
@@ -0,0 +1,379 @@
...
+
+/// TODO: remove allow(dead_code) when the constants are used or remove the constatnts
+#[allow(dead_code)]
+/// Constant storing the hash of the SRS of degree 22 used to create proof in production.
+/// This SRS is coming from the trusted setup done by Midnight and available in the following
+/// repository: https://github.com/midnightntwrk/midnight-trusted-setup.
+///
+/// If the degree of the SRS used were to change, this hash would need to be updated using
+/// the proper value available here: https://github.com/midnightntwrk/midnight-trusted-setup/blob/main/MIDNIGHT_SRS_CATALOG.md
+const SRS_HASH_K22: &str = "e8ad5eed936d657a0fb59d2a55ba19f81a3083bb3554ef88f464f5377e9b2c2f";
```
jpraynaud:
```suggestion
const MIDNIGHT_SRS_K22_HASH: &str = "e8ad5eed936d657a0fb59d2a55ba19f81a3083bb3554ef88f464f5377e9b2c2f";
```

mithril-stm/src/circuits/trusted_setup.rs
```diff
@@ -0,0 +1,379 @@
...
+#[allow(dead_code)]
+/// Constant storing URL to download the SRS of degree 22 used to create proof in production
+const SRS_URL_K22: &str = "https://srs.midnight.network/midnight-srs-2p22";
```
jpraynaud:
```suggestion
const MIDNIGHT_SRS_K22_URL: &str = "https://srs.midnight.network/midnight-srs-2p22";
```

mithril-stm/src/circuits/trusted_setup.rs
```diff
@@ -0,0 +1,379 @@
...
+            .send()?
+            .error_for_status()?;
+        let bytes = response.bytes()?;
+
+        Ok(bytes.to_vec())
+    }
+
+    /// Saves the given bytes in a temporary file then atomically moves it to the stored path
+    /// while creating the directories of the path if needed.
+    /// If the writing is interu
```
jpraynaud: This sentence is apparently incomplete.

mithril-stm/src/circuits/trusted_setup.rs
```diff
@@ -0,0 +1,379 @@
...
+const SRS_HASH_K22: &str = "e8ad5eed936d657a0fb59d2a55ba19f81a3083bb3554ef88f464f5377e9b2c2f";
+#[allow(dead_code)]
+/// Constant storing URL to download the SRS of degree 22 used to create proof in production
+const SRS_URL_K22: &str = "https://srs.midnight.network/midnight-srs-2p22";
+
+/// A structure to manage the trusted setup SRS. It stores the local path of the SRS file
+/// and information to download the file and verify integrity if it is missing.
+pub struct TrustedSetupProvider {
+    /// Path of the local SRS file
+    local_srs_path: PathBuf,
```
jpraynaud: You should probably use this field to store the local folder this would avoid having to call `.parent()` later on it.

```suggestion
    local_srs_folder_path: PathBuf,
```

mithril-stm/src/circuits/trusted_setup.rs
```diff
@@ -0,0 +1,379 @@
...
+#[allow(dead_code)]
+impl TrustedSetupProvider {
+    fn new<P: Into<PathBuf>, S: Into<String>>(
+        local_srs_path: P,
+        srs_expected_hash: S,
+        url_to_download_srs: S,
+        download_timeout_limit: Duration,
+    ) -> Self {
+        Self {
+            local_srs_path: local_srs_path.into(),
```
jpraynaud: When you build the provider the local folder path should be the `local_srs_folder_path` concatenated with the SRS subfolder (const to be created).

mithril-stm/src/circuits/trusted_setup.rs
```diff
@@ -0,0 +1,379 @@
...
+            .ok_or(anyhow!("The given path contains no parent directory!"))?;
+        std::fs::create_dir_all(parent)
+            .with_context(|| "Subdirectory creation should have succeeded!")?;
+
+        let temp_path = self.local_srs_path.with_extension(".temp");
+        // Create/clean the file
+        let mut temporary_file = File::create(&temp_path)?;
+        BufWriter::new(&mut temporary_file).write_all(srs_bytes)?;
+
+        // atomic renaming
```
jpraynaud:
```suggestion
        let mut temporary_file = File::create(&temp_path)?;
        BufWriter::new(&mut temporary_file).write_all(srs_bytes)?;
```

mithril-stm/src/circuits/trusted_setup.rs
```diff
@@ -0,0 +1,379 @@
...
+        std::fs::rename(temp_path, &self.local_srs_path)?;
+
+        Ok(())
+    }
+
+    /// Ensures the SRS file is present. If the file is missing,
+    /// downloads it, verifies its hash and stores it if the hash is valid.
+    fn ensure_srs_file_is_available(&self) -> StmResult<()> {
```
jpraynaud:
```suggestion
    fn download_srs_file_if_not_cached(&self) -> StmResult<()> {
```

mithril-stm/src/circuits/mod.rs
jpraynaud, line 1: Could you remove the `!` in the error messages?

mithril-stm/src/circuits/trusted_setup.rs
```diff
@@ -0,0 +1,379 @@
...
+            self.store_srs_bytes_to_file(&srs_bytes)
+                .with_context(|| "Saving the SRS to disk should have succeeded!")?;
+        }
+
+        Ok(())
+    }
+
+    /// Ensures the SRS file is available, downloading it if necessary
+    /// and deserializes it into memory.
+    fn check_available_file_then_load_srs(&self) -> StmResult<ParamsKZG<Bls12>> {
```
jpraynaud:
```suggestion
    fn get_trusted_setup_parameters(&self) -> StmResult<ParamsKZG<Bls12>> {
```

mithril-stm/src/circuits/trusted_setup.rs
```diff
@@ -0,0 +1,379 @@
...
+
+        ParamsKZG::read_custom(&mut reader, SerdeFormat::RawBytesUnchecked)
+            .with_context(|| "Failed to deserialize SRS from file!")
+    }
+}
+
+impl Default for TrustedSetupProvider {
+    fn default() -> Self {
+        Self::new(
+            std::env::temp_dir().join(MITHRIL_CIRCUIT_CACHE_FOLDER).join("srs"),
```
jpraynaud: The `new` function should fix the name of the subfolder:
```suggestion
            std::env::temp_dir().join(MITHRIL_CIRCUIT_CACHE_FOLDER),
```

mithril-stm/src/circuits/trusted_setup.rs
```diff
@@ -0,0 +1,379 @@
...
+    #[test]
+    fn hash_of_correct_bytes_verifies() {
+        let result = TrustedSetupProvider::new("", SRS_HASH_K1, "", Duration::from_secs(600))
+            .verify_bytes_sha256_hash(SRS_K1);
+
+        assert!(result.is_ok());
+    }
+
+    #[test]
+    fn existing_file_on_disk_skips_download_and_verification() {
```
jpraynaud, line 297: This test is a bit weak. It'd be better to use a http mock server and make sure that it is not called when a valid SRS is already on disk.

mithril-stm/src/circuits/trusted_setup.rs
```diff
@@ -0,0 +1,379 @@
...
+        std::fs::write(&srs_file, [0, 1, 2, 3, 4]).unwrap();
+
+        let result = TrustedSetupProvider::new(srs_file.path(), "", "", Duration::from_secs(600))
+            .ensure_srs_file_is_available();
+
+        assert!(result.is_ok());
+    }
+
+    #[test]
+    fn interrupted_writing_of_srs_resumes_properly_at_next_try() {
```
jpraynaud, line 322: This test should make sure that the correct complete file has been downloaded.

mithril-stm/src/circuits/trusted_setup.rs
```diff
@@ -0,0 +1,379 @@
...
+        let temp_srs_file = srs_file.path().with_extension("temp");
+        std::fs::write(&temp_srs_file, [0, 1, 2, 3, 4]).unwrap();
+
+        let result = TrustedSetupProvider::new(srs_file.path(), "", "", Duration::from_secs(600))
+            .ensure_srs_file_is_available();
+
+        assert!(result.is_ok());
+    }
+
+    mod mock_server_test {
```
jpraynaud: No need to create a submodule for that.

mithril-stm/src/circuits/mod.rs
damrobi: Moved

mithril-stm/src/circuits/mod.rs
damrobi: Updated

mithril-stm/src/circuits/trusted_setup.rs
damrobi: Updated

mithril-stm/src/circuits/trusted_setup.rs
damrobi: Updated

mithril-stm/src/circuits/trusted_setup.rs
damrobi: Fixed

mithril-stm/src/circuits/trusted_setup.rs
damrobi: Removed the comments

mithril-stm/src/circuits/trusted_setup.rs
damrobi: Updated

mithril-stm/src/circuits/mod.rs
damrobi: Removed

mithril-stm/src/circuits/trusted_setup.rs
damrobi: Updated

mithril-stm/src/circuits/trusted_setup.rs
damrobi: Removed the module

mithril-stm/src/circuits/trusted_setup.rs
damrobi: Improved the test

mithril-stm/src/circuits/mod.rs
```diff
@@ -7,10 +7,17 @@
...
 
 pub(crate) use halo2::witness::{
     CircuitInstance, CircuitMerkleTreeLeaf, CircuitWitness, MerklePath,
 };
+
+/// Constant holding the current path of the cached values related to the circuits
+const MITHRIL_CIRCUIT_CACHE_FOLDER: &str = "mithril-circuit";
+
+/// Constant holding the current path of the SRS file
+const MITHRIL_CIRCUIT_SRS_FOLDER: &str = "srs";
```
jpraynaud: You can move this one in the `trusted_setup` module.

damrobi: Done

mithril-stm/src/circuits/trusted_setup.rs
damrobi: Done

mithril-stm/src/circuits/trusted_setup.rs
damrobi: I created a constant for the subfolder and for the filename of the SRS. It can be changed later if needed

mithril-stm/src/circuits/trusted_setup.rs
damrobi: Done

mithril-stm/src/circuits/trusted_setup.rs
```diff
@@ -0,0 +1,441 @@
...
+    url_to_download_srs: String,
+    /// The timeout limit when trying to download the SRS file
+    download_timeout_limit: Duration,
+}
+
+/// TODO: remove allow(dead_code) when used
+#[allow(dead_code)]
+impl TrustedSetupProvider {
+    /// Create a new TrustedSetupProvider
+    /// Prepares a subfolder in the given `local_srs_folder_path` to put the SRS in
```
jpraynaud:
```suggestion
```

mithril-stm/src/circuits/trusted_setup.rs
```diff
@@ -0,0 +1,441 @@
...
+            }
+            .into());
+        }
+        Ok(())
+    }
+
+    /// Fetches the SRS from `self.url_to_download_srs` and returns its bytes.
+    fn download_srs_file(&self) -> StmResult<Vec<u8>> {
+        let response = reqwest::blocking::Client::builder()
+            // TODO: For now a timeout but this should be updated depending on the behavior we want
```
jpraynaud: You forgot to remove this TODO:
```suggestion
```

mithril-stm/src/circuits/trusted_setup.rs
```diff
@@ -0,0 +1,441 @@
...
+            self.store_srs_bytes_to_file(&srs_bytes)
+                .with_context(|| "Saving the SRS to disk should have succeeded.")?;
+        }
+
+        Ok(())
+    }
+
+    /// Ensures the SRS file is available, downloading it if necessary
+    /// and deserializes it into memory.
+    fn get_trusted_setup_parameters(&self) -> StmResult<ParamsKZG<Bls12>> {
```
jpraynaud: This function should probably be public:
```suggestion
    pub fn get_trusted_setup_parameters(&self) -> StmResult<ParamsKZG<Bls12>> {
```

mithril-stm/src/circuits/trusted_setup.rs
```diff
@@ -0,0 +1,441 @@
...
+};
+
+use anyhow::Context;
+use midnight_curves::Bls12;
+use midnight_proofs::{poly::kzg::params::ParamsKZG, utils::SerdeFormat};
+use sha2::{Digest, Sha256};
+
+use crate::{StmResult, circuits::MITHRIL_CIRCUIT_CACHE_FOLDER};
+
+#[allow(dead_code)]
```
jpraynaud: This can be removed:

```suggestion
```

mithril-stm/src/circuits/trusted_setup.rs
```diff
@@ -0,0 +1,441 @@
...
+    /// The timeout limit when trying to download the SRS file
+    download_timeout_limit: Duration,
+}
+
+/// TODO: remove allow(dead_code) when used
+#[allow(dead_code)]
+impl TrustedSetupProvider {
+    /// Create a new TrustedSetupProvider
+    /// Prepares a subfolder in the given `local_srs_folder_path` to put the SRS in
+    fn new<P: Into<PathBuf>, S: Into<String>, U: Into<String>>(
```
jpraynaud: This function should probably be public:
```suggestion
    pub fn new<P: Into<PathBuf>, S: Into<String>, U: Into<String>>(
```

mithril-stm/src/circuits/trusted_setup.rs
```diff
@@ -0,0 +1,441 @@
...
+            .local_srs_folder_path
+            .join(MITHRIL_CIRCUIT_SRS_FILENAME)
+            .with_extension("temp");
+        let mut temporary_file = File::create(&temp_path)?;
+        BufWriter::new(&mut temporary_file).write_all(srs_bytes)?;
+
+        std::fs::rename(
+            temp_path,
+            self.local_srs_folder_path.join(MITHRIL_CIRCUIT_SRS_FILENAME),
+        )?;
```
jpraynaud: It's better to make sure we safely rename the file and to remove the `BufWriter` which is not needed as the bytes are already in memory:
```suggestion
      let temp_path = self                                                                                                                                                                                                                                                                              
          .local_srs_folder_path                         
          .join(MITHRIL_CIRCUIT_SRS_FILENAME)                                                                                                                                                                                                                                                           
          .with_extension("temp");
      let final_path = self.local_srs_folder_path.join(MITHRIL_CIRCUIT_SRS_FILENAME);                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                        
      let mut temporary_file = File::create(&temp_path)
          .with_context(|| format!("Failed to create temporary SRS file at {temp_path:?}."))?;                                                                                                                                                                                                          
      temporary_file.write_all(srs_bytes)?;                                                                                                                                                                                                                                                             
      temporary_file
          .sync_all()                                                                                                                                                                                                                                                                                   
          .with_context(|| "Failed to fsync temporary SRS file before rename.")?;                                                                                                                                                                                                                       
      drop(temporary_file);
                                                                                                                                                                                                                                                                                                        
      std::fs::rename(&temp_path, &final_path)?;                                                                                                                                                                                                                                                        
   
      File::open(&self.local_srs_folder_path)                                                                                                                                                                                                                                                           
          .and_then(|dir| dir.sync_all())                
          .with_context(|| "Failed to fsync SRS directory after rename.")?;
```

mithril-stm/Cargo.toml
```diff
@@ -55,6 +56,7 @@ rand = { version = "0.10.1", optional = true }
 rand_chacha = { workspace = true, optional = true }
 rand_core = { workspace = true, features = ["std"] }
 rayon = { workspace = true }
+reqwest = { workspace = true, default-features = true, features = ["blocking"], optional = true }
```
jpraynaud:
```suggestion
reqwest = { workspace = true, default-features = false, features = ["blocking"], optional = true }
```

mithril-stm/Cargo.toml
```diff
@@ -30,6 +30,7 @@ future_snark = [
     "dep:rand",
     "dep:sha2",
     "dep:rand_chacha",
+    "dep:reqwest",
 ]
```
jpraynaud, line 35: The TLS backend should be configurable:
```suggestion
]
# TLS backend for the SRS download client (used by `future_snark`).                                                                                                                                                                                                                                   
# Exactly one of these must be enabled together with `future_snark`.                                                                                                                                                                                                                                  
native-tls = ["reqwest?/native-tls"]                                                                                                                                                                                                                                                                   
rustls = ["reqwest?/rustls"] 

```

mithril-stm/src/circuits/mod.rs
```diff
@@ -7,10 +7,14 @@
...
 
 #[cfg(test)]
 pub(crate) mod test_utils;
 
 pub(crate) use halo2::witness::{
     CircuitInstance, CircuitMerkleTreeLeaf, CircuitWitness, MerklePath,
 };
+
+/// Constant holding the current path of the cached values related to the circuits
+const MITHRIL_CIRCUIT_CACHE_FOLDER: &str = "mithril-circuit";
```
jpraynaud, line 20: This will create a compilation error if the TLS feature is missing:
```suggestion
const MITHRIL_CIRCUIT_CACHE_FOLDER: &str = "mithril-circuit";

#[cfg(all(                                                                                                                                                                                                                                                                                            
    feature = "future_snark",                                                                                                                                                                                                                                                                         
    not(any(feature = "rustls", feature = "native-tls"))
))]                                                                                                                                                                                                                                                                                                   
compile_error!(                                        
    "Enabling `future_snark` requires exactly one of the `rustls` or `native-tls` features."
);
```

mithril-stm/src/circuits/trusted_setup.rs
damrobi: Removed

mithril-stm/src/circuits/trusted_setup.rs
damrobi: Removed

mithril-stm/src/circuits/trusted_setup.rs
damrobi: Done

mithril-stm/src/circuits/trusted_setup.rs
damrobi: Removed

mithril-stm/src/circuits/trusted_setup.rs
damrobi: Done

mithril-stm/src/circuits/trusted_setup.rs
damrobi: Updated

mithril-stm/Cargo.toml
damrobi: Changed

mithril-stm/src/circuits/mod.rs
damrobi: Added to mod.rs

mithril-stm/Cargo.toml
damrobi: Added to toml

jpraynaud (approved):
LGTM 👍

curiecrypt (approved):
LGTM 🚀

PR #3265, opened 2026-05-13
refactor: strengthen unknown and future signed entity support

Alenar:
## Content

This PR strengthen unknown and future signed entity support:
- For future type it adds new integrations tests which runs the `mithril-aggregator` and `mithril-signer` state machines with all known signed entity types, including unstable variants.
- For unknown type it defines two new enum `SignedEntityTypeMessage` and `SignedEntityTypeDiscriminantMessage` which are almost identical to the existing entities, but with an additional `Unknown` variant which is parametrized as a "catch all" for unknown values when parsing from serde or strings.
  Those new "messages part" replace usage of their equivalent entities in all messages, allowing mithril nodes to identify gracefully if a signed entity variant is unknown to them. 

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation
  - [x] No new TODOs introduced

## Issue(s)

Relates to #3172


mithril-common/src/entities/signed_entity_type.rs
```diff
@@ -210,7 +210,7 @@ impl TryToBytes for SignedEntityType {
 }
 
 impl SignedEntityTypeDiscriminants {
-    /// Get all the discriminants
+    /// Get all the discriminants without unstable values
```
jpraynaud: This is not very clear.

jpraynaud (approved):
LGTM 👍

internal/mithril-protocol-config/src/http.rs
```diff
@@ -218,7 +227,10 @@ mod tests {
     #[tokio::test]
     async fn eventual_eventual_inconsistent_discriminants_are_removed_from_enabled_list() {
         let configuration_epoch_56 = ProtocolConfigurationMessage {
-            available_signed_entity_types: SignedEntityTypeDiscriminants::all(),
+            available_signed_entity_types: SignedEntityTypeDiscriminants::all_with_unstable()
```
turmelclem, line 230: At first I was thinking `all_with_unstable()` was production code, and I understand next that is in a extention just for test. And I was asking myself if it didn't make the test more difficult to read ?

turmelclem (approved):
LGTM 👍

mithril-common/src/messages/message_parts/signed_entity_type.rs
```diff
@@ -0,0 +1,784 @@
...
+
+impl SignedEntityTypeDiscriminantsMessage {
+    /// Get all the discriminants without unstable values
+    pub fn all() -> BTreeSet<Self> {
+        // Leverage the list from `SignedEntityTypeDiscriminants` to avoid Unknown and duplicating the filter
+        SignedEntityTypeDiscriminants::all()
+            .into_iter()
+            .map(Self::from)
+            .collect()
+    }
```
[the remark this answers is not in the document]
Alenar: I added a filter mechanism to `SignedEntityTypeDiscriminants::iter_all()`

internal/mithril-protocol-config/src/http.rs
Alenar: This is a trade-off with the "extension trait" mechanism: when called the methods looks like they are members of the extended struct or enum, but they are not.
Without an IDE you can't easily know that those are "test" methods by reading the code.

mithril-common/src/entities/signed_entity_type.rs
Alenar: I changed the comment and added a filter mechanism to those lists. They now exclude all values contained in a ` UNSTABLE_DISCRIMINANTS` const array.
This array is empty but we may add values in the future for cases such as the `CardanoBlocksTransactions` which was intentionally excluded at first.

This should make more clear what an "unstable" discriminant is and why they are excluded from `all()`.

DEV-ADR.md
```diff
@@ -23,6 +23,56 @@ To complete
 To complete
 -->
 
+### 9. Guidelines for adding a new Signed Entity
```
jpraynaud:
```suggestion
### 9. Guidelines for adding a new Signed Entity Type
```

DEV-ADR.md
```diff
@@ -23,6 +23,56 @@ To complete
...
+
+**Date:** 2026-05-19
+**Status:** Draft
+
+#### Context
+
+Adding a new Signed Entity typically requires changes across several parts of the system:
```
jpraynaud: There are also several other occurrences below:
```suggestion
Adding a new Signed Entity Type typically requires changes across several parts of the system:
```

jpraynaud (approved):
LGTM 👍

Issue #3138, opened 2026-04-01
Recursive SNARK aggregation primitives: Prepare prover input


jpraynaud:

perturbing:
For the last step in the IVC, we ideally should use [this](https://github.com/midnightntwrk/midnight-zk/pull/322) Plutus-friendly hasher. Still to do

- [ x ] Check that this is aligned with the Risc0 verifier

EDIT: we currently do not use a plutus friendly hasher (we use [this](https://github.com/midnightntwrk/midnight-zk/blob/af4bd615036dbab66fece84fd354c6d4cf7d4638/proofs/src/transcript/implementors.rs#L13) as seen [here](https://github.com/input-output-hk/mithril/blob/61c98eabdeef37bc1b473c3d1599e7419134ecb1/mithril-stm/src/circuits/halo2_ivc/tests/common/generators/proofs.rs#L109)). This hasher has an ever growing state and uses Blake2b-512 (which is not onchain). Note that newer versions of midnight-zk do have a more friendly implementation as per [this](https://github.com/midnightntwrk/midnight-zk/issues/299) closed issue.

So after an update to a newer version of midnight-zk, a trivial line change to a plutus friendly one enables the ivc transcript to run onchain. That said, blake512 can easily run in Risc0.

PR #3278, opened 2026-05-20
feat(stm): Prepare prover input - Phase 3

curiecrypt:
## Content

This PR includes _Phase 3 of the IVC prover input preparation (issue #3138)_. Lands the load-once setup, per-call epoch data decoder, and the persistent rolling state that bridges consecutive IVC proving steps.

### What this branch adds

- **`IvcSetup::load`** pulls the SRS, certificate VK, IVC VK, IVC PK, and the three fixed-base maps via four injected providers. Body is final; only the temp provider types in the signature will be swapped when the production IVC cache providers ship.
- **`EpochData::new`** decodes the certificate epoch, next Merkle root, and next protocol parameters from the 190-byte protocol-message preimage. Decoding mirrors the in-circuit `combine_bytes` step (little-endian times 256^i reduction in Fq); `CircuitBase` and `JubjubBase` both alias `midnight_curves::Fq`.
- **`IvcRollingState`** is the caller-owned bridge between consecutive proving steps. Persists chain state, IVC proof bytes, folded accumulator, and the genesis Schnorr signature. Streams through the IVC `IvcRead`/`IvcWrite` codec; on-disk schema version `0x10`. `save`/`read_from_disk` are thin file-I/O wrappers around the streaming impl.
- **Temp setup providers**: three store-prereqs providers (`TempCertificateVerifyingKeyProvider`, `TempIvcVerifyingKeyProvider`, `TempIvcProvingKeyProvider`) recompute their artifacts on demand from owned fields. SRS shared via `Arc<ParamsKZG<Bls12>>` to avoid a full ~800 MB clone at K=22. Replaced wholesale once the production cache providers land.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

### Cross-cutting changes

- `circuits/halo2_ivc/K` made `pub(crate)` so temp providers can pin the IVC circuit's domain size.
- New type aliases `CircuitVerifyingKey` / `CircuitProvingKey` in the module's `ivc_halo2_snark/mod.rs` to keep the long key types readable.
- New `IvcRead`/`IvcWrite` impls for `State` (in `io.rs`, same module) and `StandardSchnorrSignature` (next to the type in `standard_signature.rs`). `IvcRollingState`'s own codec impl lives next to the type in `rolling_state.rs`.

### Design notes worth knowing

- **Codec alignment**: rolling state uses the streaming `IvcRead`/`IvcWrite` trait, not the lib's CBOR `codec.rs` path, because (a) its inner types (`Accumulator`, `State`) are already streaming-codec-only and (b) the file is IVC-internal, not a public wire format.
- **Schema version byte (`0x10`)** is a schema version for the streaming layout, not a `CODEC_VERSION_*` codec dispatcher. Value distinct from `CODEC_VERSION_CBOR_V1 = 0x01` so a misrouted file would fail on the version check rather than producing a confusing CBOR decode error.
- **Temp providers match the real cache providers' call-style API.** 
  - The `IvcSetup::load` body and its chain of `get_*` calls **stay identical when swapping**; only the provider construction at the caller (test environment)changes. 
  - Internal fields and recompute logic in the temps do not reflect production (no caching, prereqs stored eagerly to avoid redundant VK keygen without a cache). 
  - Since real-provider construction differs, the `ivc_setup.rs` tests will be redesigned alongside the swap.
## Issue(s)

Relates to #3138


mithril-stm/src/proof_system/ivc_halo2_snark/setup.rs
```diff
@@ -40,3 +50,153 @@ pub(crate) struct IvcSetup {
...
+    /// and the IVC proving key from the supplied providers, extracts the three fixed-base
+    /// maps from the verifying keys, and assembles them into an `IvcSetup`.
+    ///
+    /// Providers are currently the temporary pure-compute ones in
+    /// `unsafe_setup_helpers`; they share the API surface the production cache providers
+    /// will expose. When the cache work lands, the temp provider types in this signature
+    /// are replaced with the real cache providers; the body stays unchanged.
+    // TODO: swap `Temp*Provider` parameters for the production IVC cache providers
+    // once they ship.
+    pub(crate) fn load(
```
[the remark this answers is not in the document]
curiecrypt: will be removed when providers ready

mithril-stm/src/proof_system/ivc_halo2_snark/unsafe_setup_helpers.rs
```diff
@@ -0,0 +1,129 @@
+//! Temporary unsafe setup helpers used by `IvcSetup::load` until the production
+//! cache providers for the recursive circuit are implemented.
+//!
+//! Each provider stores its prerequisites as owned fields and computes its artifact
+//! on demand. The caller orchestrates the chain: first the cert VK provider, then
+//! the IVC VK provider (constructed with the computed cert VK), then the IVC PK
+//! provider (constructed with both VKs).
+
+// TODO: remove this module once the production IVC cache providers ship.
+
```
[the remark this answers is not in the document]
curiecrypt: will be removed when providers ready

mithril-stm/src/proof_system/ivc_halo2_snark/unsafe_setup_helpers.rs
```diff
@@ -0,0 +1,129 @@
...
+        halo2_ivc::{K, circuit::IvcCircuit},
+    },
+    proof_system::ivc_halo2_snark::{CircuitProvingKey, CircuitVerifyingKey},
+};
+
+/// Recomputes the certificate-circuit verifying key from a shared SRS plus the
+/// `Parameters` and Merkle tree depth that pin the circuit shape.
+#[allow(dead_code)]
+pub(crate) struct TempCertificateVerifyingKeyProvider {
+    /// Shared KZG structured reference string, sized for the IVC circuit. Cloned
```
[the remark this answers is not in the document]
curiecrypt: will be removed when providers ready

mithril-stm/src/proof_system/ivc_halo2_snark/setup.rs
```diff
@@ -40,3 +50,153 @@ pub(crate) struct IvcSetup {
     /// into the new IVC folded accumulator.
     pub(crate) combined_fixed_bases: BTreeMap<String, G1Projective>,
 }
+
+#[allow(dead_code)]
+impl IvcSetup {
```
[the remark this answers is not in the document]
curiecrypt: False positive

mithril-stm/src/proof_system/ivc_halo2_snark/mod.rs
```diff
@@ -1,5 +1,22 @@
...
 mod rolling_state;
 mod setup;
+mod unsafe_setup_helpers;
+
+/// PLONK verifying key used across the IVC setup
+/// (certificate and recursive circuits share the same field/curve parameterization).
+#[allow(dead_code)]
+pub(crate) type CircuitVerifyingKey = VerifyingKey<CircuitBase, KZGCommitmentScheme<Bls12>>;
+/// PLONK proving key used by the IVC circuit.
+#[allow(dead_code)]
```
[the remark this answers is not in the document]
curiecrypt: fixed

mithril-stm/src/proof_system/ivc_halo2_snark/rolling_state.rs
```diff
@@ -1,19 +1,335 @@
...
     /// Folded accumulator the new step will build on top of
-    pub(crate) accumulator: Accumulator<BlstrsEmulation>,
+    accumulator: Accumulator<BlstrsEmulation>,
     /// Chain-specific Schnorr signature over the genesis state
-    pub(crate) genesis_signature: StandardSchnorrSignature,
+    genesis_signature: StandardSchnorrSignature,
+}
+
+#[allow(dead_code)]
+impl IvcRollingState {
```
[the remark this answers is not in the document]
curiecrypt: False positive

mithril-stm/src/proof_system/ivc_halo2_snark/setup.rs
```diff
@@ -40,3 +50,153 @@ pub(crate) struct IvcSetup {
...
+        trusted_setup_provider: &TrustedSetupProvider,
+        certificate_verifying_key_provider: &TempCertificateVerifyingKeyProvider,
+        ivc_verifying_key_provider: &TempIvcVerifyingKeyProvider,
+        ivc_proving_key_provider: &TempIvcProvingKeyProvider,
+    ) -> StmResult<Self> {
+        let srs = trusted_setup_provider.get_trusted_setup_parameters()?;
+
+        let certificate_verifying_key = certificate_verifying_key_provider.get_verifying_key()?;
+        let ivc_verifying_key = ivc_verifying_key_provider.get_verifying_key()?;
+        let ivc_proving_key = ivc_proving_key_provider.get_proving_key()?;
```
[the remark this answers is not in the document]
curiecrypt: a doc comment explaining the srs usage and temp providers added

mithril-stm/src/proof_system/ivc_halo2_snark/rolling_state.rs
```diff
@@ -1,19 +1,335 @@
...
+        } else {
+            let rolling_state = Self::genesis(genesis_signature, fixed_base_names);
+            rolling_state.save(path)?;
+            Ok(rolling_state)
+        }
+    }
+
+    /// Persists the rolling state to `path`, overwriting any existing file.
+    pub(crate) fn save(&self, path: &Path) -> StmResult<()> {
+        let file = File::create(path)
```
hjeljeli32: `File::create` truncates the existing file before writing. If the process crashes mid-write, the rolling state is irrecoverably corrupted and the prover cannot resume. IMO we can use the `NamedTempFile::new_in(parent) + persist()` pattern

mithril-stm/src/proof_system/ivc_halo2_snark/rolling_state.rs
```diff
@@ -1,19 +1,335 @@
...
+    pub(crate) fn accumulator(&self) -> &Accumulator<BlstrsEmulation> {
+        &self.accumulator
+    }
+
+    /// Returns the chain-specific Schnorr signature over the genesis state.
+    pub(crate) fn genesis_signature(&self) -> StandardSchnorrSignature {
+        self.genesis_signature
+    }
+
+    /// Loads the rolling state from `path` if it exists, otherwise bootstraps a
```
hjeljeli32: IMO, in the comment we should precise:
/// The `fixed_base_names` argument must be the keys of [`IvcSetup::combined_fixed_bases`].
/// Passing any other list will produce a genesis accumulator that does not match
/// what the circuit expects, and the first proving step will fail.

hjeljeli32 (approved):
LGTM 👍 I particularly checked the consistency with IVC pipeline. 
Just left a couple of non-blocking comments.  
🚀

mithril-stm/src/proof_system/ivc_halo2_snark/rolling_state.rs
```diff
@@ -1,19 +1,320 @@
...
+    /// genesis rolling state from the caller-supplied genesis signature and the
+    /// fixed-base names of the combined accumulator map.
+    pub(crate) fn load_or_genesis(
```
jpraynaud: I'm not convinced that we need to store/load the `IvcRollingState` which should instead be created on the fly with inputs given to the prover (previous `IvcProof` and genesis certificate).

mithril-stm/src/proof_system/ivc_halo2_snark/setup.rs
```diff
@@ -40,3 +50,149 @@ pub(crate) struct IvcSetup {
...
+    /// `unsafe_setup_helpers`; they share the API surface the production cache providers
+    /// will expose. When the cache work lands, the temp provider types in this signature
+    /// are replaced with the real cache providers; the body stays unchanged.
+    // TODO: swap `Temp*Provider` parameters for the production IVC cache providers
+    // once they ship.
+    pub(crate) fn load(
+        trusted_setup_provider: &TrustedSetupProvider,
+        ivc_verifying_key_provider: &TempIvcVerifyingKeyProvider,
+        certificate_verifying_key_provider: &TempCertificateVerifyingKeyProvider,
+        ivc_proving_key_provider: &TempIvcProvingKeyProvider,
```
jpraynaud: You should probably use a single `TmpCircuitKeyProvider` type which gives access to both proving and verification keys, and use 2 instances of it (one for non recursive circuit and another one for the recursive cicuit).

mithril-stm/src/proof_system/ivc_halo2_snark/setup.rs
```diff
@@ -40,3 +50,149 @@ pub(crate) struct IvcSetup {
...
+    use midnight_proofs::utils::SerdeFormat;
+    use rand_chacha::ChaCha20Rng;
+    use rand_core::SeedableRng;
+    use sha2::{Digest, Sha256};
+    use tempfile::tempdir;
+
+    use super::*;
+    use crate::{Parameters, circuits::halo2_ivc::K};
+
+    fn build_provider_with_unsafe_srs(base_dir: &Path) -> TrustedSetupProvider {
```
jpraynaud: It'd be a good idea to create a builder for tests only in the `mithril-stm/src/circuits/trusted_setup.rs` module, WDYT?

mithril-stm/src/proof_system/ivc_halo2_snark/setup.rs
```diff
@@ -40,3 +50,149 @@ pub(crate) struct IvcSetup {
...
+            .write_all(&srs_bytes)
+            .unwrap();
+        let expected_hash = hex::encode(Sha256::digest(&srs_bytes));
+
+        TrustedSetupProvider::new(base_dir, expected_hash, "", Duration::from_secs(600))
+    }
+
+    #[test]
+    #[ignore = "slow: generates K=19 SRS and runs keygen_pk for the IVC circuit"]
+    fn load_succeeds_with_unsafe_srs() {
```
jpraynaud: Do we want to put this test in the `slow` tier?

mithril-stm/src/proof_system/ivc_halo2_snark/rolling_state.rs
```diff
@@ -19,6 +19,18 @@ use crate::{
...
+/// other breaking shift in the byte stream produced by `IvcWrite::write`.
+///
+/// Not a `codec::CODEC_VERSION_*` codec dispatcher — that scheme is for routing
+/// public-facing types through `codec::from_versioned_bytes`. The rolling state
+/// file is IVC-internal and never goes through that path. The initial value is
+/// chosen to be clearly distinct from the lib's CBOR codec marker (`0x01`) so
+/// that a misrouted file fails on the version check rather than producing a
+/// confusing CBOR decode error.
+const ROLLING_STATE_VERSION: u8 = 0x10;
+
```
jpraynaud: This is a questionable choice. In any case, this is not needed as we don't need to serialize the `IvcRollingState`.

mithril-stm/src/proof_system/ivc_halo2_snark/epoch_data.rs
```diff
@@ -1,19 +1,124 @@
...
+    next_protocol_parameters: BaseFieldElement,
+}
+
+#[allow(dead_code)]
+impl EpochData {
+    /// Builds an `EpochData` by decoding the three announced fields from the
+    /// protocol-message preimage. The decoding mirrors the circuit's
+    /// `combine_bytes` step: bytes are interpreted little-endian against
+    /// powers-of-256, reducing modulo the base field for the 32-byte ranges.
+    pub(crate) fn new(message_preimage: [u8; PREIMAGE_SIZE]) -> StmResult<Self> {
```
jpraynaud, line 35: Maybe this can be replaced with the functions implemented by @hjeljeli32  in PR #3288: 
https://github.com/input-output-hk/mithril/blob/720bbfc6b2adb214a7d30dd1f2ae2b84df54ea36/mithril-stm/src/circuits/halo2_ivc/protocol_message.rs#L140-L170

mithril-stm/src/proof_system/ivc_halo2_snark/rolling_state.rs
curiecrypt: As @jpraynaud suggested, I removed the file operations

mithril-stm/src/proof_system/ivc_halo2_snark/rolling_state.rs
curiecrypt: fixed

mithril-stm/src/proof_system/ivc_halo2_snark/rolling_state.rs
curiecrypt: removed

mithril-stm/src/proof_system/ivc_halo2_snark/setup.rs
curiecrypt: done

mithril-stm/src/proof_system/ivc_halo2_snark/setup.rs
curiecrypt: done

mithril-stm/src/proof_system/ivc_halo2_snark/rolling_state.rs
curiecrypt: done

mithril-stm/src/proof_system/ivc_halo2_snark/setup.rs
curiecrypt: tests are working with unsafe helpers, this is why they are slow. when we have the providers, this test needs to be modified accordingly, and it wont be in the slow tier.

mithril-stm/src/proof_system/ivc_halo2_snark/rolling_state.rs
```diff
@@ -2,18 +2,190 @@
...
-    pub(crate) genesis_signature: StandardSchnorrSignature,
+    genesis_signature: StandardSchnorrSignature,
+}
+
+#[allow(dead_code)]
+impl IvcRollingState {
+    /// Builds a rolling state from the four fields produced by an IVC proving step.
+    pub(crate) fn new(
+        state: State,
+        ivc_proof: Vec<u8>,
```
jpraynaud, line 32: Maybe we need an `Option<Vec<u8>>` instead? This will clearly express the fact that the proof does not exist? Maybe we need to do the same for the other fields?

mithril-stm/src/proof_system/ivc_halo2_snark/unsafe_setup_helpers.rs
```diff
@@ -0,0 +1,123 @@
...
+/// shape.
+#[allow(dead_code)]
+pub(crate) struct TempCertificateKeyProvider {
+    /// Shared KZG structured reference string, sized for the IVC circuit. Cloned
+    /// on demand and downsized in place before certificate-circuit keygen.
+    srs: Arc<ParamsKZG<Bls12>>,
+    /// STM protocol parameters that determine the certificate circuit shape.
+    parameters: Parameters,
+    /// Merkle tree depth used by the certificate circuit.
+    merkle_tree_depth: u32,
```
jpraynaud, line 44: I'm not sure that we need to keep the Merkle tree depth in the key provider? Ideally we'll have only once circuit key provider that will support the recursive/non-recursive circuits. But you can keep it as is as it is a bit out of scope for this PR.

jpraynaud (approved):
LGTM 👍

mithril-stm/src/proof_system/ivc_halo2_snark/rolling_state.rs
curiecrypt: Actually, when I have a detailed look, I realized that it would not provides us a benefit. 
At genesis these fields aren't missing, they are documented defaults (`State::genesis()`, `trivial_acc(...)`, `Vec::new()`, matching the reference IVC loop). 
The in-circuit gadget consumes them as-is and handles genesis via `scale_by_bit(is_not_genesis, ...)`; off-circuit, `state.counter == F::ZERO` signals genesis. `Option<>` wouldn't enable a check we don't already have. WDYT?

mithril-stm/src/proof_system/ivc_halo2_snark/unsafe_setup_helpers.rs
curiecrypt: I'd skip it since we remove them soon.

mithril-stm/src/proof_system/ivc_halo2_snark/epoch_data.rs
curiecrypt: todo comment added

PR #3304, opened 2026-06-02
feat: SNARK-friendly genesis certificate primitives

jpraynaud:
## Content

This PR introduces **the additive, backward-compatible crypto foundation for a SNARK-friendly genesis certificate**, with no change yet to the certificate format or node behavior**:

- Add SNARK-friendly genesis primitives based on Schnorr over Jubjub (`StandardSchnorrSignature`, signing/verification keys), with `SchnorrSigningKey` debug output redacted.
- Promote the `genesis` crypto_helper to a sub-module folder and introduce ed25519-genesis aliases to make room for the dual-algorithm types.
- Add genesis key bundles pairing the legacy Ed25519 key with an optional Schnorr key (`GenesisSigningKeyBundle`, `GenesisVerificationKeyBundle`), with hex auto-detection between a dual bundle and a legacy single-key input.
- Wrap the two algorithms behind `GenesisSigner` and `GenesisVerifier` (Ed25519 plus optional Schnorr), exposing key loading, era-support checks, and verification-key derivation.

## Pre-submit checklist

- Branch
  - [x] Tests are provided (if possible)
  - [x] Crates versions are updated (if relevant)
  - [x] CHANGELOG file is updated (if relevant)
  - [x] Commit sequence broadly makes sense
  - [x] Key commits have useful messages
- PR
  - [x] All check jobs of the CI have succeeded
  - [x] Self-reviewed the diff
  - [x] Useful pull request description
  - [x] Reviewer requested

## Issue(s)
Relates to #3145

PR #3306, opened 2026-06-03
feat: dual-signature genesis certificate and wiring

### LINKED ISSUE #3145: Implement SNARK-friendly genesis certificate
To be activated with Lagrange era

- [ ] Change signature from Ed25519 to Jubjub
- [ ] Add the signature of the circuit verification key in the Genesis certificate?

jpraynaud:
## Content

This PR includes **the dual-signature genesis certificate for the Lagrange era and wires it through the nodes, while staying fully backward compatible with the legacy Ed25519 genesis under Pythagoras**:

- Introduce a `GenesisDualSignature` certificate variant carrying both the legacy Ed25519 signature and the Schnorr signature, and add era-aware signing on `GenesisSigner` to produce it. The certificate hash feeds only the Ed25519 bytes, so a Lagrange genesis certificate hashes byte-identically to a legacy one and existing clients keep walking the
chain across the era boundary.
- Make the genesis certificate producer and verifier era-aware (Pythagoras accepts an Ed25519-only key, Lagrange requires the Schnorr half), and accept the new `GenesisVerifier` in the certificate verifier.
- Extend the certificate message wire format with optional `aggregate_verification_key_snark` and `genesis_schnorr_signature` fields (omitted or empty outside Lagrange genesis, so the format stays a strict superset of the previous one), and document them in `openapi.yaml`.
- Adapt the aggregator genesis subcommands (`generate-keypair`, `export`, `sign`, `import`, `bootstrap`) to the dual signature, and add an `upgrade-key-to-dual` subcommand to migrate an existing Ed25519 key to a dual bundle.
- Adapt the client certificate verifier to the new `GenesisVerifier`.
- Update the genesis runbook for the dual ceremony.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Comments

This PR is the second part following https://github.com/input-output-hk/mithril/pull/3304

## Issue(s)
Closes #3145

### DIFF
--- a/mithril-aggregator/src/commands/genesis_command.rs
+++ b/mithril-aggregator/src/commands/genesis_command.rs
@@ -7,9 +7,13 @@ use serde::Deserialize;
 use slog::{Logger, debug};
 
 use mithril_cardano_node_chain::chain_observer::ChainObserverType;
+#[cfg(not(feature = "future_snark"))]
+use mithril_common::crypto_helper::GenesisEd25519VerificationKey;
+#[cfg(feature = "future_snark")]
+use mithril_common::crypto_helper::GenesisVerifier;
 use mithril_common::{
     StdResult,
-    crypto_helper::{GenesisEd25519VerificationKey, GenesisSigner},
+    crypto_helper::GenesisSigner,
     entities::{HexEncodedGenesisSecretKey, HexEncodedGenesisVerificationKey, SupportedEra},
 };
 use mithril_doc::{Documenter, StructDoc};
@@ -239,6 +243,12 @@ pub struct ImportGenesisSubCommand {
     /// Genesis Verification Key
     #[clap(long)]
     genesis_verification_key: HexEncodedGenesisVerificationKey,
+
+    /// Mithril era to use for the genesis certificate
+    ///
+    /// Optional when only one era exists, required when multiple eras are supported.
+    #[clap(long)]
+    mithril_era: Option<SupportedEra>,
 }
 
 impl ImportGenesisSubCommand {
@@ -257,7 +267,7 @@ impl ImportGenesisSubCommand {
             "Genesis import signed payload from {}",
             self.signed_payload_path.to_string_lossy()
         );
-        let mithril_era = resolve_mithril_era(None)?;
+        let mithril_era = resolve_mithril_era(self.mithril_era)?;
         let mut dependencies_builder =
             DependenciesBuilder::new(root_logger.clone(), Arc::new(config.clone()));
         let dependencies = dependencies_builder
@@ -270,14 +280,45 @@ impl ImportGenesisSubCommand {
         let genesis_tools = GenesisTools::from_dependencies(dependencies)
             .await
             .with_context(|| "genesis-tools: initialization error")?;
+
+        self.run_import(&genesis_tools, mithril_era).await
+    }
+
+    #[cfg(not(feature = "future_snark"))]
+    async fn run_import(
+        &self,
+        genesis_tools: &GenesisTools,
+        _mithril_era: SupportedEra,
+    ) -> StdResult<()> {
+        let verification_key =
+            GenesisEd25519VerificationKey::from_json_hex(&self.genesis_verification_key)?;
         genesis_tools
-            .import_payload_signature(
-                &self.signed_payload_path,
-                &GenesisEd25519VerificationKey::from_json_hex(&self.genesis_verification_key)?,
-            )
+            .import_payload_signature(&self.signed_payload_path, &verification_key)
             .await
-            .with_context(|| "genesis-tools: import error")?;
-        Ok(())
+            .with_context(|| "genesis-tools: import error")
+    }
+
+    #[cfg(feature = "future_snark")]
+    async fn run_import(
+        &self,
+        genesis_tools: &GenesisTools,
+        mithril_era: SupportedEra,
+    ) -> StdResult<()> {
+        let verifier = GenesisVerifier::try_from_hex(&self.genesis_verification_key)?;
+        verifier.ensure_supports_era(mithril_era)?;
+        match mithril_era {
+            SupportedEra::Lagrange => genesis_tools
+                .import_dual_payload_signature(&self.signed_payload_path, &verifier)
+                .await
+                .with_context(|| "genesis-tools: import error"),
+            _ => genesis_tools
+                .import_payload_signature(
+                    &self.signed_payload_path,
+                    &verifier.ed25519.to_verification_key(),
+                )
+                .await
+                .with_context(|| "genesis-tools: import error"),
+        }
     }
 
     pub fn extract_config(command_path: String) -> HashMap<String, StructDoc> {
@@ -438,6 +479,21 @@ mod tests {
                 "Should error when multiple eras exist and none is provided"
             );
         }
+    }
+
+    #[test]
+    fn import_subcommand_parses_era_flag() {
+        let cmd = ImportGenesisSubCommand::try_parse_from([
+            "import",
+            "--signed-payload-path",
+            "/tmp/signed",
+            "--genesis-verification-key",
+            "deadbeef",
+            "--mithril-era",
+            &SupportedEra::Lagrange.to_string(),
+        ])
+        .expect("CLI parse should succeed");
+        assert_eq!(cmd.mithril_era, Some(SupportedEra::Lagrange));
     }
 
     #[test]
--- a/mithril-aggregator/src/commands/genesis_command.rs
+++ b/mithril-aggregator/src/commands/genesis_command.rs
@@ -395,7 +395,7 @@ impl GenerateKeypairGenesisSubCommand {
             self.target_path.to_string_lossy()
         );
 
-        GenesisTools::create_and_save_genesis_keypair(&self.target_path)
+        GenesisTools::create_and_save_genesis_keypair(&self.target_path, SupportedEra::Pythagoras)
             .with_context(|| "genesis-tools: keypair generation error")?;
 
         Ok(())
--- a/mithril-aggregator/src/commands/genesis_command.rs
+++ b/mithril-aggregator/src/commands/genesis_command.rs
@@ -298,6 +298,12 @@ pub struct SignGenesisSubCommand {
     /// Genesis Secret Key Path
     #[clap(long)]
     genesis_secret_key_path: PathBuf,
+
+    /// Mithril era to use for the genesis certificate
+    ///
+    /// Optional when only one era exists, required when multiple eras are supported.
+    #[clap(long)]
+    mithril_era: Option<SupportedEra>,
 }
 
 impl SignGenesisSubCommand {
@@ -308,11 +314,13 @@ impl SignGenesisSubCommand {
             self.to_sign_payload_path.to_string_lossy(),
             self.target_signed_payload_path.to_string_lossy()
         );
+        let mithril_era = resolve_mithril_era(self.mithril_era)?;
 
         GenesisTools::sign_genesis_certificate(
             &self.to_sign_payload_path,
             &self.target_signed_payload_path,
             &self.genesis_secret_key_path,
+            mithril_era,
         )
         .await
         .with_context(|| "genesis-tools: sign error")?;
@@ -432,6 +440,23 @@ mod tests {
                 "Should error when multiple eras exist and none is provided"
             );
         }
+    }
+
+    #[test]
+    fn sign_subcommand_parses_era_flag() {
+        let cmd = SignGenesisSubCommand::try_parse_from([
+            "sign",
+            "--to-sign-payload-path",
+            "/tmp/to-sign",
+            "--target-signed-payload-path",
+            "/tmp/signed",
+            "--genesis-secret-key-path",
+            "/tmp/genesis.sk",
+            "--mithril-era",
+            &SupportedEra::Lagrange.to_string(),
+        ])
+        .expect("CLI parse should succeed");
+        assert_eq!(cmd.mithril_era, Some(SupportedEra::Lagrange));
     }
 
     #[tokio::test]
--- a/mithril-aggregator/src/commands/genesis_command.rs
+++ b/mithril-aggregator/src/commands/genesis_command.rs
@@ -592,18 +592,27 @@ mod tests {
         assert_eq!(cmd.mithril_era, Some(SupportedEra::Lagrange));
     }
 
-    #[tokio::test]
-    async fn generate_keypair_subcommand_parses_era_flag() {
-        let target = temp_dir!().join("keys");
-        std::fs::create_dir_all(&target).unwrap();
+    #[test]
+    fn generate_keypair_subcommand_parses_era_flag() {
         let cmd = GenerateKeypairGenesisSubCommand::try_parse_from([
             "generate-keypair",
             "--target-path",
-            target.to_str().unwrap(),
+            "/tmp/keys",
             "--mithril-era",
             &SupportedEra::Pythagoras.to_string(),
         ])
-        .unwrap();
+        .expect("CLI parse should succeed");
+        assert_eq!(cmd.mithril_era, Some(SupportedEra::Pythagoras));
+    }
+
+    #[tokio::test]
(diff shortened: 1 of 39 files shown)


mithril-common/src/entities/certificate.rs
```diff
@@ -118,30 +147,48 @@ impl Certificate {
...
+        Ok(hex::encode(hasher.finalize()))
+    }
+
+    /// Infallible [Self::try_compute_hash] for construction-time hashing, where the inputs were
+    /// just built and validated. Panics only on a genuinely impossible serialization bug; the
+    /// verification path uses [Self::try_compute_hash] so it cannot crash the process.
+    pub fn compute_hash(&self) -> String {
+        self.try_compute_hash()
+            .expect("certificate hashing is infallible for a well-formed certificate")
     }
```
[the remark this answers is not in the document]
jpraynaud: Fixed

mithril-common/src/certificate_chain/certificate_genesis.rs
```diff
@@ -65,7 +55,13 @@ impl CertificateGenesisProducer {
...
         let genesis_concatenation_avk =
             genesis_aggregate_verification_key_for_concatenation.to_json_hex()?;
-        let mut protocol_message = ProtocolMessage::new();
+        let mut protocol_message = match mithril_era {
+            SupportedEra::Pythagoras => ProtocolMessage::new(),
+            #[cfg(feature = "future_snark")]
+            SupportedEra::Lagrange => ProtocolMessage::new_rigid(),
+            #[cfg(not(feature = "future_snark"))]
+            SupportedEra::Lagrange => ProtocolMessage::new(),
+        };
```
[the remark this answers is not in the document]
jpraynaud: Fixed

mithril-common/src/crypto_helper/genesis/signer.rs
```diff
@@ -111,6 +114,66 @@ impl GenesisSigner {
...
+        let ed25519_signature = self.ed25519.sign(protocol_message.to_message().as_bytes());
+        match mithril_era {
+            SupportedEra::Pythagoras => {
+                Ok(CertificateSignature::GenesisSignature(ed25519_signature))
+            }
+            #[cfg(feature = "future_snark")]
+            SupportedEra::Lagrange => {
+                let schnorr_signer =
+                    self.schnorr.as_ref().ok_or(GenesisBundleError::LegacySigningKey)?;
+                let preimage = protocol_message.rigid_preimage();
```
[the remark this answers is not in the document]
jpraynaud: Fixed

mithril-common/src/entities/certificate.rs
```diff
@@ -114,34 +143,52 @@ impl Certificate {
...
+        Ok(hex::encode(hasher.finalize()))
     }
 
-    /// Tell if the certificate is a genesis certificate
+    /// Infallible [Self::try_compute_hash] for construction-time hashing, where the inputs were
+    /// just built and validated. Panics only on a genuinely impossible serialization bug; the
+    /// verification path uses [Self::try_compute_hash] so it cannot crash the process.
+    pub fn compute_hash(&self) -> String {
+        self.try_compute_hash()
+            .expect("certificate hashing is infallible for a well-formed certificate")
```
Alenar: I think this method should either: be removed or transformed into a test only extension trait. Doing so would ensure that it's not used in business code (which we should not need as `new` was transformed into `try_new`).

mithril-aggregator/src/commands/genesis_command.rs
```diff
@@ -406,11 +482,52 @@ impl GenerateKeypairGenesisSubCommand {
...
+}
+
+#[cfg(feature = "future_snark")]
+impl UpgradeKeyToDualGenesisSubCommand {
+    pub async fn execute(&self, root_logger: Logger) -> StdResult<()> {
+        debug!(root_logger, "UPGRADE-KEY-TO-DUAL GENESIS command");
+        println!(
+            "Upgrade legacy genesis keypair from {} to dual bundle at {}",
+            self.legacy_secret_key_path.to_string_lossy(),
+            self.target_path.to_string_lossy()
```
Alenar: I think we made a mistake in this files, all prints of path should use `display` instead of `to_string_lossy`.
```suggestion
            self.legacy_secret_key_path.display(),
            self.target_path.display()
```

mithril-aggregator/src/commands/genesis_command.rs
```diff
@@ -434,6 +551,97 @@ mod tests {
...
+            target.to_str().unwrap(),
+        ])
+        .unwrap();
+
+        cmd.execute(TestLogger::stdout())
+            .await
+            .expect("upgrade should succeed");
+        assert!(target.join("genesis.sk").exists());
+        assert!(target.join("genesis.vk").exists());
+    }
```
Alenar: Tests of command parameters should be separated from commands execution, btw isn't it the intended intent here ? With the tests names alone `generate_keypair_subcommand_parses_era_flag` should not execute and `upgrade_key_to_dual_subcommand_writes_bundle_files` should construct its parameters using directly the `GenesisCommandConfiguration` struct.

mithril-aggregator/src/database/record/certificate.rs
```diff
@@ -19,6 +21,70 @@ use mithril_persistence::{
...
+    /// Recovers the SNARK half from the JSON wrapper when present; otherwise treats the value as
+    /// a legacy Ed25519-only genesis signature or a multi-signature, depending on the record.
+    fn split_message_signatures(record: &CertificateRecord) -> PersistedRecordSignatures {
+        if record.parent_certificate_id.is_some() {
+            return PersistedRecordSignatures::MultiSignature(record.signature.clone());
+        }
+        if let Ok(payload) = serde_json::from_str::<Self>(&record.signature) {
+            return PersistedRecordSignatures::GenesisDual(payload);
+        }
+        PersistedRecordSignatures::Genesis(record.signature.clone())
```
Alenar: _nitpick_ I think using a simple if/else is more readable, as it avoid the `return` keyword and put every case on the same level.
```suggestion
        if record.parent_certificate_id.is_some() {
            PersistedRecordSignatures::MultiSignature(record.signature.clone());
        }
        else if let Ok(payload) = serde_json::from_str::<Self>(&record.signature) {
            PersistedRecordSignatures::GenesisDual(payload);
        } else {
            PersistedRecordSignatures::Genesis(record.signature.clone())
        }
```

mithril-aggregator/src/database/record/certificate.rs
```diff
@@ -19,6 +21,70 @@ use mithril_persistence::{
...
+            return Ok(CertificateSignature::GenesisDualSignature(ed25519, schnorr));
+        }
+        Ok(CertificateSignature::GenesisSignature(raw.try_into()?))
+    }
+
+    /// Split a stored certificate record into its wire-format signature.
+    ///
+    /// Recovers the SNARK half from the JSON wrapper when present; otherwise treats the value as
+    /// a legacy Ed25519-only genesis signature or a multi-signature, depending on the record.
+    fn split_message_signatures(record: &CertificateRecord) -> PersistedRecordSignatures {
```
Alenar: _nitpick-performance_ If you change the signature to the following to let the caller move the signature:

```suggestion
    fn split_message_signatures(
        have_parent: bool,
        signature: HexEncodedKey,
    ) -> PersistedRecordSignatures {
```
then you can avoid all `clone` call in this method body.

Since this trade have value since:
- it's is small
- this method will be called for a lot of certificates when verifying a chain
- this method is private (no external impact)

mithril-aggregator/src/database/record/certificate.rs
```diff
@@ -225,6 +304,25 @@ impl TryFrom<CertificateRecord> for Certificate {
...
+                }
+                PersistedRecordSignatures::GenesisDual(payload) => {
+                    (String::new(), payload.ed25519, payload.schnorr)
+                }
+            };
+        #[cfg(not(feature = "future_snark"))]
+        let (multi_signature, genesis_signature) = if value.parent_certificate_id.is_none() {
+            (String::new(), value.signature.clone())
+        } else {
+            (value.signature.clone(), String::new())
```
Alenar: You should partial move like before
```suggestion
        let (multi_signature, genesis_signature) = if value.parent_certificate_id.is_none() {
            (String::new(), value.signature)
        } else {
            (value.signature, String::new())
```

mithril-aggregator/src/database/record/certificate.rs
```diff
@@ -19,6 +21,70 @@ use mithril_persistence::{
...
+    GenesisDual(PersistedDualGenesisSignature),
+}
+
+#[cfg(feature = "future_snark")]
+impl PersistedDualGenesisSignature {
+    /// Parse a persisted genesis signature column value.
+    ///
+    /// Attempts to decode the JSON wrapper (Lagrange dual signature) and falls back to the raw
+    /// Ed25519 hex string used by Pythagoras-era records.
+    fn parse_genesis_signature(raw: &str) -> Result<CertificateSignature, StdError> {
```
Alenar, line 65: **_⚠️ this is me thinking out loud, feel free to resolve this comment_**

I wonder if we should avoid the double parsing.

To do so there's multiple ways, but all that I kind thinks of requires a database change:

The `PersistedRecordSignatures` enum could be refactored has a enum without values named `SignatureKind` or `SignatureType`, instead a new `signature_kind` field and database column would be added to the `CertificateMessage`/table.

This mean adding a migration which would populate the new column based on if the certificate have a parent certificate or not (yes `MultiSignature`, false `Genesis`).

A variant would be a nullable `genesis_signature_kind` column (so no `MultiSignature`), in the migration it would be better to populate it with `Genesis` for certificates without parent but not mandatory (when hydrating a null + no parent certificate could fallback to `Genesis`).

advantages:
- allow direct parsing to the right target type
- (consequence) no need to parse twice with a fallback
- no changes to `Genesis` and `MultiSignature` fields in existing db

disadvantages:
- new column, so it requires a migration (easy enough)
- all certificates in db are impacted, this column which will be null for 99% of records (very bad)

Given that this double parsing is only for the genesis, only when reading it to the full `Certificate` type (so it will be read almost only when the aggregator validate its own chain), I think it's fine and that solving it would do more harm than good.

mithril-aggregator/src/services/certificate_chain_synchronizer/synchronizer_service.rs
```diff
@@ -129,7 +129,10 @@ impl MithrilCertificateChainSynchronizer {
         loop {
             let parent_certificate = self
                 .certificate_verifier
-                .verify_certificate(&certificate, &self.genesis_verifier.to_verification_key())
+                .verify_certificate(
+                    &certificate,
+                    &self.genesis_verifier.to_ed25519_verification_key(),
+                )
```
Alenar, line 135: This is really confusing:
Isn't this incompatible with snark ? Will a follower aggregator be able to synchronize against a snark aggregator ?

mithril-common/src/certificate_chain/certificate_genesis.rs
```diff
@@ -99,77 +95,95 @@ impl CertificateGenesisProducer {
...
         genesis_avk: ProtocolAggregateVerificationKey,
-        genesis_signature: GenesisEd25519Signature,
+        signature: CertificateSignature,
         mithril_era: SupportedEra,
     ) -> StdResult<Certificate> {
-        let protocol_version = PROTOCOL_VERSION.to_string();
-        let initiated_at = Utc::now();
-        let sealed_at = Utc::now();
-        let signers = vec![];
+        if matches!(signature, CertificateSignature::MultiSignature(..)) {
```
Alenar: Is it possible to avoid this guard ?
Maybe using three function like this:
```rust
    pub fn create_legacy_genesis_certificate<T: Into<String>>(
        &self,
        protocol_parameters: ProtocolParameters,
        network: T,
        epoch: Epoch,
        genesis_avk: ProtocolAggregateVerificationKey,
        genesis_signature: GenesisEd25519Signature,
        mithril_era: SupportedEra,
    ) -> StdResult<Certificate> {
    self.create_genesis_certificate_internal(protocol_parameters, network, epoch, genesis_avk, genesis_signature, None, mithril_era)
}

    pub fn create_genesis_certificate<T: Into<String>>(
        &self,
        protocol_parameters: ProtocolParameters,
        network: T,
        epoch: Epoch,
        genesis_avk: ProtocolAggregateVerificationKey,
        genesis_signature: GenesisEd25519Signature,
        genesis_signature_snark: GenesisSnarkSignature,
        signature: CertificateSignature,
        mithril_era: SupportedEra,
    ) -> StdResult<Certificate> {
    self.create_genesis_certificate_internal(protocol_parameters, network, epoch, genesis_avk, genesis_signature, Some(genesis_signature_snark), mithril_era)
}

    // private
    fn create_genesis_certificate_internal<T: Into<String>>(
        &self,
        protocol_parameters: ProtocolParameters,
        network: T,
        epoch: Epoch,
        genesis_avk: ProtocolAggregateVerificationKey,
        genesis_signature: GenesisEd25519Signature,
        genesis_signature_snark: Option<GenesisSnarkSignature>,
        signature: CertificateSignature,
        mithril_era: SupportedEra,
    ) -> StdResult<Certificate> { ... }
```

Alenar (commented):
LGTM, with a caveat:

I feel that the gate behind `future_snark` is in contradiction with the mithril era mechanism.
When using a build without this feature all Lagrande era path are disabled / fallback to the previous era, this defeat the purpose of the mithril era which is to release a code that will be triggered at the same time across all nodes.

Would it be possible to relax the usage of the `future_snark`, keeping it in existing function that are impacted by it but removing it from the primitives and tooling around it ?

Maybe we should document some of those types has "unstable" as a trade-off, or maybe this is too soon to remove this feature and we only want the real support to comes in a future distribution.

This is an open question that does not need to be addressed in this PR.

mithril-common/src/entities/certificate.rs
jpraynaud: Agreed, I removed it.

mithril-aggregator/src/commands/genesis_command.rs
jpraynaud: Fixed in genesis, era and database commands :+1:

mithril-aggregator/src/database/record/certificate.rs
jpraynaud: Fixed

mithril-aggregator/src/database/record/certificate.rs
jpraynaud: Fixed

mithril-aggregator/src/database/record/certificate.rs
jpraynaud: Fixed

mithril-aggregator/src/commands/genesis_command.rs
jpraynaud: Fixed

mithril-common/src/certificate_chain/certificate_genesis.rs
jpraynaud: Fixed

mithril-aggregator/src/services/certificate_chain_synchronizer/synchronizer_service.rs
jpraynaud: The IVC proof created, once verified, will prove the validity of the full certificate chain up to the genesis certificate. This means that the verification of the full certificate chain will be done only for the concatenation proof system. Supporting a follower aggregator is a longer term feature as the leader aggregator will remain on concatenation proof: we will probably need to have follower aggregator able to synchronize on another follower aggregator. This will be addressed later when the IVC SNARK proofs are stable :+1:

mithril-aggregator/src/database/record/certificate.rs
jpraynaud: Yes, this part is a bit tricky and there are no solution without trade-off. We could re-challenge that in the future.

Alenar (approved):

turmelclem (approved):
LGTM 👍

PR #3313, opened 2026-06-05
feat(stm):  Prepare prover input - Phase 4

curiecrypt:
## Content

Phase 4 of the IVC prover input preparation. Implements `IvcProverInput::prepare`, refactors supporting types in response to review feedback, and adds asset-driven tests with full golden-artifact comparisons.

### What this branch adds

- **`IvcProverInput::prepare`** as one orchestrating function with a top-level genesis short-circuit: at `step_counter == 0` the base case skips certificate and previous-IVC verifier-prep entirely and passes the rolling state's trivial accumulator through unchanged. Non-genesis steps verify the certificate proof, validate the epoch transition (out-of-range, first-cert-after-genesis must be next-epoch, same-epoch lookahead consistency), and fold the certificate and previous-IVC accumulators into the chain's accumulator.
- Mechanics live on the data owners: `IvcSetup::{certificate_collapsed_accumulator, trivial_previous_ivc_proof_collapsed_accumulator, previous_ivc_proof_collapsed_accumulator}` and `IvcRollingState::{verify_genesis_signature, previous_ivc_proof_public_inputs, new_step_counter}`.
- **`IvcCircuitError`** gains `InvalidEpochTransition { kind: EpochTransitionErrorKind, chain_epoch }`, `StepCounterOverflow`, and `CertificateVerifyingKeyMismatch`. `EpochTransitionErrorKind` enumerates `OutOfRange`, `SameEpochLookaheadMismatch`, and `FirstCertificateAfterGenesisMustBeNextEpoch`.
- **`ProtocolMessagePreimage`** (in `circuits/halo2_ivc/types.rs`) now exposes on-demand `current_epoch()`, `next_merkle_tree_commitment()`, and `next_protocol_parameters()` accessors returning IVC domain types directly. The earlier parallel `DecodedProtocolMessage` type is removed.
- **Asset structs** redesigned: `FirstCertificateInEpochAsset` (formerly `FirstStepCertAsset`); the previously generic `RecursiveStepOutputAsset` is split into three named types (`GenesisStepOutputAsset`, `NextEpochStepOutputAsset`, `FollowingCertificateInEpochAsset`) each paired with its own loader/writer through a shared `StepOutputAsset` trait. Fields use fixed `[u8; N]` arrays (with zero placeholders at genesis) instead of `Vec<u8>`. `avk_merkle_root` is renamed to `aggregate_verification_key_merkle_root`.
- **Generator** updates persist incidental data and add a new `generate_first_step_cert_asset` entry point under `#[ignore]`. All committed `.bin` assets regenerated to match the new shapes.
- **Seven `mod slow` tests** sharing one `IvcSetup` via `OnceLock`: genesis, same-epoch, next-epoch happy paths plus invalid SNARK proof, invalid genesis signature, invalid epoch transition, step counter overflow. Happy-path assertions compare the full `next_state` and `next_accumulator` (byte equality) against committed golden artifacts plus witness fields against the asset's certificate data. Negative tests downcast errors and `matches!` against specific variants. Tests are flagged `#[ignore]` so CI skips them by default; opt-in via `cargo test -- --ignored`.

### Deleted

- `IvcRollingState::from_previous_proof`. The constructor copied `IvcProof::proof_bytes` (Blake2b-transcript, externally verifiable) into a field that the next IVC step consumes under the Poseidon transcript, a category error. Resolved cross-PR with #3315.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

### Cross-cutting changes

- `pub(crate)` bumps on `tests::common` and asset-generator constants (`QUORUM_SIZE`, `SIGNER_COUNT`, `TOTAL_STAKE`).
- `#[cfg(test)] SnarkProof::from_parts(...)` so tests can wrap stored proof bytes without keygen, plus `pub(crate) SnarkProof::circuit_verification_key()` accessor for the off-circuit VK mismatch check.
- `#[cfg(test)]` re-export of `CircuitVerificationKey`; `#[derive(Debug, PartialEq, Eq)]` on `State` and `Witness`; `#[derive(Debug)]` on `IvcProverInput`.
- `u64_wrapper!::from_field` gated `#[cfg(test)]` (used only by test infrastructure).

### Design notes

- **No production API changed for testing.** Fixture uses the existing `unsafe_setup_helpers` providers; constants come from the asset generator so cert/IVC VKs match committed proofs.
- **Slow tier opt-in.** Slow tests are `#[ignore]`-flagged so default `cargo test` runs skip them on CI; opt-in via `--ignored`. Switching to `key_cache` later is a ~5-10 line swap per fixture.
- **No transcript role confusion.** `IvcRollingState::ivc_proof` is exclusively the Poseidon-transcript proof that the next step's in-circuit verifier consumes; `IvcProof::proof_bytes` is the externally-verifiable Blake2b-transcript proof. The two are never bridged.

## Issue(s)

Relates to #3138


mithril-stm/src/circuits/halo2_ivc/tests/common/generators/asset_generation.rs
```diff
@@ -713,6 +741,69 @@ pub(crate) fn generate_same_epoch_step_output_asset(
     );
 }
 
+/// Generates and writes the first-step certificate asset. The certificate is produced
+/// against `previous_state = State::genesis()`, so the asset captures the very first
+/// non-genesis cert in the chain — the input scenario for `IvcProverInput::prepare` at
+/// `step_counter == 0`.
+pub(crate) fn generate_first_step_cert_asset(setup: &AssetGenerationSetup, paths: &AssetPaths) {
```
hjeljeli32, line 749: As you added a new asset, can you kindly update the README in assets folder about the role of this asset and the commands to generate it?

mithril-stm/src/proof_system/ivc_halo2_snark/prover_input.rs
```diff
@@ -18,3 +36,658 @@ pub(crate) struct IvcProverInput {
...
+        let certificate_accumulator = setup.certificate_accumulator(dual_msm);
+        let previous_ivc_proof_accumulator = if is_genesis {
+            setup.trivial_previous_ivc_proof_accumulator()
+        } else {
+            setup.previous_ivc_proof_accumulator(
+                rolling_state.ivc_proof().as_bytes(),
+                &rolling_state.previous_ivc_proof_public_inputs(global),
+            )?
+        };
+        let mut next_accumulator = Accumulator::accumulate(&[
```
hjeljeli32, line 175: I think genesis needs a separate path before verifier-preparing the certificate proof.

The genesis/base-case generator uses `CertificateProofBytes::empty()`, `IvcProofBytes::empty()`, and sets `next_accumulator = current_accumulator.clone()` from `trivial_acc(&combined_fixed_base_names)`. It does not verifier-prepare a certificate proof and does not fold a certificate accumulator into the chain accumulator.

Here, `snark_proof.prepare_and_check(...)` is called unconditionally before `is_genesis` is handled, and later the code always folds `[current, certificate, previous_ivc]`. Replacing only the genesis certificate accumulator with a trivial one would still leave genesis requiring a valid certificate proof, which is not the base-case model.

I think the fix should be:
- at `step_counter == 0`, skip certificate proof verifier-preparation;
- skip previous IVC proof verifier-preparation;
- return the current/trivial chain accumulator directly as `next_accumulator`;
- use the genesis witness/preimage expected by the genesis base case.

Then `step_counter == 1` becomes the first real certificate step.

mithril-stm/src/proof_system/ivc_halo2_snark/prover_input.rs
```diff
@@ -18,3 +36,658 @@ pub(crate) struct IvcProverInput {
...
+            let asset_setup = shared_asset_setup();
+            let first_step =
+                load_embedded_first_step_cert_asset().expect("first step cert asset should load");
+
+            let global = build_global();
+            let combined_names: Vec<String> = setup.combined_fixed_bases.keys().cloned().collect();
+            let rolling_state =
+                IvcRollingState::genesis(asset_setup.genesis_signature, &combined_names);
+
+            let snark_proof = wrap_snark_proof(first_step.certificate_proof.clone().into_vec());
```
hjeljeli32, line 368: The circuit selects `global.genesis_message` at genesis, then hashes `witness.message_preimage` and asserts equality.

This test passes `first_step.message_preimage` from a normal first certificate. That preimage hashes to the first certificate message, not necessarily `global.genesis_message`. So prepare can return an `IvcProverInput` that looks valid to the test but cannot satisfy the IVC circuit.

mithril-stm/src/proof_system/ivc_halo2_snark/rolling_state.rs
hjeljeli32: Can we clarify the transcript role of `previous_proof.proof_bytes` here?

`IvcRollingState` later feeds this value into `IvcSetup::previous_ivc_proof_accumulator`, which verifier-prepares it with the Poseidon transcript. That is correct for the recursive/self proof consumed by the IVC circuit.

However `IvcProof::proof_bytes` is documented as “Externally-verifiable proof bytes (Blake2b transcript)”. If this field really stores the external Blake2b proof, then this path will consume it with the wrong transcript. If it stores the recursive Poseidon proof, then the comment/type is misleading.

mithril-stm/src/proof_system/ivc_halo2_snark/prover_input.rs
```diff
@@ -18,3 +36,658 @@ pub(crate) struct IvcProverInput {
...
+        snark_proof: SnarkProof<D>,
+        message: &[u8],
+        aggregate_verification_key_for_snark: &AggregateVerificationKeyForSnark<D>,
+        global: &Global,
+        decoded_protocol_message: &DecodedProtocolMessage,
+        rolling_state: &IvcRollingState,
+        setup: &IvcSetup,
+    ) -> StmResult<Self> {
+        // Verify the SNARK proof and prepare the dual MSM for the certificate accumulator.
+        let dual_msm = snark_proof.prepare_and_check(
```
hjeljeli32: I think this should use, or at least be checked against, `setup.certificate_verifying_key`.

`snark_proof.prepare_and_check(...)` verifier-prepares the certificate proof using the verification key embedded in `SnarkProof`. But the IVC circuit verifies the certificate proof with the certificate VK carried by `Global`, and the prepared certificate accumulator is later normalized with `setup.certificate_fixed_bases`.

If the `SnarkProof` was built with a different certificate VK than `setup.certificate_verifying_key`, this can produce a `DualMSM` for one VK and then extract/collapse it using fixed bases for another VK. That accumulator would not match what the IVC circuit computes.

For IVC correctness, I think `prepare` should either:
- verifier-prepare the certificate proof with `setup.certificate_verifying_key`, or
- explicitly reject when the proof’s embedded VK differs from `setup.certificate_verifying_key` / `global.certificate_verification_key`.

mithril-stm/src/proof_system/ivc_halo2_snark/prover_input.rs
```diff
@@ -18,3 +36,658 @@ pub(crate) struct IvcProverInput {
...
+        let is_next_epoch = certificate_epoch == chain_epoch + EpochNumber::new(1).as_field();
+        let is_genesis = rolling_state.state().step_counter == StepCounter::ZERO;
+
+        // Verify the genesis signature if at genesis.
+        if is_genesis {
+            rolling_state.verify_genesis_signature(global)?;
+        }
+
+        // Non-genesis steps must advance by zero or one epoch. Genesis bypasses this check.
+        if !is_genesis && !is_same_epoch && !is_next_epoch {
```
hjeljeli32: This epoch-transition check is missing the circuit's special first-non-genesis rule.

In `gadget.rs`, when `state.step_counter == 1`, the circuit requires the current certificate to be a next-epoch certificate: `is_valid = is_not_first || is_next_epoch`

So the first certificate after the genesis base case cannot be same-epoch.

Here, `prepare` only rejects non-genesis steps that are neither same-epoch nor next-epoch. That means `prepare` can accept a same-epoch certificate when `rolling_state.state().step_counter == 1`, but the IVC circuit will reject the resulting witness/public inputs.

I think prepare should mirror the gadget and reject:
```
if rolling_state.state().step_counter == StepCounter::new(1) && !is_next_epoch {
    return Err(...);
}
```

mithril-stm/src/proof_system/ivc_halo2_snark/prover_input.rs
```diff
@@ -18,3 +36,658 @@ pub(crate) struct IvcProverInput {
...
+            accumulator: midnight_circuits::verifier::Accumulator<BlstrsEmulation>,
+            genesis_signature: StandardSchnorrSignature,
+        ) -> IvcRollingState {
+            IvcRollingState::new(state, ivc_proof, accumulator, genesis_signature)
+        }
+
+        // ----- Scenario 1: genesis step -----
+
+        #[test]
+        fn prepare_at_genesis_produces_advanced_state_and_witness() {
```
hjeljeli32, line 357: Could we strengthen these tests to compare the prepared output against the committed golden artifacts, not only selected fields?

For instance we can assert (just suggestions, please ignore them if they don't make sense):
- full `input.next_state == step.next_state`, not only selected fields;
- full `input.next_accumulator == step.next_accumulator`;
- witness fields match the certificate data: certificate message, certificate Merkle-tree commitment, protocol-message preimage, and genesis signature.

hjeljeli32 (commented):
Good job @curiecrypt 👍 
I just left some comments.

mithril-stm/src/circuits/halo2_ivc/types.rs
```diff
@@ -68,7 +66,7 @@ macro_rules! u64_wrapper {
                 F::from(self.0)
             }
 
-            #[cfg(test)]
+            #[allow(dead_code)]
```
jpraynaud: Is this still needed?

mithril-stm/src/proof_system/ivc_halo2_snark/prover_input.rs
```diff
@@ -18,3 +36,658 @@ pub(crate) struct IvcProverInput {
...
+        fn build_rolling_state(
+            state: State,
+            ivc_proof: crate::circuits::halo2_ivc::types::IvcProofBytes,
+            accumulator: midnight_circuits::verifier::Accumulator<BlstrsEmulation>,
+            genesis_signature: StandardSchnorrSignature,
+        ) -> IvcRollingState {
+            IvcRollingState::new(state, ivc_proof, accumulator, genesis_signature)
+        }
+
+        // ----- Scenario 1: genesis step -----
```
jpraynaud: Can you remove all of these comments?

mithril-stm/src/circuits/halo2_ivc/tests/common/asset_readers.rs
```diff
@@ -57,6 +60,25 @@ pub(crate) struct VerificationContextAsset {
     pub(crate) certificate_verifying_key: MidnightVK,
 }
 
+/// Stored data of the first certificate produced from `State::genesis()`. Used to test
+/// `IvcProverInput::prepare` at the genesis transition (where the rolling state's
+/// `step_counter` is zero and no previous IVC proof exists).
+#[derive(Debug)]
+pub(crate) struct FirstStepCertAsset {
```
jpraynaud: The name is not very clear, maybe `FirstCertificateInEpochAsset` is better.

mithril-stm/src/circuits/halo2_ivc/tests/common/asset_readers.rs
```diff
@@ -57,6 +60,25 @@ pub(crate) struct VerificationContextAsset {
...
+    /// Certificate proof bytes (consumed by `prepare` via `SnarkProof`).
+    pub(crate) certificate_proof: CertificateProofBytes,
+    /// Expected next state after `prepare` advances by one step.
+    pub(crate) next_state: State,
+    /// 32-byte SHA-256 hash that the certificate proof committed to.
+    pub(crate) message: Vec<u8>,
+    /// 190-byte protocol-message preimage; `DecodedProtocolMessage::new(preimage)` decodes the four
+    /// epoch fields.
+    pub(crate) message_preimage: Vec<u8>,
```
jpraynaud: If you need to specify the length of the arrays, it'd better to use the types for that instead of the comments.

mithril-stm/src/circuits/halo2_ivc/tests/common/asset_readers.rs
```diff
@@ -57,6 +60,25 @@ pub(crate) struct VerificationContextAsset {
...
+    /// 32-byte canonical encoding of the AVK merkle root the certificate proof
+    /// committed to.
+    pub(crate) avk_merkle_root: [u8; 32],
```
jpraynaud:
```suggestion
    pub(crate) aggregate_verification_key_merkle_root: [u8; 32],
```

mithril-stm/src/circuits/halo2_ivc/tests/common/asset_readers.rs
```diff
@@ -57,6 +60,25 @@ pub(crate) struct VerificationContextAsset {
...
+}
+
 /// Stored output of extending the recursive chain by one more step.
 #[derive(Debug)]
 pub(crate) struct RecursiveStepOutputAsset {
```
jpraynaud: Same here, `FollowingCertificateInEpochAsset` or `TailCertificateInEpochAsset` would be better.

mithril-stm/src/circuits/halo2_ivc/tests/common/asset_readers.rs
```diff
@@ -68,6 +90,14 @@ pub(crate) struct RecursiveStepOutputAsset {
     pub(crate) next_state: State,
     /// Stored certificate proof consumed by the recursive step.
     pub(crate) certificate_proof: CertificateProofBytes,
+    /// 32-byte SHA-256 hash that the certificate proof committed to. Empty at genesis.
+    pub(crate) message: Vec<u8>,
+    /// 190-byte protocol-message preimage; `DecodedProtocolMessage::new(preimage)` decodes the four
+    /// epoch fields. Empty at genesis.
+    pub(crate) message_preimage: Vec<u8>,
```
jpraynaud: Same remark to use the types to enforce the length of arrays.

mithril-stm/src/circuits/halo2_ivc/tests/common/asset_readers.rs
```diff
@@ -419,3 +479,55 @@ pub(crate) fn store_recursive_step_output_asset(
     })?;
     Ok(())
 }
+
+/// Loads a first-step certificate asset from the committed binary asset layout.
+fn load_first_step_cert_asset_from_reader<R: Read>(
```
jpraynaud: Same remark concerning namings of this funciton and below.

mithril-stm/src/circuits/halo2_ivc/errors.rs
```diff
@@ -30,6 +30,29 @@ pub enum IvcCircuitError {
...
+    StepCounterOverflow { current: u64 },
+
+    /// Off-circuit step transition: a same-epoch certificate's preimage announces a
+    /// next-epoch lookahead that disagrees with the chain's previous state. All certs
+    /// in the same epoch must encode the same next-epoch fields.
+    #[error(
+        "IvcProverInput::prepare: same-epoch cert at chain epoch {chain_epoch} announces \
+         a next-epoch lookahead that does not match the chain state"
+    )]
+    SameEpochLookaheadMismatch { chain_epoch: u64 },
```
jpraynaud: I don't really understand what this error does and why we need to have 2 different errors to characterize that there is a mismatch of epochs.

mithril-stm/src/proof_system/ivc_halo2_snark/decoded_protocol_message.rs
```diff
@@ -9,12 +10,11 @@ use crate::{
...
-/// Per-call external input carrying the certificate's epoch identifier and the
-/// upcoming epoch's announcements.
-// TODO: remove this allow dead_code directive when the IVC prover consumes this epoch data
-#[allow(dead_code)]
+/// The certificate's protocol-message preimage together with the three fields decoded
+/// from its fixed byte ranges. Consumed as a per-call external input by the IVC prover
+/// input preparation step.
 #[derive(Clone)]
-pub(crate) struct EpochData {
+pub(crate) struct DecodedProtocolMessage {
```
jpraynaud: I still feel that this struct is a bit weird. Why don't we use a `ProtocolMessagePreimage` type that can extract when needed the epoch, Merkle tree commitment and protocol parameters when needed?

mithril-stm/src/proof_system/ivc_halo2_snark/rolling_state.rs
```diff
@@ -98,6 +108,37 @@ impl IvcRollingState {
...
     }
+
+    /// Verifies the chain's genesis Schnorr signature against the genesis message and
+    /// verification key carried in `global`. Called at the first proving step only.
+    pub(crate) fn verify_genesis_signature(&self, global: &Global) -> StmResult<()> {
+        self.genesis_signature.verify(
+            &[BaseFieldElement::from(global.genesis_message.as_field())],
+            &global.genesis_verification_key,
+        )?;
+        Ok(())
```
jpraynaud:
```suggestion
        self.genesis_signature.verify(
            &[BaseFieldElement::from(global.genesis_message.as_field())],
            &global.genesis_verification_key,
        )
```

mithril-stm/src/proof_system/ivc_halo2_snark/setup.rs
```diff
@@ -66,22 +75,24 @@ impl IvcSetup {
...
+    /// cache design).
     // TODO: swap `Temp*Provider` parameters for the production IVC cache providers
     // once they ship.
     pub(crate) fn load(
         trusted_setup_provider: &TrustedSetupProvider,
         certificate_key_provider: &TempCertificateKeyProvider,
         ivc_key_provider: &TempIvcKeyProvider,
     ) -> StmResult<Self> {
+        // Pull the full SRS only long enough to derive the verifier-side params; the SRS
+        // itself drops at end of scope and never lives in the returned setup.
```
jpraynaud: This comment is not needed.

mithril-stm/src/proof_system/ivc_halo2_snark/setup.rs
```diff
@@ -106,6 +117,47 @@ impl IvcSetup {
             combined_fixed_bases,
         })
     }
+
+    /// Wrap the certificate proof's prepared `DualMSM` into a collapsed accumulator on
+    /// the certificate circuit's fixed bases.
+    pub(crate) fn certificate_accumulator(
```
jpraynaud: We need to be careful with namings related to the accumulator as we can not recover an uncollapsed accumulator from a collapsed one:
```suggestion
    pub(crate) fn certificate_collapsed_accumulator(
```

mithril-stm/src/proof_system/ivc_halo2_snark/setup.rs
```diff
@@ -106,6 +117,47 @@ impl IvcSetup {
...
+            self.combined_fixed_bases.keys().cloned().collect();
+        trivial_acc(&combined_fixed_base_names)
+    }
+
+    /// Off-circuit verify of the previous step's IVC proof, returning the collapsed
+    /// accumulator the in-circuit IVC verifier gadget would have produced on the same
+    /// proof. Used at every non-genesis step.
+    pub(crate) fn previous_ivc_proof_accumulator(
+        &self,
+        proof_bytes: &[u8],
```
jpraynaud:
```suggestion
        ivc_proof_bytes: &[u8],
```

mithril-stm/src/proof_system/ivc_halo2_snark/setup.rs
```diff
@@ -106,6 +117,47 @@ impl IvcSetup {
...
+    pub(crate) fn trivial_previous_ivc_proof_accumulator(&self) -> Accumulator<BlstrsEmulation> {
+        let combined_fixed_base_names: Vec<String> =
+            self.combined_fixed_bases.keys().cloned().collect();
+        trivial_acc(&combined_fixed_base_names)
+    }
+
+    /// Off-circuit verify of the previous step's IVC proof, returning the collapsed
+    /// accumulator the in-circuit IVC verifier gadget would have produced on the same
+    /// proof. Used at every non-genesis step.
+    pub(crate) fn previous_ivc_proof_accumulator(
```
jpraynaud:
```suggestion
    pub(crate) fn previous_ivc_proof_collapsed_accumulator(
```

mithril-stm/src/proof_system/ivc_halo2_snark/prover_input.rs
```diff
@@ -18,3 +36,658 @@ pub(crate) struct IvcProverInput {
     /// Folded accumulator the new step's IVC proof commits to.
     pub(crate) next_accumulator: Accumulator<BlstrsEmulation>,
 }
+
+// TODO: remove this allow dead_code directive when the IVC prover consumes this input
+#[allow(dead_code)]
+impl IvcProverInput {
+    pub(crate) fn prepare<D: MembershipDigest>(
```
jpraynaud, line 48: This function is too long/complex and should be split in sub-functions that will be easier to read and that rae unit tested.

mithril-stm/src/proof_system/ivc_halo2_snark/prover_input.rs
```diff
@@ -18,3 +36,658 @@ pub(crate) struct IvcProverInput {
...
+    ) -> StmResult<Self> {
+        // Verify the SNARK proof and prepare the dual MSM for the certificate accumulator.
+        let dual_msm = snark_proof.prepare_and_check(
+            message,
+            aggregate_verification_key_for_snark,
+            &setup.srs_verifier_params,
+        )?;
+
+        let chain_epoch = rolling_state.state().current_epoch.as_field();
+        let certificate_epoch = decoded_protocol_message.current_epoch().0;
```
jpraynaud: The names are not very clear about what they represent and in particular the temporarility.

mithril-stm/src/proof_system/ivc_halo2_snark/prover_input.rs
```diff
@@ -18,3 +36,658 @@ pub(crate) struct IvcProverInput {
...
+        Ok(IvcProverInput {
+            witness,
+            next_state,
+            next_accumulator,
+        })
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
jpraynaud, line 198: All the `OnceLock` are not going to have any effect as tests are run in parallel from different binaries

mithril-stm/src/proof_system/ivc_halo2_snark/prover_input.rs
```diff
@@ -18,3 +36,658 @@ pub(crate) struct IvcProverInput {
...
+    use super::*;
+
+    // Asset-driven tests for `IvcProverInput::prepare`. Slow tier because the shared
+    // `IvcSetup` runs real `keygen` once per test binary (real cache providers are not
+    // yet wired in).
+    mod slow {
```
jpraynaud, line 201: All these tests are too long and are considerably increasing the duration of the CI tests. We need to find a way to avoid that inflation, probably by creating smaller functions that can be unit tested.

jpraynaud (commented):
@curiecrypt we can probably also get rid of the `ivc_halo2_snark/unsafe_setup_helpers.rs` module and replace temporary types with final types.

mithril-stm/src/circuits/halo2_ivc/tests/common/generators/asset_generation.rs
curiecrypt: README updated, thanks

mithril-stm/src/circuits/halo2_ivc/tests/common/asset_readers.rs
curiecrypt: array lengths specified, thanks

mithril-stm/src/circuits/halo2_ivc/tests/common/asset_readers.rs
curiecrypt: done

mithril-stm/src/proof_system/ivc_halo2_snark/rolling_state.rs
curiecrypt: fixed

mithril-stm/src/proof_system/ivc_halo2_snark/setup.rs
curiecrypt: removed

mithril-stm/src/circuits/halo2_ivc/tests/common/asset_readers.rs
curiecrypt: done

mithril-stm/src/circuits/halo2_ivc/tests/common/asset_readers.rs
curiecrypt: done

mithril-stm/src/circuits/halo2_ivc/tests/common/asset_readers.rs
curiecrypt: done

mithril-stm/src/circuits/halo2_ivc/tests/common/asset_readers.rs
curiecrypt: renamed

mithril-stm/src/proof_system/ivc_halo2_snark/setup.rs
curiecrypt: renamed

mithril-stm/src/proof_system/ivc_halo2_snark/setup.rs
curiecrypt: renamed

mithril-stm/src/proof_system/ivc_halo2_snark/setup.rs
curiecrypt: renamed

mithril-stm/src/circuits/halo2_ivc/errors.rs
curiecrypt: two error variants merged.

mithril-stm/src/proof_system/ivc_halo2_snark/decoded_protocol_message.rs
curiecrypt: struct deleted and `ProtocolMessagePreimage` is used instead

mithril-stm/src/proof_system/ivc_halo2_snark/prover_input.rs
jpraynaud: This will be done in https://github.com/input-output-hk/mithril/issues/3319

mithril-stm/src/proof_system/ivc_halo2_snark/prover_input.rs
jpraynaud: This will be done in https://github.com/input-output-hk/mithril/issues/3319

mithril-stm/src/proof_system/ivc_halo2_snark/prover_input.rs
jpraynaud: This will be done in https://github.com/input-output-hk/mithril/issues/3319

mithril-stm/src/proof_system/ivc_halo2_snark/prover_input.rs
jpraynaud: This will be done in https://github.com/input-output-hk/mithril/issues/3319

mithril-stm/src/proof_system/ivc_halo2_snark/prover_input.rs
curiecrypt: Done. New top-level `if is_genesis { ... return ... }` branch in prepare, verifies only the chain's genesis Schnorr signature, builds the base-case witness/state from `protocol_message_preimage`, returns `rolling_state.accumulator()` unchanged. Cert verification and previous-IVC verifier-prep are skipped entirely at genesis.

mithril-stm/src/proof_system/ivc_halo2_snark/prover_input.rs
curiecrypt: Done. New `IvcCircuitError::CertificateVerifyingKeyMismatch` variant in errors, and new `SnarkProof::circuit_verification_key()` accessor. 
Check in prepare (non-genesis path) compares proof's verification key against setup's verification key and rejects on mismatch before `prepare_and_check`.

mithril-stm/src/proof_system/ivc_halo2_snark/prover_input.rs
curiecrypt: Done. New `EpochTransitionErrorKind::FirstCertificateAfterGenesisMustBeNextEpoch` variant in errors. 
Check in prepare rejects when 
```rust
rolling_state.state().step_counter == StepCounter::new(1) && !is_next_epoch
```
mirroring the circuit's `is_valid = is_not_first || is_next_epoch`.

mithril-stm/src/proof_system/ivc_halo2_snark/prover_input.rs
curiecrypt: Done. Added `PartialEq + Eq` derives to `State` and `Witness`. Added `accumulator_bytes` test helper for canonical byte equality. Rewrote scenarios genesis, same-epoch, next-epoch to assert:
- full next_state equality against the golden artifact, 
- full next_accumulator byte equality, 
- witness field comparisons against the asset's certificate data.

Also, note that all tests are flagged as ignore until key cache integration is completed.

mithril-stm/src/circuits/halo2_ivc/types.rs
curiecrypt: it is only used in tests so I replaced by test flag again.

mithril-stm/src/proof_system/ivc_halo2_snark/prover_input.rs
curiecrypt: removed

mithril-stm/src/proof_system/ivc_halo2_snark/rolling_state.rs
curiecrypt: Done, deleted `IvcRollingState::from_previous_proof`

mithril-stm/src/proof_system/ivc_halo2_snark/prover_input.rs
curiecrypt: Done. The related test scenarios now use `build_genesis_protocol_message_preimage(asset_setup)` instead of `first_step.message_preimage` (the latter hashes to the first-cert message, not `global.genesis_message`).

mithril-stm/src/proof_system/ivc_halo2_snark/prover_input.rs
```diff
@@ -18,3 +33,631 @@ pub(crate) struct IvcProverInput {
...
+                &global,
+                &epoch_data,
+                &rolling_state,
+                setup,
+            );
+
+            assert!(
+                result.is_err(),
+                "prepare should reject an invalid genesis signature"
+            );
```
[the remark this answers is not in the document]
curiecrypt: resolved

mithril-stm/src/circuits/halo2_ivc/types.rs
```diff
@@ -5,7 +5,11 @@
 
 use ff::Field;
```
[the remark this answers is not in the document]
curiecrypt: disagree.

mithril-stm/src/proof_system/ivc_halo2_snark/prover_input.rs
```diff
@@ -18,3 +33,645 @@ pub(crate) struct IvcProverInput {
...
+        use tempfile::tempdir;
+
+        use crate::{
+            MithrilMembershipDigest, Parameters,
+            circuits::{
+                halo2_ivc::{
+                    K,
+                    errors::{EpochTransitionErrorKind, IvcCircuitError},
+                    io::Write as IvcWrite,
+                    tests::common::{
```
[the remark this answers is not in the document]
curiecrypt: disagree

jpraynaud (approved):

mithril-stm/src/circuits/halo2_ivc/tests/common/generators/asset_generation.rs
```diff
@@ -699,20 +719,91 @@ pub(crate) fn generate_same_epoch_step_output_asset(
...
         "generate_same_epoch_step_output: done in {:?}",
         total_start.elapsed()
     );
 }
 
+/// Generates and writes the first-step certificate asset. The certificate is produced
+/// against `previous_state = State::genesis()`, so the asset captures the very first
+/// non-genesis cert in the chain — the input scenario for `IvcProverInput::prepare` at
+/// `step_counter == 0`.
+pub(crate) fn generate_first_step_cert_asset(setup: &AssetGenerationSetup, paths: &AssetPaths) {
```
hjeljeli32, line 749: Small naming/documentation suggestion: the comment says the certificate is produced against `previous_state = State::genesis()`, but the code below uses `build_genesis_base_case_next_state(...)`.

So This asset is the first real non-genesis certificate after the genesis base-case state. Could we adjust the wording here, and also the asset README, to say something like “first post-genesis certificate”? That would avoid confusing this asset with the genesis base-case itself.

mithril-stm/src/proof_system/ivc_halo2_snark/setup.rs
```diff
@@ -106,6 +115,50 @@ impl IvcSetup {
...
+
+    /// Off-circuit verify of the previous step's IVC proof, returning the collapsed
+    /// accumulator the in-circuit IVC verifier gadget would have produced on the same
+    /// proof. Used at every non-genesis step.
+    pub(crate) fn previous_ivc_proof_collapsed_accumulator(
+        &self,
+        ivc_proof_bytes: &[u8],
+        public_inputs: &[CircuitBase],
+    ) -> StmResult<Accumulator<BlstrsEmulation>> {
+        let dual_msm = verify_and_prepare_accumulator(
```
hjeljeli32, line 151: Minor refactor point, that we can handle in #3319: this helper is now used for both certificate proofs and previous IVC proofs. The implementation itself is generic enough for both roles. The certificate/IVC distinction is made by the caller.

Now this helper lives in `certificate_proof.rs`, the docs describe only certificate proofs, and all failures map to `CertificateProofRejected`. I did this mistake previously 🙏 

Could we later move/rename this to something role-neutral and let the certificate/previous-IVC wrappers map failures to role-specific errors?

hjeljeli32 (approved):
LGTM 👍 
Left non-blocking comments.

mithril-stm/src/circuits/halo2_ivc/tests/common/generators/asset_generation.rs
curiecrypt: will do in 3319

mithril-stm/src/proof_system/ivc_halo2_snark/setup.rs
curiecrypt: will do in 3319

Issue #3295, opened 2026-06-01
Remove Mithril nodes support in `mithril-devnet`


jpraynaud:
## Why
We want to remove the support of the Mithril nodes in `mithril-devnet` which is a legacy feature that is difficult to maintain. The `mithril-devnet` must only be responsible to run Cardano and DMQ nodes to be used by the end to end tests.

## What
Remove support for the Mithril nodes in `mithril-devnet`.

## How

- [x] Remove the Mithril nodes related parts in commands:
  - [x] `log`
  - [x] `query`
  - [x] `run`
  - [x] `stop`
  - [x] `visualize` (can be removed completely)
- [x] ~~Remove the `NODES=` option~~ (kept for cardano and the optional DMQ)
- [x] Clean the `mkfiles-init.sh` file
- [x] Clean the `mkfiles-query.sh` file
- [x] Clean the `mkfiles-start.sh` file
- [x] Remove the `mkfiles_docker.sh` script
- [x] Remove the `Dockerfile` (not `Dockerfile.ci`) in:
  - [x] `mithril-client-cli`
  - [x] `mithril-aggregator`
  - [x] `mithril-signer`
  - [x] `mithril-relay`
- [x] Update README for `mithril-devnet` (and mention the end to end tests with `--run-only` option)
- [x] Make sure that the backward compatibility test still works
- [x] Rename the `mithril-devnet` folder to `cardano-devnet`

PR #3314, opened 2026-06-08
Removing nightly docker build of mithril legacy images

turmelclem:
## Content

- removing nightly docker build of mithril legacy images
- silently kill dmq-node with non-failing return code in e2e test script

Relates to #3295


.github/workflows/docker-builds.yml
jpraynaud, line 1: You probably need to remove the `docker-builds` entry in `.github/workflows/nightly-dispatcher.yml`

jpraynaud (approved):
LGTM

Alenar (approved):

Issue #3139, opened 2026-04-01
Recursive SNARK aggregation primitives: Create SNARK proof with circuit


jpraynaud:
(accumulator, non recursive SNARK to create first, ...)

PR #3315, opened 2026-06-08
implement `verify` IVC  proof over blake2b/poseidon

### LINKED ISSUE #3140: Recursive SNARK aggregation primitives: Verify SNARK proof
(genesis verification key, ...)

perturbing:
## Content

- commit 1 : which implements the verifier logic
- commit 2 : which creates a type for the setup to not depend on the big SRS.
- commit 3 : add test to e2e test the `IvcProof::verify()`
- commit 4: refactor `IvcProof` to be generic over transcript hash type
- commit 5: add test-only IvcVerifierSetup constructor from live SRS
- commit 6: document Global/IvcVerifierSetup key-consistency invariant on verify

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Issue(s)

This PR closes [this](https://github.com/input-output-hk/mithril/issues/3140) issue and contains two commits.

### DIFF
--- a/mithril-stm/Cargo.toml
+++ b/mithril-stm/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mithril-stm"
-version = "0.10.33"
+version = "0.10.34"
 edition = { workspace = true }
 authors = { workspace = true }
 homepage = { workspace = true }
--- a/mithril-stm/Cargo.toml
+++ b/mithril-stm/Cargo.toml
@@ -19,6 +19,7 @@ default = ["num-integer-backend"]
 benchmark-internals = []
 # Enable experimental SNARK support 
 future_snark = [
+    "dep:blake2b_simd",
     "dep:ff",
     "dep:group",
     "dep:midnight-circuits",
@@ -44,6 +45,7 @@ rug-backend = ["rug/default"]
 [dependencies]
 anyhow = { workspace = true }
 blake2 = "0.10.6"
+blake2b_simd = { version = "1.0.4", optional = true }
 # Enforce blst portable feature for runtime detection of Intel ADX instruction set.
 blst = { version = "0.3.16", features = ["portable"] }
 ciborium = "0.2.2"
--- a/mithril-stm/src/circuits/halo2_ivc/errors.rs
+++ b/mithril-stm/src/circuits/halo2_ivc/errors.rs
@@ -8,6 +8,22 @@ pub enum IvcCircuitError {
     #[error("Certificate proof verification rejected")]
     CertificateProofRejected,
 
+    /// Blake2b transcript could not be prepared for KZG opening verification.
+    #[error("Recursive IVC proof rejected: Blake2b transcript preparation failed")]
+    RecursiveProofTranscriptPreparationFailed,
+
+    /// Blake2b transcript contained unexpected trailing bytes after preparation.
+    #[error("Recursive IVC proof rejected: Blake2b transcript was not fully consumed")]
+    RecursiveProofTranscriptNotFullyConsumed,
+
+    /// KZG opening equations (dual MSM check) did not verify.
+    #[error("Recursive IVC proof rejected: KZG opening check failed")]
+    RecursiveProofKzgOpeningFailed,
+
+    /// Folded accumulator pairing equation did not verify.
+    #[error("Recursive IVC proof rejected: accumulator pairing check failed")]
+    RecursiveProofAccumulatorFailed,
+
     /// IVC verification key degree does not match the IVC circuit degree constant K.
     #[error(
         "IvcCircuitData::validate_ivc_verification_key_degree failed: expected k={expected}, got k={actual}"
--- a/mithril-stm/src/circuits/halo2_ivc/errors.rs
+++ b/mithril-stm/src/circuits/halo2_ivc/errors.rs
@@ -8,22 +8,6 @@ pub enum IvcCircuitError {
     #[error("Certificate proof verification rejected")]
     CertificateProofRejected,
 
-    /// Blake2b transcript could not be prepared for KZG opening verification.
-    #[error("Recursive IVC proof rejected: Blake2b transcript preparation failed")]
-    RecursiveProofTranscriptPreparationFailed,
-
-    /// Blake2b transcript contained unexpected trailing bytes after preparation.
-    #[error("Recursive IVC proof rejected: Blake2b transcript was not fully consumed")]
-    RecursiveProofTranscriptNotFullyConsumed,
-
-    /// KZG opening equations (dual MSM check) did not verify.
-    #[error("Recursive IVC proof rejected: KZG opening check failed")]
-    RecursiveProofKzgOpeningFailed,
-
-    /// Folded accumulator pairing equation did not verify.
-    #[error("Recursive IVC proof rejected: accumulator pairing check failed")]
-    RecursiveProofAccumulatorFailed,
-
     /// IVC verification key degree does not match the IVC circuit degree constant K.
     #[error(
         "IvcCircuitData::validate_ivc_verification_key_degree failed: expected k={expected}, got k={actual}"
--- a/mithril-stm/src/circuits/halo2_ivc/tests/off_circuit/proof_verification.rs
+++ b/mithril-stm/src/circuits/halo2_ivc/tests/off_circuit/proof_verification.rs
@@ -12,6 +12,8 @@
 //! Negative tests for `IvcProof::verify` cover tampered proof bytes, a mismatched
 //! state, and a mismatched accumulator — each asserting that `verify()` returns `Err`.
 
+use std::marker::PhantomData;
+
 use midnight_circuits::types::Instantiable;
 
 use crate::circuits::halo2_ivc::{
@@ -147,14 +149,15 @@ fn ivc_proof_verify_accepts_stored_recursive_step_output() {
         combined_fixed_bases: verification_context.combined_fixed_bases,
     };
 
-    let proof = IvcProof {
+    let proof = IvcProof::<blake2b_simd::State> {
         proof_bytes: step_output.ivc_proof,
         state: step_output.next_state,
         accumulator: step_output.next_accumulator,
+        _hash: PhantomData,
     };
 
     proof
-        .verify_blake2b(&global, &verifier_setup)
+        .verify(&global, &verifier_setup)
         .expect("stored recursive step output should pass IvcProof::verify");
 }
 
@@ -171,14 +174,15 @@ fn ivc_proof_verify_rejects_tampered_proof_bytes() {
     let mid = tampered_bytes.len() / 2;
     tampered_bytes[mid] ^= 0xff;
 
-    let proof = IvcProof {
+    let proof = IvcProof::<blake2b_simd::State> {
         proof_bytes: IvcProofBytes::new(tampered_bytes),
         state: step_output.next_state,
         accumulator: step_output.next_accumulator,
+        _hash: PhantomData,
     };
 
     assert!(
-        proof.verify_blake2b(&global, &verifier_setup).is_err(),
+        proof.verify(&global, &verifier_setup).is_err(),
         "tampered proof bytes should be rejected by IvcProof::verify"
     );
 }
@@ -194,14 +198,15 @@ fn ivc_proof_verify_rejects_mismatched_state() {
     let same_epoch = load_embedded_following_certificate_in_epoch_asset()
         .expect("same-epoch step output asset should load");
 
-    let proof = IvcProof {
+    let proof = IvcProof::<blake2b_simd::State> {
         proof_bytes: step_output.ivc_proof,
         state: same_epoch.next_state,
         accumulator: step_output.next_accumulator,
+        _hash: PhantomData,
     };
 
     assert!(
-        proof.verify_blake2b(&global, &verifier_setup).is_err(),
+        proof.verify(&global, &verifier_setup).is_err(),
         "state from a different proof should be rejected by IvcProof::verify"
     );
 }
@@ -217,35 +222,37 @@ fn ivc_proof_verify_rejects_mismatched_accumulator() {
     let same_epoch = load_embedded_following_certificate_in_epoch_asset()
         .expect("same-epoch step output asset should load");
 
-    let proof = IvcProof {
+    let proof = IvcProof::<blake2b_simd::State> {
         proof_bytes: step_output.ivc_proof,
         state: step_output.next_state,
         accumulator: same_epoch.next_accumulator,
+        _hash: PhantomData,
     };
 
     assert!(
-        proof.verify_blake2b(&global, &verifier_setup).is_err(),
+        proof.verify(&global, &verifier_setup).is_err(),
         "accumulator from a different proof should be rejected by IvcProof::verify"
     );
 }
 
 #[test]
 fn ivc_proof_verify_rejects_poseidon_proof_bytes() {
-    // `IvcProof::verify` always reads the transcript with Blake2b. Feeding it
-    // Poseidon-transcript proof bytes (from `recursive_chain_state`) must fail
-    // because the two transcript formats are not interchangeable.
+    // Constructing an `IvcProof<blake2b_simd::State>` with Poseidon-transcript bytes
+    // and verifying it with the Blake2b path must fail: the two transcript formats
+    // are not interchangeable.
     let (global, verifier_setup) = build_proof_verifier_context();
     let chain_state = load_embedded_recursive_chain_state_asset()
         .expect("recursive chain state asset should load");
 
-    let proof = IvcProof {
+    let proof = IvcProof::<blake2b_simd::State> {
         proof_bytes: chain_state.ivc_proof,
         state: chain_state.state,
         accumulator: chain_state.accumulator,
+        _hash: PhantomData,
     };
 
     assert!(
-        proof.verify_blake2b(&global, &verifier_setup).is_err(),
-        "Poseidon proof bytes should be rejected by the Blake2b-based IvcProof::verify"
+        proof.verify(&global, &verifier_setup).is_err(),
+        "Poseidon proof bytes should be rejected by IvcProof::<Blake2b>::verify"
     );
 }
--- a/mithril-stm/src/circuits/halo2_ivc/tests/off_circuit/proof_verification.rs
+++ b/mithril-stm/src/circuits/halo2_ivc/tests/off_circuit/proof_verification.rs
@@ -7,21 +7,49 @@
 //! These tests add the same combined check for:
 //!   - the same-epoch Blake2b proof (`same_epoch_step_output`)
 //!   - the chain-state Poseidon proof (`recursive_chain_state`)
+//!   - the `IvcProof::verify` high-level API path (next-epoch Blake2b proof)
+//!
+//! Negative tests for `IvcProof::verify` cover tampered proof bytes, a mismatched
+//! state, and a mismatched accumulator — each asserting that `verify()` returns `Err`.
 
 use midnight_circuits::types::Instantiable;
 
 use crate::circuits::halo2_ivc::{
     AssignedAccumulator,
+    state::Global,
     tests::common::{
         asset_readers::{
             load_embedded_following_certificate_in_epoch_asset,
-            load_embedded_recursive_chain_state_asset, load_embedded_verification_context_asset,
+            load_embedded_next_epoch_step_output_asset,
+            load_embedded_recursive_chain_state_asset,
+            load_embedded_verification_context_asset,
         },
+        generators::{build_asset_generation_setup, build_recursive_global},
         helpers::{
             verify_prepare_blake2b_recursive_proof, verify_prepare_poseidon_recursive_proof,
         },
     },
+    types::IvcProofBytes,
 };
+use crate::proof_system::ivc_halo2_snark::{proof::IvcProof, verifier_setup::IvcVerifierSetup};
+
+fn build_proof_verifier_context() -> (Global, IvcVerifierSetup) {
+    let ctx =
+        load_embedded_verification_context_asset().expect("verification context asset should load");
+    let setup = build_asset_generation_setup();
+    let global = build_recursive_global(
+        &setup,
+        &ctx.certificate_verifying_key,
+        &ctx.recursive_verifying_key,
+    );
+    let verifier_setup = IvcVerifierSetup {
+        verifier_params: ctx.verifier_params,
+        tau_g2: ctx.verifier_tau_in_g2,
+        ivc_verifying_key: ctx.recursive_verifying_key,
+        combined_fixed_bases: ctx.combined_fixed_bases,
+    };
+    (global, verifier_setup)
+}
 
 #[test]
 fn same_epoch_proof_passes_dual_msm_check_and_accumulator_check() {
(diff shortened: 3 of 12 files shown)


mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
```diff
@@ -31,3 +49,58 @@ pub(crate) struct IvcProof {
...
+// TODO: remove this allow dead_code directive when IvcProof::verify is called
+#[allow(dead_code)]
+impl IvcProof {
+    /// Verifies the IVC proof (Blake2b transcript) and its folded accumulator.
+    ///
+    /// This is the externally-verifiable form of the proof emitted at the end of
+    /// each proving step. It checks:
+    /// 1. The Blake2b KZG opening equations against the IVC verifying key.
+    /// 2. The folded accumulator pairing equation.
+    pub(crate) fn verify(
```
hjeljeli32, line 99: This verifier implementation makes `IvcProof` clearly represent the externally-verifiable IVC proof (Blake2b transcript proof). Good for this `verify` method 👍 

However, IMO there is an integration risk with the prover-side rolling state: `IvcRollingState::from_previous_proof` copies `IvcProof::proof_bytes` into a field documented as the previous IVC proof under the Poseidon transcript, which is the recursive/self-proof role consumed by the next proving step.

mithril-stm/src/proof_system/ivc_halo2_snark/verifier_setup.rs
```diff
@@ -0,0 +1,156 @@
...
+            tau_g2,
+            ivc_verifying_key,
+            combined_fixed_bases,
+        })
+    }
+
+    /// Derive from an already-built [`IvcSetup`], reusing its precomputed fixed bases.
+    ///
+    /// Avoids recomputing fixed bases from scratch when a proving session is already running.
+    pub(crate) fn from_ivc_setup(ivc_setup: &IvcSetup) -> StmResult<Self> {
```
hjeljeli32, line 84: I am worried that we silently mixe two setup sources. This function reuses `ivc_setup.ivc_verifying_key` and `ivc_setup.combined_fixed_bases`, but it reads `verifier_params` and `tau_g2` from the embedded production verifier params. If `ivc_setup` was built from another SRS, for example the current unsafe/deterministic test SRS, the returned verifier setup combines keys/fixed bases from one setup with KZG verifier parameters from another.

Could we for instance:
- derive `verifier_params` and `tau_g2` from `ivc_setup.srs` in this constructor; or
- make this constructor explicitly production-only and reject/non-expose it for non-production setups?

mithril-stm/src/proof_system/ivc_halo2_snark/verifier_setup.rs
```diff
@@ -0,0 +1,156 @@
...
+    pub(crate) fn read_embedded_params() -> StmResult<(ParamsVerifierKZG<Bls12>, G2Affine)> {
+        let verifier_params = ParamsVerifierKZG::<Bls12>::read(
+            &mut &IVC_VERIFIER_PARAMS_BYTES[..],
+            SerdeFormat::RawBytesUnchecked,
+        )
+        .with_context(|| "Failed to read embedded IVC verifier params")?;
+
+        // `ParamsVerifierKZG.s_g2` is not publicly accessible; tau_g2 is extracted by
+        // re-reading the leading G2 element from the same bytes (s_g2 is serialized first).
+        let tau_g2: G2Affine = G2Projective::read(
```
hjeljeli32: Small suggested simplification: with the current `midnight-proofs` version, `ParamsVerifierKZG` exposes `s_g2()` publicly.

So after reading `verifier_params`, we can derive `tau_g2` with something like:
`let tau_g2 = verifier_params.s_g2().into();`
This avoids depending on the raw serialization layout a second time.

WDYT?

mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
```diff
@@ -31,3 +49,58 @@ pub(crate) struct IvcProof {
...
+        &self,
+        global: &Global,
```
hjeljeli32, line 101: This verifier API receives `global` and `verifier_setup` independently, so production callers need to preserve the same invariant: `global` must be built from the same certificate and IVC verifying keys used to construct `IvcVerifierSetup`.

Could we document this requirement on `verify` / `IvcVerifierSetup`, or provide a constructor that derives both from the same key material? Because if they differ, verification should fail through the public-input proof check or accumulator check, but the caller will only see a generic `RecursiveProofRejected`, which may be hard to diagnose.

hjeljeli32 (commented):
Good job @perturbing 👍 
Left some comments.

mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
perturbing: Yeah, this is indeed a true gap that we currently have.

I believe @curiecrypt and @jpraynaud had a discussion about this yesterday.

But, as you also say, this is a proven issue: there we should have the option to target the Blake2b or Poseidon path.

The latter is for continuing the rolling state, and the former is for "finalizing" that state to be verified elsewhere.

I believe that the Poseidon verifier is implemented in the recursive step in #3313

mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
```diff
@@ -2,12 +2,30 @@
 
 use std::sync::Arc;
 
-use midnight_circuits::verifier::{Accumulator, BlstrsEmulation};
+use group::Group;
+use midnight_circuits::{
+    types::Instantiable,
+    verifier::{Accumulator, AssignedAccumulator, BlstrsEmulation},
+};
```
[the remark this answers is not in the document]
perturbing: As per this run https://github.com/input-output-hk/mithril/actions/runs/27195607817/job/80286370381

This was a false positive...

Instantiable is restored, the reviewer's feedback was incorrect; the trait must be in scope for `AssignedAccumulator::as_public_input` to resolve.

mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
```diff
@@ -31,3 +46,58 @@
...
+    /// 2. The folded accumulator pairing equation.
+    pub(crate) fn verify(
+        &self,
+        global: &Global,
+        verifier_setup: &IvcVerifierSetup,
+    ) -> StmResult<()> {
+        let public_inputs: Vec<CircuitBase> = [
+            global.as_public_input(),
+            self.state.as_public_input(),
+            AssignedAccumulator::as_public_input(&self.accumulator),
```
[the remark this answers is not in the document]
perturbing: This is pertaining to the old version of the file. The new verions rolled back here https://github.com/input-output-hk/mithril/compare/d431916940cb08ae255c47028dda34466be8ed5a..b1cf6c70223f453279ef104fc0c6396ad2c22f3b

fixes the wrong suggestion from copilot here https://github.com/input-output-hk/mithril/pull/3315#discussion_r3372061340

mithril-stm/src/proof_system/ivc_halo2_snark/verifier_setup.rs
perturbing: I missed that! Good point, much cleaner this way :) 

fixed in https://github.com/input-output-hk/mithril/compare/b1cf6c70223f453279ef104fc0c6396ad2c22f3b..d25f98818bafb2f285ac3a8396257f7ad288e12f

mithril-stm/src/proof_system/ivc_halo2_snark/verifier_setup.rs
perturbing: Hmmm, so I think that the embedded bytes are the _authoritative_ production value. The prover and verifier must agree on the same SRS. So an SRS update always requires updating the hardcoded bytes too. Meaning in production, deriving from `ivc_setup.srs` or reading the embedded constant should yield identical results.

Now I see your trouble, as this will not be compatible with the test SRS. But perhaps then the question is: how in general do we handle this conflict of wanting to hardcode values for production, but also have test use other values?

mithril-stm/src/proof_system/ivc_halo2_snark/verifier_setup.rs
```diff
@@ -0,0 +1,146 @@
...
+//! loading the full KZG SRS (hundreds of MB). The KZG verifier parameters are
+//! embedded as a compile-time constant and deserialized on construction.
+
+use std::collections::BTreeMap;
+
+use anyhow::Context;
+use midnight_curves::{Bls12, G1Projective, G2Affine};
+use midnight_proofs::{
+    poly::kzg::params::ParamsVerifierKZG,
+    utils::{SerdeFormat, helpers::ProcessedSerdeObject},
```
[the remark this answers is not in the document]
perturbing: https://github.com/input-output-hk/mithril/compare/d25f98818bafb2f285ac3a8396257f7ad288e12f..74949847d1543e291b84e51d44017fa97047d534

Fixed

mithril-stm/Cargo.toml
jpraynaud, line 88: This dev dependency can probably be removed now:
```suggestion

```

mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
```diff
@@ -31,3 +49,58 @@ pub(crate) struct IvcProof {
...
+        if !dual_msm.check(&verifier_setup.verifier_params) {
+            return Err(IvcCircuitError::RecursiveProofRejected.into());
+        }
+
+        if !self
+            .accumulator
+            .check(&verifier_setup.tau_g2, &verifier_setup.combined_fixed_bases)
+        {
+            return Err(IvcCircuitError::RecursiveProofRejected.into());
+        }
```
jpraynaud: As all these verification steps all return the same error, debugging is going to be complicated. We have few options here:
- Add a context to the errors with `with_context` call after the `map_err`
- Create distinct variant errors for each step (preferred)
- Maybe we could keep some information about the source error in these variants

mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
jpraynaud, line 99: This function should be unit-tested.

jpraynaud: You're right @hjeljeli32, this `IvcRollingState::from_previous_proof` does not make sense as the transcript hashes are different. We should simply remove it as it is only currently used in the tests.

mithril-stm/src/proof_system/ivc_halo2_snark/verifier_setup.rs
```diff
@@ -0,0 +1,143 @@
...
+/// Serialized `ParamsVerifierKZG` from the Midnight production trusted SRS.
+///
+/// `ParamsVerifierKZG` serializes only `s_g2` — a single G2 point from the power-of-tau
+/// ceremony. This value is independent of the circuit degree `k`, so both the non-recursive
+/// and IVC circuits use the same bytes (see `SNARK_VERIFIER_PARAMS_BYTES` in
+/// `halo2_snark/unsafe_helpers.rs`).
+///
+/// Regenerate by running the `ivc_verifier_params_match_trusted_srs` test (requires SRS
+/// download), serializing `srs.verifier_params()` with `SerdeFormat::RawBytesUnchecked`.
+const IVC_VERIFIER_PARAMS_BYTES: [u8; 192] = [
```
jpraynaud: This constant should be named `KZG_VERIFIER_PARAMS` and shared between recursive and non-recurisve verifier setups.

mithril-stm/src/proof_system/ivc_halo2_snark/verifier_setup.rs
jpraynaud: I agree with @hjeljeli32. If we need such a function for tests it must be behind a `#[cfg(test)]` and its comment should state a "for tests only" mention.

mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
jpraynaud: @curiecrypt could you remove it in PR https://github.com/input-output-hk/mithril/pull/3313 and acknowledge in this thread once donde?

perturbing: Just for clarification (also for myself): so the ticket #3313 implements the proving inputs for making the proof. It is ticket #3139 that implements the creation of either the Poseidon or Blake2b-based proof.

And just to reconfirm: it is indeed bad that we have one object that represents two things. That is, the same type `IvcProof::proof_bytes` can now represent a proof over either transcript methods.

Can you clarify what is disliked about the `IvcRollingState::from_previous_proof`?

Besides that, it has the hidden assumption that the proof bytes are always Poseidon-based. Which is bad, but note that the preparation for the next proof step verifies these bytes, so if they are Blake2b-based, it fails.

jpraynaud: > Just for clarification (also for myself): so the ticket #3313 implements the proving inputs for making the proof. It is ticket #3139 that implements the creation of either the Poseidon or Blake2b-based proof.

Yes, that's the ticket that will take care of creating the IVC proofs

> And just to reconfirm: it is indeed bad that we have one object that represents two things. That is, the same type `IvcProof::proof_bytes` can now represent a proof over either transcript methods.

I agree this is not an ideal situation. We could use a generic transcript hash type to the `IvcProof<TranscriptHash>` to avoid confusion. @curiecrypt @hjeljeli32 WDYT?

> Can you clarify what is disliked about the `IvcRollingState::from_previous_proof`?

The `IvcRollingState` is supposed to carry only bytes of a proof created with Poseidon transcript hash as it is used to extend the certificate chain for the following epoch. It will be one of the output of the prover at one epoch and become one of  the input of the prover at the next epoch (and that's its only purpose). I don't see any situation where we would need to use an `IvcProof` to create it (or maybe in tests? In that case, it must be properly gated). Removing the function would avoid the situation where someone uses an `IvcProof` created with a Blake2 hash as an input to the rolling state (which would ultimately not work).

perturbing: > I agree this is not an ideal situation. We could use a generic transcript hash type to the IvcProof<TranscriptHash> to avoid confusion. @curiecrypt @hjeljeli32 WDYT?

Regarding this, as discussed today, it is not that bad. We now implement `verify()` using blake2b over `IvcProof`, and as discussed, a Poseidon-based one is also needed. But note this is an assumption. Because this verification will be performed in the circuit. That is, in the proof creation (either a Poseidon or Blake2b next ivc proof), we can either successfully prove the next step or not. This step includes verification of the previous proof over Poseidon!

TLDR: We do not need to change the `IvcProof` object to represent that. I'll add a comment that the verify implementation will fail if it is called over a Poseidon-based. I'll also rename `.verify() -> .verify_blake2b()` :+1:

mithril-stm/Cargo.toml
perturbing: `blake2b_simd::State` is referenced directly in test code at `generators/proofs.rs:109` and line `151`

jpraynaud: Yes, but is now part of the dependencies so it's not needed to repeat it.

mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
perturbing: Fixed in https://github.com/input-output-hk/mithril/pull/3315/changes/7d37a2b70c94d87408efbf924e3853b52789feec

mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
perturbing: That is done in https://github.com/input-output-hk/mithril/pull/3315/changes/7d37a2b70c94d87408efbf924e3853b52789feec

resolving this

mithril-stm/src/proof_system/ivc_halo2_snark/verifier_setup.rs
perturbing: thanks fixed in https://github.com/input-output-hk/mithril/pull/3315/changes/7d37a2b70c94d87408efbf924e3853b52789feec

mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
perturbing: This is for #3139 , as we need proofs to verify :)

perturbing: Opening this again: verification is indeed implicit in proving the next step (as this would fail). That said, this is computationally expensive. Better to fail fast. 

This is why in this PR we will make the IVC proof generic over the hash `IvcProof<TranscriptHash>`.

The verify function then depends on either one.

perturbing: Fixed in dae7c93b7e2fcdc8663ae2290ccd89a0c49fb716

mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
perturbing: I believe we cannot make such a constructor, so I added documentation :+1: 

Fixed in 21307eb0c4e1a06bbc6d601505969ca8740a8e50

mithril-stm/src/proof_system/ivc_halo2_snark/verifier_setup.rs
perturbing: fixed in b11cd1bc01ee703d5a1fcd2ab57b187a4ae7c692

mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
curiecrypt: @perturbing @jpraynaud @hjeljeli32 done in #3313: removed `IvcRollingState::from_previous_proof` along with its two unit tests (the only call sites in the workspace)

mithril-stm/src/circuits/halo2_ivc/tests/off_circuit/proof_verification.rs
jpraynaud, line 1: These tests should probably live in the `ivc_halo2_sark/proof` module.

mithril-stm/src/circuits/halo2_ivc/errors.rs
```diff
@@ -9,6 +9,22 @@ pub enum IvcCircuitError {
...
+    RecursiveProofTranscriptNotFullyConsumed,
+
+    /// KZG opening equations (dual MSM check) did not verify.
+    #[error("Recursive IVC proof rejected: KZG opening check failed")]
+    RecursiveProofKzgOpeningFailed,
+
+    /// Folded accumulator pairing equation did not verify.
+    #[error("Recursive IVC proof rejected: accumulator pairing check failed")]
+    RecursiveProofAccumulatorFailed,
+
```
jpraynaud: These error should live in an error module in the `ivc_halo2_snark` module.

mithril-stm/src/proof_system/halo2_snark/unsafe_helpers.rs
```diff
@@ -103,30 +103,11 @@ pub(crate) struct SnarkVerifierSetup {
...
-    134, 204, 16, 108, 51, 54, 245, 143, 236, 224, 30, 118, 109, 196, 20, 125, 56, 227, 25, 54, 16,
-    90, 73, 68, 203, 89,
-];
-
 impl SnarkVerifierSetup {
     /// Build the verifier setup from the embedded constant verifier params bytes.
     pub(crate) fn try_new() -> StmResult<Self> {
         let verifier_params = ParamsVerifierKZG::<Bls12>::read(
-            &mut &SNARK_VERIFIER_PARAMS_BYTES[..],
+            &mut &crate::proof_system::KZG_VERIFIER_PARAMS[..],
```
jpraynaud: Can you put the import of the constant at the top of the module?

mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
```diff
@@ -21,13 +39,80 @@ pub(crate) struct IvcProver<R: RngCore + CryptoRng> {
...
 /// IVC proof emitted at the end of a proving step.
+///
+/// `H` is the transcript hash used to produce this proof and must be used to verify it.
+/// It is a zero-cost phantom: no `H`-dependent data is stored, but it prevents accidentally
+/// verifying a Poseidon-produced proof via the Blake2b path and vice versa.
 // TODO: remove this allow dead_code directive when the IVC prover emits this proof
 #[allow(dead_code)]
-pub(crate) struct IvcProof {
-    /// Externally-verifiable proof bytes (Blake2b transcript).
+pub(crate) struct IvcProof<H> {
```
```suggestion
pub(crate) struct IvcProof<H: TranscriptHash> {
```

mithril-stm/src/proof_system/ivc_halo2_snark/rolling_state.rs
```diff
@@ -67,8 +67,8 @@ impl IvcRollingState {
     /// The chain state, proof bytes, and folded accumulator are pulled from the
     /// previous proof; the genesis signature is the chain-specific constant
     /// supplied once at session start.
-    pub(crate) fn from_previous_proof(
-        previous_proof: &IvcProof,
+    pub(crate) fn from_previous_proof<H>(
```
jpraynaud: Do we still need to keep this function which is not used at the moment?

mithril-stm/src/proof_system/ivc_halo2_snark/verifier_setup.rs
```diff
@@ -0,0 +1,148 @@
...
+/// The certificate and IVC verifying keys used to build this struct must match those used to
+/// build the [`Global`] passed to [`IvcProof::verify`]. A mismatch silently produces wrong
+/// public inputs and will cause verification to fail with
+/// [`IvcCircuitError::RecursiveProofKzgOpeningFailed`].
+///
+/// [`Global`]: crate::circuits::halo2_ivc::state::Global
+/// [`IvcProof::verify`]: crate::proof_system::ivc_halo2_snark::proof::IvcProof::verify
+// TODO: remove this allow dead_code directive when IvcVerifierSetup is wired into STM
+#[allow(dead_code)]
+pub(crate) struct IvcVerifierSetup {
```
curiecrypt, line 42: These fields are pub(crate) and the tests just build the struct directly (see proof_verification.rs:198 and :232), so none of the constructors actually get called anywhere. I grepped and `try_new`, `from_ivc_setup`, and `from_ivc_setup_with_srs` have zero callers. That means the cert+ivc fixed-base merge in `try_new` never runs in any test, even though that's the key-consistency logic you added in commit 6. So the `allow(dead_code)` here is really hiding three untested constructors, not just a type that is waiting to be wired up.

Also building by literal lets you pair `verifier_params` with a `tau_g2` from a different source, which is exactly what the constructors are trying to prevent by deriving `tau_g2 = s_g2()` together.

Could we make the fields private and route the tests through `from_ivc_setup_with_srs`? Then the merge logic gets exercised end to end. If some test genuinely needs to plug in the embedded asset parts by hand, a small `#[cfg(test)] from_parts()` would at least make that explicit.

mithril-stm/src/proof_system/mod.rs
```diff
@@ -21,7 +21,29 @@ mod concatenation;
...
 #[cfg(feature = "future_snark")]
-mod ivc_halo2_snark;
+/// Serialized `ParamsVerifierKZG` (i.e. `s_g2`) from the Midnight production trusted SRS.
+///
+/// `ParamsVerifierKZG` serializes only `s_g2` — a single G2 point from the power-of-tau
+/// ceremony. This value is independent of the circuit degree `k`, so all circuits
+/// (non-recursive and IVC) share it.
+///
+/// Regenerate by running the `verifier_setup_matches_trusted_srs` test, serializing
+/// `srs.verifier_params()` with `SerdeFormat::RawBytesUnchecked`.
```
curiecrypt: I think the following would be more accurate?
```suggestion
/// Regenerate by extracting `srs.verifier_params()` from `TrustedSetupProvider` and
/// serializing with `SerdeFormat::RawBytesUnchecked`; the `verifier_setup_matches_trusted_srs`
/// test (in `halo2_snark`) asserts the embedded bytes still match the trusted SRS.

```

mithril-stm/src/circuits/halo2_ivc/tests/off_circuit/proof_verification.rs
```diff
@@ -94,3 +126,135 @@ fn chain_state_proof_passes_dual_msm_check_and_accumulator_check() {
...
+        proof.verify(&global, &verifier_setup).is_err(),
+        "state from a different proof should be rejected by IvcProof::verify"
+    );
+}
+
+#[test]
+fn ivc_proof_verify_rejects_mismatched_accumulator() {
+    // Substituting the accumulator from a different proof step corrupts the public
+    // inputs (accumulator is part of them) and also fails the pairing equation
+    // check, so `verify` must return `Err` on both counts.
```
curiecrypt: The comment says it fails "on both counts," but this can't reach the accumulator check. The accumulator is folded into `public_inputs`, so a mismatched one fails at `dual_msm.check` first and returns `RecursiveProofKzgOpeningFailed`.  `RecursiveProofAccumulatorFailed` is never hit by any test. 
Two things: 
- could we switch the four negatives from `.is_err()` to `expect_err()` + a downcast on the specific variant? Right now an asset-load or setup error would pass too, so a broken verify could stay green.  
- could we add a case that actually trips the accumulator branch, e.g. a `verifier_setup` with a mismatched `tau_g2` that still passes the KZG check? If that's not feasible yet, worth noting as a known gap rather than leaving this test implying it is covered.

mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
```diff
@@ -21,13 +39,80 @@ pub(crate) struct IvcProver<R: RngCore + CryptoRng> {
...
+
+        let mut transcript = CircuitTranscript::<H>::init_from_bytes(self.proof_bytes.as_bytes());
+
+        let dual_msm = prepare::<CircuitBase, KZGCommitmentScheme<Bls12>, CircuitTranscript<H>>(
+            &verifier_setup.ivc_verifying_key,
+            &[&[G1Projective::identity()]],
+            &[&[&public_inputs]],
+            &mut transcript,
+        )
+        .map_err(|_| IvcCircuitError::RecursiveProofTranscriptPreparationFailed)?;
```
curiecrypt: Nit: here we drop the underlying PlonkError, this is fine to leave, but a `.with_context(|| ...)` would keep the source in the chain for debugging. Matches what `SnarkProof::verify` already does, so no strong opinion.

mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
```diff
@@ -21,13 +39,80 @@ pub(crate) struct IvcProver<R: RngCore + CryptoRng> {
...
-    /// Externally-verifiable proof bytes (Blake2b transcript).
+pub(crate) struct IvcProof<H> {
+    /// Externally-verifiable proof bytes.
     pub(crate) proof_bytes: IvcProofBytes,
     /// Chain state the proof commits to.
     pub(crate) state: State,
     /// Folded accumulator the proof commits to.
     pub(crate) accumulator: Accumulator<BlstrsEmulation>,
+    /// Phantom marker tying the proof to its transcript hash type.
+    pub(crate) _hash: PhantomData<H>,
```
curiecrypt: why `_hash` here? we can remove the underscore?

mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
```diff
@@ -21,13 +39,80 @@ pub(crate) struct IvcProver<R: RngCore + CryptoRng> {
 }
 
 /// IVC proof emitted at the end of a proving step.
+///
+/// `H` is the transcript hash used to produce this proof and must be used to verify it.
+/// It is a zero-cost phantom: no `H`-dependent data is stored, but it prevents accidentally
```
curiecrypt, line 45: Related to my comment on verifier setup:
Could we add a `pub(crate) fn new(...)` that the prover emits through with the same `H` it used, and have the tests go through that instead of the literal? 
Then the phantom actually means what the comment says.

mithril-stm/src/proof_system/ivc_halo2_snark/rolling_state.rs
curiecrypt: this is removed in pr 3313

mithril-stm/src/circuits/halo2_ivc/errors.rs
perturbing: Fixed in c15ccfe7f6bd3e7405a4fd673ef6e7fd1c19b44e

mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
perturbing: Fixed in 45a23dc761193cffbac9a727e6e697fde76917cb

mithril-stm/src/circuits/halo2_ivc/tests/off_circuit/proof_verification.rs
perturbing: Fixed in b2d54976e4af2050bba32e44ed008effc4bae934

mithril-stm/src/proof_system/halo2_snark/unsafe_helpers.rs
perturbing: Fixed in https://github.com/input-output-hk/mithril/pull/3315/changes/a961205b64b40c4a0c84a258fb3a3ebb7133f8be

mithril-stm/src/proof_system/ivc_halo2_snark/verifier_setup.rs
perturbing: Fixed in [57e4e0dfa1033b2fa799c1b8359b930e5adcdf46](https://github.com/input-output-hk/mithril/pull/3315/changes/a2faff500b697a3f50d9d2a1c9c813e06c0c6e55)

mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
perturbing: Convention https://doc.rust-lang.org/rust-by-example/generics/phantom.html

(Note that AI suggested this for me). Shall I remove it?

mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
perturbing: fixed in https://github.com/input-output-hk/mithril/pull/3315/changes/13679d68909ccc6a294d41600812faef62bccd80

mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
perturbing: Changed it in https://github.com/input-output-hk/mithril/pull/3315/changes/13679d68909ccc6a294d41600812faef62bccd80

mithril-stm/src/circuits/halo2_ivc/tests/off_circuit/proof_verification.rs
perturbing: Thanks, fixed in 677aced3973634d6ffbcb36cc52f25c4d95ba2f5

mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
```diff
@@ -21,13 +40,322 @@ pub(crate) struct IvcProver<R: RngCore + CryptoRng> {
...
+/// verifying a Poseidon-produced proof via the Blake2b path and vice versa.
 // TODO: remove this allow dead_code directive when the IVC prover emits this proof
 #[allow(dead_code)]
-pub(crate) struct IvcProof {
-    /// Externally-verifiable proof bytes (Blake2b transcript).
-    pub(crate) proof_bytes: IvcProofBytes,
+pub(crate) struct IvcProof<H: TranscriptHash> {
+    /// Externally-verifiable proof bytes.
```
hjeljeli32, line 50: The type is now generic over H, so it can also represent a Poseidon proof.  “IVC proof bytes produced with transcript hash H” would be clearer.

hjeljeli32 (approved):
LGTM 🚀

mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
perturbing: What is your suggestion? To change the comment? We now have
```rust
/// IVC proof emitted at the end of a proving step.
///
/// `H` is the transcript hash used to produce this proof and must be used to verify it.
/// It is a zero-cost phantom: no `H`-dependent data is stored, but it prevents accidentally
/// verifying a Poseidon-produced proof via the Blake2b path and vice versa.
// TODO: remove this allow dead_code directive when the IVC prover emits this proof
#[allow(dead_code)]
pub(crate) struct IvcProof<H: TranscriptHash> {
    /// Externally-verifiable proof bytes.
    proof_bytes: IvcProofBytes,
    /// Chain state the proof commits to.
    state: State,
    /// Folded accumulator the proof commits to.
    accumulator: Accumulator<BlstrsEmulation>,
    /// Phantom marker tying the proof to its transcript hash type.
    hash: PhantomData<H>,
}
```
Note that the struct doc does include this "IVC proof bytes produced with transcript hash H"

mithril-stm/src/proof_system/ivc_halo2_snark/verifier_setup.rs
```diff
@@ -0,0 +1,205 @@
...
+        &self.tau_g2
+    }
+
+    pub(crate) fn ivc_verifying_key(&self) -> &CircuitVerifyingKey {
+        &self.ivc_verifying_key
+    }
+
+    pub(crate) fn combined_fixed_bases(&self) -> &BTreeMap<String, G1Projective> {
+        &self.combined_fixed_bases
+    }
```
jpraynaud, line 166: Even if these functions are only `pub(crate)`, it'd be nice to have doc comments for them as well.

jpraynaud (approved):
LGTM 🚀

mithril-stm/src/proof_system/ivc_halo2_snark/verifier_setup.rs
perturbing: Fixed in https://github.com/input-output-hk/mithril/pull/3315/changes/7f5e41da9fbb2e36336117009123c248a42fc5ec

curiecrypt (approved):
LGTM 👍

PR #3316, opened 2026-06-08
refactor(end-to-end): promote scenario to commands, switch epoch length based wait duration, rework and cleanup assertions

Alenar:
## Content

This PR includes three major changes to the end-to-end runners:

- Scenario are now clap sub-commands, meaning to run run-only you need to pass `run-only` instead of `--run-only` and it have to be passed at the end of the command
- All wait have been changed from hardcoded values to multiple for epoch length based duration like "a tenth of an epoch" (computed from the `cardano_slot_length` and `cardano_epoch_length` args)
- The `assertions` module, which contains shared, have been renamed to `toolkit` and overhauled from an list of static method into a tree of structures (split by subdomains) to allow them to hold a context.

### Additional details

About the command line arguments:
- `Args` struct have been renamed to `Cli` to avoid a collision with `clap::Args`
- most of the `Cli` values they have been split into multiple structures:
  - `NetworkTopologyArgs`: number of aggregators, signers, do we use the mithril relay, the dmq, ...
  - `MithrilArgs`: Mithril state machines run interval, the start era, ...
  - `CardanoDevnetArgs`: Devnet scripts location, Cardano node version, slot/epoch lenght, ...
  - they are flattened into the `Cli` struct, meaning there's no functional change for them for users

About scenario:
- their "scripts" have been moved to a new `scenario` folder module
- if no scenario command subcommand is given, `Full` is used
- some arguments have been moved from the main argument list to the subcommands:
  - to `Full`: `check_client_cli_snapshot_converter` has it was only used by that scenario, `signed_entity_types` has the upcoming `minimal` scenario won't allow for more than one
  - to `RunOnly`: `signed_entity_types` same cause as in the `Full` scenario

About the refactor of the assertion module into `toolkit`:
- scoping `check` and `wait` into struct allowed named simplification (e.g. no need to repeat "check" in method names has you will use them through their structures: `toolkit.check.certificate.is_creating_certificate_with_enough_signers` instead of `toolkit.check.certificate.check_is_creating_certificate_with_enough_signers`)
- new `wait_for_latest_artifact` utility method was added to remove duplication between `wait_for_artifact` methods in each signed entity type dedicated check toolkit 
- To avoid duplication with the upcoming `minimal` scenario, check steps of each signed entities have been aggregated into `is_certified_and_verified` methods
- Methods that were fetching artifact individually based on a hash have been removed, as their value was low, the only additional check that they were doing (asserting a minimal value for artifact epoch) have been transferred to new `check_artifact` methods
- `check_artifact` naming is willingly broad to allow future addition of additional checks

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation
  - [x] No new TODOs introduced

## Issue(s)

Relates to #3151


turmelclem (approved):
LGTM 👍

.github/workflows/backward-compatibility.yml
```diff
@@ -155,6 +155,7 @@ jobs:
               --cardano-node-version ${{ matrix.cardano_node_version }} \
               --cardano-slot-length 0.25 \
               --cardano-epoch-length 45.0 \
+              full \
```
jpraynaud, line 158: This will likely break runs with previous versions of the e2e. Maybe we should keep the `full` implicit here?

mithril-test-lab/mithril-end-to-end/src/main.rs
```diff
@@ -19,18 +19,24 @@ use tokio::{
...
 
+/// Default signed entity types used by scenarios that support multiple entities, such as Full and RunOnly.
+const DEFAULT_SIGNED_ENTITY_TYPES: &str =
+    "CardanoTransactions,CardanoBlocksTransactions,CardanoStakeDistribution,CardanoDatabase";
+
 /// Tests args
 #[derive(Parser, Debug, Clone)]
-pub struct Args {
-    /// Available commands
+pub struct Cli {
```
jpraynaud, line 36: It'd be nice to print the groups of parsed args at the beginning of the e2e test.

jpraynaud (approved):
LGTM 🚀

.github/workflows/backward-compatibility.yml
Alenar: Good catch but we can't make it implicit: the `--signed-entity-types` that come right after is now an argument of the `full` subcommand, if `full` is missing then clap will raise an error as it will be considered "unknown".

What can we do ?

- checking the runner version in this step to adapt the parameters (adding significant complexity which could be offset by using a script)
- remove the `--signed-entity-types` parameter, but this would reduce the workflow usefulness
- nothing and instead use the "Use workflow from (<branch/tag>)" parameter already available when triggering a workflow manually

I think that doing nothing is fine, with the backcompat rules integrated into the e2e runner we do not need that much to use previous versions.

jpraynaud: OK, let's use the appropriate branch/tag for earlier versions of the e2e :+1:

mithril-test-lab/mithril-end-to-end/src/main.rs
Alenar: Done, FYI if no scenario is given the log will print `scenario: None`, it's possible to print the default fallback but it would requires more complex code.

PR #3317, opened 2026-06-08
Package only one ledger state snapshot for Cardano node version above 10.4

### LINKED ISSUE #3298: Update ledger state snapshot selection in artifacts
## Why
Today we package the last two ledger state snapshots to make sure that we can fast bootstrap the Cardano node with the oldest one if the most recent is incomplete. We want to use a different selection rule that will be more efficient to capture the most recent complete snapshot and avoid serving two ledger state snapshots.

## What
Include only the most recent complete ledger state snapshot in the ancillary files.

## How

- [x] For a Cardano node version above `10.4`, package only one ledger state snapshot (which is complete)
- [x] For a Cardano node version strictly below `10.4`, package the two most recent ledger state snapshots
- [ ] Make sure that there is no flakiness introduced in the fast bootstrap for 7 days:
  - [ ] Daily manual tests
  - [ ] No failures in the nightly tests

turmelclem:
## Content

Package only one ledger state snapshot (instead of two) for Cardano node version above 10.4 (version introducing support of UTxO-HD) 

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation
  - [x] No new TODOs introduced

## Issue(s)

this PR relates to or closes #3298

### DIFF
--- a/mithril-aggregator/src/artifact_builder/cardano_database.rs
+++ b/mithril-aggregator/src/artifact_builder/cardano_database.rs
@@ -131,6 +131,8 @@ mod tests {
 
     use super::*;
 
+    const DUMMY_CARDANO_NODE_VERSION: Version = Version::new(1, 2, 3);
+
     fn get_test_directory(dir_name: &str) -> PathBuf {
         TempDir::create("cardano_database", dir_name)
     }
@@ -195,6 +197,7 @@ mod tests {
 
         let snapshotter = Arc::new(
             CompressedArchiveSnapshotter::new(
+                DUMMY_CARDANO_NODE_VERSION,
                 cardano_db.get_dir().to_path_buf(),
                 test_dir.join("ongoing_snapshots"),
                 CompressionAlgorithm::Gzip,
--- a/mithril-aggregator/src/artifact_builder/cardano_database_artifacts/immutable.rs
+++ b/mithril-aggregator/src/artifact_builder/cardano_database_artifacts/immutable.rs
@@ -282,6 +282,7 @@ mod tests {
         entities::TemplateUri,
         test::{TempDir, assert_equivalent, equivalent_to},
     };
+    use semver::Version;
 
     use crate::services::ancillary_signer::MockAncillarySigner;
     use crate::services::{CompressedArchiveSnapshotter, DumbSnapshotter, MockSnapshotter};
@@ -289,6 +290,8 @@ mod tests {
     use crate::tools::file_archiver::FileArchiver;
 
     use super::*;
+
+    const DUMMY_CARDANO_NODE_VERSION: Version = Version::new(1, 2, 3);
 
     fn fake_uploader(
         archive_paths: Vec<&str>,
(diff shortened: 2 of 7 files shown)


mithril-aggregator/src/artifact_builder/cardano_database_artifacts/immutable.rs
```diff
@@ -290,6 +291,8 @@ mod tests {
 
     use super::*;
 
+    const CARDANO_NODE_VERSION_WITHOUT_HTXO_HD_SUPPORT: Version = Version::new(10, 4, 0);
```
Alenar: This bring more context than needed into those tests, you can keep that constant but name it in a way that say "do not worry, it does not matter here". 
```suggestion
    const DUMMY_CARDANO_NODE_VERSION: Version = Version::new(1, 2, 3);
```

mithril-aggregator/src/artifact_builder/cardano_database.rs
```diff
@@ -131,6 +131,8 @@ mod tests {
 
     use super::*;
 
+    const CARDANO_NODE_VERSION_WITHOUT_HTXO_HD_SUPPORT: Version = Version::new(10, 4, 0);
```
Alenar: This bring more context than needed into those tests, you can keep that constant but name it in a way that say "do not worry, it does not matter here". 
```suggestion
    const DUMMY_CARDANO_NODE_VERSION: Version = Version::new(1, 2, 3);
```

mithril-aggregator/src/dependency_injection/builder/protocol/artifacts.rs
```diff
@@ -143,7 +143,14 @@ impl DependenciesBuilder {
                 let ongoing_snapshot_directory =
                     self.configuration.get_snapshot_dir()?.join("pending_snapshot");
 
+                let cardano_node_version = Version::parse(&self.configuration.cardano_node_version())
+                    .map_err(|e| DependenciesBuilderError::Initialization {
+                        message: format!("Could not parse configuration setting 'cardano_node_version' value '{}' as Semver.", self.configuration.cardano_node_version()),
+                        error: Some(e.into()),
+                    })?;
+
```
Alenar: This duplicate the same parsing and error handling above (in `build_signed_entity_service`).

It's fine at is but I would rather add a `parse_cardano_node_version` method on the `DependenciesBuilder` struct in the `mithril-aggregator/src/dependency_injection/builder/mod.rs` file.

An alternative would be to add the `parse_cardano_node_version` method directly as a blanket on the `ConfigurationSource` trait, something like this:

```rust
    fn parsed_cardano_node_version(&self) -> StdResult<semver::Version> {
        Version::parse(&self.cardano_node_version())
           .with_context(||format!("Could not parse configuration setting 'cardano_node_version' value '{}' as Semver.", self.configuration.cardano_node_version()) )
    }
```

Alenar (approved):
LGTM

jpraynaud (approved):
LGTM 👍 

Can you switch back to draft mode until we can merge it (i.e. after the cut for the next distribution)?

PR #3325, opened 2026-06-11
feat(ci): e2e light scenario for snark running in nightly ci

### LINKED ISSUE #3151: Optimize e2e tests for SNARK

Alenar:
## Content

This PR includes a new `minimal` scenario in the end-to-end runner and a `test-e2e` workflow which is run in the nightly dispatcher.

### `mithril-end-to-end` changes 

- New scenario: `minimal`, 
  - A light scenario to be used with compute intensive mode such as with Snark as it needs significantly less epochs to finish
  - differences compared to the `full` scenario:
    - it does not check propagation of protocol parameters and stakes distribution changes 
    - it does not support era switching
    - it only support activation of a single signed entity type (in addition of the default `MithrilStakeDistribution`)
  - What does not changes compared to the `full` scenario:
    - it run the same check procedures against activated signed entity types
    - it support multiple aggregators

- New check for `full` and `minimal`: `is_creating_certificate_with_min_epoch`, run before any artifact validation to ensure that the aggregator certificate chain is still running and is not in a "epoch gap" state (this is better than finding it when checking mithril stake distributions).

- enhanced `attempt!` macro to support providing a predicate to check the returned value and wait if it's not satisfied.

- better type enforcement by using `AggregateSignatureType` instead of string where applicable

### CI changes

- new `test-e2e.yml` workflow
  - it support both workflow call and dispatch
  - key differences from the e2e integrated in `ci.yml`:
    - it build the runner and the nodes itself (no-reuse of built artifact)
    - it by default build with `--features future_snark`
    - it run the two scenario `minimal` and `full`
    - scenarios are run with slower network (1s slot length, 60 slot per epoch vs 0.1 and 30)

- integrate `test-e2e.yml` to the nightly dispatcher

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation
  - ~~No new TODOs introduced~~ One todo is added to advise to add `Snark` case in the new `test-e2e` workflow when they are ready

## Issue(s)

Closes #3151

### DIFF
--- a/.github/workflows/nightly-dispatcher.yml
+++ b/.github/workflows/nightly-dispatcher.yml
@@ -23,9 +23,18 @@ jobs:
   test-client:
     uses: ./.github/workflows/test-client.yml
 
+  test-e2e:
+    uses: ./.github/workflows/test-e2e.yml
+
   notify-on-failure:
     uses: ./.github/workflows/test-notify-on-failure.yml
     needs:
-      [backward-compatibility, aggregator-stress-test, test-client, test-rust]
+      [
+        backward-compatibility,
+        aggregator-stress-test,
+        test-client,
+        test-rust,
+        test-e2e,
+      ]
     if: failure()
     secrets: inherit
--- a/.github/workflows/test-e2e.yml
+++ b/.github/workflows/test-e2e.yml
@@ -1,0 +1,284 @@
+name: End-to-end tests
+
+on:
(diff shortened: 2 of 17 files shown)


jpraynaud (approved):
LGTM 👍

turmelclem (approved):
LGTM 👍

PR #3326, opened 2026-06-11
refactor(stm): Refactor prover input preparation

curiecrypt:
## Content

Refactors `IvcProverInput::prepare` for readability and faster test execution. 
- **Split `prepare` into five free helpers** in a new `prover_input_helpers.rs` module: 
   - `determine_transition`, 
   - `verify_certificate_proof`, 
   - `build_snark_message_and_decode_fields`, 
   - `build_next_state`, 
   - `build_next_accumulator`. 
- Genesis dispatches to a new `prepare_genesis` impl method.
- **Removed the `OnceLock` cache** by consolidating the four slow integration scenarios into one `prepare_integration_scenarios` test that builds `IvcSetup` once into a local. 
- **Renamed variables for temporality clarity**, including 
   - `chain_epoch` to `last_committed_epoch`, 
   - `certificate_epoch` to `incoming_certificate_epoch`, 
   - `dual_msm` to `certificate_dual_msm`, 
   - `snark_proof` to `certificate_proof`, 
   - `message` to `certificate_message_bytes`, 
   - `certificate_message` to `certificate_message_hash`. 
- Error struct fields renamed to match. The naming now separates SNARK-layer vocabulary (types, byte forms) from Mithril certificate-layer vocabulary (roles, decoded values).
- **Exposed `transition_type` on `IvcProverInput`** so downstream consumers can branch on the step's classification without re-running `determine_transition`.

Latest changes:
- split `try_compute` into `try_compute_transition_type` (epoch classification)  and `assert_correct_parameters ` (commitment and parameter validation)
- Genesis handling was moved inline into `proof.rs`

## Pre-submit checklist

- Branch
  - [x] Tests are provided (if possible)
  - [ ] Crates versions are updated (if relevant)
  - [ ] CHANGELOG file is updated (if relevant)
  - [x] Commit sequence broadly makes sense
  - [x] Key commits have useful messages
- PR
  - [x] All check jobs of the CI have succeeded
  - [x] Self-reviewed the diff
  - [x] Useful pull request description
  - [x] Reviewer requested

## Issue(s)

Closes #3319

Issue #3319, opened 2026-06-10
Recursive SNARK aggregation primitives: Refactor preparation of prover input


jpraynaud:
## Why
We need to refactor the preparation of the IVC prover input to make the code more readable and maintainable. Also we need to accelerate the tests which are too long.

## What
Refactor the IVC prover for readability and maintainability.

## How

- [x] Split the `prepare<D: MembershipDigest>` function in logical sub-functions which are unit tested
- [x] Make sure that variable names are clear enough to understand what temporality they represent
- [x] Remove the `OnceLock` cache which is not efficient with tests execution parallelism
- [x] Update the [SNARK Book](https://hackmd.io/EeCC7vaPQoWdmMVAkyDQpQ?view#IVC-Prover-Input-Preparation) to reflect the changes made

curiecrypt:
See https://github.com/input-output-hk/mithril/pull/3313#discussion_r3392268360 and https://github.com/input-output-hk/mithril/pull/3313#discussion_r3392308730

PR #3327, opened 2026-06-12
feat(mithril-stm): implement `IvcProver::prove`

### LINKED ISSUE #3139: Recursive SNARK aggregation primitives: Create SNARK proof with circuit

### LINKED ISSUE #3337: Move slow tests to slow tier
## Why
Some tests in the `halo2_snark` needs to be moved to the slow tests tier as they are above the threshold.

## What
Move new slow tests to the slow test tier.

## How 
- `proof_system::halo2_snark::proof::tests::fails_with_insufficient_signatures`  
- `proof_system::halo2_snark::proof::tests::fails_with_empty_signatures`

hjeljeli32:
## Content

This PR completes **Phase 1** and **Phase 2** of issue #3139 and incorporates follow-on review changes and CI optimizations.

### Changes

- **`IvcProvingSetup`** renamed to **`IvcProverSetup`** and extended to store the full KZG SRS
- **`IvcProver::prove`** redesigned: takes `rolling_state: Option<&IvcRollingState>` and `genesis_bootstrap: Option<IvcGenesisBootstrapInput>`; always returns `(IvcProof<blake2b_simd::State>, Option<IvcRollingState>)`
- **`IvcGenesisBootstrapInput`**: new struct carrying the genesis Schnorr signature and protocol-message preimage for the bootstrap path; will be populated from `AncillaryGenesisData` at the aggregator call site (issue #3141)
- **`run_genesis_step`**: private helper that runs the internal genesis IVC step before the first public Epoch 1 call
- **`prove_with_transcript`** moved to `IvcProof` as a private associated function
- **`IvcRollingState::is_genesis()`** and **`is_next_epoch(certificate_epoch)`** predicates added
- **`IvcProofError`** extended with `ProofGenerationFailed` and `InvalidProvingContext`
- **`IvcProof::verify`** added; adapted from @perturbing's work in #3315

### Tests

- **Three slow prover scenarios** in `proof_system/ivc_halo2_snark/proof.rs` covering bootstrap, next-epoch, and same-epoch paths; each verifies all produced proofs via `IvcProof::verify`
- **Slow tests use embedded assets**: certificate proofs are loaded from pre-generated binary assets instead of being regenerated on every run
- **Two-tier split**: bootstrap and same-epoch remain in `mod slow` (regular CI); next-epoch is promoted to `mod very_slow` and excluded from PR CI (~17 min additional saving when slow tests are triggered)
- `proof_system::halo2_snark::fails_with_insufficient_signatures` and `fails_with_empty_signatures`  moved to `mod slow` tier

### CI

- Slow-test filter extended to cover `proof_system::ivc_halo2_snark`
- IVC slow and very-slow test groups run serially (`max-threads = 1`) to avoid resource contention on CI runners
- `filter-slow-tests.sh` updated to exclude `very_slow` tests from the regular slow tier fragment

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation
  - [x] No new TODOs introduced

## Issue(s)
Closes https://github.com/input-output-hk/mithril/issues/3139
Closes https://github.com/input-output-hk/mithril/issues/3337


mithril-stm/src/circuits/halo2_ivc/errors.rs
```diff
@@ -46,9 +46,9 @@ pub enum IvcCircuitError {
     StepCounterOverflow { current: u64 },
 
     /// Off-circuit step transition: the certificate proof's embedded verifying key does
-    /// not match the certificate verifying key carried by `IvcSetup`.
+    /// not match the certificate verifying key carried by `IvcProvingSetup`.
```
jpraynaud: Maybe this should be named `IvcProverSetup` for alignment with the other types?

mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
```diff
@@ -1,33 +1,946 @@
...
+    }
+}
+
+/// Transcript-generic IVC proof generation helper.
+///
+/// Calls `create_proof` with the committed-instance layout the IVC circuit expects:
+/// `&[&[&[], public_inputs]]` (one circuit, one instance group, empty committed
+/// instance, then the field-element public inputs). Returns the finalised transcript
+/// bytes on success.
+fn prove_with_transcript<H: TranscriptHash>(
```
jpraynaud: This could be an associated function of `IvcProof`.

mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
```diff
@@ -1,33 +1,946 @@
...
+        message: &[u8],
+        aggregate_verification_key: &AggregateVerificationKeyForSnark<D>,
+        global: &Global,
+        protocol_message_preimage: &ProtocolMessagePreimage,
+        rolling_state: &IvcRollingState,
+    ) -> StmResult<(
+        Option<IvcProof<blake2b_simd::State>>,
+        Option<IvcRollingState>,
+    )> {
+        let is_genesis = rolling_state.state().step_counter == StepCounter::ZERO;
```
jpraynaud: This could be implemented in `IvcRollingState::is_genesis()` to avoid code duplication.

@curiecrypt this probably applies to your work on https://github.com/input-output-hk/mithril/issues/3319

mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
```diff
@@ -1,33 +1,946 @@
...
+        let chain_epoch = rolling_state.state().current_epoch;
+        let certificate_epoch = protocol_message_preimage.current_epoch();
+        let is_next_epoch =
+            certificate_epoch.as_field() == chain_epoch.as_field() + EpochNumber::new(1).as_field();
```
jpraynaud: Same here, this could be implemented in `IvcRollingState::is_next_epoch(certificate.epoch)`.

@curiecrypt this also probably applies to your work on https://github.com/input-output-hk/mithril/issues/3319

mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
```diff
@@ -1,33 +1,946 @@
...
+    /// genesis — [`IvcProverInput::prepare`] ignores them on the genesis path. A future
+    /// refactor may wrap them in `Option<IvcCertificateInput>`.
+    pub(crate) fn prove<D: MembershipDigest>(
+        &mut self,
+        snark_proof: SnarkProof<D>,
+        message: &[u8],
+        aggregate_verification_key: &AggregateVerificationKeyForSnark<D>,
+        global: &Global,
+        protocol_message_preimage: &ProtocolMessagePreimage,
+        rolling_state: &IvcRollingState,
```
jpraynaud: This would probably be better, like that you can check whether it is a genesis (and in that case use the `IvcRollingState::genesis()` function or return an error):
```suggestion
        rolling_state: Option<&IvcRollingState>,
```

mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
```diff
@@ -1,33 +1,946 @@
...
+    ) -> StmResult<(
+        Option<IvcProof<blake2b_simd::State>>,
```
jpraynaud: There should always be an `IvcProof` returned in case of success, otherwise we should return an error:
```suggestion
        IvcProof<blake2b_simd::State>,
```

mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
```diff
@@ -1,33 +1,946 @@
...
+    )> {
+        let is_genesis = rolling_state.state().step_counter == StepCounter::ZERO;
+        let chain_epoch = rolling_state.state().current_epoch;
+        let certificate_epoch = protocol_message_preimage.current_epoch();
+        let is_next_epoch =
+            certificate_epoch.as_field() == chain_epoch.as_field() + EpochNumber::new(1).as_field();
+
+        // Prepare the witness, next state, and folded next accumulator.
+        // prepare() borrows snark_proof; snark_proof is still owned afterward.
+        let input = IvcProverInput::prepare(
```
jpraynaud: For clarity and better distinguish it from public input:
```suggestion
        let prever_input = IvcProverInput::prepare(
```

mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
```diff
@@ -1,33 +1,946 @@
...
+    ///   nothing to hand back to an external verifier yet.
+    /// - **Same-epoch** (`cert_epoch == chain_epoch`): one Blake2b proof only. The rolling
+    ///   state is unchanged; the caller retains the existing [`IvcRollingState`].
+    /// - **Next-epoch** (`cert_epoch == chain_epoch + 1`): two proofs — a Poseidon proof
+    ///   that advances the rolling state, and a Blake2b proof returned to the caller.
+    ///
+    /// Note: `snark_proof`, `message`, and `aggregate_verification_key` are unused at
+    /// genesis — [`IvcProverInput::prepare`] ignores them on the genesis path. A future
+    /// refactor may wrap them in `Option<IvcCertificateInput>`.
+    pub(crate) fn prove<D: MembershipDigest>(
```
jpraynaud, line 241: @hjeljeli32 @curiecrypt @perturbing I feel that there is a difficulty here with the bootstrap of the recursion 🤔 

That is the way the creation of the certificates works:
- Epoch 0: the genesis certificate is created
- Epoch 1: the recursion is bootstrapped on the genesis
- Epoch 2+: the recursion continues from previous epoch

But the current implementation, it looks like the expectation is that the recursion is bootstrapped at Epoch 0.

We should maybe include this recursion bootstrap from genesis as a preliminary step of the proving or we will need to deal with that in the aggregate signature which is less natural. WDYT?

mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
```diff
@@ -1,33 +1,946 @@
...
+                input.next_accumulator.clone(),
+                rolling_state.genesis_signature(),
+            ))
+        } else {
+            None
+        };
+
+        // Genesis only seeds the rolling state; no Blake2b external proof is returned.
+        // Same-epoch and next-epoch steps produce a Blake2b proof for the external verifier.
+        let external_proof = if is_genesis {
```
```suggestion
        let proof = if is_genesis {
```

jpraynaud (commented):
Thanks @hjeljeli32 , overall this looks good 👍 

However, I have spotted a friction point in the way the recursion bootstraps from genesis and left a comment for that below.

mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
curiecrypt: Yes, as I told in Thursday meeting, I created an enum for transition and I will expose that with prepare function. (Either as an output or in the prover input struct itself)

mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
curiecrypt: yes

mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
hjeljeli32: @curiecrypt @perturbing 
Sure @jpraynaud I'll follow the "policy" for the certificate creation: 
- genesis cert = Schnorr signature only at Epoch 0, 
- first prove() at Epoch 1 bootstraps internally.

A couple of things to align on before implementing:

- **Output shape**: in that case the return type should be `(IvcProof<blake2b_simd::State>, Option<IvcRollingState>)` rather than bare `IvcProof`, since same-epoch already returns a Blake2b proof with no rolling-state advance. Genesis is the only branch currently returning `None` for the external proof, which your redesign moves inside.

- **Bootstrap input**: when `rolling_state = None`, the internal genesis step needs more than just the Schnorr signature. `prepare()` also uses `protocol_message_preimage` at genesis. So the bootstrap input would be something like:

```
struct IvcGenesisBootstrapInput {
    genesis_signature: StandardSchnorrSignature,
    genesis_protocol_message_preimage: ProtocolMessagePreimage,
}
```
How should this reach `prove()`? A second optional parameter alongside `rolling_state`, or an enum replacing the raw Option?

hjeljeli32: @jpraynaud 

One clarification on the `genesis_protocol_message_preimage`
In the current API, `prove()` takes `protocol_message_preimage` as a top-level parameter. At the genesis call, the **caller** naturally provides the genesis preimage there. `prepare_genesis()` uses it to set the lookahead fields in the genesis output state (`next_merkle_tree_commitment`, `next_protocol_parameters`, `current_epoch`).

Under the new design, the caller skips Epoch 0 entirely and calls `prove()` for the first time at Epoch 1, passing the Epoch 1 preimage. When `prove()` internally bootstraps the genesis step, it calls `prepare_genesis()` but the only preimage in scope is the Epoch 1 one. The genesis preimage loses its natural slot.

.config/nextest.toml
jpraynaud, line 1: Maybe we can keep it simple with only one `ivc-snark-sequential` group.

.github/workflows/scripts/filter-slow-tests.sh
```diff
@@ -41,8 +41,10 @@ generate_nextest_allowed_slow_filterset() {
       local module_prefix="${entry##*#}"
 
       if echo "$changed_files" | grep -q "${watch_path}"; then
-          # Uses glob notation from the nextest filterset DSL
-          local fragment="test(#${module_prefix}*slow::*)"
+          # Uses glob notation from the nextest filterset DSL. `very_slow`
+          # tests are excluded from the regular slow tier; they only run when
+          # the caller asks for all tests.
+          local fragment="(test(#${module_prefix}*slow::*) and not test(#*very_slow::*))"
```
jpraynaud: Let's keep only `slow` tests and narrow the conditions to trigger them in the declaration of `SLOW_MITHRIL_STM_TESTS` (adding a line for each sub-module should be enough): like that very slow tests will be triggered only when the associated module is modified (and will always be run in the nigjhtly tests).

mithril-stm/src/proof_system/ivc_halo2_snark/errors.rs
```diff
@@ -19,4 +19,16 @@ pub(crate) enum IvcProofError {
...
     #[error("IVC proof rejected: accumulator pairing check failed")]
     AccumulatorFailed,
+
+    /// The PLONK prover (`create_proof`) failed internally.
+    #[error("IVC proof generation failed: {0}")]
+    ProofGenerationFailed(String),
+
+    /// `prove()` was called with an invalid combination of `rolling_state` and
+    /// `genesis_bootstrap`: exactly one must be `Some`, and `rolling_state` must not carry a
+    /// genesis state (`step_counter == 0`). Use `genesis_bootstrap` for the first certificate.
```
jpraynaud: This looks too restrictive IMO.

mithril-stm/src/proof_system/ivc_halo2_snark/prover_setup.rs
jpraynaud, line 1: Can you rename the module to `prover_setup.rs`?

mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
```diff
@@ -2,43 +2,70 @@
...
     /// Randomness source used during proof generation.
     pub(crate) rng: R,
 }
 
+/// Bootstrap input for the first [`IvcProver::prove`] call in an IVC chain.
+///
+/// Passed as `genesis_bootstrap: Some(...)` when `rolling_state = None`. Supplies the
+/// data needed to run the internal genesis IVC step before processing the first certificate.
+///
+// TODO(#3141): Wire genesis bootstrap inputs into the aggregator call site.
```
jpraynaud: This TODO can be removed.

mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
```diff
@@ -137,6 +164,235 @@ where
...
+    /// Exactly one of `rolling_state` and `genesis_bootstrap` must be `Some`:
+    ///
+    /// - `rolling_state = Some(rs), genesis_bootstrap = None`: normal step. `rs` carries the
+    ///   previous step's output. The transition type (same-epoch / next-epoch) is determined
+    ///   from the certificate epoch vs the chain epoch recorded in `rs`.
+    /// - `rolling_state = None, genesis_bootstrap = Some(bootstrap)`: genesis bootstrap.
+    ///   Called at the first certificate (Epoch 1). Internally runs a genesis IVC step using
+    ///   `bootstrap.genesis_signature` and `bootstrap.genesis_protocol_message_preimage`,
+    ///   then immediately runs the Epoch 1 step with the supplied certificate inputs.
+    ///   Returns the Epoch 1 Blake2b proof and the updated rolling state.
```
jpraynaud: That seems too restrictive and would imply genesis detection in the caller. It's be better to always have a `IvcGenesisBootstrapInput` and an optional `Option<IvcRollingState>`.

mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
```diff
@@ -137,6 +164,235 @@ where
...
+    #[allow(clippy::too_many_arguments)]
+    pub(crate) fn prove<D: MembershipDigest>(
+        &mut self,
+        snark_proof: SnarkProof<D>,
+        message: &[u8],
+        aggregate_verification_key: &AggregateVerificationKeyForSnark<D>,
+        global: &Global,
+        protocol_message_preimage: &ProtocolMessagePreimage,
+        rolling_state: Option<&IvcRollingState>,
+        // TODO(#3141): Wire genesis_bootstrap into the aggregator call site.
```
jpraynaud: This TODO can be removed.

jpraynaud (approved):
LGTM 👍 

Except the modifications regarding the slow tests where narrower conditions for triggering them should replace the "very slow" tier.

mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
```diff
@@ -137,6 +164,235 @@ where
...
+        genesis_bootstrap: Option<IvcGenesisBootstrapInput>,
+    ) -> StmResult<(IvcProof<blake2b_simd::State>, Option<IvcRollingState>)> {
+        // Exactly one of rolling_state or genesis_bootstrap must be Some.
+        // rolling_state must not be a genesis state (step_counter == 0): the bootstrap path
+        // is the only supported entry point for the first certificate.
+        if rolling_state.is_some() == genesis_bootstrap.is_some()
+            || rolling_state.is_some_and(|rs| rs.is_genesis())
+        {
+            return Err(IvcProofError::InvalidProvingContext.into());
```
curiecrypt: Should we add a test `prove_rejects_invalid_context` to cover both-`Some`, both-`None`, and `Step-with-genesis-state`?

curiecrypt (approved):
LGTM. The genesis bootstrap flow and the two-transcript handling read cleanly.

Just a heads-up on coordination with #3326: I'll take care of the seam on my side after this merges. On rebase I'll switch prove to consume the `transition_type` that the refactored prepare exposes instead of recomputing `is_next_epoch`, route `try_compute` through your new `IvcRollingState` predicates so they stay the single source of truth, and absorb the prepare borrow, the `IvcSetup` rename, and the duplicate slow-test allowlist line. I also left one small optional suggestion about a fast negative test for the `InvalidProvingContext` guard.

damrobi (approved):
LGTM 👍

perturbing (approved):

PR #3291, opened 2026-05-28
feature: custom genesis verification key support in explorer

Alenar:
## Content

This PR modify the explorer to allow users to provide custom genesis verification key.

- Update redux store settings slice:
  - aggregators are now a list of object instead of a list of url
  - add a `genesisVerificationKey` field to aggregator
- Supersede `aggregators-list.js` with file auto-generated at build time based on the networks declared in `mithril-infra/configuration`
  - Notes: `dev-*` networks are excluded from auto-gen and aggregators are sorted (mainnet first then preprod, ...)
- Add a `Genesis verification key` field in the `AddAggregatorModal`
- Adapt url search query:
  - Change update of the url to add the genesis vkey (only if the aggregator is custom)
  - Adapt parsing of the aggregator to read the genesis vkey (if provided)
- Block verification of the certificate chain or certification of blocks and transactions if the current aggregator does not have a genesis vkey set
- Remove remote fetching of genesis vkeys (as they are not either embedded at built time or provided by users)

Also it updates the explorer and fixes lingering issues:
- update dependencies
- fix new eslint warnings, in particular a lot of `react-hooks/set-state-in-effect` which required significant changes for impacted components
- migrate all CommonJs export (`module.exports { ... }` ) to Es Module (`export { ... }`)
- Fix hydration warnings at startup by adding a loading state at startup while the redux store is loading

### Context:

- Extract of `aggregators-list.generated.js` (as it's gitignored)

```js
export const defaultAggregators = [
  {
    "name": "release-mainnet",
    "url": "https://aggregator.release-mainnet.api.mithril.network/aggregator",
    "genesisVerificationKey": "5b3139312c36362c3134302c3138352c3133382c31312c3233372c3230372c3235302c3134342c32372c322c3138382c33302c31322c38312c3135352c3230342c31302c3137392c37352c32332c3133382c3139362c3231372c352c31342c32302c35372c37392c33392c3137365d"
  },
  {
    "name": "release-preprod",
    "url": "https://aggregator.release-preprod.api.mithril.network/aggregator",
    "genesisVerificationKey": "5b3132372c37332c3132342c3136312c362c3133372c3133312c3231332c3230372c3131372c3139382c38352c3137362c3139392c3136322c3234312c36382c3132332c3131392c3134352c31332c3233322c3234332c34392c3232392c322c3234392c3230352c3230352c33392c3233352c34345d"
  },
];

export default defaultAggregators;
```

- Updated add aggregator modal:
image

[...]

PR #3336, opened 2026-06-15
fix(explorer): unresponsive UI

### LINKED ISSUE #3328: Switch of aggregator does not work in explorer
## Why
The switch of aggregator does not work in the unstable version of the explorer. 

## What
Investigate the problem and create a fix.

## How

- [x] Investigate the source of the problem: this is due to storage limit of the local storage of the browser
- [x] Create a fix

Alenar:
## Content

This PR fixes a cache issue in the explorer pools slice:
When saving retrieved pools from an aggregator, the `poolsSlice` reducer retrieved existing pools if any by calling  `poolsForAggregator` and passing to it the url string, but since the last refactor (#3291) this method expected an object instead of a string, making it returns null.

This lead to a cascading effect:
1.  The reducer would never re-use existing keys, saving each the whole pools list each time an aggregator if switched in the UI
2. This evergrowing state is saved into the local storage
3. Given that for mainnet the pools list is ~1MB, it would quickly hit the storage api browser limit (10Mb for firefox, [see mdn](https://developer.mozilla.org/en-US/docs/Web/API/Storage_API/Storage_quotas_and_eviction_criteria#web_storage))
4. When the storage api limit is hit, all changes sent to the redux store (including current aggregator, auto-refresh, ...) would fails because the configuration persistence to local storage fails

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation
  - [x] No new TODOs introduced

## Issue(s)

Closes #3328

### DIFF
--- a/.github/workflows/actions/deploy-terraform-infrastructure/action.yml
+++ b/.github/workflows/actions/deploy-terraform-infrastructure/action.yml
@@ -338,8 +338,11 @@ runs:
     - name: Init Terraform
       shell: bash
       working-directory: mithril-infra
+      env:
+        GOOGLE_APPLICATION_CREDENTIALS: ./google-application-credentials.json
+        TF_REGISTRY_CLIENT_TIMEOUT: 60
       run: |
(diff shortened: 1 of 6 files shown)


jpraynaud (approved):
LGTM 👍

PR #3342, opened 2026-06-17
Improve balance build/run time for future snark tests in CI

perturbing:
## Content

This PR targets to optimise the CI build / run time balance for the future snark slow tests.

That is, if we run the slow tests, there is a valuable time decrease in running the test by letting the compiler optimising the underlying used libs like `blst`/`ff` and others expensive crypto heave functions like FFT over the scalar field.

This optimisation at compile time increases build time, so empericly we need to check if the total time the test takes improves. Also, having such flags enabled for all tests is undesired, because they will definitly not decrease their run time. This is why we target only the future snark tests for the nightly and in PR's that change the circuit.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Comments

An overview of the benchmarks of the CI:

- 1. An initial clean run with no changes was forced by an empty commit + the `run-slow-tests` label
- 2. A run with `opt-level = 3` profile on the future-snark tests only + the `run-slow-tests` label
- 3. A run with `opt-level = 3` profile on the future-snark test + no label
- 4. A run with `opt-level = 3` profile on the future snark test and filter on which jobs to use it + no label
- 5. A run with `opt-level = 3` profile on the future snark test and filter on which jobs to use it + the `run-slow-tests` label

| Test | Build time future-snark test | Run time future-snark test | Total time of CI |
|---|---|---|---|
| [1. Initial clean run + label](https://github.com/input-output-hk/mithril/actions/runs/27678632115/job/81889890565) | 3m 19s | 2h 19m 18s | 2h 24m 6s |
| [2. With opt profile + label](https://github.com/input-output-hk/mithril/actions/runs/27688270378/job/81892321522) | 13m 31s (+307.5%) | 56m 1s (-59.8%) | 1h 13m 25s (-49.1%) |
| [3. With opt profile](https://github.com/input-output-hk/mithril/actions/runs/27694046049/job/81912355872) | 12m 56s | 1m 58s | 18m 42s |
| [4. With opt profile + changed filter](https://github.com/input-output-hk/mithril/actions/runs/27744714584/job/82080081138) | 2m 47s | 6m 42s | 11m 0s |
| [5. With opt profile + changed filter + label](https://github.com/input-output-hk/mithril/actions/runs/27745540394/job/82082801512) | 13m 25s | 55m 40s | 1h 13m 26s |

We see that in test 1 (no changes) the building time is fast for all tests, but the full future snark tests take a long time.

In test 2 and 3 we see what the `opt-level = 3` profile gains in time wrt test 1. Build time increase (with our without the label forcing the long test), and if the label is there, the long testing time is halved (outweighing the increased build time).

In test 4 and 5 we see that we can have our cake an eat it to; have best of both worlds. That is, when we do not use the label, the build is quick and the rest is quick since it skips the slow tests. And in the case we have the label, the build time increase but the decreased testing time outweights this (effectifly decreasing the total CI time).

Test 4 and 5 are ran over the current changes.

## Issue(s)

Relates to https://github.com/input-output-hk/mithril/issues/3343

### DIFF
--- a/.github/workflows/test-rust.yml
+++ b/.github/workflows/test-rust.yml
@@ -17,13 +17,19 @@ jobs:
         include:
           - os: ubuntu-24.04
             test-args: --features full,unstable --workspace
+            cargo-profile: ci-tests
           - os: ubuntu-24.04
             test-args: --features future_snark -p mithril-stm -p mithril-common -p mithril-aggregator -p mithril-signer -p mithril-relay
             artifact-suffix: -future-snark
+            # Dedicated profile that optimizes (opt-level 3) the crypto deps and `mithril-stm`
+            # so the slow SNARK tests run faster, at the cost of a longer (cache-amortized) build.
+            cargo-profile: ci-tests-snark
           - os: ubuntu-24.04-arm
             test-args: --features full,unstable --workspace
+            cargo-profile: ci-tests
           # Exclude nodes not officially supported on Windows and macOS (only mithril-client is supported)
           - os: macos-14
+            cargo-profile: ci-tests
             test-args: >
               --features full,unstable --workspace
               --exclude mithril-aggregator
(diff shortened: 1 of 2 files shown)


Alenar (approved):
LGTM

jpraynaud (approved):
LGTM 👍

PR #3348, opened 2026-06-19
feat(stm): wire IVC snark prover

### LINKED ISSUE #3141: Recursive SNARK aggregation primitives: Wire SNARK proof in aggregate signature

damrobi:
## Content

This PR includes the wiring of the IVC prove and verify function inside the `Clerk` and `AggregateSignature` structures. Those structures can now compute and represent a proof generated by the IVC circuit along with all the information necessary to continue the computations or to verify the proof.

Changes:
- `IvcRollingState` serialization: now derives Serialize/Deserialize.
- `IvcVerifierData`: New struct in `verifier_setup.rs` that contains the genesis message, genesis Schnorr verifying key, and the certificate and IVC circuit verifying keys needed to verify a proof chain.
- `AncillaryProverData` / `AncillaryVerifierData`: Enum wrappers around `IvcRollingState` and `IvcVerifierData` for now.
- `AggregateSignature::IvcSnark`: New variant wired through `verify` (and `batch_verify`).
- Clerk: Added `ivc_prover_input_preparation_and_prove` function to drive a single IVC step and implement the IVC variant of `Clerk::aggregate_signatures_with_type`.

The PR also contains some fixes for bugs that appeared during end-to-end testing. 

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation
  - [x] No new TODOs introduced

## Issue(s)

Closes #3141

### DIFF
--- a/mithril-common/src/protocol/ancillary.rs
+++ b/mithril-common/src/protocol/ancillary.rs
@@ -103,7 +103,7 @@ mod tests {
             protocol_message.rigid_preimage().as_slice()
         );
         assert_eq!(
-            genesis_data.genesis_message_preimage().0,
+            genesis_data.genesis_message_preimage().to_vec(),
             genesis.protocol_message.rigid_preimage().as_slice()
         );
         assert!(genesis_data.genesis_schnorr_signature().is_some());
--- a/mithril-common/src/protocol/ancillary.rs
+++ b/mithril-common/src/protocol/ancillary.rs
@@ -60,6 +60,8 @@ pub fn build_ancillary_proof_input(
 
 #[cfg(all(test, feature = "future_snark"))]
 mod tests {
+    use std::ops::Deref;
+
     use rand_chacha::ChaCha20Rng;
     use rand_core::SeedableRng;
 
@@ -103,7 +105,7 @@ mod tests {
             protocol_message.rigid_preimage().as_slice()
         );
         assert_eq!(
-            genesis_data.genesis_message_preimage().to_vec(),
+            genesis_data.genesis_message_preimage().deref(),
             genesis.protocol_message.rigid_preimage().as_slice()
         );
         assert!(genesis_data.genesis_schnorr_signature().is_some());
--- a/mithril-common/src/protocol/ancillary.rs
+++ b/mithril-common/src/protocol/ancillary.rs
@@ -103,7 +103,7 @@ mod tests {
             protocol_message.rigid_preimage().as_slice()
         );
         assert_eq!(
-            genesis_data.genesis_message_preimage(),
+            genesis_data.genesis_message_preimage().0,
             genesis.protocol_message.rigid_preimage().as_slice()
         );
         assert!(genesis_data.genesis_schnorr_signature().is_some());
--- a/mithril-stm/CHANGELOG.md
+++ b/mithril-stm/CHANGELOG.md
@@ -16,7 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `Clerk::aggregate_signatures_with_type` IVC variant is completed.
 - `AggregateSignature::verify` IVC variant is completed.
 - `IvcProof::verify` now also checks that the input message is the same as the one used to create the proof.
-- More fixed the how the circuit verification keys are handled and computed.
+- More fixes: downsizing the srs properly for the ivc keys, proper deserialization of `VerifyingKey`, carrying the `AncillaryVerifierData` in the certificate.
 
 ## 0.10.38 (06-17-2026)
 
--- a/mithril-stm/CHANGELOG.md
+++ b/mithril-stm/CHANGELOG.md
@@ -5,6 +5,19 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## 0.10.39 (06-24-2026)
+
+### Added
+
+- Added wiring of the IVC prover and verifier in the `Clerk` and `AggregateSignature`.
+
+### Changed
+
+- `Clerk::aggregate_signatures_with_type` IVC variant is completed.
+- `AggregateSignature::verify` IVC variant is completed.
+- `IvcProof::verify` now also checks that the input message is the same as the one used to create the proof.
+- More fixed the how the circuit verification keys are handled and computed.
+
 ## 0.10.38 (06-17-2026)
 
 ### Added
--- a/mithril-stm/src/circuits/halo2_ivc/certificate_proof.rs
+++ b/mithril-stm/src/circuits/halo2_ivc/certificate_proof.rs
@@ -7,6 +7,7 @@
 //! proof, so it can be wrapped into a certificate accumulator for IVC
 //! aggregation.
 
+use anyhow::Context;
 use group::Group;
 
 use midnight_circuits::hash::poseidon::PoseidonState;
@@ -55,14 +56,17 @@ pub(crate) fn verify_and_prepare_accumulator(
         &[&[public_inputs]],
         &mut transcript,
     )
-    .map_err(|_| IvcCircuitError::CertificateProofRejected)?;
+    .map_err(|_| IvcCircuitError::CertificateProofRejected)
+    .with_context(|| "Error during the preparation of the dual msm.")?;
 
     transcript
         .assert_empty()
-        .map_err(|_| IvcCircuitError::CertificateProofRejected)?;
+        .map_err(|_| IvcCircuitError::CertificateProofRejected)
+        .with_context(|| "Transcript is not empty after the dual msm prepare.")?;
 
     if !dual_msm.clone().check(verifier_params) {
-        return Err(IvcCircuitError::CertificateProofRejected.into());
+        return Err(IvcCircuitError::CertificateProofRejected)
+            .with_context(|| "Dual msm check failed");
     }
 
     Ok(dual_msm)
--- a/mithril-stm/src/circuits/halo2_ivc/certificate_proof.rs
+++ b/mithril-stm/src/circuits/halo2_ivc/certificate_proof.rs
@@ -43,6 +43,8 @@ pub(crate) fn verify_and_prepare_accumulator(
     let mut transcript =
         CircuitTranscript::<PoseidonState<CircuitBase>>::init_from_bytes(proof_bytes);
 
+    println!("Preparing the dual msm");
+
     let dual_msm = prepare::<
         CircuitBase,
         KZGCommitmentScheme<Bls12>,
@@ -57,13 +59,18 @@ pub(crate) fn verify_and_prepare_accumulator(
     )
     .map_err(|_| IvcCircuitError::CertificateProofRejected)?;
 
+    println!("Checking the transcript");
+
     transcript
         .assert_empty()
         .map_err(|_| IvcCircuitError::CertificateProofRejected)?;
 
+    println!("Checking the dualMSM");
+
     if !dual_msm.clone().check(verifier_params) {
         return Err(IvcCircuitError::CertificateProofRejected.into());
     }
+    println!("DONE Checking the dualMSM");
 
     Ok(dual_msm)
 }
--- a/mithril-stm/src/circuits/halo2_ivc/certificate_proof.rs
+++ b/mithril-stm/src/circuits/halo2_ivc/certificate_proof.rs
@@ -43,8 +43,6 @@ pub(crate) fn verify_and_prepare_accumulator(
     let mut transcript =
         CircuitTranscript::<PoseidonState<CircuitBase>>::init_from_bytes(proof_bytes);
 
-    println!("Preparing the dual msm");
-
     let dual_msm = prepare::<
         CircuitBase,
         KZGCommitmentScheme<Bls12>,
@@ -59,18 +57,13 @@ pub(crate) fn verify_and_prepare_accumulator(
     )
     .map_err(|_| IvcCircuitError::CertificateProofRejected)?;
 
-    println!("Checking the transcript");
-
     transcript
         .assert_empty()
         .map_err(|_| IvcCircuitError::CertificateProofRejected)?;
 
-    println!("Checking the dualMSM");
-
     if !dual_msm.clone().check(verifier_params) {
         return Err(IvcCircuitError::CertificateProofRejected.into());
     }
-    println!("DONE Checking the dualMSM");
 
     Ok(dual_msm)
 }
--- a/mithril-stm/src/proof_system/ivc_halo2_snark/errors.rs
+++ b/mithril-stm/src/proof_system/ivc_halo2_snark/errors.rs
@@ -33,9 +33,9 @@ pub(crate) enum IvcProofError {
     )]
     InvalidProvingContext,
 
-    /// Mismatch between the protocol messages
+    /// Mismatch between the messages
     #[error(
-        "IVC proof rejected: the protocol message used to create the proof is different from the input message"
+        "IVC proof rejected: the message used to create the proof is different from the input message"
     )]
-    InvalidProtocolMessage,
+    InvalidMessage,
 }
--- a/mithril-stm/src/proof_system/ivc_halo2_snark/errors.rs
+++ b/mithril-stm/src/proof_system/ivc_halo2_snark/errors.rs
@@ -32,4 +32,10 @@ pub(crate) enum IvcProofError {
         "IVC prover called with invalid context: rolling_state must not be a genesis (step_counter == 0) state; pass None to bootstrap from genesis"
     )]
     InvalidProvingContext,
+
+    /// Mismatch between the protocol messages
+    #[error(
+        "IVC proof rejected: the protocol message used to create the proof is different from the input message"
+    )]
+    InvalidProtocolMessage,
 }
--- a/mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
+++ b/mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
@@ -835,6 +835,11 @@ mod tests {
             );
             let epoch1_preimage = wrap_protocol_message_preimage(&first_step.message_preimage);
             let bootstrap = genesis_bootstrap(&ctx.asset_setup);
+            println!(
+                "Genesis message preimage: {:?}",
+                bootstrap.genesis_protocol_message_preimage
+            );
+            println!("epoch1_preimage: {:?}", epoch1_preimage);
 
             let mut prover = IvcProver {
                 ivc_setup: Arc::clone(&ctx.ivc_setup),
@@ -960,6 +965,8 @@ mod tests {
             );
             let preimage = wrap_protocol_message_preimage(&step.message_preimage);
 
+            println!("preimage next certificate: {:?}", preimage);
+
             let mut prover = IvcProver {
                 ivc_setup: Arc::clone(&ctx.ivc_setup),
                 rng: OsRng,
@@ -1051,8 +1058,8 @@ mod tests {
                 verification_context,
             };
 
-            run_bootstrap_path(&ctx);
-            run_next_epoch_path(&ctx);
+            // run_bootstrap_path(&ctx);
+            // run_next_epoch_path(&ctx);
             run_same_epoch_path(&ctx);
         }
 
--- a/mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
+++ b/mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
@@ -551,6 +551,27 @@ mod tests {
     }
 
     #[test]
+    fn ivc_proof_to_from_bytes_round_trip() {
+        let step_output = load_embedded_next_epoch_step_output_asset()
+            .expect("recursive step output asset should load");
+
+        let proof = IvcProof::<blake2b_simd::State>::new(
+            step_output.ivc_proof,
+            step_output.next_state,
+            step_output.next_accumulator,
+        );
+
+        let bytes = proof.to_bytes().expect("serialization should not fail");
+        let restored = IvcProof::<blake2b_simd::State>::from_bytes(&bytes)
+            .expect("deserialization should not fail");
+
+        assert_eq!(
+            bytes,
+            restored.to_bytes().expect("re-serialization should not fail")
+        );
+    }
+
+    #[test]
     fn ivc_proof_message_verification_accepts_correct_message() {
         let step_output = load_embedded_next_epoch_step_output_asset()
             .expect("recursive step output asset should load");
@@ -674,7 +695,35 @@ mod tests {
     }
 
     #[test]
-    fn ivc_proof_verify_rejects_tampered_message_bytes() {
+    fn ivc_proof_verify_rejects_tampered_message_bytes_with_correct_input_message() {
+        let (global, verifier_setup) = build_proof_verifier_context();
+        let mut step_output = load_embedded_next_epoch_step_output_asset()
+            .expect("recursive step output asset should load");
+
+        let msg = &[
+            22, 148, 87, 37, 149, 0, 124, 10, 156, 94, 108, 6, 78, 59, 239, 80, 126, 213, 158, 211,
+            191, 213, 128, 70, 128, 30, 235, 80, 192, 191, 159, 67,
+        ];
+        step_output.next_state.message = MessageHash::ZERO;
+
+        let proof = IvcProof::<blake2b_simd::State>::new(
+            step_output.ivc_proof,
+            step_output.next_state,
+            step_output.next_accumulator,
+        );
+
+        let err = proof
+            .verify(msg, &global, &verifier_setup)
+            .expect_err("different protocol message should be rejected by IvcProof::verify");
+        assert_eq!(
+            err.downcast_ref::<IvcProofError>(),
+            Some(&IvcProofError::KzgOpeningFailed),
+            "different protocol message must fail the KZG opening check, got: {err}"
+        );
+    }
+
+    #[test]
+    fn ivc_proof_verify_rejects_tampered_message_bytes_with_tampered_input_message() {
         let (global, verifier_setup) = build_proof_verifier_context();
         let mut step_output = load_embedded_next_epoch_step_output_asset()
             .expect("recursive step output asset should load");
@@ -773,6 +822,7 @@ mod tests {
         let chain_state = load_embedded_recursive_chain_state_asset()
             .expect("recursive chain state asset should load");
 
+        // I don't know what the correct bytes are...
         let msg = &[
             22, 148, 87, 37, 149, 0, 124, 10, 156, 94, 108, 6, 78, 59, 239, 80, 126, 213, 158, 211,
             191, 213, 128, 70, 128, 30, 235, 80, 192, 191, 159, 67,
--- a/mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
+++ b/mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
@@ -33,6 +33,7 @@ use crate::{
             types::{CertificateProofBytes, IvcProofBytes, ProtocolMessagePreimage},
         },
     },
+    codec,
     proof_system::ivc_halo2_snark::{
         CircuitProvingKey,
         errors::IvcProofError,
@@ -45,8 +46,6 @@ use crate::{
 };
 
 /// Per-session IVC prover handle.
-// TODO: remove this allow dead_code directive when the IVC prover is wired into STM
-#[allow(dead_code)]
 pub(crate) struct IvcProver<R: RngCore + CryptoRng> {
     /// Shared, cached setup (SRS, verifying keys, proving key, fixed-base maps).
     pub(crate) ivc_setup: Arc<IvcProverSetup>,
@@ -58,8 +57,6 @@ pub(crate) struct IvcProver<R: RngCore + CryptoRng> {
 ///
 /// Always supplied to [`IvcProver::prove`] by reference; used only when `rolling_state = None`
 /// (the first certificate) to run the internal genesis IVC step before processing it.
-// TODO: remove this allow dead_code directive when IvcProver::prove is wired into STM
-#[allow(dead_code)]
 pub(crate) struct IvcGenesisBootstrapInput {
     /// Schnorr half of the Lagrange-era dual genesis signature (Ed25519 + Schnorr). Carried
     /// forward through every rolling state for in-circuit verification of the genesis message.
@@ -71,12 +68,14 @@ pub(crate) struct IvcGenesisBootstrapInput {
     pub(crate) genesis_protocol_message_preimage: ProtocolMessagePreimage,
 }
 
+/// Fails if the genesis Schnorr signature is absent
+/// or if the message preimage is not exactly 190 bytes.
 impl TryFrom<&AncillaryGenesisData> for IvcGenesisBootstrapInput {
     type Error = anyhow::Error;
     fn try_from(ancillary_genesis_data: &AncillaryGenesisData) -> StmResult<Self> {
         let genesis_signature = ancillary_genesis_data
             .genesis_schnorr_signature()
-            .ok_or_else(|| anyhow!("Missing genesis signature!"))?;
+            .ok_or_else(|| anyhow!("Missing genesis Schnorr signature."))?;
 
         let genesis_protocol_message_preimage: [u8; 190] =
             ancillary_genesis_data.genesis_message_preimage().try_into()?;
@@ -93,8 +92,6 @@ impl TryFrom<&AncillaryGenesisData> for IvcGenesisBootstrapInput {
 /// `H` is the transcript hash used to produce this proof and must be used to verify it.
 /// It is a zero-cost phantom: no `H`-dependent data is stored, but it prevents accidentally
 /// verifying a Poseidon-produced proof via the Blake2b path and vice versa.
-// TODO: remove this allow dead_code directive when the IVC prover emits this proof
-#[allow(dead_code)]
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct IvcProof<H: TranscriptHash> {
     /// Externally-verifiable proof bytes.
@@ -113,7 +110,6 @@ impl<H: TranscriptHash> IvcProof<H> {
     ///
     /// `H` is inferred from the prover's own type parameter, so the proof's hash type
     /// is bound to the hash used to produce it without any runtime check.
-    #[cfg_attr(not(test), allow(dead_code))]
     pub(crate) fn new(
         proof_bytes: IvcProofBytes,
         state: State,
@@ -126,10 +122,24 @@ impl<H: TranscriptHash> IvcProof<H> {
             hash: PhantomData,
         }
     }
+
+    /// Converts a IvcProof to CBOR bytes with a version prefix.
+    pub fn to_bytes(&self) -> StmResult<Vec<u8>> {
+        codec::to_cbor_bytes(self)
+    }
+
+    /// Deserialise an IVC proof from bytes.
+    pub fn from_bytes(bytes: &[u8]) -> StmResult<Self> {
+        if codec::has_cbor_v1_prefix(bytes) {
+            codec::from_cbor_bytes(&bytes[1..])
+        } else {
+            Err(anyhow::anyhow!(
+                "IvcProof: unsupported encoding, expected a CBOR v1 prefix"
+            ))
+        }
+    }
 }
 
-// TODO: remove this allow dead_code directive when IvcProof::verify is called
-#[allow(dead_code)]
 impl<H: TranscriptHash> IvcProof<H>
 where
     CircuitBase: Sampleable<H> + Hashable<H>,
@@ -239,8 +249,6 @@ fn ensure_advanceable_rolling_state(rolling_state: Option<&IvcRollingState>) -> 
     Ok(())
 }
 
-// TODO: remove this allow dead_code directive when the IVC prover is wired into STM
-#[allow(dead_code)]
 impl<R: RngCore + CryptoRng> IvcProver<R> {
     /// Advances the IVC chain by one step.
     ///
@@ -1051,8 +1059,8 @@ mod tests {
                 verification_context,
             };
 
-            // run_bootstrap_path(&ctx);
-            // run_next_epoch_path(&ctx);
+            run_bootstrap_path(&ctx);
+            run_next_epoch_path(&ctx);
             run_same_epoch_path(&ctx);
         }
 
--- a/mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
+++ b/mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
@@ -492,6 +492,21 @@ mod tests {
 
     use super::IvcProof;
 
+    const STEP_OUTPUT_MSG: [u8; 32] = [
+        22, 148, 87, 37, 149, 0, 124, 10, 156, 94, 108, 6, 78, 59, 239, 80, 126, 213, 158, 211,
+        191, 213, 128, 70, 128, 30, 235, 80, 192, 191, 159, 67,
+    ];
+
+    const SAME_EPOCH_MSG: [u8; 32] = [
+        147, 84, 244, 74, 250, 60, 153, 155, 8, 94, 236, 150, 53, 39, 132, 61, 99, 153, 192, 207,
+        20, 90, 16, 130, 216, 12, 87, 134, 230, 4, 190, 175,
+    ];
+
+    const CHAIN_STATE_MSG: [u8; 32] = [
+        253, 10, 116, 221, 249, 84, 222, 35, 101, 84, 229, 73, 90, 91, 97, 173, 36, 63, 47, 98,
+        189, 1, 99, 75, 183, 186, 225, 31, 226, 29, 121, 122,
+    ];
+
     fn build_proof_verifier_context() -> (Global, IvcVerifierSetup) {
         let ctx = load_embedded_verification_context_asset()
             .expect("verification context asset should load");
@@ -520,11 +535,6 @@ mod tests {
         let step_output = load_embedded_next_epoch_step_output_asset()
             .expect("recursive step output asset should load");
 
-        let msg = &[
-            22, 148, 87, 37, 149, 0, 124, 10, 156, 94, 108, 6, 78, 59, 239, 80, 126, 213, 158, 211,
-            191, 213, 128, 70, 128, 30, 235, 80, 192, 191, 159, 67,
-        ];
-
         let setup = build_asset_generation_setup();
         let global = build_recursive_global(
             &setup,
@@ -546,7 +556,7 @@ mod tests {
         );
 
         proof
-            .verify(msg, &global, &verifier_setup)
+            .verify(&STEP_OUTPUT_MSG, &global, &verifier_setup)
             .expect("stored recursive step output should pass IvcProof::verify");
     }
 
@@ -576,11 +586,6 @@ mod tests {
         let step_output = load_embedded_next_epoch_step_output_asset()
             .expect("recursive step output asset should load");
 
-        let msg = &[
-            22, 148, 87, 37, 149, 0, 124, 10, 156, 94, 108, 6, 78, 59, 239, 80, 126, 213, 158, 211,
-            191, 213, 128, 70, 128, 30, 235, 80, 192, 191, 159, 67,
-        ];
-
         let proof = IvcProof::<blake2b_simd::State>::new(
             step_output.ivc_proof,
             step_output.next_state,
@@ -588,7 +593,7 @@ mod tests {
         );
 
         proof
-            .check_input_message_matches_state_message(msg)
+            .check_input_message_matches_state_message(&STEP_OUTPUT_MSG)
             .expect("Correct message should be accepted by verification function");
     }
 
@@ -597,10 +602,8 @@ mod tests {
         let step_output = load_embedded_next_epoch_step_output_asset()
             .expect("recursive step output asset should load");
 
-        let wrong_msg = &[
-            21, 148, 87, 37, 149, 0, 124, 10, 156, 94, 108, 6, 78, 59, 239, 80, 126, 213, 158, 211,
-            191, 213, 128, 70, 128, 30, 235, 80, 192, 191, 159, 67,
-        ];
+        let mut wrong_msg = STEP_OUTPUT_MSG;
+        wrong_msg[0] ^= 0xff;
 
         let proof = IvcProof::<blake2b_simd::State>::new(
             step_output.ivc_proof,
@@ -609,7 +612,7 @@ mod tests {
         );
 
         let err = proof
-            .check_input_message_matches_state_message(wrong_msg)
+            .check_input_message_matches_state_message(&wrong_msg)
             .expect_err("wrong message should be rejected by verification function");
         assert_eq!(
             err.downcast_ref::<IvcProofError>(),
@@ -625,10 +628,8 @@ mod tests {
         let step_output = load_embedded_next_epoch_step_output_asset()
             .expect("recursive step output asset should load");
 
-        let wrong_msg = &[
-            21, 148, 87, 37, 149, 0, 124, 10, 156, 94, 108, 6, 78, 59, 239, 80, 126, 213, 158, 211,
-            191, 213, 128, 70, 128, 30, 235, 80, 192, 191, 159, 67,
-        ];
+        let mut wrong_msg = STEP_OUTPUT_MSG;
+        wrong_msg[0] ^= 0xff;
 
         let setup = build_asset_generation_setup();
         let global = build_recursive_global(
@@ -651,7 +652,7 @@ mod tests {
         );
 
         let err = proof
-            .verify(wrong_msg, &global, &verifier_setup)
+            .verify(&wrong_msg, &global, &verifier_setup)
             .expect_err("tampered message should be rejected by IvcProof::verify");
         assert_eq!(
             err.downcast_ref::<IvcProofError>(),
@@ -669,11 +670,6 @@ mod tests {
         let step_output = load_embedded_next_epoch_step_output_asset()
             .expect("recursive step output asset should load");
 
-        let msg = &[
-            22, 148, 87, 37, 149, 0, 124, 10, 156, 94, 108, 6, 78, 59, 239, 80, 126, 213, 158, 211,
-            191, 213, 128, 70, 128, 30, 235, 80, 192, 191, 159, 67,
-        ];
-
         let mut tampered_bytes = step_output.ivc_proof.as_bytes().to_vec();
         let mid = tampered_bytes.len() / 2;
         tampered_bytes[mid] ^= 0xff;
@@ -685,7 +681,7 @@ mod tests {
         );
 
         let err = proof
-            .verify(msg, &global, &verifier_setup)
+            .verify(&STEP_OUTPUT_MSG, &global, &verifier_setup)
             .expect_err("tampered proof bytes should be rejected by IvcProof::verify");
         assert_eq!(
             err.downcast_ref::<IvcProofError>(),
@@ -700,10 +696,6 @@ mod tests {
         let mut step_output = load_embedded_next_epoch_step_output_asset()
             .expect("recursive step output asset should load");
 
-        let msg = &[
-            22, 148, 87, 37, 149, 0, 124, 10, 156, 94, 108, 6, 78, 59, 239, 80, 126, 213, 158, 211,
-            191, 213, 128, 70, 128, 30, 235, 80, 192, 191, 159, 67,
-        ];
         step_output.next_state.message = MessageHash::ZERO;
 
         let proof = IvcProof::<blake2b_simd::State>::new(
@@ -713,7 +705,7 @@ mod tests {
         );
 
         let err = proof
-            .verify(msg, &global, &verifier_setup)
+            .verify(&STEP_OUTPUT_MSG, &global, &verifier_setup)
             .expect_err("different protocol message should be rejected by IvcProof::verify");
         assert_eq!(
             err.downcast_ref::<IvcProofError>(),
@@ -760,11 +752,6 @@ mod tests {
         let same_epoch = load_embedded_following_certificate_in_epoch_asset()
             .expect("same-epoch step output asset should load");
 
-        let msg = &[
-            147, 84, 244, 74, 250, 60, 153, 155, 8, 94, 236, 150, 53, 39, 132, 61, 99, 153, 192,
-            207, 20, 90, 16, 130, 216, 12, 87, 134, 230, 4, 190, 175,
-        ];
-
         let proof = IvcProof::<blake2b_simd::State>::new(
             step_output.ivc_proof,
             same_epoch.next_state,
@@ -772,7 +759,7 @@ mod tests {
         );
 
         let err = proof
-            .verify(msg, &global, &verifier_setup)
+            .verify(&SAME_EPOCH_MSG, &global, &verifier_setup)
             .expect_err("state from a different proof should be rejected by IvcProof::verify");
         assert_eq!(
             err.downcast_ref::<IvcProofError>(),
@@ -792,18 +779,13 @@ mod tests {
         let same_epoch = load_embedded_following_certificate_in_epoch_asset()
             .expect("same-epoch step output asset should load");
 
-        let msg = &[
-            22, 148, 87, 37, 149, 0, 124, 10, 156, 94, 108, 6, 78, 59, 239, 80, 126, 213, 158, 211,
-            191, 213, 128, 70, 128, 30, 235, 80, 192, 191, 159, 67,
-        ];
-
         let proof = IvcProof::<blake2b_simd::State>::new(
             step_output.ivc_proof,
             step_output.next_state,
             same_epoch.next_accumulator,
         );
 
-        let err = proof.verify(msg, &global, &verifier_setup).expect_err(
+        let err = proof.verify(&STEP_OUTPUT_MSG, &global, &verifier_setup).expect_err(
             "accumulator from a different proof should be rejected by IvcProof::verify",
         );
         assert_eq!(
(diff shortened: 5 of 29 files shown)


damrobi:
There are a few points to clarify so let me know if you have comments regarding them:
 - Implementation change (match branch for IVC) in `mithril-common/src/certificate_chain/certificate_verifier.rs`
 - double implementation of Serialize for `MidnightVK` one in `halo2_snark/circuit_verification_key.rs` and one in `ivc_halo2_snark/verifier_setup.rs` (I was thinking of removing the one in 'ivc_halo2_snark)

mithril-stm/src/proof_system/ivc_halo2_snark/verifier_setup.rs
```diff
@@ -167,6 +171,140 @@ impl IvcVerifierSetup {
     }
 }
 
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct IvcVerifierData {
+    genesis_message: MessageHash,
+    genesis_schnorr_verification_key: SchnorrVerificationKey,
```
jpraynaud, line 178: We don't want the genesis verification key in the IvcVerifierData as this would be a serious security issue (as anyone could create a fake certificate chain).

mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
```diff
@@ -65,19 +68,38 @@ pub(crate) struct IvcGenesisBootstrapInput {
...
+/// Fails if the genesis Schnorr signature is absent
+/// or if the message preimage is not exactly 190 bytes.
+impl TryFrom<&AncillaryGenesisData> for IvcGenesisBootstrapInput {
+    type Error = anyhow::Error;
+    fn try_from(ancillary_genesis_data: &AncillaryGenesisData) -> StmResult<Self> {
+        let genesis_signature = ancillary_genesis_data
+            .genesis_schnorr_signature()
+            .ok_or_else(|| anyhow!("Missing genesis Schnorr signature."))?;
+
+        let genesis_protocol_message_preimage: [u8; 190] =
```
```suggestion
        let genesis_protocol_message_preimage: [u8; PREIMAGE_SIZE] =
```

mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
```diff
@@ -101,10 +122,24 @@ impl<H: TranscriptHash> IvcProof<H> {
...
+    /// Deserialise an IVC proof from bytes.
+    pub fn from_bytes(bytes: &[u8]) -> StmResult<Self> {
+        if codec::has_cbor_v1_prefix(bytes) {
+            codec::from_cbor_bytes(&bytes[1..])
+        } else {
+            Err(anyhow::anyhow!(
+                "IvcProof: unsupported encoding, expected a CBOR v1 prefix"
+            ))
+        }
+    }
```
jpraynaud, line 144: We need golden tests for the `to_bytes`/`from_bytes` (at least roundtrip) here and in all other occurrences in this PR.

mithril-stm/src/proof_system/ivc_halo2_snark/prover_setup.rs
```diff
@@ -89,7 +85,8 @@ impl IvcProverSetup {
         certificate_key_provider: &TempCertificateKeyProvider,
         ivc_key_provider: &TempIvcKeyProvider,
     ) -> StmResult<Self> {
-        let srs = trusted_setup_provider.get_trusted_setup_parameters()?;
+        let mut srs = trusted_setup_provider.get_trusted_setup_parameters()?;
+        srs.downsize(K);
```
jpraynaud, line 88: @hjeljeli32 heads-up for your PR here.

mithril-stm/src/proof_system/ivc_halo2_snark/unsafe_setup_helpers.rs
```diff
@@ -60,17 +58,27 @@ impl TempCertificateKeyProvider {
     }
 
     /// Builds the certificate circuit, clones the shared SRS so it can be
-    /// downsized in place, and runs `zk::setup_vk` to produce the verifying key.
-    pub(crate) fn get_verifying_key(&self) -> StmResult<CircuitVerifyingKey> {
+    /// downsized in place, and runs `zk::setup_vk` to produce the [MidnightVK].
+    ///
+    /// The [MidnightVK] carries the circuit architecture, which is required to
+    /// deserialize the verifying key against the correct constraint system.
+    pub(crate) fn get_midnight_verifying_key(&self) -> StmResult<MidnightVK> {
```
jpraynaud, line 65: @hjeljeli32 another heads-up for your PR.

mithril-stm/src/protocol/aggregate_signature/clerk.rs
```diff
@@ -114,6 +137,49 @@ impl<D: MembershipDigest> Clerk<D> {
...
+                        "Signatures failed to aggregate for type {}",
+                        AggregateSignatureType::Snark
+                    )
+                })?;
+
+                // A same-epoch step does not advance the rolling state, so the prover returns
+                // none: carry the input rolling state forward so the next certificate keeps
+                // building on it.
+                let previous_prover_data = ancillary_input.prover_data().cloned();
+
```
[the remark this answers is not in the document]
damrobi: Done

mithril-stm/src/proof_system/ivc_halo2_snark/unsafe_setup_helpers.rs
```diff
@@ -60,17 +58,27 @@ impl TempCertificateKeyProvider {
...
+
+    /// Builds the certificate circuit, clones the shared SRS so it can be
+    /// downsized in place, and runs `zk::setup_vk` to produce the verifying key.
+    pub(crate) fn get_verifying_key(&self) -> StmResult<CircuitVerifyingKey> {
+        Ok(self.get_midnight_verifying_key()?.vk().clone())
     }
 
     /// Recomputes the verifying key (the temp layer has no cache, so the VK is
     /// re-derived here) and runs `zk::setup_pk` to produce the proving key.
+    #[allow(dead_code)]
```
jpraynaud: If it's not needed, you can remove it.

mithril-stm/src/proof_system/ivc_halo2_snark/unsafe_setup_helpers.rs
```diff
@@ -110,14 +116,18 @@ impl TempIvcKeyProvider {
     /// and runs `keygen_vk_with_k` at the IVC circuit's domain size `K`.
     pub(crate) fn get_verifying_key(&self) -> StmResult<CircuitVerifyingKey> {
         let ivc_circuit_data = IvcCircuitData::unknown(&self.certificate_verifying_key)?;
-        Ok(keygen_vk_with_k(self.srs.as_ref(), &ivc_circuit_data, K)?)
+        let mut ivc_srs = (*self.srs).clone();
+        ivc_srs.downsize(K);
```
jpraynaud, line 108: Maybe it's worth checking that this is hamrless if no downsizing is needed.

mithril-stm/src/protocol/aggregate_signature/clerk.rs
```diff
@@ -165,3 +231,195 @@ impl<D: MembershipDigest> Clerk<D> {
...
+#[cfg(feature = "future_snark")]
+fn load_ivc_prover_setup(
+    parameters: Parameters,
+    merkle_tree_depth: u32,
+) -> StmResult<(Arc<IvcProverSetup>, MidnightVK)> {
+    let cache_key = (parameters.to_bytes()?, merkle_tree_depth);
+
+    if let Some(cached) = IVC_PROVER_SETUP_CACHE.lock().unwrap().get(&cache_key) {
+        return Ok(cached.clone());
+    }
```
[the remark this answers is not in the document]
damrobi: Done for now but this part might be replaced by the PR #3349

mithril-stm/src/protocol/aggregate_signature/clerk.rs
```diff
@@ -165,3 +231,195 @@ impl<D: MembershipDigest> Clerk<D> {
...
+        &trusted_setup_provider,
+        &certificate_key_provider,
+        &ivc_key_provider,
+    )?);
+
+    let cached = (ivc_setup, certificate_midnight_verifying_key);
+    IVC_PROVER_SETUP_CACHE
+        .lock()
+        .unwrap()
+        .insert(cache_key, cached.clone());
```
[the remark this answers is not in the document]
damrobi: Same as above

mithril-stm/src/protocol/aggregate_signature/signature.rs
```diff
@@ -136,7 +160,6 @@ impl<D: MembershipDigest> AggregateSignature<D> {
         parameters: &Parameters,
         ancillary_verifier_data: Option<AncillaryVerifierData>,
     ) -> StmResult<()> {
-        let _ = ancillary_verifier_data;
         match self {
             AggregateSignature::Concatenation(concatenation_proof) => concatenation_proof.verify(
```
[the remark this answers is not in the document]
damrobi: Solved using a reference like in the previous comment.

mithril-stm/src/protocol/aggregate_signature/ancillary_data.rs
```diff
@@ -36,15 +43,26 @@ impl AncillaryProverData {
             ))
         }
     }
+
+    /// Returns the wrapped IvcRollingState of an AncillaryProverData.
+    #[cfg(feature = "future_snark")]
+    pub fn as_ivc_rolling_state(&self) -> IvcRollingState {
+        match self {
+            Self::IvcSnark(state) => state.clone(),
```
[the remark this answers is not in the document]
damrobi: Updated to return a reference.

mithril-stm/src/protocol/aggregate_signature/ancillary_data.rs
```diff
@@ -64,6 +82,14 @@ impl AncillaryVerifierData {
             ))
         }
     }
+
+    /// Returns the wrapped IvcVerifierData of an AncillaryVerifierData.
+    #[cfg(feature = "future_snark")]
+    pub fn as_ivc_verifier_data(&self) -> IvcVerifierData {
+        match self {
+            Self::IvcSnark(state) => state.clone(),
+        }
```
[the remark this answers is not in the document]
damrobi: Updated to return a reference.

mithril-stm/src/proof_system/ivc_halo2_snark/verifier_setup.rs
```diff
@@ -167,6 +171,138 @@ impl IvcVerifierSetup {
...
+        self.genesis_schnorr_verification_key
+    }
+
+    pub(crate) fn certificate_circuit_verification_key(&self) -> CircuitVerifyingKey {
+        self.certificate_circuit_verification_key.vk().clone()
+    }
+
+    pub(crate) fn ivc_circuit_verification_key(&self) -> CircuitVerifyingKey {
+        self.ivc_circuit_verification_key.clone()
+    }
```
jpraynaud, line 240: Could you add doc comments here?

mithril-stm/src/protocol/aggregate_signature/ancillary_data.rs
```diff
@@ -8,15 +8,22 @@ use serde::{Deserialize, Serialize};
 use crate::StmResult;
 use crate::codec;
 #[cfg(feature = "future_snark")]
+use crate::proof_system::IvcRollingState;
+#[cfg(feature = "future_snark")]
+use crate::proof_system::ivc_halo2_snark::verifier_setup::IvcVerifierData;
```
jpraynaud: These 2 lines can be combined to avoid too many repetitions of `#[cfg(feature = "future_snark")]`

mithril-stm/src/protocol/aggregate_signature/ancillary_data.rs
```diff
@@ -36,15 +43,26 @@ impl AncillaryProverData {
             ))
         }
     }
+
+    /// Returns the wrapped IvcRollingState of an AncillaryProverData.
+    #[cfg(feature = "future_snark")]
+    pub fn as_ivc_rolling_state(&self) -> &IvcRollingState {
```
jpraynaud: Otherwise this is not future proof:
```suggestion
    pub fn as_ivc_rolling_state(&self) -> Option<&IvcRollingState> {
```

mithril-stm/src/protocol/aggregate_signature/ancillary_data.rs
```diff
@@ -64,6 +82,14 @@ impl AncillaryVerifierData {
             ))
         }
     }
+
+    /// Returns the wrapped IvcVerifierData of an AncillaryVerifierData.
+    #[cfg(feature = "future_snark")]
+    pub fn as_ivc_verifier_data(&self) -> &IvcVerifierData {
```
jpraynaud:
```suggestion
    pub fn as_ivc_verifier_data(&self) -> Option<&IvcVerifierData> {
```

mithril-stm/src/protocol/aggregate_signature/clerk.rs
```diff
@@ -114,6 +137,49 @@ impl<D: MembershipDigest> Clerk<D> {
                     )
                 })
             }
+            #[cfg(feature = "future_snark")]
+            AggregateSignatureType::IvcSnark => {
+                let clerk = self
```
jpraynaud:
```suggestion
                let snark_clerk = self
```

mithril-stm/src/protocol/aggregate_signature/clerk.rs
```diff
@@ -165,3 +231,199 @@ impl<D: MembershipDigest> Clerk<D> {
...
+    snark_proof: SnarkProof<D>,
+    msg: &[u8],
+    clerk: &SnarkClerk,
+    ancillary_input: AncillaryProofInput,
+) -> StmResult<(
+    IvcProof<blake2b_simd::State>,
+    Option<IvcRollingState>,
+    IvcVerifierData,
+)> {
+    let protocol_message_preimage_bytes: [u8; 190] =
```
jpraynaud:
```suggestion
    let protocol_message_preimage_bytes: [u8; PREIMAGE_SIZE] =
```

mithril-stm/src/protocol/aggregate_signature/clerk.rs
```diff
@@ -165,3 +231,199 @@ impl<D: MembershipDigest> Clerk<D> {
...
+
+    let genesis_verifying_key = genesis_data
+        .genesis_schnorr_verification_key()
+        .cloned()
+        .ok_or_else(|| anyhow!("Missing genesis verifying key from ancillary data"))?;
+
+    let genesis_preimage = genesis_data.genesis_message_preimage();
+    let genesis_preimage_hash: [u8; 32] = Sha256::digest(genesis_preimage).into();
+    let genesis_message_field_elem = BaseFieldElement::from_raw(&genesis_preimage_hash)?.0;
+    let genesis_message = MessageHash::from_field(genesis_message_field_elem);
```
jpraynaud: Don't we already have some conversion available from value to field element?

mithril-stm/src/protocol/aggregate_signature/clerk.rs
```diff
@@ -165,3 +231,199 @@ impl<D: MembershipDigest> Clerk<D> {
...
+
+    let (ivc_prover_setup, certificate_midnight_verifying_key) =
+        load_ivc_prover_setup(clerk.parameters, MERKLE_TREE_DEPTH_FOR_SNARK)?;
+    let certificate_circuit_verifying_key = certificate_midnight_verifying_key.vk().clone();
+    let ivc_circuit_verifying_key = ivc_prover_setup.ivc_verifying_key.clone();
+
+    let rolling_state = ancillary_input
+        .prover_data()
+        .map(|prover_data| prover_data.as_ivc_rolling_state());
```
jpraynaud: The computation of the rolling state is a bit awkward, maybe it the previous rolling state from the previous certificate should be an input to this function.

mithril-stm/src/protocol/aggregate_signature/clerk.rs
```diff
@@ -165,3 +231,199 @@ impl<D: MembershipDigest> Clerk<D> {
...
+
+    let genesis_bootstrap = &genesis_data.try_into()?;
+
+    // This is only used to get the root of the tree so maybe we can replace
+    // the avk inputs by the root directly?
```
jpraynaud: This comment should be dealt with and removed.

mithril-stm/src/protocol/aggregate_signature/clerk.rs
```diff
@@ -165,3 +231,199 @@ impl<D: MembershipDigest> Clerk<D> {
...
+}
+
+/// Process-wide cache of the IVC prover setup keyed by the certificate circuit shape.
+///
+/// The setup (SRS plus the certificate and IVC circuit keys) only depends on the protocol
+/// parameters and the Merkle tree depth, so it is computed once per shape and reused across
+/// certificates instead of being rebuilt on every proof. The certificate [MidnightVK] is cached
+/// alongside it because it is needed to expose the verifier data.
+#[cfg(feature = "future_snark")]
+type IvcProverSetupCache = HashMap<(Vec<u8>, u32), (Arc<IvcProverSetup>, MidnightVK)>;
```
jpraynaud: This is probably not the best place for this cache, which will likely be removed in the future with @hjeljeli32's PR. It'd be better to isolate this in another sub-module with a TODO explaining it is temporary.

mithril-stm/src/protocol/aggregate_signature/error.rs
```diff
@@ -53,4 +53,8 @@ pub enum AggregateSignatureError {
     /// The proof system is unknown
     #[error("Unknown proof system: {0}")]
     UnknownProofSystem(String),
+
+    /// Missing ancillary data to verify the IVC proof
+    #[error("Missing ancillary data to verify the IVC proof.")]
+    MissingAncillaryData,
```
jpraynaud:
```suggestion
    MissingAncillaryVerifierData,
```

mithril-stm/src/protocol/aggregate_signature/clerk.rs
jpraynaud, line 1: I noticed that we always return an `IvcRollingState` (which will be stored in the certificate chain when not needed) when the certificate is not the first of the epoch.

mithril-stm/src/proof_system/ivc_halo2_snark/prover_setup.rs
```diff
@@ -131,6 +128,8 @@ impl IvcProverSetup {
     /// gadget zeros the prepared accumulator via `scale_by_bit(is_not_genesis, ...)`;
     /// off-circuit we construct a fresh trivial accumulator over the combined fixed-base
     /// names.
+    // TODO: remove this allow dead_code directive when the function is used or remove the function
+    #[allow(dead_code)]
     pub(crate) fn trivial_previous_ivc_proof_collapsed_accumulator(
```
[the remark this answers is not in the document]
damrobi: Function removed

mithril-stm/src/proof_system/ivc_halo2_snark/unsafe_setup_helpers.rs
```diff
@@ -60,17 +58,27 @@ impl TempCertificateKeyProvider {
...
+    /// Builds the certificate circuit, clones the shared SRS so it can be
+    /// downsized in place, and runs `zk::setup_vk` to produce the verifying key.
+    pub(crate) fn get_verifying_key(&self) -> StmResult<CircuitVerifyingKey> {
+        Ok(self.get_midnight_verifying_key()?.vk().clone())
     }
 
     /// Recomputes the verifying key (the temp layer has no cache, so the VK is
     /// re-derived here) and runs `zk::setup_pk` to produce the proving key.
+    #[allow(dead_code)]
     pub(crate) fn get_proving_key(&self) -> StmResult<CircuitProvingKey> {
```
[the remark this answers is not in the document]
damrobi: Function removed

mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
damrobi: Changed

mithril-stm/src/proof_system/ivc_halo2_snark/unsafe_setup_helpers.rs
damrobi: Downsizing to the same or higher degree has no impact but it might have a small cost so I think it's worth avoiding it if possible.

mithril-stm/src/proof_system/ivc_halo2_snark/verifier_setup.rs
damrobi: Added

mithril-stm/src/protocol/aggregate_signature/ancillary_data.rs
damrobi: Done

mithril-stm/src/protocol/aggregate_signature/ancillary_data.rs
damrobi: I added the Option return. The match statement can be completed when we add more variants

mithril-stm/src/protocol/aggregate_signature/ancillary_data.rs
damrobi: Same as above

mithril-stm/src/protocol/aggregate_signature/clerk.rs
damrobi: Done

mithril-stm/src/proof_system/ivc_halo2_snark/verifier_setup.rs
curiecrypt: Agree. On the verify path, I suggest pinning `genesis_message` and `genesis_schnorr_verification_key` to the chain's trusted genesis (not per-certificate ancillary data), and assert the certificate/IVC verifying keys match the embedded production keys.

mithril-stm/src/protocol/aggregate_signature/signature.rs
```diff
@@ -151,6 +174,30 @@ impl<D: MembershipDigest> AggregateSignature<D> {
                 let verifier_setup = SnarkVerifierSetup::try_new()?;
                 snark_proof.verify(msg, snark_avk, &verifier_setup.verifier_params)
             }
+            #[cfg(feature = "future_snark")]
+            AggregateSignature::IvcSnark(ivc_proof) => {
```
curiecrypt, line 179: I could not see any negative test that `IvcSnark` verify fails with `MissingAncillaryData` when ancillary data is absent. Should we add?

mithril-stm/src/proof_system/ivc_halo2_snark/rolling_state.rs
```diff
@@ -134,6 +132,35 @@ impl IvcRollingState {
...
 }
 
+pub(crate) mod midnight_accumulator_serde {
+    use midnight_circuits::verifier::{Accumulator, BlstrsEmulation};
+    use midnight_proofs::utils::SerdeFormat;
+    use serde::{Deserializer, Serializer};
+
+    use crate::circuits::halo2_ivc::io::{Read, Write};
+
+    /// Serialization function based on the write function of the MidnightVK
```
curiecrypt: The doc comment says it wraps `MidnightVK`, but it serializes `Accumulator<BlstrsEmulation>`. I think we should revise the comment accordingly

mithril-stm/src/protocol/aggregate_signature/clerk.rs
```diff
@@ -165,3 +231,199 @@ impl<D: MembershipDigest> Clerk<D> {
...
+/// advance it (the prover returns none), so the input rolling state carried from the parent
+/// certificate is reused for the next certificate.
+#[cfg(feature = "future_snark")]
+fn resolve_ivc_ancillary_prover_data(
+    next_rolling_state: Option<IvcRollingState>,
+    previous_prover_data: Option<AncillaryProverData>,
+) -> StmResult<AncillaryProverData> {
+    match next_rolling_state {
+        Some(rolling_state) => Ok(AncillaryProverData::IvcSnark(rolling_state)),
+        None => previous_prover_data.ok_or_else(|| anyhow!("missing rolling state")),
```
curiecrypt: A more explaining error message could be better here.

mithril-stm/src/protocol/aggregate_signature/clerk.rs
damrobi: I tried to create a type for the `GenesisMessagePreimage` to abstract this away. Let me know what you think. The type is in `aggregate_signature/mod.rs` for now.

mithril-stm/src/proof_system/ivc_halo2_snark/verifier_setup.rs
```diff
@@ -167,6 +171,138 @@ impl IvcVerifierSetup {
...
+
+    pub(crate) fn genesis_message(&self) -> MessageHash {
+        self.genesis_message
+    }
+
+    pub(crate) fn genesis_schnorr_verification_key(&self) -> SchnorrVerificationKey {
+        self.genesis_schnorr_verification_key
+    }
+
+    pub(crate) fn certificate_circuit_verification_key(&self) -> CircuitVerifyingKey {
```
curiecrypt: Instead of returning clones for these getters, could we return reference?

mithril-stm/src/protocol/aggregate_signature/clerk.rs
damrobi: I removed the comment

mithril-stm/src/protocol/aggregate_signature/signature.rs
```diff
@@ -151,6 +174,30 @@ impl<D: MembershipDigest> AggregateSignature<D> {
...
+                    ivc_verifier_data.genesis_message(),
+                    ivc_verifier_data.genesis_schnorr_verification_key(),
+                    &certificate_verifying_key,
+                    &ivc_verifying_key,
+                );
+
+                let verifier_setup =
+                    IvcVerifierSetup::try_new(&certificate_verifying_key, ivc_verifying_key)?;
+
+                ivc_proof.verify(global, &verifier_setup)
```
curiecrypt: The IVC verify arm ignores `msg`. Nothing binds the proof's committed `state.message` to the certificate's `signed_message`.

I think we should recompute the expected message field element from `msg` the way the prover does (`MessageHash` from the protocol-message hash), pass it into `IvcProof::verify`, and assert it equals ivc_proof's state.message before accepting.

mithril-stm/src/protocol/aggregate_signature/clerk.rs
damrobi: I moved it to a temp module in the ivc_halo2_snark module

mithril-stm/src/protocol/aggregate_signature/error.rs
damrobi: Fixed

mithril-stm/src/protocol/aggregate_signature/signature.rs
hjeljeli32: Extending @curiecrypt's point about `msg`: this IVC branch also does not bind the proof’s final public state to the `avk` or `parameters` passed to `verify`.

The recursive proof commits to `state.message`, `state.merkle_tree_commitment`, and `state.protocol_parameters`, but here we only verify the proof against `global`/setup. Since `mithril-common` can stop chain verification once an `IvcSnark` certificate verifies, I think we should also assert that the final state matches the certificate being accepted:

- `state.message` matches the current certificate signed message
- `state.merkle_tree_commitment` matches the current SNARK AVK Merkle root
- `state.protocol_parameters` matches the current certificate protocol-parameters hash

Otherwise a valid IVC proof can be accepted without proving it corresponds to this certificate’s public fields.

mithril-stm/src/protocol/aggregate_signature/signature.rs
```diff
@@ -59,6 +71,8 @@ impl AggregateSignatureType {
             AggregateSignatureType::Concatenation => false,
             #[cfg(feature = "future_snark")]
             AggregateSignatureType::Snark => false,
+            #[cfg(feature = "future_snark")]
+            AggregateSignatureType::IvcSnark => true,
```
hjeljeli32, line 75: I’m not sure this shortcut is sound for same-epoch IVC certificates yet.

`IvcProver::prove` returns no new rolling state for same-epoch steps, and the clerk carries the previous rolling state forward. That means the persisted recursive state does not include the same-epoch certificate that was just produced. But `mithril-common` uses this predicate to stop walking the previous certificate chain entirely.

Could we either make same-epoch steps advance/persist the recursive rolling state as well, or keep this as `false` until we have a test showing that an IVC same-epoch certificate really certifies all previous same-epoch certificates too?

A useful negative test would be: create two same-epoch certificates, tamper the parent, then verify the child through `verify_certificate_chain`; it should fail if the shortcut is sound.

hjeljeli32 (commented):
Left a couple of suggestions 👍

mithril-stm/src/proof_system/ivc_halo2_snark/unsafe_setup_helpers.rs
damrobi: Removed

mithril-stm/src/protocol/aggregate_signature/clerk.rs
damrobi: Updated

mithril-stm/src/proof_system/ivc_halo2_snark/rolling_state.rs
damrobi: I fixed the comments

mithril-stm/src/protocol/aggregate_signature/clerk.rs
damrobi: I updated the returned error

mithril-stm/src/proof_system/ivc_halo2_snark/verifier_setup.rs
damrobi: I updated the functions to return references

mithril-stm/src/protocol/aggregate_signature/signature.rs
damrobi: Good point! We can also do it directly in the `AggregateSignature::verify` function since the hashed message comes from the state inside the proof itself. For now I will add it the to `IvcProof::verify` as you mentioned

mithril-stm/src/protocol/aggregate_signature/signature.rs
damrobi: I think I understand what you mean but I'm not sure, essentially a certificate that does not advances the epoch does not certify other such certificate? I'm not sure it is in opposition with this function goal for now but I might be wrong.
@jpraynaud @curiecrypt what do you think?

mithril-stm/src/protocol/aggregate_signature/signature.rs
damrobi: Ok so I should extend the checks to the avk and parameters on top of the message. I'll try to add that part

mithril-stm/src/proof_system/ivc_halo2_snark/verifier_setup.rs
jpraynaud: In this PR we'll go only with using a genesis verification key provided by the end user. We'll work an authentication/whitelisting of circuit verification keys after.

mithril-stm/src/protocol/aggregate_signature/signature.rs
jpraynaud: Good catch @curiecrypt! The verify function indeed lacks the message as input.

The verification should be done this way:
- The `IvcProof::verify` function gets the `msg: &[u8]` as input
- The `message" MessageHash` of the `State` must be recomputed from the other inputs (which represent the pre-image)
- The re-computed message must match the value given as input (after conversion to the field element)
- Otherwise, the verify function returns an error
- The IVC circuit proof bytes are verified as it is done today against the `Global` state and the verifier setup.

jpraynaud: As soon as the `message: MessageHash` that we recompute from the preimage fields matches the message that is verified, it means that the aggregate verification key and the protocol parameters match.

mithril-stm/src/protocol/aggregate_signature/signature.rs
jpraynaud: The IVC proof validates the full certificate chain since genesis. This means that this proof is enough and that a certificate chain is valid as soon as an IVC certificate is verified (this is exactly the feature we want to implement).

The rolling state is what will be needed to make another round of recursion at the following epoch and is not used during the verification of the proof of the certificate, it is only stored in the ancillary prover data of the certificate for convenience when gathering the information to be delivered to as input of the prover and in the case of synchronization of the certificate chain between 2 aggregators running the IVC proofs.

mithril-common/src/protocol/ancillary.rs
```diff
@@ -103,7 +103,7 @@ mod tests {
             protocol_message.rigid_preimage().as_slice()
         );
         assert_eq!(
-            genesis_data.genesis_message_preimage(),
+            genesis_data.genesis_message_preimage().0,
```
jpraynaud: This shouldn't change IMO.

mithril-stm/src/proof_system/ivc_halo2_snark/errors.rs
```diff
@@ -32,4 +32,10 @@ pub(crate) enum IvcProofError {
         "IVC prover called with invalid context: rolling_state must not be a genesis (step_counter == 0) state; pass None to bootstrap from genesis"
     )]
     InvalidProvingContext,
+
+    /// Mismatch between the protocol messages
+    #[error(
+        "IVC proof rejected: the protocol message used to create the proof is different from the input message"
+    )]
+    InvalidProtocolMessage,
```
jpraynaud:
```suggestion
    /// Mismatch between the messages
    #[error(
        "IVC proof rejected: the message used to create the proof is different from the input message"
    )]
    InvalidMessage,
```

mithril-stm/src/protocol/aggregate_signature/signature.rs
damrobi: In the end, we only check the message since it depends on the avk and parameters so it should be enough. I think there is a possibility that the verifier tries to call the verify with an avk and parameters that don't match the message in which case the proof would still validate but I'm not sure it is a scenario worth considering. Especially because the verifier can check on his own for this mismatch. Let me know what you think

mithril-stm/src/proof_system/ivc_halo2_snark/temp_ivc_prover_cache.rs
```diff
@@ -0,0 +1,72 @@
+#[cfg(feature = "future_snark")]
```
jpraynaud, line 2:
```suggestion
// TODO: remove when the `IvcSnarkSetup` is final
#[cfg(feature = "future_snark")]
```

mithril-stm/src/protocol/aggregate_signature/signature.rs
damrobi: I added this part along with some tests

mithril-stm/src/protocol/aggregate_signature/clerk.rs
damrobi: I tried to make the computation better by moving everything to the utility function. Let me know what you think.

mithril-stm/src/protocol/aggregate_signature/mod.rs
```diff
@@ -11,8 +11,27 @@ pub use ancillary_data::{
...
+#[cfg(feature = "future_snark")]
+use sha2::{Digest, Sha256};
 pub use signature::{AggregateSignature, AggregateSignatureType};
 
+#[cfg(feature = "future_snark")]
+use crate::{BaseFieldElement, StmResult, circuits::halo2_ivc::types::MessageHash};
+
+#[cfg(feature = "future_snark")]
+#[derive(Clone, Debug)]
+pub struct GenesisMessagePreimage(pub Vec<u8>);
```
jpraynaud: This applies to any preimage:
```suggestion
pub struct MessagePreimage(pub Vec<u8>);
```

mithril-stm/src/protocol/aggregate_signature/signature.rs
```diff
@@ -151,6 +175,31 @@ impl<D: MembershipDigest> AggregateSignature<D> {
...
+                    ivc_verifier_data.certificate_circuit_verification_key();
+
+                let ivc_verifying_key = ivc_verifier_data.ivc_circuit_verification_key();
+
+                let global = Global::new(
+                    ivc_verifier_data.genesis_message(),
+                    ivc_verifier_data.genesis_schnorr_verification_key(),
+                    certificate_verifying_key,
+                    ivc_verifying_key,
+                );
```
jpraynaud:
```suggestion
                let global = Global::new(
                    ivc_verifier_data.genesis_message(),
                    ivc_verifier_data.genesis_schnorr_verification_key(),
                    ivc_verifier_data.certificate_circuit_verification_key(),
                    ivc_verifier_data.ivc_circuit_verification_key(),
                );
```

mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
```diff
@@ -161,6 +202,33 @@ where
 
         Ok(())
     }
+
+    /// Verifies that the input protocol message is the same as the one used to generate
+    /// the proof.
+    ///
+    /// Returns an error if the input message has a wrong format, cannot be converted to a field
+    /// element or is different from the message store in the proof state.
+    fn input_message_matches_state_message(&self, msg: &[u8]) -> StmResult<()> {
```
```suggestion
    fn check_input_message_matches_state_message(&self, msg: &[u8]) -> StmResult<()> {
```

mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
```diff
@@ -1029,5 +1249,54 @@ mod tests {
             run_next_epoch_path(&ctx);
             run_same_epoch_path(&ctx);
         }
+
+        // The IVC circuit is degree `K`, but production builds its setup from the larger degree-22
+        // SRS. Keygen and the stored proving SRS must both downsize to `K`, otherwise their Lagrange
+        // basis differs and proofs do not verify. A larger unsafe SRS shares the smaller one's tau,
+        // so a correctly downsized setup reproduces the embedded degree-`K` assets exactly.
+        #[test]
+        fn ivc_setup_downsizes_keys_and_srs_to_the_circuit_degree() {
```
jpraynaud: Is this the good place for this test?

mithril-stm/src/protocol/aggregate_signature/clerk.rs
```diff
@@ -114,6 +129,40 @@ impl<D: MembershipDigest> Clerk<D> {
...
+                        snark_proof,
+                        msg,
+                        snark_clerk,
+                        &ancillary_input,
+                    )?;
+
+                Ok((
+                    AggregateSignature::IvcSnark(Box::new(ivc_proof)),
+                    AncillaryProofOutput::new(
+                        Some(ancillary_prover_data),
```
jpraynaud: This should be set to `None` when the proof is not for the first certificate of the epoch. The `ancillary_prover_data` returned by the `ivc_prover_input_preparation_and_prove` function should probably be an `Option<AncillaryProverData>`.

mithril-stm/src/protocol/aggregate_signature/signature.rs
damrobi: I added a test for that

mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
damrobi: I add the missing round trip for IvcProof, I will add the golden ones later

mithril-stm/src/protocol/aggregate_signature/signature.rs
damrobi: I added it

mithril-stm/src/proof_system/ivc_halo2_snark/errors.rs
damrobi: Done

mithril-stm/src/proof_system/ivc_halo2_snark/temp_ivc_prover_cache.rs
damrobi: Added

mithril-stm/src/protocol/aggregate_signature/mod.rs
damrobi: Done

mithril-stm/src/protocol/aggregate_signature/signature.rs
damrobi: Done

mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
damrobi: Done

mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
damrobi: True, I moved it to the temporary module for now since it uses Temp key providers

mithril-stm/src/protocol/aggregate_signature/clerk.rs
```diff
@@ -132,49 +131,34 @@ impl<D: MembershipDigest> Clerk<D> {
...
-                println!("First certificate proof generated, starting the IVC proof");
-
-                // A same-epoch step does not advance the rolling state, so the prover returns
-                // none: carry the input rolling state forward so the next certificate keeps
-                // building on it.
-                let previous_prover_data = ancillary_input.prover_data().cloned();
-
-                let (ivc_proof, next_rolling_state, verifier_data) =
+                // ancillary_prover_data is Some() when moving to the next epoch and None otherwise
+                let (ivc_proof, ancillary_prover_data, ancillary_verifier_data) =
```
jpraynaud:
```suggestion
                let (ivc_proof, next_ancillary_prover_data, ancillary_verifier_data) =
```

mithril-stm/src/protocol/aggregate_signature/clerk.rs
```diff
@@ -132,49 +131,34 @@ impl<D: MembershipDigest> Clerk<D> {
...
 
-                println!("First certificate proof generated, starting the IVC proof");
-
-                // A same-epoch step does not advance the rolling state, so the prover returns
-                // none: carry the input rolling state forward so the next certificate keeps
-                // building on it.
-                let previous_prover_data = ancillary_input.prover_data().cloned();
-
-                let (ivc_proof, next_rolling_state, verifier_data) =
+                // ancillary_prover_data is Some() when moving to the next epoch and None otherwise
```
jpraynaud: The doc comment of the `ivc_prover_input_preparation_and_prove` function should be enough:
```suggestion
```

mithril-stm/src/protocol/aggregate_signature/clerk.rs
```diff
@@ -229,178 +213,86 @@ impl<D: MembershipDigest> Clerk<D> {
     }
 }
 
+/// Prepares the IVC prover inputs from `ancillary_input`, loads the cached prover setup,
```
jpraynaud: It'd be better to have the prover setup loaded out of the function IMO.

mithril-stm/src/protocol/aggregate_signature/clerk.rs
```diff
@@ -229,178 +213,86 @@ impl<D: MembershipDigest> Clerk<D> {
...
-
-    println!("Getting the genesis_preimage");
-
-    let genesis_preimage = ancillary_input.genesis_data().genesis_message_preimage();
-
-    println!("Converting the genesis_preimage to field element");
-    // TODO: clean this to have proper call to SHA256
-    // the try_from function applies the Sha256 hash
-    let genesis_message_field_elem = BaseFieldElement::try_from(genesis_preimage)?.0;
+        .ok_or_else(|| anyhow!("Missing genesis verifying key from ancillary data"))?;
```
jpraynaud: You should return a typed error for that, and add a test.

mithril-stm/src/protocol/aggregate_signature/clerk.rs
```diff
@@ -229,178 +213,86 @@ impl<D: MembershipDigest> Clerk<D> {
...
+    let (ivc_prover_setup, certificate_midnight_verifying_key) =
+        load_ivc_prover_setup(clerk.parameters, MERKLE_TREE_DEPTH_FOR_SNARK)?;
+    let certificate_circuit_verifying_key = certificate_midnight_verifying_key.vk().clone();
+    let ivc_circuit_verifying_key = ivc_prover_setup.ivc_verifying_key.clone();
 
-    let srs = Arc::new(trusted_setup_provider.get_trusted_setup_parameters()?);
-    println!("Creating the certificate_key_provider from srs");
+    // Get a rolling state if there is some ancillary prover data and some rolling state
+    // otherwise get None
+    let rolling_state = ancillary_input
```
jpraynaud: If there is some prover data and it can't be converted to an `IvcRollingState` it should retrun an error (and be tested).

mithril-stm/src/protocol/aggregate_signature/clerk.rs
```diff
@@ -229,178 +213,86 @@ impl<D: MembershipDigest> Clerk<D> {
...
-        rolling_state.as_ref(),
+        rolling_state,
     )?;
 
-    Ok((ivc_proof, next_rolling_state, verifier_data))
-}
+    // A same-epoch step does not advance the rolling state, so the prover returns
+    // none: carry the input rolling state forward so the next certificate keeps
+    // building on it.
+    let ancillary_prover_data = next_rolling_state.map(AncillaryProverData::IvcSnark);
```
jpraynaud:
```suggestion
    let next_ancillary_prover_data = 
```

mithril-stm/src/protocol/aggregate_signature/mod.rs
```diff
@@ -11,8 +11,34 @@ pub use ancillary_data::{
...
+#[cfg(feature = "future_snark")]
+use sha2::{Digest, Sha256};
 pub use signature::{AggregateSignature, AggregateSignatureType};
 
+#[cfg(feature = "future_snark")]
+use crate::{BaseFieldElement, StmResult, circuits::halo2_ivc::types::MessageHash};
+
+#[cfg(feature = "future_snark")]
+#[derive(Clone, Debug)]
+pub struct MessagePreimage(pub Vec<u8>);
```
jpraynaud: It's worth moving this implementation to a new `preimage.rs` submodule and maybe to add some golden tests.

mithril-stm/src/protocol/aggregate_signature/clerk.rs
```diff
@@ -165,3 +212,87 @@ impl<D: MembershipDigest> Clerk<D> {
...
+    let genesis_message = genesis_data.genesis_message_preimage().try_into()?;
+
+    let (ivc_prover_setup, certificate_midnight_verifying_key) =
+        load_ivc_prover_setup(clerk.parameters, MERKLE_TREE_DEPTH_FOR_SNARK)?;
+    let certificate_circuit_verifying_key = certificate_midnight_verifying_key.vk().clone();
+    let ivc_circuit_verifying_key = ivc_prover_setup.ivc_verifying_key.clone();
+
+    // Get a rolling state if there is some ancillary prover data and some rolling state
+    // otherwise get None
+    let rolling_state = ancillary_input
```
jpraynaud:
```suggestion
    let current_rolling_state = ancillary_input
```

mithril-stm/src/protocol/aggregate_signature/clerk.rs
```diff
@@ -165,3 +212,87 @@ impl<D: MembershipDigest> Clerk<D> {
...
+
+    let genesis_message = genesis_data.genesis_message_preimage().try_into()?;
+
+    let (ivc_prover_setup, certificate_midnight_verifying_key) =
+        load_ivc_prover_setup(clerk.parameters, MERKLE_TREE_DEPTH_FOR_SNARK)?;
+    let certificate_circuit_verifying_key = certificate_midnight_verifying_key.vk().clone();
+    let ivc_circuit_verifying_key = ivc_prover_setup.ivc_verifying_key.clone();
+
+    // Get a rolling state if there is some ancillary prover data and some rolling state
+    // otherwise get None
```
jpraynaud:
```suggestion
```

mithril-stm/src/protocol/aggregate_signature/clerk.rs
```diff
@@ -165,3 +212,87 @@ impl<D: MembershipDigest> Clerk<D> {
...
+        &avk,
+        &global,
+        &ProtocolMessagePreimage(protocol_message_preimage_bytes),
+        genesis_bootstrap,
+        rolling_state,
+    )?;
+
+    // A same-epoch step does not advance the rolling state, so the prover returns
+    // none: carry the input rolling state forward so the next certificate keeps
+    // building on it.
```
jpraynaud:
```suggestion
```

mithril-stm/src/protocol/aggregate_signature/signature.rs
```diff
@@ -379,6 +463,55 @@ impl<D: MembershipDigest> AggregateSignature<D> {
 mod tests {
     use super::*;
 
+    mod ivc_proof {
```
jpraynaud, line 467: You can remove the gates below:
```suggestion
    #[cfg(feature = "future_snark")]
    mod ivc_proof {
```

mithril-stm/CHANGELOG.md
```diff
@@ -5,6 +5,19 @@ All notable changes to this project will be documented in this file.
...
+### Added
+
+- Added wiring of the IVC prover and verifier in the `Clerk` and `AggregateSignature`.
+
+### Changed
+
+- `Clerk::aggregate_signatures_with_type` IVC variant is completed.
+- `AggregateSignature::verify` IVC variant is completed.
+- `IvcProof::verify` now also checks that the input message is the same as the one used to create the proof.
+- More fixed the how the circuit verification keys are handled and computed.
```
jpraynaud: This is not very clear.

mithril-stm/src/protocol/aggregate_signature/mod.rs
```diff
@@ -11,8 +11,34 @@ pub use ancillary_data::{
...
+        Ok(MessageHash::from_field(genesis_message_field_elem))
+    }
+}
+
+#[cfg(feature = "future_snark")]
+impl MessagePreimage {
+    pub fn to_vec(&self) -> Vec<u8> {
+        self.0.clone()
+    }
+}
```
jpraynaud: You should implement `Deref` here to return a `&[u8]` or a `take(self) -> Vec<u8>`.

mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
```diff
@@ -475,10 +546,120 @@ mod tests {
...
+
+    #[test]
+    fn ivc_proof_message_verification_accepts_correct_message() {
+        let step_output = load_embedded_next_epoch_step_output_asset()
+            .expect("recursive step output asset should load");
+
+        let msg = &[
+            22, 148, 87, 37, 149, 0, 124, 10, 156, 94, 108, 6, 78, 59, 239, 80, 126, 213, 158, 211,
+            191, 213, 128, 70, 128, 30, 235, 80, 192, 191, 159, 67,
+        ];
```
jpraynaud: Can you create a const for the correct message and alter it with e.g. a bit flip when you need to make it tampered in these tests?

jpraynaud (approved):
LGTM 👍 

I left few comments to be addressed before merging.

mithril-stm/src/protocol/aggregate_signature/clerk.rs
damrobi: This was updated to return a `None` coming from the `IvcProver::prove()` function when staying in the same epoch

mithril-stm/src/protocol/aggregate_signature/clerk.rs
damrobi: This was updated

mithril-stm/src/protocol/aggregate_signature/clerk.rs
damrobi: I moved the prover setup creation outside the function in the `aggregate_signature_with_type`

mithril-stm/src/protocol/aggregate_signature/clerk.rs
damrobi: Done

mithril-stm/src/protocol/aggregate_signature/mod.rs
damrobi: Done

mithril-stm/src/protocol/aggregate_signature/clerk.rs
damrobi: Done

mithril-stm/src/protocol/aggregate_signature/clerk.rs
damrobi: Removed

mithril-stm/src/protocol/aggregate_signature/clerk.rs
damrobi: Removed

mithril-stm/src/protocol/aggregate_signature/signature.rs
damrobi: Done

mithril-stm/src/protocol/aggregate_signature/mod.rs
damrobi: I added the Deref implementation

mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
damrobi: Done

mithril-stm/CHANGELOG.md
damrobi: I tried to be more explicit

mithril-stm/src/protocol/aggregate_signature/mod.rs
jpraynaud: This type is a bit in conflict with the existing `ProtocolMessagePreimage`. It's be good to sort that out before merging.

mithril-stm/src/protocol/aggregate_signature/clerk.rs
damrobi: Added both

hjeljeli32 (approved):
LGTM 👍

mithril-stm/src/protocol/aggregate_signature/mod.rs
damrobi: I renamed the `MessagePreimage` to `GenesisMessagePreimage` for now. This struct can be replaced by `ProtocolMessagePreimage` modulo some modification to convert some vecs to `[u8; PREIMAGE_SIZE]`

mithril-stm/src/protocol/aggregate_signature/clerk.rs
damrobi: Error and test added

mithril-stm/src/protocol/aggregate_signature/clerk.rs
damrobi: Removed

mithril-stm/src/protocol/aggregate_signature/clerk.rs
damrobi: Done

mithril-common/src/protocol/ancillary.rs
```diff
@@ -103,7 +105,7 @@ mod tests {
             protocol_message.rigid_preimage().as_slice()
         );
         assert_eq!(
-            genesis_data.genesis_message_preimage().to_vec(),
+            genesis_data.genesis_message_preimage().deref(),
```
jpraynaud, line 108: The derefed object can be retrieved directly like this: 
```suggestion
            &genesis_data.genesis_message_preimage(),
```

mithril-stm/src/circuits/halo2_ivc/certificate_proof.rs
```diff
@@ -55,14 +56,17 @@ pub(crate) fn verify_and_prepare_accumulator(
         &[&[public_inputs]],
         &mut transcript,
     )
-    .map_err(|_| IvcCircuitError::CertificateProofRejected)?;
+    .map_err(|_| IvcCircuitError::CertificateProofRejected)
```
jpraynaud, line 59: It'd be also good for debugging to have the detail of the error added to the `CertificateProofRejected`.

mithril-stm/src/protocol/aggregate_signature/ancillary_data.rs
```diff
@@ -18,10 +18,13 @@ use crate::{StmResult, codec};
 /// Holds the prover-side state needed to produce the next certificate. Carried in the certificate,
 /// hashed into the certificate hash and transmitted in the certificate message. Variants map to the
 /// aggregate signature types that require prover data.
+#[cfg_attr(test, allow(clippy::large_enum_variant))]
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub enum AncillaryProverData {
     #[cfg(feature = "future_snark")]
     IvcSnark(IvcRollingState),
+    #[cfg(all(feature = "future_snark", test))]
+    FutureVariant,
```
jpraynaud, line 27:
```suggestion
    Future,
```

mithril-stm/src/protocol/aggregate_signature/clerk.rs
```diff
@@ -294,5 +307,154 @@ fn ivc_prover_input_preparation_and_prove<D: MembershipDigest>(
...
+            ivc_fixed_bases: BTreeMap::new(),
+            combined_fixed_bases: BTreeMap::new(),
+        });
+
+        (ivc_setup, midnight_vk)
+    }
+
+    #[test]
+    fn ivc_prover_input_preparation_fails_when_prover_data_carries_no_ivc_rolling_state() {
+        use crate::SchnorrVerificationKey;
```
jpraynaud, line 377: Could you move all imports to the beginning of the module

PR #3349, opened 2026-06-22
refactor(mithril-stm): refactor and stabilize the SNARK setups

### LINKED ISSUE #3300: Refactor unsafe SNARK setup
## Why
  
  The SNARK setups (recursive and non recursive) need to be refactored and stabilized.
  
  ## What
  
  Refactor and stabilize the SNARK setups (recursive and non recursive).
  
  ## How

- [x] Align the implementations of `SnarkSetup` and `IvcProverSetup`:
  - [x] Align names (rename `SnarkSetup` → `SnarkProverSetup`)
  - [x] Align structures (shared concepts only; keep IVC-specific fields)
  - [x] Non-recursive already has both a prover and verifier setup (`SnarkSetup` + `SnarkVerifierSetup`)
- [x] Rename `proof_system/halo2_snark/unsafe_helpers.rs` to `setup.rs` and drop the test-only unsafe SRS plumbing
- [x] Remove `ivc_halo2_snark/unsafe_setup_helpers.rs` (`TempCertificateKeyProvider`, `TempIvcKeyProvider`) and re-point every `load` call site (`prover_setup.rs`, `prover_input.rs`, `proof.rs`)
- [x] Add the (de)serialization of the verifying/proving keys:
  - [x] `CircuitKeySerialization` trait + impls in the circuit modules; `CircuitVerificationKey` stays in proof_system; raw recursive aliases moved to `circuits/halo2_ivc`
  - [x] Typed access from `CircuitKeyCache`
- [x] `RecursiveCircuitKeyProvider` downsizes keygen to the IVC circuit degree; `IvcProverSetup` stores the downsized SRS
- [x] Rename `K` → `RECURSIVE_CIRCUIT_DEGREE`
- [x] Wire all the tests with deterministic proving setup (and avoid flakiness with correct folder setup)

**Realign:** the cache/provider plumbing above is reworked into one compute-on-miss provider per circuit, with per-circuit key newtypes and a shared content-keyed test cache.

- [x] Turn the circuit key cache into a verification key provider:
  - [x] Add a `CircuitKeyGenerator` trait
  - [x] Add a generic compute-on-miss `CircuitVerificationKeyProvider<C>` alongside `CircuitKeyCache`
  - [x] Make the key-pair store atomic (per-writer-unique temp file; rename proving key then verifying key)
- [x] Compute the certificate key through the verification key provider:
  - [x] Add `NonRecursiveCircuitVerifyingKey` / `NonRecursiveCircuitProvingKey` newtypes
  - [x] `impl CircuitKeyGenerator for StmCertificateCircuit`
  - [x] Route `SnarkProverSetup` through the provider; delete the `get_or_build_snark_keys_with_disk_cache` free function
  - [x] Reconcile `CircuitVerificationKey` with `NonRecursiveCircuitVerifyingKey` (byte-compatible `SnarkProof`; final removal of `CircuitVerificationKey` deferred to the recursive step — #3348's `IvcVerifierData` still uses its serde)
- [x] Use the provider on the recursive side (rename `IvcProverSetup` → `IvcSnarkProverSetup`):
  - [x] Add `RecursiveCircuitVerifyingKey` / `RecursiveCircuitProvingKey` newtypes
  - [x] `impl CircuitKeyGenerator for IvcCircuitData`
  - [x] Rewire `load`: certificate VK → `downsize(RECURSIVE_CIRCUIT_DEGREE)` → recursive provider via `for_recursive_circuit(certificate_vk)`
  - [x] The recursive compute path lives in the `CircuitKeyGenerator` impl (no separate `key_providers.rs` — the #3348 rebase already dropped our wrapper providers)
  - [x] Complete the Step-1 rename (deferred): fold `CircuitKeyCache` into `CircuitVerificationKeyProvider` and delete `circuits/key_cache.rs` (no `CircuitKeyCache` references remain)
  - [x] Delete #3348's `unsafe_setup_helpers.rs` (`Temp*KeyProvider`s); keep and rework `temp_ivc_prover_cache.rs`'s process-wide setup memo to build via the production providers (no behaviour change to the `clerk.rs` path)
  - [x] Finish the certificate-VK reconciliation deferred from Step 2: migrate `IvcVerifierData` serde to the key wrappers and delete `proof_system/halo2_snark/circuit_verification_key.rs`, guarded by byte-for-byte verifier-data roundtrip tests
  - [x] Ripple the key newtypes through the prover/verifier consumers at the proof-system/setup boundaries; keep raw `VerifyingKey`/`ProvingKey` internal to `circuits::halo2_ivc` where Halo2 APIs require them
- [x] Share a content-keyed key cache across the slow tests:
  - [x] Re-add the empty-golden skip in `get_verification_key` (moved here from Step 1; it regressed `different_parameters_prove_and_verify_fails`) — prerequisite for the shared cache
  - [x] Add `fs2` dev-dependency; one shared content-keyed cache under `target/` with a per-config lock
  - [x] Route the slow certificate/IVC setup tests through the shared cache (replace the per-test isolation)

hjeljeli32:
## Content

This PR refactors and stabilizes both SNARK setups (the non-recursive certificate circuit and the recursive IVC circuit). It replaces the test-only "unsafe" setup plumbing and the duplicated key-computation paths with a single compute-on-miss key provider backed by a typed on-disk cache, introduces per-circuit key newtypes with their own (de)serialization, aligns the prover/verifier setup types and names across both circuits, and makes the test setup deterministic, race-free, and key-reusing.

### Changes

- **Key provider + cache**: `CircuitVerificationKeyProvider<C>` owns the on-disk key cache and computes the key pair on a miss through the circuit's `CircuitKeyGenerator`; the standalone key cache was folded into the provider. The cached verifying key is checked for staleness against the embedded production verifying-key bytes.
- **Per-circuit key newtypes**: `NonRecursiveCircuit{Verifying,Proving}Key` and `RecursiveCircuit{Verifying,Proving}Key` wrap the Midnight and raw PLONK key types, each with its own key (de)serialization. The raw `PlonkVerifyingKey`/`PlonkProvingKey` aliases remain only at the Halo2/verifier boundary.
- **Non-recursive setup**: the certificate verifying key is computed through the verification key provider; renamed `SnarkSetup` → `SnarkProverSetup`.
- **Recursive setup**: renamed `IvcProverSetup` → `IvcSnarkProverSetup`; it loads a single SRS, derives the certificate verifying key through the provider, downsizes the SRS to the recursive circuit degree, builds the recursive provider from that certificate key, and derives the IVC key pair — aligning the setup types and names across both circuits.
- **IVC setup memo**: renamed the process-wide IVC prover-setup memo `temp_ivc_prover_cache` → `ivc_prover_setup_cache` (it memoizes the assembled setup on the clerk path, not key computation).
- **Naming**: renamed the IVC circuit degree constant to `RECURSIVE_CIRCUIT_DEGREE`.

### Tests

- The slow SNARK tests share a single content-keyed key cache under the Cargo target directory (keyed by the protocol parameters, Merkle-tree depth, unsafe-SRS identity, and the embedded production verifying keys), so the first test to need a configuration pays key generation and the rest — and later runs — reuse it. An exclusive lock serializes cold-start key generation across parallel test processes.
- The certificate key is shared across the non-recursive setup builders for a given configuration; `SnarkProof::try_new_with_unsafe_srs` is documented as being only for the deterministic unsafe SRS.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation
  - [x] No new TODOs introduced

## Issue(s)
Closes https://github.com/input-output-hk/mithril/issues/3300


mithril-stm/src/proof_system/halo2_snark/circuit_verification_key.rs
```diff
@@ -24,18 +23,13 @@ impl CircuitVerificationKey {
     // TODO: remove this once the 'to_bytes' function is used.
     #[allow(dead_code)]
```
damrobi: Can we remove this `dead_code` or not yet?

mithril-stm/src/proof_system/halo2_snark/setup.rs
```diff
@@ -0,0 +1,326 @@
...
+    /// downsized to the exact circuit degree before key derivation (required by `keygen_vk`).
+    pub(crate) fn try_new_with_srs(
+        params: &Parameters,
+        merkle_tree_depth: u32,
+        mut srs: ParamsKZG<Bls12>,
+        key_cache: &CircuitKeyCache,
+    ) -> StmResult<Self> {
+        let circuit = StmCertificateCircuit::try_new(params, merkle_tree_depth)?;
+        let circuit_degree = MidnightCircuit::from_relation(&circuit).min_k();
+        zk::downsize_srs_for_relation(&mut srs, &circuit);
```
damrobi, line 68: I think this operation does almost nothing when the degree of the srs equals the degree of the circuit but do we want to add a condition to downsize only when necessary?

mithril-stm/src/circuits/halo2/key_serialization.rs
jpraynaud, line 1: The trait could be more general (and also splitted), live upper in the `mithril-stm/codec.rs`, and be implemented here for the verification keys:

```rust
/// Traits for serializing to bytes
pub trait TryToBytes {
    /// Try to convert to a bytes vector.
    fn to_bytes_vec(&self) -> StmResult<Vec<u8>>;
}

/// Traits for deserializing from bytes
pub trait TryFromBytes: Sized {
    /// Try to convert from a bytes slice.
    fn try_from_bytes(bytes: &[u8]) -> StmResult<Self>;
}
```

mithril-stm/src/circuits/halo2/key_serialization.rs
```diff
@@ -0,0 +1,67 @@
...
+
+    fn deserialize_key(bytes: &[u8]) -> StmResult<Self> {
+        let mut reader = bytes;
+        MidnightPK::<StmCertificateCircuit>::read(&mut reader, KEY_SERDE_FORMAT)
+            .with_context(|| "Failed to deserialize the certificate proving key")
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
jpraynaud, line 54: It's probably worth adding some golden tests on top of the roundtrips.

damrobi (approved):
LGTM overall 👍
I think @jpraynaud comments on the PR #3348 about downsizing the srs will apply to the key provider of the IVC
A more general remark, are we planning to store/download the srs of degree 19 for the IVC circuit instead of downsizing? The downsizing operation appears to be pretty slow sometimes.

mithril-stm/src/circuits/key_cache.rs
```diff
@@ -8,33 +8,41 @@ use std::{
...
 /// Outcome of [`CircuitKeyCache::validate`].
 pub enum CacheState {
-    /// On-disk VK matches the expected bytes and the PK file is also present.
-    Valid,
     /// No VK on disk, or the cache was stale and has been cleared.
     Empty,
+    /// On-disk VK matches the expected bytes, but the PK file is absent (partial cache).
+    VerificationKeyOnly,
+    /// On-disk VK matches the expected bytes and the PK file is also present.
+    ValidKeyPair,
```
jpraynaud: I don't see why we'd need to make the cache more complex here by adding these 2 states. You can implement the `get_verification_key` and `get_proving_key` functions and let them return an error if the something goes wrong during the deserialization.

mithril-stm/src/circuits/halo2_ivc/mod.rs
```diff
@@ -64,16 +66,23 @@ pub(crate) mod types;
...
 
-type E = <S as SelfEmulation>::Engine;
+pub(crate) type E = <S as SelfEmulation>::Engine;
 type CBase = <C as CircuitCurve>::Base;
 
 type NG = NativeGadget<F, P2RDecompositionChip<F>, NativeChip<F>>;
 
+/// PLONK verifying key shared by the certificate and recursive circuits
+/// (both use the same field/curve parameterization).
+pub(crate) type PlonkVerifyingKey = VerifyingKey<F, KZGCommitmentScheme<E>>;
```
jpraynaud: Why not use the `MidnightVK` which is the ultimate format that we need?

mithril-stm/src/proof_system/halo2_snark/proof.rs
```diff
@@ -73,13 +76,20 @@ impl<D: MembershipDigest> SnarkProof<D> {
     /// caller-supplied SRS instead of the trusted production SRS, so a proof can be reconstructed
     /// against the same unsafe SRS it was generated with.
     #[cfg(test)]
-    pub(crate) fn try_new_with_srs(
+    pub(crate) fn try_new_with_srs<T: Into<std::path::PathBuf>>(
```
jpraynaud: Could you move the imports at the top of the module?
```suggestion
    pub(crate) fn try_new_with_srs<T: Into<PathBuf>>(
```

mithril-stm/src/proof_system/halo2_snark/setup.rs
```diff
@@ -0,0 +1,326 @@
...
+struct SnarkProverSetupCacheKey {
+    circuit_degree: u32,
+    k: u64,
+    m: u64,
+    merkle_tree_depth: u32,
+}
+
+type SnarkProverSetupKeyPair = Arc<(MidnightVK, MidnightPK<StmCertificateCircuit>)>;
+
+static SNARK_PROVER_SETUP_KEY_CACHE: LazyLock<
```
jpraynaud: There is already a cache for the verification keys, and the trusted setup, what is the purpose of this other cache?

mithril-stm/src/circuits/halo2_ivc/mod.rs
curiecrypt: In the recursive circuit, we do not use the wrapping structure MidnightVK, we directly use the Plonk verifying key

mithril-stm/src/circuits/halo2/key_serialization.rs
curiecrypt: Also, do we need the roundtrips for PK as well?

mithril-stm/src/circuits/halo2_ivc/key_serialization.rs
```diff
@@ -0,0 +1,84 @@
...
+            &mut reader,
+            KEY_SERDE_FORMAT,
+            (),
+        )
+        .with_context(|| "Failed to deserialize the recursive PLONK proving key")
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
curiecrypt, line 67: similar to the non recursive one, do we need the rountrip for PK?

mithril-stm/src/proof_system/ivc_halo2_snark/key_providers.rs
```diff
@@ -0,0 +1,109 @@
...
+    cache: CircuitKeyCache,
+}
+
+#[allow(dead_code)]
+impl RecursiveCircuitKeyProvider {
+    pub(crate) fn new(cache: CircuitKeyCache) -> Self {
+        Self { cache }
+    }
+
+    /// Returns the IVC verifying and proving keys, deriving them on a cache miss. The SRS must
```
curiecrypt: `RecursiveCircuitKeyProvider::get_key_pair` requires an already-downsized SRS by documentation only, while the sibling cert provider downsizes internally. 
Should we add an `assert srs.max_k() == RECURSIVE_CIRCUIT_DEGREE` at entry or downsize a clone internally to mirror the cert provider?

mithril-stm/src/circuits/key_cache.rs
```diff
@@ -49,39 +57,96 @@ fn remove_if_exists(path: &Path) -> StmResult<()> {
...
+    ) -> StmResult<()> {
+        if let Some(parent) = self.verification_key_path.parent() {
+            fs::create_dir_all(parent)
+                .with_context(|| "Failed to create the circuit key cache directory")?;
+        }
+        fs::write(&self.verification_key_path, verification_key_bytes)
+            .with_context(|| "Failed to write the verification key to the cache")?;
+        if let Err(error) = fs::write(&self.proving_key_path, proving_key_bytes) {
+            let _ = fs::remove_file(&self.verification_key_path);
+            return Err(error).with_context(|| "Failed to write the proving key to the cache");
```
curiecrypt: Should we add a test for PK-write-failure rollback?

curiecrypt (commented):
Thanks @hjeljeli32, this PR will resolve a lot of issues we have been struggling.
I have a suggestion that would also avoid some ambiguotions we might have and @jpraynaud 's comment about using MidnightVK.

The certificate and recursive circuits share the same raw key type (`VerifyingKey<F, KZGCommitmentScheme<E>>`), and `MidnightVK::vk()` hands back that same type. Since `PlonkVerifyingKey` is only an alias for it, the `CircuitKeySerialization` impl is effectively defined on the shared raw type, and `deserialize_key` always rebuilds with `IvcCircuitData`. It works today because only recursive keys are ever deserialized raw, but nothing stops a certificate key (same type) from being decoded as the IVC circuit, and that would only surface at proving or verification time.

Proposal: give each circuit its own clearly named key types, so the names answer the "why not MidnightVK" question and the serialization impls become unambiguous.

```rust
// circuits/halo2 (non-recursive)
pub struct NonRecursiveCircuitVerifyingKey(MidnightVK);
pub struct NonRecursiveCircuitProvingKey(MidnightPK<StmCertificateCircuit>);

// circuits/halo2_ivc (recursive)
pub struct RecursiveCircuitVerifyingKey(VerifyingKey<F, KZGCommitmentScheme<E>>);
pub struct RecursiveCircuitProvingKey(ProvingKey<F, KZGCommitmentScheme<E>>);
```

A few notes so this lands cleanly:

1. Make the recursive pair real newtypes, not type aliases. A renamed alias is still the shared raw type, so it reads better but does not stop a certificate key from matching the recursive impl. If you want to keep them as aliases, add a degree check (`== RECURSIVE_CIRCUIT_DEGREE`) inside the recursive `deserialize_key` so a wrong-circuit decode fails loudly instead of silently.
2. `circuit_verification_key.rs` is not dead code. `CircuitVerificationKey` is the serde-embedded VK field in `SnarkProof` and exposes `get_midnight_vk()` and `from_bytes()`, both of which are used. So this is a relocate and rename into the non-recursive keys file, carrying `new`, `from_bytes`, `get_midnight_vk`, the serde derive, and the `midnight_vk_serde` module. It is not a delete.
3. Once the impl lives on `NonRecursiveCircuitVerifyingKey`, `midnight_vk_serde` and `CircuitKeySerialization for MidnightVK` encode the same "MidnightVK to RawBytes" logic. Point the `#[serde(with = ...)]` at `serialize_key` and `deserialize_key` so there is a single source of truth.
4. `CircuitVerificationKey::to_bytes` is only used by a `#[cfg(test)]` golden test, so gate it with `#[cfg(test)]` rather than `#[allow(dead_code)]`.
5. `MidnightPK` is generic, so the non-recursive proving key wraps `MidnightPK<StmCertificateCircuit>`.

This also sets up the later move to generic `TryToBytes` / `TryFromBytes` traits in `codec.rs`: distinct newtypes avoid conflicting impls on the shared foreign key type.

We can have a discussion about my proposal and decide on the best choice.

mithril-stm/src/proof_system/halo2_snark/circuit_verification_key.rs
hjeljeli32: Good catch, it is actually used by the snark_proof_to_from_bytes round-trip test, so I kept the method but gated it #[cfg(test)]

mithril-stm/src/proof_system/halo2_snark/setup.rs
hjeljeli32: `downsize_srs_for_relation` just calls `srs.downsize_from_circuit(...)` → ultimately `ParamsKZG::downsize`, which already early-returns if `self.max_k() == new_k`

mithril-stm/src/circuits/key_cache.rs
hjeljeli32: Sure I'll collapse the states so `get_verification_key/get_proving_key` read and deserialize directly.

mithril-stm/src/proof_system/halo2_snark/setup.rs
hjeljeli32: @jpraynaud 
My understanding was that the clerk builds a fresh SnarkProverSetup on every aggregation and does not keep it, so repeated aggregations rebuild the certificate keys each time.
My intention with the cache was to keep the derived keys in memory so those repeated builds within the process skip deriving them and reading them back from disk.
After your remark and looking more at the code, I realized that the on-disk key cache already avoids the expensive part, deriving the keys, and persists them across restarts, while the SRS cache covers the SRS. I will remove it.

mithril-stm/src/proof_system/ivc_halo2_snark/key_providers.rs
hjeljeli32: Actually `downsize_srs_for_relation` just calls `srs.downsize_from_circuit(...)` → ultimately `ParamsKZG::downsize`, which already early-returns if `self.max_k() == new_k`

hjeljeli32:
> Thanks @hjeljeli32, this PR will resolve a lot of issues we have been struggling. I have a suggestion that would also avoid some ambiguotions we might have and @jpraynaud 's comment about using MidnightVK.
> 
> The certificate and recursive circuits share the same raw key type (`VerifyingKey<F, KZGCommitmentScheme<E>>`), and `MidnightVK::vk()` hands back that same type. Since `PlonkVerifyingKey` is only an alias for it, the `CircuitKeySerialization` impl is effectively defined on the shared raw type, and `deserialize_key` always rebuilds with `IvcCircuitData`. It works today because only recursive keys are ever deserialized raw, but nothing stops a certificate key (same type) from being decoded as the IVC circuit, and that would only surface at proving or verification time.
> 
> Proposal: give each circuit its own clearly named key types, so the names answer the "why not MidnightVK" question and the serialization impls become unambiguous.
> 
> ```rust
> // circuits/halo2 (non-recursive)
> pub struct NonRecursiveCircuitVerifyingKey(MidnightVK);
> pub struct NonRecursiveCircuitProvingKey(MidnightPK<StmCertificateCircuit>);
> 
> // circuits/halo2_ivc (recursive)
> pub struct RecursiveCircuitVerifyingKey(VerifyingKey<F, KZGCommitmentScheme<E>>);
> pub struct RecursiveCircuitProvingKey(ProvingKey<F, KZGCommitmentScheme<E>>);
> ```
> 
> A few notes so this lands cleanly:
> 
> 1. Make the recursive pair real newtypes, not type aliases. A renamed alias is still the shared raw type, so it reads better but does not stop a certificate key from matching the recursive impl. If you want to keep them as aliases, add a degree check (`== RECURSIVE_CIRCUIT_DEGREE`) inside the recursive `deserialize_key` so a wrong-circuit decode fails loudly instead of silently.
> 2. `circuit_verification_key.rs` is not dead code. `CircuitVerificationKey` is the serde-embedded VK field in `SnarkProof` and exposes `get_midnight_vk()` and `from_bytes()`, both of which are used. So this is a relocate and rename into the non-recursive keys file, carrying `new`, `from_bytes`, `get_midnight_vk`, the serde derive, and the `midnight_vk_serde` module. It is not a delete.
> 3. Once the impl lives on `NonRecursiveCircuitVerifyingKey`, `midnight_vk_serde` and `CircuitKeySerialization for MidnightVK` encode the same "MidnightVK to RawBytes" logic. Point the `#[serde(with = ...)]` at `serialize_key` and `deserialize_key` so there is a single source of truth.
> 4. `CircuitVerificationKey::to_bytes` is only used by a `#[cfg(test)]` golden test, so gate it with `#[cfg(test)]` rather than `#[allow(dead_code)]`.
> 5. `MidnightPK` is generic, so the non-recursive proving key wraps `MidnightPK<StmCertificateCircuit>`.
> 
> This also sets up the later move to generic `TryToBytes` / `TryFromBytes` traits in `codec.rs`: distinct newtypes avoid conflicting impls on the shared foreign key type.
> 
> We can have a discussion about my proposal and decide on the best choice.

@jpraynaud @damrobi 
Thanks @curiecrypt, really good proposal. It fixes a real issue: `PlonkVerifyingKey / PlonkProvingKey `are just aliases over the shared raw Midnight key type, so the codec lives on that shared type and always deserializes with `IvcCircuitData`. As you said, nothing stops a certificate key (same type via `MidnightVK::vk()`) from being decoded as the IVC circuit, and that will fail only at proving/verification.

IMO, the newtype direction is the right fix: it makes the cross-circuit decode impossible by construction, gives each key its own codec. A couple of your points are already in this PR (the generic `TryToBytes/TryFromBytes` traits in `codec.rs`, and `to_bytes` gated with #[cfg(test)])

The one challenge is scope: the aliases are used in ~37 places across 9 files, and it touches the same IVC area @damrobi is changing in #3348, so folding it into this PR would  risk conflicts. My suggestion: do the newtypes as a follow-up ticket, coordinated with @damrobi once #3348 is merged. WDYT?

mithril-stm/src/proof_system/halo2_snark/setup.rs
```diff
@@ -115,45 +91,30 @@ impl SnarkVerifierSetup {
...
+    // The proving key is only read once the verifying key is present, so an orphan proving key left
+    // by a failed store (which removes the verifying key) is recomputed rather than surfaced as a
+    // deserialization error.
+    if let Some(verification_key) = disk_cache.get_verification_key::<MidnightVK>()?
+        && let Some(proving_key) =
+            disk_cache.get_proving_key::<MidnightPK<StmCertificateCircuit>>()?
     {
-        return Ok(pair);
+        return Ok((verification_key, proving_key));
     }
```
jpraynaud: This should be a guarantee of the `CircuitKeyCache` (and I believe it is once the `validate` function has been called).

mithril-stm/src/circuits/halo2/keys.rs
```diff
@@ -0,0 +1,239 @@
...
+pub(crate) struct NonRecursiveCircuitProvingKey(MidnightPK<StmCertificateCircuit>);
+
+impl NonRecursiveCircuitVerifyingKey {
+    /// Wraps a Midnight verifying key.
+    pub(crate) fn new(midnight_vk: MidnightVK) -> Self {
+        Self(midnight_vk)
+    }
+
+    /// Borrows the wrapped Midnight verifying key.
+    pub(crate) fn midnight_vk(&self) -> &MidnightVK {
```
jpraynaud, line 35: This could be implemented with `Deref` trait.

mithril-stm/src/circuits/halo2/keys.rs
```diff
@@ -0,0 +1,239 @@
...
+    ) -> Result<MidnightVK, D::Error> {
+        let bytes: Vec<u8> = serde::Deserialize::deserialize(deserializer)?;
+        MidnightVK::read(&mut bytes.as_slice(), SerdeFormat::RawBytes)
+            .map_err(serde::de::Error::custom)
+    }
+}
+
+impl NonRecursiveCircuitProvingKey {
+    /// Borrows the wrapped Midnight proving key, for proof generation.
+    pub(crate) fn midnight_pk(&self) -> &MidnightPK<StmCertificateCircuit> {
```
jpraynaud, line 80: This also could be implemented with `Deref` trait.

mithril-stm/src/circuits/halo2/keys.rs
```diff
@@ -0,0 +1,239 @@
...
+        &self.0
+    }
+}
+
+impl CircuitVerificationKeyProvider<StmCertificateCircuit> {
```
jpraynaud: Maybe this should live in the `CircuitVerificationKeyProvider` module?

mithril-stm/src/circuits/halo2/keys.rs
```diff
@@ -0,0 +1,239 @@
...
+                .expect("deserialize should succeed");
+        assert_eq!(
+            proving_key_bytes,
+            restored_proving_key.to_bytes_vec().unwrap(),
+            "proving key bytes must be stable across a round trip"
+        );
+    }
+
+    #[test]
+    fn provider_returns_cached_verifying_key_on_hit() {
```
jpraynaud: This test looks like you are trying to test underlying implementation details and can probably be removed.

mithril-stm/src/circuits/halo2_ivc/certificate_proof.rs
```diff
@@ -38,7 +40,7 @@ use crate::{
 pub(crate) fn verify_and_prepare_accumulator(
     proof_bytes: &[u8],
     public_inputs: &[CircuitBase],
-    circuit_verification_key: &CircuitVerifyingKey,
+    circuit_verification_key: &PlonkVerifyingKey,
```
jpraynaud: We should probably use the wrapper type instead of the `PlonkVerifyingKey` here?

mithril-stm/src/circuits/halo2_ivc/mod.rs
```diff
@@ -65,16 +69,23 @@ pub(crate) mod tests;
...
 
-type E = <S as SelfEmulation>::Engine;
+pub(crate) type E = <S as SelfEmulation>::Engine;
 type CBase = <C as CircuitCurve>::Base;
 
 type NG = NativeGadget<F, P2RDecompositionChip<F>, NativeChip<F>>;
 
+/// Raw PLONK verifying key used directly by the recursive circuit — its own key and the certificate
+/// verifying key it consumes. The certificate circuit itself wraps this key type as `MidnightVK`.
+pub(crate) type PlonkVerifyingKey = VerifyingKey<F, KZGCommitmentScheme<E>>;
```
jpraynaud: I'm not completely convinced that we want to add another name for designating the verification key as it makes the code hard to read and maintain.

mithril-stm/src/circuits/key_generator.rs
jpraynaud, line 1: You can remove the `circuit_` prefix` in this sub-module name as it is redundant with the module name.

mithril-stm/src/circuits/key_provider.rs
jpraynaud, line 1: You can rename this module to `key_provider.rs`.

mithril-stm/src/circuits/circuit_verification_key_provider.rs
```diff
@@ -0,0 +1,563 @@
...
+use crate::StmResult;
+use crate::codec::{TryFromBytes, TryToBytes};
+
+use super::MITHRIL_CIRCUIT_CACHE_FOLDER;
+use super::circuit_key_generator::CircuitKeyGenerator;
+
+/// Provides a circuit's verifying and proving keys: an on-disk cache (with staleness detection)
+/// plus the [`CircuitKeyGenerator`] circuit that computes them on a miss.
+#[allow(dead_code)] // consumed by the certificate and recursive setups
+pub(crate) struct CircuitVerificationKeyProvider<C: CircuitKeyGenerator> {
```
jpraynaud: This name would be more accurate (my bad for suggesting to include `Verification` in it):
```suggestion
pub(crate) struct CircuitKeyProvider<C: CircuitKeyGenerator> {
```

mithril-stm/src/proof_system/halo2_snark/proof.rs
```diff
@@ -59,47 +65,69 @@ impl<D: MembershipDigest> SnarkProof<D> {
...
-    /// Test-only variant of [`Self::try_new`] that derives the circuit verification key from a
-    /// caller-supplied SRS instead of the trusted production SRS, so a proof can be reconstructed
-    /// against the same unsafe SRS it was generated with.
+    /// Test-only variant of [`Self::try_new`] that derives the circuit verification key from the
+    /// deterministic unsafe SRS the proof was generated with, instead of the trusted production SRS.
+    /// It is only for that seed-pinned unsafe SRS: the shared key cache identifies the SRS by its
+    /// fixed seed, not by inspecting the supplied SRS object, so an unrelated SRS would mis-key it.
     #[cfg(test)]
-    pub(crate) fn try_new_with_srs(
+    pub(crate) fn try_new_with_unsafe_srs(
```
jpraynaud:
```suggestion
    pub(crate) fn try_new_with_srs(
```

mithril-stm/src/circuits/test_utils/key_cache.rs
jpraynaud, line 1: I'm a bit confused about this module:
- The need is to create a synchronization mechanism based on the file system
- It's very procedural
- It's too complex given our need

It should be named something like `file_mutex.rs`, provide a `FileMutex` struct which can be create an instance with a `lock_id` (for any lock id there is one file on the system). Then we can use it with a similar pattern as the `Mutex` like `FileMutex::new(lock_id).lock()` that will create the exclusive read access to the file (and thus give access to one process and block all the others, preventing the race conditions we have).

mithril-stm/src/circuits/key_provider.rs
```diff
@@ -0,0 +1,563 @@
...
+                }
+            };
+
+        if let Err(error) = fs::rename(&proving_key_temp, &self.proving_key_path) {
+            let _ = fs::remove_file(&proving_key_temp);
+            let _ = fs::remove_file(&verification_key_temp);
+            return Err(error).with_context(|| "Failed to store the proving key in the cache");
+        }
+        if let Err(error) = fs::rename(&verification_key_temp, &self.verification_key_path) {
+            let _ = fs::remove_file(&verification_key_temp);
```
damrobi, line 186: We don't want to remove the proving key file in that case ?

mithril-stm/src/proof_system/halo2_snark/proof.rs
```diff
@@ -45,7 +51,7 @@ pub struct SnarkProof<D: MembershipDigest> {
     // Note: this field is temporary and will be removed once the circuit setup is stable
     merkle_tree_depth: u32,
     /// Circuit verification key
-    circuit_verification_key: CircuitVerificationKey,
+    circuit_verification_key: NonRecursiveCircuitVerifyingKey,
```
damrobi, line 45: Should we continue to store the verification key with the proof or try to get it from the cache when trying to verify the proof?

mithril-stm/src/proof_system/halo2_snark/setup.rs
```diff
@@ -0,0 +1,188 @@
...
+/// This includes the Structured Reference String (SRS), the compiled circuit, and the
+/// proving and verification keys derived from them.
+pub struct SnarkProverSetup {
+    /// KZG Structured Reference String.
+    pub(crate) srs: ParamsKZG<Bls12>,
+    /// Compiled STM circuit.
+    pub(crate) circuit: StmCertificateCircuit,
+    /// Verification key for the SNARK proof.
+    pub(crate) verification_key: NonRecursiveCircuitVerifyingKey,
+    /// Proving key for the SNARK proof.
```
damrobi, line 43: This might not be needed anymore if we remove the verification key from the proof but if we leave it this can stay.

mithril-stm/src/circuits/key_provider.rs
jpraynaud, line 1: The code of this module is hard to read and could probably benefit from a state machine like the previous implementation did. Also most the accessor functions are not pure.

mithril-stm/src/proof_system/halo2_snark/setup.rs
```diff
@@ -0,0 +1,188 @@
...
+    /// provider. The circuit is taken from the provider (single source); the SRS must have
+    /// `max_k >= circuit.min_k()` and is downsized to that circuit's degree and stored for proving;
+    /// the key pair is read from, or derived through, the provider.
+    pub(crate) fn try_new_with_srs(
+        mut srs: ParamsKZG<Bls12>,
+        provider: &CircuitVerificationKeyProvider<StmCertificateCircuit>,
+    ) -> StmResult<Self> {
+        let circuit = provider.circuit().clone();
+        zk::downsize_srs_for_relation(&mut srs, &circuit);
+        let (verification_key, proving_key) = provider.key_pair(&srs)?;
```
damrobi, line 69: This `key_pair()` function can clone the srs to downsize it again if the keys are not available. Maybe it's worth checking if we can remove the second downsizing as it demands cloning the full SRS.

mithril-stm/src/proof_system/ivc_halo2_snark/verifier_setup.rs
```diff
@@ -49,7 +51,7 @@ pub(crate) struct IvcVerifierSetup {
     /// `s_g2` (tau·G2) extracted from the embedded params; passed to the accumulator check.
     tau_g2: G2Affine,
     /// Verifying key of the IVC circuit.
-    ivc_verifying_key: CircuitVerifyingKey,
+    ivc_verifying_key: PlonkVerifyingKey,
```
damrobi: I think this should be the new type `RecursiveCircuitVerifyingKey`.

damrobi (commented):
Left a few non blocking comments I think. I haven't had time to look at the tests in details so I will do so on Monday and approve

mithril-stm/src/proof_system/halo2_snark/setup.rs
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+    /// Builds a `SnarkProverSetup` from a caller-supplied SRS and certificate verification key
+    /// provider. The circuit is taken from the provider (single source); the SRS must have
+    /// `max_k >= circuit.min_k()` and is downsized to that circuit's degree and stored for proving;
+    /// the key pair is read from, or derived through, the provider.
+    pub(crate) fn try_new_with_srs(
+        mut srs: ParamsKZG<Bls12>,
+        provider: &CircuitVerificationKeyProvider<StmCertificateCircuit>,
+    ) -> StmResult<Self> {
```
jpraynaud: Actually, I think that the inputs should be both a `TrustedSetupProvider` and a  `CircuitKeyProvider`, and that the `TrustedSetupProvider` should have a test-only constructor that creates the unsafe SRS. This would make the design much cleaner and would make the code for tests leaner and more readable.

mithril-stm/src/proof_system/ivc_halo2_snark/prover_setup.rs
```diff
@@ -55,48 +58,40 @@ pub(crate) struct IvcProverSetup {
...
-    /// verifying and proving keys come from the key providers, which today carry their own
-    /// SRS reference. The caller must ensure both SRS sources agree. A mismatch produces an
-    /// internally inconsistent `IvcProverSetup` that surfaces only at proving or verification time.
-    /// In practice: route all providers through the same `TrustedSetupProvider` instance
-    /// (temp design) or the same canonical trusted setup source (production cache design).
-    // TODO: swap `Temp*Provider` parameters for the production IVC cache providers
-    // once they ship.
+    /// `recursive_provider_factory` builds the recursive provider once the certificate verifying key
+    /// is known; production passes [`CircuitVerificationKeyProvider::for_recursive_circuit`].
     pub(crate) fn load(
```
jpraynaud, line 77: We should try to align as much as possible the function names between recursive and non-recursive.

mithril-stm/src/proof_system/ivc_halo2_snark/prover_setup.rs
```diff
@@ -55,48 +58,40 @@ pub(crate) struct IvcProverSetup {
...
         trusted_setup_provider: &TrustedSetupProvider,
-        certificate_key_provider: &TempCertificateKeyProvider,
-        ivc_key_provider: &TempIvcKeyProvider,
+        certificate_provider: &CircuitVerificationKeyProvider<StmCertificateCircuit>,
+        recursive_provider_factory: impl FnOnce(
+            &NonRecursiveCircuitVerifyingKey,
+        ) -> StmResult<
+            CircuitVerificationKeyProvider<IvcCircuitData>,
+        >,
```
jpraynaud: Why not inject the non-recursive `CircuitKeyProvider` in the recursive one and provide only the recursive one to this function instead? This would avoid this complex constructor.

mithril-stm/src/proof_system/ivc_halo2_snark/prover_setup.rs
```diff
@@ -145,62 +140,82 @@ impl IvcProverSetup {
...
+pub(crate) fn build_unsafe_ivc_setup(
+    parameters: crate::Parameters,
+    merkle_tree_depth: u32,
+) -> StmResult<IvcSnarkProverSetup> {
+    use crate::circuits::{
+        halo2::NON_RECURSIVE_CIRCUIT_VERIFICATION_KEY_FOR_PRODUCTION,
+        halo2_ivc::RECURSIVE_CIRCUIT_VERIFICATION_KEY_FOR_PRODUCTION,
+        test_utils::key_cache::with_shared_key_cache,
+        trusted_setup::{UNSAFE_SRS_SEED, build_provider_with_unsafe_srs},
+    };
```
jpraynaud: Can you avoid imports in function body?

mithril-stm/src/proof_system/ivc_halo2_snark/verifier_setup.rs
```diff
@@ -311,6 +276,109 @@ mod tests {
         );
     }
 
+    /// The verifying-key newtypes must not change the `IvcVerifierData` wire format. This mirrors the
+    /// bare-key shape (a `MidnightVK` and a `PlonkVerifyingKey`, each via a raw-bytes `#[serde(with)]`
+    /// module), serializes it through the same codec, then asserts the current struct decodes those
+    /// bytes and re-encodes them byte-for-byte.
+    mod wire_format_compatibility {
```
jpraynaud: I don't really understand what this tests do.

mithril-stm/src/proof_system/ivc_halo2_snark/ivc_prover_setup_cache.rs
jpraynaud, line 1: This module should stay as temporary or better be removed as it is only here to avoid re-computing SRS and circuit keys constantly.

mithril-stm/src/proof_system/halo2_snark/proof.rs
hjeljeli32: Good question @damrobi  🙂 My understanding is that now the proof carries the VK alongside params and merkle_tree_depth until the production params/depth are fixed, the certificate VK varies per configuration, so it can't be a single known constant yet.

Once the setup stabilizes, the VK should collapse to the production constant and verify would use that directly. So the plan would be to drop params, merkle_tree_depth, and the VK together in that stabilization step. Does that match how you see it going for production or do you see another path?

mithril-stm/src/proof_system/halo2_snark/setup.rs
hjeljeli32: Yes I agree👍 So it'll go away together with the proof's VK in the stabilization cleanup.

mithril-stm/src/circuits/halo2/keys.rs
hjeljeli32: @jpraynaud 
These accessors were intentionally named rather than Deref: because Deref would transparently expose the inner key's whole API and hide the explicit raw-key extraction the newtypes were introduced to make visible. If you'd prefer a standard conversion trait over named methods, we can use AsRef rather than Deref. WDYT?

mithril-stm/src/circuits/halo2/keys.rs
hjeljeli32: Same answer as above

mithril-stm/src/proof_system/ivc_halo2_snark/prover_setup.rs
hjeljeli32: Thanks @jpraynaud. I tried it out and want to share what I found.

The tricky part is the order things happen in. The recursive circuit is built from the certificate VK. But we only get that certificate key after load has loaded the SRS and computed it. So the recursive provider can't be handed to load ready-made: it can only be built once we're already inside load, with the SRS in hand. That's what the closure is doing today: it's the little "build the recursive provider once the certificate key is known" step.

I can definitely do what you're describing: bundle the certificate provider together with the recursive provider's config into one type and pass that to load, so load takes two arguments with no closure. That does read more nicely. The honest trade-off is that it moves that little build step from a closure into a method on the new type rather than removing it the ordering above forces the step to live somewhere.

Do you have a different shape in mind that I might be missing?

mithril-stm/src/proof_system/ivc_halo2_snark/ivc_prover_setup_cache.rs
hjeljeli32: @jpraynaud 
This cache is for the production proving path. It keeps the assembled IVC prover setup (the SRS plus the circuit keys) in memory after the first build, so we reuse it instead of rebuilding it on every proof.

That said, a process-global cache isn't the cleanest permanent thing. How would you like to change it?

mithril-stm/src/circuits/halo2/keys.rs
jpraynaud: We can keep the name accessors :+1:

mithril-stm/src/circuits/halo2/keys.rs
jpraynaud: OK, let's keep the named accessors.

mithril-stm/src/proof_system/ivc_halo2_snark/ivc_prover_setup_cache.rs
jpraynaud: We already have the `TrustedSetupProvider` and the `CircuitKeyProvider` which provide cache. I'm not convinced that we need to add another layer of cache at this stage. Also the implementation, I'm not convinced by the procedural implementation. I propose that we assess if we need another layer of cache and implement it in another PR as this one is already very dense.

mithril-stm/src/circuits/halo2/keys.rs
```diff
@@ -0,0 +1,242 @@
...
+        Self(midnight_vk)
+    }
+
+    /// Borrows the wrapped Midnight verifying key.
+    pub(crate) fn midnight_vk(&self) -> &MidnightVK {
+        &self.0
+    }
+}
+
+impl AsPlonkVerifyingKey for NonRecursiveCircuitVerifyingKey {
```
jpraynaud: Could have this been done with a `From` or `AsRef` trait?

mithril-stm/src/circuits/halo2_ivc/tests/common/generators/proofs.rs
```diff
@@ -43,7 +45,7 @@ where
 
 /// Verifies a recursive proof and returns the prepared MSM.
 fn verify_prepare_ivc_with_transcript<H: TranscriptHash>(
-    verifying_key: &VerifyingKey<F, KZGCommitmentScheme<E>>,
+    verifying_key: &impl AsPlonkVerifyingKey,
```
jpraynaud: IMO static dispatch is enough here.

mithril-stm/src/circuits/halo2_ivc/certificate_proof.rs
```diff
@@ -37,7 +38,7 @@ use crate::{
 pub(crate) fn verify_and_prepare_accumulator(
     proof_bytes: &[u8],
     public_inputs: &[CircuitBase],
-    circuit_verification_key: &CircuitVerifyingKey,
+    circuit_verification_key: &impl AsPlonkVerifyingKey,
```
jpraynaud: Same remark as above, this would be better with static dispatch.

mithril-stm/src/circuits/halo2_ivc/state.rs
```diff
@@ -216,11 +218,14 @@ pub(crate) fn trivial_acc(fixed_base_names: &[String]) -> Accumulator<S> {
 
 pub(crate) fn fixed_bases_and_names(
     vk_name: &str,
-    vk: &VerifyingKey<F, KZGCommitmentScheme<E>>,
+    vk: &impl AsPlonkVerifyingKey,
```
jpraynaud: Same remark as above about static dispatch.

mithril-stm/src/circuits/test_utils/file_mutex.rs
```diff
@@ -0,0 +1,80 @@
...
+            fs::create_dir_all(parent)?;
+        }
+        let lock_file = File::create(&self.lock_file_path)?;
+        lock_file.lock_exclusive()?;
+        Ok(FileMutexGuard { lock_file })
+    }
+}
+
+/// Holds the exclusive lock until dropped, like a [`std::sync::MutexGuard`].
+#[must_use = "the lock is released as soon as the guard is dropped"]
```
jpraynaud: This `must_use` can be removed.

mithril-stm/src/circuits/test_utils/key_cache.rs
jpraynaud, line 1: This key cache module is not really needed and could be implemented with a constructor of the `FileMutex`.

mithril-stm/src/proof_system/halo2_snark/proof.rs
jpraynaud: This could be placed in a new variant of the `AncillaryVerifierData`. 

Let's do that in another PR as this one is already dense.

mithril-stm/src/proof_system/halo2_snark/setup.rs
```diff
@@ -0,0 +1,221 @@
...
+    }
+}
+
+/// Builds a [`SnarkProverSetup`] from a deterministic, oversized unsafe SRS, exercising the real `load`
+/// path without the production SRS. Mirrors `build_unsafe_ivc_setup` on the recursive side. Shared by
+/// the slow certificate tests through a content-keyed cache keyed by the protocol parameters,
+/// Merkle-tree depth, the unsafe SRS seed, and the production verifying key as a circuit-version salt,
+/// so the certificate key is computed once and reused across tests and runs.
+#[cfg(test)]
+pub(crate) fn build_unsafe_certificate_setup(
```
jpraynaud: I'd be much easier to read if this was an associated function of the `SnarkProverSetup`. Also the name is not very clear and could be `build_for_test`.

mithril-stm/src/proof_system/ivc_halo2_snark/prover_setup.rs
```diff
@@ -145,62 +145,76 @@ impl IvcProverSetup {
...
+/// `load` path without the production SRS. Shared by the slow IVC tests through a content-keyed cache
+/// keyed by the protocol parameters, Merkle-tree depth, the unsafe SRS identity (degree and seed), and
+/// the production verifying keys as a circuit-version salt, so the recursive keys — the dominant cost —
+/// are computed once and reused across tests and runs.
 #[cfg(test)]
-mod tests {
-    use std::sync::Arc;
-
-    use tempfile::tempdir;
+pub(crate) fn build_unsafe_ivc_setup(
```
jpraynaud: Same remark as for the non recursive case, this function should be associated to the `IvcSnarkProverSetup` type.

mithril-stm/src/circuits/halo2_ivc/key_serialization.rs
```diff
@@ -0,0 +1,99 @@
...
+        )
+        .expect("production recursive verifying key bytes should deserialize");
+        let first = verifying_key.to_bytes_vec().expect("serialize should succeed");
+        let restored = VerifyingKey::<F, KZGCommitmentScheme<E>>::try_from_bytes(&first)
+            .expect("re-deserialize should succeed");
+        let second = restored.to_bytes_vec().expect("re-serialize should succeed");
+        assert_eq!(
+            first, second,
+            "recursive verifying key bytes must be stable across a round trip"
+        );
```
damrobi: Not sure it's needed since we start by deserializing the const bytes, but then it becomes similar to the following test

damrobi (approved):
LGTM 👍 
Just one small remark regarding a test

mithril-stm/src/proof_system/ivc_halo2_snark/prover_setup.rs
damrobi: Regarding this, I was also wondering if we could reuse the `TrustedSetupProvider` created in the `SnarkProver` called to generate the proof since they are both `Default`. Not sure if there is a real gain since the first call will fill the cache if it is empty and it would probably require implementing more function on the `SnarkProver`.

jpraynaud: I think we need to create a `RecursiveCircuitKeyGenerator` from a non recursive key provider and that implements the `KeyGenerator` trait. This would remove the need for the lambda and would help mutualize the SRS:
```rust
struct RecursiveCircuitKeyGenerator {
    non_recursive_key_provider: KeyProvider<StmCertificateCircuit>,
}

impl RecursiveCircuitKeyGenerator {
    pub fn new(non_recursive_key_provider: KeyProvider<StmCertificateCircuit>) -> Self {
        Self {
            non_recursive_key_provider,
        }
    }

    pub fn certificate_verifying_key(
        &self,
        srs: &ParamsKZG<Bls12>,
    ) -> StmResult<NonRecursiveCircuitVerifyingKey> {
        self.non_recursive_key_provider.verification_key(srs)
    }
}

impl KeyGenerator for RecursiveCircuitKeyGenerator {
    type VerifyingKey = RecursiveCircuitVerifyingKey;
    type ProvingKey = RecursiveCircuitProvingKey;

    fn generate_key_pair(
        &self,
        srs: &ParamsKZG<Bls12>,
    ) -> StmResult<(Self::VerifyingKey, Self::ProvingKey)> {
        let certificate_verifying_key = self.non_recursive_key_provider.verification_key(srs)?;
        IvcCircuitData::unknown(&certificate_verifying_key)?.generate_key_pair(srs)
    }
}
```

Like that the load function has this signature:
```rust
pub(crate) fn load(
      trusted_setup_provider: &TrustedSetupProvider,
      key_provider: &KeyProvider<RecursiveCircuitKeyGenerator>,
) -> StmResult<Self> 
```

jpraynaud (approved):
LGTM 🚀

PR #3353, opened 2026-06-25
Make kes_bytes file creation/write atomic for concurrent access

### LINKED ISSUE #3329: Flaky CI on hydra run of `ci/hydra-build:x86_64-linux.mithril-common`
## Why

All tests that call `setup_temp_directory_for_signer` share the same base path:

```text
mithril_test/tests_setup/mithril_crypto_helper_material/{party_id}
````

KES key files in this shared directory are written non-atomically: `to_file` first calls `fs::File::create`, which truncates the file to empty, and then writes the JSON content in a second step.

A concurrent test that reads the file between those two steps sees an empty file and panics with:

```text
EOF while parsing a value at line 1 column 0
```

The result is that some CI runs fail this test due to race conditions (see, for example, [this](https://github.com/input-output-hk/mithril/runs/80893343380) one on an unrelated part of the code base)

## What

The test `certificate_chain::certificate_verifier::tests::verify_genesis_certificate_fails_if_is_not_genesis`, and potentially any other test that reads KES key files from the shared temp directory, fails intermittently when the Hydra runner executes tests in parallel.

Rerunning the job succeeds, masking the underlying data race.

## How

- [x] In `fixture_builder.rs`, replace the direct `to_file` calls for `kes.sk` and `opcert.cert` with a write-to-temp-then-rename pattern.
- [x] Use `std::process::id()` or a random suffix in the temp filename to avoid collisions between parallel processes.
- [x] `fs::rename` is atomic on POSIX filesystems, so concurrent readers will always see either no file or a complete one.
- [x] Verify the same pattern is not needed elsewhere by auditing other callers of `to_file` that write into a shared temp directory.
- [ ] Add a comment at the write site explaining why the atomic rename is necessary.

turmelclem:
## Content

Make kes_bytes file creation/write atomic for concurrent access, especially to fix CI test flakiness

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation
  - [x] No new TODOs introduced

## Issue(s)

this PR closes #3329

### DIFF
--- a/mithril-common/src/crypto_helper/cardano/codec.rs
+++ b/mithril-common/src/crypto_helper/cardano/codec.rs
@@ -15,6 +15,7 @@ use anyhow::{Context, anyhow};
 use hex::FromHex;
 use kes_summed_ed25519::kes::Sum6Kes;
 use kes_summed_ed25519::traits::KesSk;
+use rand_core::RngCore;
 use serde::de::DeserializeOwned;
 use serde::{Deserialize, Serialize};
 use serde_with::{As, Bytes};
@@ -24,6 +25,7 @@ use std::path::Path;
 use thiserror::Error;
 
 use crate::StdError;
+use crate::test::crypto_helper::SerDeShelleyFileFormatTestExtension;
 
 /// We need to create this struct because the design of Sum6Kes takes
 /// a reference to a mutable pointer. It is therefore not possible to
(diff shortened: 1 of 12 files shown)


mithril-common/src/crypto_helper/cardano/codec.rs
```diff
@@ -49,30 +50,12 @@ struct ShelleyFileFormat {
...
-            .with_context(|| "SerDeShelleyFileFormat can not unserialize json data")
-            .map_err(CodecParseError)?;
+use crate::test::crypto_helper::SerDeShelleyFileFormatTestExtension;
 
-        Self::from_cbor_hex(&file.cbor_hex)
-    }
-
-    /// Serialize a type `T: Serialize + DeserializeOwned` to file following Cardano
-    /// Shelley file format.
+impl<T> SerDeShelleyFileFormatTestExtension for T
```
Alenar: Can you move this impl after the trait definition?

mithril-common/src/crypto_helper/cardano/codec.rs
```diff
@@ -180,7 +194,7 @@ impl<'a> TryFrom<&'a mut Sum6KesBytes> for Sum6Kes<'a> {
 #[cfg(test)]
 mod test {
     use super::*;
-    use crate::test::TempDir;
+    use crate::test::{TempDir, crypto_helper::SerDeShelleyFileFormatTestExtension};
```
Alenar: It should not be needed since it's imported in the parent context.

Alenar (approved):
LGTM

jpraynaud (approved):
LGTM 👍

Issue #3333, opened 2026-06-15
Enhance aggregator state machine to support `Blocked` state


jpraynaud:
## Why
We want to enhance the state machine of the aggregator to better support when there is no genesis certificate yet or in case of epoch gap. This will avoid numerous logs which are not needed in these situations

## What
Update the state machine of the aggregator to support blocked state for the rest of the epoch

## How 

- [x] Add a new `Blocked` state (with a reason field which is an enum) in the aggregator state machine
- [x] The state machine transitions out of this state upon epoch change and gets back to `Idle`
- [x] In case of missing genesis go to this new state (from `Idle` state)
- [x] In case of genesis epoch (genesis certificate was created during this epoch) go to this new state (from `Idle` state)
- [x] In case of epoch gap, go to this new state (from `Idle` state)
- [x] Add/update integration tests in aggregator
- [x] Update documentation

PR #3372, opened 2026-07-06
feat(aggregator): add state machine `blocked` state

Alenar:
## Content

This PR add a new state to the aggregator state machine, `Blocked`, with three variants:
- `NoGenesis`: when the aggregator have yet to start a chain
- `GenesisEpoch`: when the genesis certificate was issued in the current epoch, meaning that single signatures can't be handled because the AVK in the genesis certificate will only be in effect in the next epoch
- `EpochGap`: when the certificate chain validation failed because of an epoch gap between the last certificate and the actual epoch, meaning that the chain must be re-genesis

This cleanup all errors from the logs of the Mithril Aggregator when starting a new chain, notably in end to end tests, allowing easier logs review.

### Changes

Refactor aggregator state machine:
- simplify the main `cycle` method by extracting every state handling into dedicated methods (`cycle_idle`, `cycle_ready`, ...)
- split `cycle_idle` in two branchs, `cycle_idle_leader` and `cycle_leader_follower`, to avoid mixing their specificity (with shared methods to avoid duplication)
- rename `AggregatorRuntime::get_state` to `state_label` to better reflect that it does not returns a string with the full state
- refactor `AggregatorRuntime::try_transition_from_idle_to_ready`: renamed to `execute_epoch_initialization_tasks`, and move its guard condition to a new method on the `IdleState`
- simplify tests by adding shared methods that configure the mocks for all methods used with epoch initialization tasks 
- additional improvements: remove some unneeded allocation/cloning, fix and align rustdoc grammars

New state machine / `AggregatorRuntime` state: `Blocked`:
- New transition `Idle -> Blocked (Epoch gap)`: triggered by reading the error returned by the chain validation or by a follower chain synchronization (if the leader has an epoch gap as well)
- New transition `Idle -> Blocked (No Genesis)`: triggered by fetching if a genesis certificate was issued or not, if not the chain is not validated and this state is returned
- New transition `Idle -> Blocked (Genesis epoch)`: triggered by fetching the epoch of the last issued genesis certificate if it's the same as the actual epoch from the latest timepoint.

Refactor `MithrilCertificateChainSynchronizer` to avoid sending an error if the remote aggregator do not have a chain, instead a log is written and the synchronization is skipped.
This is change is required for the follower aggregator, else it would cycle with an error to `idle` instead of transitioning to `blocked (no genesis)`.

Adapt integration tests scenario:
- a lot of them were using their tooling shortcut to generate certificates at the same epoch of the genesis, which is not possible anymore with the `Blocked (genesis-epoch)` state
- some of them were relying on the fact that before genesis the state machine would keep cycling to `idle` with an error

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation
  - [x] No new TODOs introduced

## Comments

Should we add a transition from `Blocked (no-genesis)` to `Blocked (genesis-epoch)` ? This would be more correct if we write the genesis certificate without shutting down the aggregator (limited interest since we shut down the aggregator in both e2e and our runbook).

A long warning is still remaining ion the case of the startup of a new chain: failed single signature authentication.

This is due to the same problem than the one handled by `Blocked - Genesis Epoch` (no avk available that can identified signers), but the `/resgister-signatures` route doesn't have currently access to the data that would allow to identify that case and reject the signature more gracefully.

## Issue(s)

Relates to #3333


mithril-aggregator/src/runtime/state_machine.rs
```diff
@@ -92,12 +139,17 @@ impl AggregatorRuntime {
...
-            AggregatorState::Signing(_) => "signing".to_string(),
+    /// Return the label of the actual state of the state machine.
+    pub fn state_label(&self) -> &'static str {
+        match &self.state {
+            AggregatorState::Idle(_) => "idle",
+            AggregatorState::Blocked(state) => match state.reason {
+                BlockedReason::NoGenesis => "blocked-no-genesis",
+                BlockedReason::GenesisEpoch => "blocked-genesis-epoch",
+                BlockedReason::EpochGap { .. } => "blocked-epoch-gap",
+            },
```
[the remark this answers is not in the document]
Alenar: False positive

mithril-aggregator/src/services/certificate_chain_synchronizer/synchronizer_service.rs
```diff
@@ -5,7 +5,8 @@
 //!    - If false, fetch the latest local genesis certificate in database
 //!       - If it's found, fetch the remote Genesis certificate
 //!          - If it's different from the local genesis, continue synchronization
-//!          - If it's the same, abort with an `Ok`
+//!          - If it's the same, skip synchronization with `AlreadyUpToDate`
+//!          - If it's not found, skip synchronization with `EmptyRemoteCertificateChain`
```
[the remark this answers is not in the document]
Alenar: Fixed

mithril-aggregator/src/runtime/state_machine.rs
```diff
@@ -137,169 +189,305 @@ impl AggregatorRuntime {
...
+            .synchronize_follower_aggregator_certificate_chain(force_sync)
+            .await?;
+        // Refetch last genesis certificate epoch after synchronization
+        let last_genesis_certificate_epoch = self.runner.last_genesis_certificate_epoch().await?;
+
+        self.transition_from_idle(
+            last_time_point,
+            last_genesis_certificate_epoch,
+            chain_validity_result,
+        )
```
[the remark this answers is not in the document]
Alenar: Synchronization also validate the chain

turmelclem (approved):
LGTM 👍

mithril-aggregator/src/runtime/state_machine.rs
```diff
@@ -213,77 +214,71 @@ impl AggregatorRuntime {
...
```
jpraynaud, line 285: The transition is done with the `chain_validity_result` computed before the re-sync of the certificate chain. It should be updated after the re-sync otherwise the follower aggregator will have to wait another epoch even if the re-sync is successful.

mithril-aggregator/src/runtime/state_machine.rs
jpraynaud, line 199: The counter is incremented even if the state is `Blocked` and in epoch gap. This means that an errorish state would be counted as success. This is a bit of a tricky case, but I'm not sure that it's worth changing the behavior. WDYT?

jpraynaud (commented):
LGTM 👍

mithril-aggregator/src/runtime/state_machine.rs
Alenar: Good catch, but fixing it have cost:
- either we re-validate the chain after sync (easy but triple validation, twice from the certifier, once from the synchronization)
- or we modify the synchronizer to include epoch gap check (it already validate the chain otherwise)
- or we extract the epoch gap check to avoid re-validating after sync (but should we sync if leader have an epoch gap ?)

mithril-aggregator/src/runtime/state_machine.rs
Alenar: IMO it's not an strictly an error state so keeping it as is is acceptable, the alternative would be either not registering it or adding a new counter.

mithril-aggregator/src/runtime/state_machine.rs
Alenar: Fixed, see the the commits at the end of the branch starting with ["refactor(aggregator): extract `CertificateEpochGap` error a dedicated module"](https://github.com/input-output-hk/mithril/pull/3372/commits).

jpraynaud (approved):
LGTM 🚀

turmelclem (approved):

PR #3400, opened 2026-07-08
 feat(stm): upgrade midnight zk 2.3.3 and update the circuit keys

### LINKED ISSUE #3375: Upgrade midnight-zk to `2.3.3`
## Why
We want to upgrade the Midnight zk library to latest released version `2.3.3`.

## What

Upgrade the Midnight zk library to latest released version `2.3.3`.

## How

- [x] Upgrade midnight-zk [`2.3.3`](https://crates.io/crates/midnight-zk-stdlib/2.3.3)
- [x] Update the circuit verification keys with this [runbook](https://github.com/input-output-hk/mithril/blob/main/docs/runbook/update-circuit-keys/README.md)
- [x] Create a summary of changes that have an impact for Mithril SNARK
- [x] Benchmark performances gain:
  - [x] Circuit degree
    - The circuit degrees did not change
  - [x] Proving time
    - Proving time looks similar to before
  - [x] Proof size

damrobi:
## Content

This PR includes the upgrade to midnight zk 2.3.3 and some fixes that come with it.

Upgrades `mithril-stm` to the latest `midnight-zk` library versions and adapts all affected code.

**Main API changes adapted:**

- `ForeignEccChip` renamed to `ForeignWeierstrassEccChip`; `configure` now requires explicit `nb_parallel_range_check` and `max_bit_length` parameters (previously hardcoded internally)
- `Accumulator` construction: `dual_msm.into()` + `extract_fixed_bases` → `Accumulator::from_dual_msm(dual_msm, prefix, fixed_bases)`, which populates fixed-base scalars on both `lhs` and `rhs` at construction time
- `accumulator.check(&tau_g2, fixed_bases)` → `accumulator.check(&verifier_params, fixed_bases)`; `tau_g2` is no longer a standalone field in `IvcVerifierSetup`
- `MockProver::run` no longer takes an explicit `k` parameter; `min_k()` removed, replaced by `k()`
- `srs.downsize_from_circuit()` replaces `zk::downsize_srs_for_relation`
- `verifier_gadget.prepare()`: committed instances no longer require a name
- `ZkStdLibArch` gained two new fields (`p256`, `curve25519`), bumping the serialization version from 1 to 2 — embedded verifying key bytes regenerated accordingly
- `Relation` trait now requires an `Error` associated type: Generic error for now, will be modified in a later PR
- `extract_fixed_bases`  removed as done in `from_dual_msm` function

**Consequences:**

- Accumulator serialization format changed: old persisted `IvcRollingState` bytes are not compatible with this version
- All IVC test assets regenerated
- Embedded production circuit keys updated

**Tests:**

- All `fixed_base_extraction` tests updated and re-enabled for the new `resolve_fixed_bases` semantics, including inverting the direction of the public input length assertion (resolution now grows the encoding rather than shrinking it)
- Added golden test (`recursive_proof_accumulator_serialized_bytes_are_stable`) and generator to detect future accumulator format changes at the byte level

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Issue(s)

Closes #3375


mithril-stm/src/circuits/halo2_ivc/tests/off_circuit/fixed_base_extraction.rs
```diff
@@ -19,6 +19,7 @@ use crate::circuits::halo2_ivc::{
 };
 
 #[test]
+#[ignore]
 fn extract_fixed_bases_reduces_public_input_length_by_n_times_field_elements_per_base_point() {
```
hjeljeli32: @damrobi 
The old flow was two steps 
- dual_msm.into() put all bases in the variable part, 
- then extract_fixed_bases() pulled the fixed ones out (shrinking as_public_input, which is what this test measured). 
In the new library that separation happens inside from_dual_msm(...), so the accumulator our helper returns is already separated before the test sees it. With the extract_fixed_bases call gone, the two measurements now read the same accumulator twice → before - after == 0, and the assertion expects a positive number. Hence the fail/ignore.

Fix: measure the same fact from the other side. The new API exposes resolve_fixed_bases(), which does the inverse (moves fixed bases back into the variable part, so the encoding grows). So we call it as the observed step and assert the encoding grows by count × field_elements_per_base_point.

Remark: the expected count is not recursive_fixed_bases.len(). When I first tried that, it failed 320 vs 325: the fixed-base map names one base this proof doesn't actually reference, and from_dual_msm only separates the terms it finds. So I derive the count from the accumulator's real separated scalars instead:

```
let fixed_base_scalar_count = accumulator.lhs().fixed_base_scalars().len()
    + accumulator.rhs().fixed_base_scalars().len();
// ...
let before = AssignedAccumulator::as_public_input(&accumulator).len();
accumulator.resolve_fixed_bases(&recursive_fixed_bases);
let after  = AssignedAccumulator::as_public_input(&accumulator).len();
assert_eq!(after - before, fixed_base_scalar_count * field_elements_per_base_point);
```

If you agree, We need to rename the test to resolve_fixed_bases_grows_public_input_length_by_n_times_field_elements_per_base_point since the direction flipped 😄 
On my side it passes.

mithril-stm/src/circuits/halo2_ivc/tests/off_circuit/fixed_base_extraction.rs
```diff
@@ -75,12 +73,11 @@ fn extracted_fixed_base_scalars_are_non_trivial() {
 fn extract_fixed_bases_does_not_affect_lhs() {
```
hjeljeli32: @damrobi 
Claude flagged this and I think he is right:
Same root cause as the extract_fixed_bases_reduces… comment above 👆 but this one it wasn't ignored: with the extract_fixed_bases call removed, the two lhs() snapshots read the same untouched accumulator, so it just asserts X == X and passes while testing nothing.

Fix is the same shape as Test 1: reinsert a real mutation (resolve_fixed_bases) between the two snapshots, then assert lhs is unchanged. 

One thing: the new resolve_fixed_bases touches both lhs and rhs (the old extract was rhs-only), so "lhs unchanged" is only valid if lhs has no fixed bases. We confirmed it does not (all fixed bases sit on rhs), and the test passes. 

IMO we should also rename to resolve_fixed_bases_does_not_affect_lhs. 
It passes on my side.

mithril-stm/src/circuits/halo2_ivc/tests/off_circuit/fixed_base_extraction.rs
```diff
@@ -94,13 +91,14 @@ fn extract_fixed_bases_does_not_affect_lhs() {
 #[should_panic(
     expected = "fixed_bases should not contain keys (names) that appear in self.fixed_base_scalars"
 )]
+#[ignore]
 fn extracting_same_fixed_bases_twice_panics() {
```
hjeljeli32: @damrobi
This one's premise is gone rather than broken. It was a #[should_panic] test guarding the old footgun: calling extract_fixed_bases() twice panicked (the names were already extracted). The new resolve_fixed_bases() moves the named scalars into the variable part and then clears its map, so a second call has nothing left to do: it's a safe no-op and can never raise that panic (its only panic path is a different "Base not provided" message). A #[should_panic] whose body runs cleanly = a failing test, hence the ignore.

Fix: either we delete it, or repurpose it to lock down the behaviour that actually changed a second resolve_fixed_bases()
```
accumulator.resolve_fixed_bases(&recursive_fixed_bases);
let after_first = AssignedAccumulator::as_public_input(&accumulator);
accumulator.resolve_fixed_bases(&recursive_fixed_bases);
let after_second = AssignedAccumulator::as_public_input(&accumulator);
assert_eq!(after_first, after_second);

```
If you decide to repurpose it, we can rename to resolving_fixed_bases_twice_is_a_harmless_no_op and drop #[should_panic] + #[ignore].

On my side it passes.

mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
```diff
@@ -803,6 +799,7 @@ mod tests {
     }
 
     #[test]
+    #[ignore]
     fn ivc_proof_verify_rejects_wrong_tau_g2() {
```
hjeljeli32: @damrobi 
The old flow passed tau_g2 as a separate parameter only to the accumulator check, while verifier_params (with the correct s_g2) went to the KZG opening check. That separation is exactly what let the test inject a wrong tau, pass the KZG check, and fail specifically at the accumulator pairing (AccumulatorFailed).

The upgrade folded tau_g2 into verifier_params, and verify() now feeds the same verifier_params to both checks. So there's no independent tau to set wrong anymore — and if you corrupt the tau inside verifier_params, the KZG opening check (which runs first) catches it → you'd get KzgOpeningFailed, never AccumulatorFailed. The exact bug this test guarded is now architecturally impossible.

So I don't think there's a faithful "fix" I'd propose two parts:

1. Delete ivc_proof_verify_rejects_wrong_tau_g2 its scenario was designed out by the unification.

2. If we still want to keep coverage of the AccumulatorFailed branch, add a new test that reaches it through the one input only the accumulator check consumes, the fixed bases (the KZG check ignores them):

```
// negate every fixed base: KZG opening check still passes, accumulator pairing breaks
let wrong_fixed_bases = ctx
    .combined_fixed_bases
    .into_iter()
    .map(|(name, base)| (name, -base))
    .collect();
let verifier_setup = IvcVerifierSetup::from_parts(
    ctx.verifier_params, ctx.recursive_verifying_key, wrong_fixed_bases,
);
let err = proof.verify(&STEP_OUTPUT_MSG, &global, &verifier_setup).expect_err(...);
assert_eq!(err.downcast_ref::<IvcProofError>(), Some(&IvcProofError::AccumulatorFailed));

```
This new test checks a different property (fixed-base integrity, not tau). Wdyt?

damrobi: I agree with your assessment, I will add the test you mention

mithril-stm/src/circuits/halo2_ivc/tests/off_circuit/fixed_base_resolution.rs
```diff
@@ -19,42 +19,41 @@ use crate::circuits::halo2_ivc::{
...
+    accumulator.resolve_fixed_bases(&recursive_fixed_bases);
     let encoding_length_after_extraction = AssignedAccumulator::as_public_input(&accumulator).len();
 
     assert_eq!(
-        encoding_length_before_extraction - encoding_length_after_extraction,
-        fixed_base_count * field_elements_per_base_point,
+        encoding_length_after_extraction - encoding_length_before_extraction,
+        fixed_base_scalar_count * field_elements_per_base_point,
         "extract_fixed_bases should reduce as_public_input length by field_elements_per_base_point per extracted base"
     );
```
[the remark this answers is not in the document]
damrobi: Fixed

mithril-stm/src/circuits/halo2_ivc/tests/off_circuit/fixed_base_extraction.rs
```diff
@@ -72,15 +71,15 @@ fn extracted_fixed_base_scalars_are_non_trivial() {
 }
 
 #[test]
-fn extract_fixed_bases_does_not_affect_lhs() {
+fn resolved_fixed_bases_does_not_affect_lhs() {
     // extract_fixed_bases only touches rhs (right-hand side); the lhs (left-hand side)
     // MSM encoding must be identical before and after the call.
     let (mut accumulator, recursive_fixed_bases) =
```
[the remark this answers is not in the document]
damrobi: Fixed

mithril-stm/src/circuits/halo2_ivc/tests/off_circuit/fixed_base_extraction.rs
```diff
@@ -91,16 +90,16 @@ fn extract_fixed_bases_does_not_affect_lhs() {
 }
```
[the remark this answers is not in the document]
damrobi: Fixed

mithril-stm/src/circuits/halo2/tests/golden/helpers.rs
```diff
@@ -570,12 +570,12 @@ pub(crate) fn setup_stm_circuit_env(
...
     {
-        let stm_circuit = MidnightCircuit::from_relation(&relation);
+        let stm_circuit = MidnightCircuit::from_relation(&relation, None);
         println!("\n=== STM circuit case: {case_name} ===");
         println!("circuit degree (selected) {circuit_degree}");
         println!("k {k}");
-        println!("min_k {:?}", stm_circuit.min_k());
-        println!("{:?}", zk::cost_model(&relation));
+        println!("min_k {:?}", stm_circuit.k());
+        println!("{:?}", zk::cost_model(&relation, None));
```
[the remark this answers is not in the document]
damrobi: Fixed

mithril-stm/src/circuits/halo2_ivc/tests/off_circuit/fixed_base_extraction.rs
```diff
@@ -91,16 +90,16 @@ fn extract_fixed_bases_does_not_affect_lhs() {
...
-    expected = "fixed_bases should not contain keys (names) that appear in self.fixed_base_scalars"
-)]
-fn extracting_same_fixed_bases_twice_panics() {
+fn resolving_fixed_bases_twice_is_a_harmless_no_op() {
     // Attempting to extract the same fixed bases a second time violates the
     // contract (names already present in fixed_base_scalars) and must panic.
     // This test documents the current behaviour; it will be updated when
     // extract_fixed_bases is changed to return a Result.
     let (mut accumulator, recursive_fixed_bases) =
         build_unextracted_recursive_proof_accumulator_from_assets();
```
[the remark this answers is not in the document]
damrobi: Fixed

mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
```diff
@@ -195,7 +195,7 @@ where
         }
 
         if !self.accumulator.check(
```
kitounliu, line 197: I did some quick tests. The combined verification of dual_msm and accumulator now takes 7.421 ms which is slightly faster than the separately verification  (8.677 ms). So it is probably worth combining the checking now

mithril-stm/src/circuits/halo2_ivc/config.rs
```diff
@@ -79,19 +81,33 @@ pub fn configure_ivc_circuit(meta: &mut ConstraintSystem<NativeField>) -> IvcCon
             .expect("column counts pre-validated by validate_column_counts"),
     );
 
+    let nb_parallel_range_check = NB_ARITH_COLS - 1;
+    let max_bit_length = RECURSIVE_CIRCUIT_DEGREE - 1;
```
hjeljeli32, line 85: @damrobi
Small thing but worth a look before we regen keys: max_bit_length is a parameter of the range-check decomposition chip: the library doc calls it "the maximum bit-length each lookup supports" (field/foreign/util.rs:193). I am not sure if it is related to the circuit degree 🤔

hjeljeli32 (approved):
LGTM 👍 I just left a question about max_bit_length which I am not sure how to set

mithril-stm/src/circuits/halo2/circuit.rs
```diff
@@ -293,6 +294,8 @@ impl Relation for StmCertificateCircuit {
             nr_pow2range_cols: 2,
```
kitounliu, line 294: a smaller `nr_pow2range_cols` means more rows but smaller proof size. It is always good to check if nr_pow2range_cols is still optimal when updating midnight-zk. When adjusting the parameters, StmCertificateCircuit degree needs to be kept under 22 and ivc circuit degree under 19

damrobi: I tried to lower `nr_pow2range_cols` to 1 but it grows the certificate circuit to degree 23

mithril-stm/src/circuits/halo2_ivc/tests/common/generators/asset_generation.rs
```diff
@@ -158,9 +159,8 @@ fn build_recursive_chain_snapshot(
             &public_inputs,
         );
         assert!(dual_msm.clone().check(&context.universal_verifier_params));
-
-        let mut proof_accumulator: Accumulator<RecursiveEmulation> = dual_msm.into();
-        proof_accumulator.extract_fixed_bases(recursive_fixed_bases);
+        let mut proof_accumulator: Accumulator<RecursiveEmulation> =
+            Accumulator::from_dual_msm(dual_msm, IVC_VERIFICATION_KEY_NAME, recursive_fixed_bases);
```
jpraynaud: I find it weird to reuse the circuit verification key name as prefix. It'd be more readable to create another constant specifically for that.

mithril-stm/src/circuits/halo2/circuit.rs
jpraynaud: It's probably something we want to include in the runbook @damrobi?

jpraynaud (commented):
LGTM 👍

mithril-stm/src/circuits/halo2_ivc/tests/common/generators/asset_generation.rs
damrobi: Those are the constants we use when extracting the fixed bases from the verification keys [here](https://github.com/input-output-hk/mithril/blob/513d88afe5621a45bae8ce514b66a600642b533b/mithril-stm/src/proof_system/ivc_halo2_snark/prover_setup.rs#L87-L94). The constants are only used for that so I can rename them to something like `CERTIFICATE_FIXED_BASES_PREFIX`?

mithril-stm/src/circuits/halo2/circuit.rs
damrobi: I added a mention in the runbook, let me know what you think

mithril-stm/src/circuits/halo2_ivc/config.rs
damrobi: From what I understand, setting the `max_bit_length` to degree - 1 (its maximal value) can allow to reduce the number of rows by fitting more bits of the witnesses in the range check table rows. @kitounliu can you confirm that? Should the value be independent from the degree?

mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
damrobi: Maybe we can fit that in another PR, WDYT @jpraynaud ?

mithril-stm/src/circuits/halo2_ivc/config.rs
kitounliu: yes, that is correct @damrobi. Just need to be careful when finding the smallest RECURSIVE_CIRCUIT_DEGREE as any higher degree can also generate proofs now

mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
jpraynaud: Yes let's do that in another PR :+1:

Issue #3452, opened 2026-08-03
Fix flakiness of e2e tests in the CI


jpraynaud:
## Why

The e2e tests fail intermittently in the CI with `Timeout exhausted waiting for Certificate for epoch XX, no certificate found with epoch >= XX`. 

Here is a list of occurences in the CI:
- https://github.com/IntersectMBO/mithril/actions/runs/30268111482/job/89985446600#step:6:737
- https://github.com/IntersectMBO/mithril/actions/runs/30268111482/job/89985446315#step:6:766
- https://github.com/IntersectMBO/mithril/actions/runs/30278226513/job/90019562236#step:6:1268
- https://github.com/IntersectMBO/mithril/actions/runs/30374444928/job/90328388998#step:6:723
- https://github.com/IntersectMBO/mithril/actions/runs/30374444928/job/90328389046#step:6:1785
- https://github.com/IntersectMBO/mithril/actions/runs/30379691520/job/90346046444#step:6:1817
- https://github.com/IntersectMBO/mithril/actions/runs/30453954566/job/90584419636#step:6:1296
- https://github.com/IntersectMBO/mithril/actions/runs/30456461437/job/90593095939#step:6:735
- https://github.com/IntersectMBO/mithril/actions/runs/30461622912/job/90610833554#step:6:707
- https://github.com/IntersectMBO/mithril/actions/runs/30461622912/job/90610833770#step:6:776
- https://github.com/IntersectMBO/mithril/actions/runs/30557735551/job/90924517606#step:6:729

## What

Fix the flakiness in the CI which creates an epoch gap and leads to the error message witnessed.

## How

- [x] Investigate the source of flakiness
- [x] Fix the source of flakiness
- [x] Verify stability with batch runs of the e2e test

PR #3475, opened 2026-08-06
fix(e2e): fix flakiness with minimal scenario

jpraynaud:
## Content

This PR includes a **fix for the flakiness of the e2e test in `minimal` mode, where certificates could be sealed with an incomplete signer set**.

The new protocol parameters are crafted to avoid having one signer reach the quorum alone which lead to the `Certificate is not signed by expected number of signers: 1 < 2` error. The success rate witnessed on my local environment increased from 75% to 100% with the new protocol parameters.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Issue(s)
Relates to #3452

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4601,7 +4601,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-end-to-end"
-version = "0.5.10"
+version = "0.5.11"
 dependencies = [
  "anyhow",
  "async-recursion",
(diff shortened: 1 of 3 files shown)


Alenar (approved):
LGTM, what about the stability of networks with more than two signers?

turmelclem (approved):
LGTM 👍

Issue #3435, opened 2026-07-27
Create deterministic immutable compressed archives


jpraynaud:
## Why

IPFS content addressing (CID) requires byte identical archives across nodes, but non deterministic compression produces a different CID for the same immutable data.

## What

Produce deterministic compressed archives of the immutable files so that identical input always yields an identical CID.

## How

- [x] Deprecate/remove the Gzip compression for immutable files
- [x] Identify the sources of non determinism in the current archive creation (timestamps, entry ordering, metadata, compression level)
- [x] Make the archive creation deterministic (https://docs.rs/tar/latest/tar/enum.HeaderMode.html)
- [x] Verify that identical inputs produce identical CIDs across nodes
- [x] Add tests covering the determinism of the archives
- [x] Lock versions of crate creating tar and compressor crates?
- [x] How to support breaking changes in the future:
  - Manual operation for aggregator operators (command/process to recompute the immutable archives which avoids service disruption)
  - Dev book to support the upgrade of tar and zstd versions

PR #3479, opened 2026-08-13
refactor(aggregator): deterministic snapshots archives

Alenar:
## Content

This PR harden the archives building of the `mithril-aggregator` in order to makes them bytes identical for a given content.

It extract the `FileArchiver` api from the Aggregator into introduces a new library, `mithril-file-archiver`, with additional changes to ensure deterministic, byte-stable `.tar.zst` archives creations.

### Main Changes

New Library Added:

- Introduces `mithril-file-archiver` as a standalone library for generating deterministic archives. Like the previous api, it supports archiving files, directories, or serializable data with compression.

  - Includes support for configurable Zstandard compression parameters.
  - Ensures reproducibility and determinism across platforms.
  - Refactored the appender system around a `ArchiveEntry` enum that hold one entry of the supported type (file, directory, or data), `ArchiveEntry` are `Ord` and the existing appender now return them instead of directly writting to the tar archive, this allows `AppenderChain` to be deterministic as long as there's no entries overlap between the chained appenders.
  - Comprehensive tests for functionality, compatibility, cross-platform reproducibility, and golden master validation were implemented in the new library.

- Existing File Archiver Logic Removed: The previously internal `file_archiver` and associated modules (`appender`, `tools`, etc.) in the `mithril-aggregator` folder have been moved to the new library.
  - Updated `mithril-aggregator` to integrate `mithril-file-archiver` for tasks like snapshot generation and ancillary file handling.
  - Adjusted tests to use the new library's provided abstractions and utilities (e.g., `FileArchiver::new_with_default_parameters` and `FileArchiveTestExtension`).  - Existing tests in mithril-aggregator were adapted accordingly.
  - Removed direct dependencies on `zstd` and `tar` in `mithril-aggregator`.

- Documentation:
  - Enhanced documentation to describe `mithril-file-archiver`, its API, and usage constraints.

- Other:
  - `follow_symlinks` support is now disabled in the `FileArchiver` (as recommended in their [doc](https://docs.rs/tar/0.4.46/tar/struct.Builder.html#method.follow_symlinks))
  - some unused dependencies were removed from the aggregator: `tempfile`, and `mithril-test-http-server` 

### How deterministic archives were achieved

- Changes to the tar header:
  - switch to [`HeaderMode::Deterministic`](https://docs.rs/tar/latest/tar/enum.HeaderMode.html#variant.Deterministic): this set the `mtime` to a fix value (timestamp [`1153704088`](https://github.com/composefs/tar-rs/blob/0.4.46/src/header.rs#L24), date:  Jul 23, 2006), the `guid` and `gid` to `0`, and the permissions to `644` for files and `755` for directories (see the code [here](https://github.com/composefs/tar-rs/blob/0.4.46/src/header.rs#L797))
  - `sparse` file support is disabled as it's not consistent across systems and the zstandard will compress the sparse part itself anyway
- A normalization path is now run on the `AppenderEntries` entries list:
  - removal of trailing `/` from directory names (or `\` on Windows)
  - removal of leading `./` from path
  - the entries list is now sorted using rust default sort
- changes to `AppenderData` metadata:
  - a deterministic `mtime` is set instead of the current time: randomly choose to `1767225600` (2026-01-01 00:00:00 UTC)
- Removal of `AppenderDirAll`: the order of the files and folder is not stable across systems,  if needed, it could be re-created as a constructor of `AppenderEntry` (`from_directory`) which would do the same job of recursively adding all entries but would then leverage its normalization pass
- Reworking the appender to output a list of sortable `ArchiveEntry` allowed to add commutativity and associativity to the `ChainAppender`, as long as there's no overlapping path between chained appenders, else the rightmost take priority

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation
  - [x] No new TODOs introduced

## Issue(s)

Relates to #3435


turmelclem (approved):
LGTM 👍

internal/mithril-file-archiver/Makefile
jpraynaud, line 1: This `Makefile` is not run by the `make` commands at the root of the repository.

internal/mithril-file-archiver/src/api.rs
```diff
@@ -272,16 +275,25 @@ impl FileArchiver {
 
         Ok(())
     }
+
+    fn configure_tar_builder<W: std::io::Write>(builder: &mut tar::Builder<W>) {
+        builder.mode(HeaderMode::Deterministic);
```
jpraynaud, line 285: This does not guarantee that archives will ultimately be deterministic across platforms as the tar library does not normalize the permissions under the hood:
- for Unix: `let fs_mode = if meta.is_dir() || (0o100 & meta.mode() == 0o100) { 0o755 } else { 0o644 };`
- for Windows: `let fs_mode = if meta.is_dir() { 0o755 } else { 0o644 };`

It'd probably be better to enforce the permissions in the headers.

internal/mithril-file-archiver/src/entities.rs
```diff
@@ -1,12 +1,36 @@
...
+
+impl Default for ZstandardCompressionParameters {
+    fn default() -> Self {
+        Self {
+            level: 9,
+            number_of_workers: 4,
+        }
+    }
+}
+
```
jpraynaud, line 25: Do we still want to provide custom parameters for the Zstandard compression?

jpraynaud (approved):
LGTM 🔥

internal/mithril-file-archiver/src/entities.rs
Alenar: That's an open question, I think we can keep it for custom aggregators.

internal/mithril-file-archiver/src/api.rs
Alenar: I documented the behavior difference between unix and windows

PR #3480, opened 2026-08-13
feat(stm): add missing off circuit checks

### LINKED ISSUE #3381: Add missing IVC off-circuit checks
## Why

The off-circuit checks run before proving and reject a prover input that would produce a proof that cannot verify. They do not cover all the in-circuit constraints: on the next epoch transition, the chaining of the protocol parameters is not checked, so a mismatch is only caught after a full and expensive proving run.

## What

Complete the off-circuit checks of the epoch transition so that they mirror the in-circuit constraints, starting with the chaining of the protocol parameters.

## How

- [x] Add missing off-circuit check for chaining of protocol parameter (see [here](https://input-output-rnd.slack.com/archives/C09RY8KEYUF/p1782753724223729))
- [x] Audit the in-circuit constraints and list the checks that have no off-circuit counterpart
- [x] Add the missing off-circuit checks found by the audit
- [x] Add unit tests covering the accepted and rejected transitions for each new check
- [x] Update the SNARK book if needed

damrobi:
## Content

This PR includes the addition of some missing off circuit checks. The goal of those checks is to prevent a prover from trying to create a proof that will fail to verify since the proof creation is a very expensive operations.

## Changes

- Validation of the protocol message hash: it was not done off circuit yet but is done at every step in circuit. Recompute the hash of the protocol message preimage and compares it to the MessageHash received.
- Check of the validity of the genesis verification key: verifies the key is an actual point on the elliptic curve. The check is done at every step (even post genesis) in circuit.
- Check that the protocol parameters do not evolve from one epoch to the next: check that `protocol_parameters == next_protocol_parameters` for the values coming from the rolling state. An update of the parameters would break the recursive proof.
- Moved all the checks to before the generation of the non recursive proof to allow a failure before the heavy computation: This implied a big restructure of the code around the call to the IVC prover

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Issue(s)

Closes #3381

### DIFF
--- a/mithril-stm/src/circuits/halo2_ivc/state.rs
+++ b/mithril-stm/src/circuits/halo2_ivc/state.rs
@@ -1,5 +1,6 @@
 use serde::{Deserialize, Serialize};
 
+use crate::StmResult;
 use crate::circuits::halo2::keys::NonRecursiveCircuitVerifyingKey;
 use crate::circuits::halo2_ivc::keys::RecursiveCircuitVerifyingKey;
 use crate::signature_scheme::{SchnorrVerificationKey, StandardSchnorrSignature};
@@ -115,14 +116,15 @@ pub(crate) struct Global {
 }
 
 impl Global {
-    #[allow(dead_code)]
     pub(crate) fn new(
         genesis_message: MessageHash,
         genesis_verification_key: SchnorrVerificationKey,
         certificate_verification_key: &NonRecursiveCircuitVerifyingKey,
         ivc_verification_key: &RecursiveCircuitVerifyingKey,
-    ) -> Self {
-        Global {
+    ) -> StmResult<Self> {
+        genesis_verification_key.is_valid()?;
+
+        Ok(Global {
             genesis_message,
             genesis_verification_key,
             certificate_circuit_verification_key_representation:
@@ -133,7 +135,7 @@ impl Global {
                 IvcCircuitVerificationKeyRepresentation::from_field(
                     ivc_verification_key.as_ref().transcript_repr(),
                 ),
-        }
+        })
     }
 
     pub(crate) fn as_public_input(&self) -> Vec<NativeField> {
--- a/mithril-stm/src/circuits/halo2_ivc/state.rs
+++ b/mithril-stm/src/circuits/halo2_ivc/state.rs
@@ -201,3 +201,42 @@ pub(crate) struct AssignedWitness {
     // Protocol message preimage bytes
     pub(crate) message_preimage: Vec<AssignedByte<NativeField>>,
 }
+
+#[cfg(test)]
+mod tests {
+    use crate::{
+        circuits::{
+            halo2::NON_RECURSIVE_CIRCUIT_VERIFICATION_KEY_FOR_PRODUCTION,
+            halo2_ivc::RECURSIVE_CIRCUIT_VERIFICATION_KEY_FOR_PRODUCTION,
+        },
+        codec::TryFromBytes,
+        signature_scheme::SchnorrSignatureError,
+    };
+
+    use super::*;
+
+    #[test]
+    fn new_rejects_invalid_genesis_verification_key() {
+        let certificate_verifying_key = NonRecursiveCircuitVerifyingKey::try_from_bytes(
+            NON_RECURSIVE_CIRCUIT_VERIFICATION_KEY_FOR_PRODUCTION,
+        )
+        .expect("production certificate verifying key bytes should deserialize");
+        let ivc_verifying_key = RecursiveCircuitVerifyingKey::try_from_bytes(
+            RECURSIVE_CIRCUIT_VERIFICATION_KEY_FOR_PRODUCTION,
+        )
+        .expect("production IVC verifying key bytes should deserialize");
+
+        let err = Global::new(
+            MessageHash::ZERO,
+            SchnorrVerificationKey::default(),
+            &certificate_verifying_key,
+            &ivc_verifying_key,
+        )
+        .expect_err("Global::new should reject an invalid genesis verification key");
+
+        assert!(matches!(
+            err.downcast_ref::<SchnorrSignatureError>(),
+            Some(SchnorrSignatureError::PointIsNotPrimeOrder(..))
+        ));
+    }
+}
--- a/mithril-stm/src/proof_system/ivc_halo2_snark/prover_input.rs
+++ b/mithril-stm/src/proof_system/ivc_halo2_snark/prover_input.rs
@@ -15,8 +15,8 @@ use crate::{
     },
     proof_system::ivc_halo2_snark::{
         prover_input_helpers::{
-            IvcTransitionType, build_next_accumulator, build_next_state,
-            create_snark_message_for_next_state, verify_certificate_proof,
+            IvcTransitionType, assert_message_hash_matches_preimage, build_next_accumulator,
+            build_next_state, create_snark_message_for_next_state, verify_certificate_proof,
         },
         prover_setup::IvcSnarkProverSetup,
         rolling_state::IvcRollingState,
@@ -66,15 +66,17 @@ impl IvcProverInput {
             transition_type,
         )?;
 
+        let (certificate_message_hash, certificate_merkle_tree_commitment) =
+            create_snark_message_for_next_state(aggregate_verification_key_for_snark, message)?;
+
+        assert_message_hash_matches_preimage(certificate_message_hash, protocol_message_preimage)?;
+
         let certificate_dual_msm = verify_certificate_proof(
             certificate_proof,
             message,
             aggregate_verification_key_for_snark,
             prover_setup,
         )?;
-
-        let (certificate_message_hash, certificate_merkle_tree_commitment) =
-            create_snark_message_for_next_state(aggregate_verification_key_for_snark, message)?;
 
         let next_state = build_next_state(
             transition_type,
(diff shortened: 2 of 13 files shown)


mithril-stm/src/protocol/aggregate_signature/clerk.rs
```diff
@@ -290,37 +258,113 @@ fn ivc_prover_input_preparation_and_prove<D: MembershipDigest>(
...
+            prepare_checks(msg, &avk, &protocol_message_preimage, rolling_state)?;
+        }
+        None => {
+            let combined_fixed_base_names: Vec<String> =
+                ivc_prover_setup.combined_fixed_bases.keys().cloned().collect();
+            let genesis_rolling_state = IvcRollingState::genesis(
+                genesis_bootstrap.genesis_signature,
+                &combined_fixed_base_names,
+            );
+            prepare_genesis_checks(&genesis_rolling_state, &protocol_message_preimage, &global)?;
```
hjeljeli32: @damrobi I think the genesis check uses the wrong preimage. If I understood well, protocol_message_preimage is the preimage of the current certificate. I think we should pass &genesis_bootstrap.genesis_protocol_message_preimage

mithril-stm/src/protocol/aggregate_signature/clerk.rs
```diff
@@ -290,37 +258,113 @@ fn ivc_prover_input_preparation_and_prove<D: MembershipDigest>(
...
         genesis_verifying_key,
         &certificate_verifying_key,
-        &ivc_verifying_key,
-    );
+        &ivc_prover_setup.ivc_verifying_key,
+    )?;
+
+    match current_rolling_state {
+        Some(rolling_state) => {
+            rolling_state.assert_protocol_parameters_unchanged()?;
```
hjeljeli32: Would it make sense to move this check into IvcProverInput::prepare_checks? Keeping it here means direct IvcProver::prove / IvcProverInput::prepare callers bypass this invariant.

hjeljeli32 (commented):
Good initial work 👍 . I left 2 comments.

mithril-stm/src/protocol/aggregate_signature/clerk.rs
damrobi: Good catch, updated!

mithril-stm/src/protocol/aggregate_signature/clerk.rs
damrobi: I think it is fine to have it separated from the `prove` since it's mainly here to prevent an aggregator from running a prove that will ultimately fail but it is also valid to have it inside the `prove` function. If you prefer I can move it inside the prove function

mithril-stm/src/protocol/aggregate_signature/clerk.rs
```diff
@@ -290,37 +259,117 @@ fn ivc_prover_input_preparation_and_prove<D: MembershipDigest>(
...
+
+            let avk = snark_clerk
+                .compute_aggregate_verification_key_for_snark::<MithrilMembershipDigest>();
+            off_circuit_checks(msg, &avk, &protocol_message_preimage, rolling_state)?;
+        }
+        None => {
```
hjeljeli32: Nice restructure, this reads much better 👍 . 
But, when current_rolling_state is None, IvcProver::prove bootstraps the genesis step and then still runs a full certificate step, effective_rolling_state at proof.rs:384-387, followed by the unconditional IvcProverInput::prepare at proof.rs:391. So on the first-ever aggregation only the genesis checks run here, and the certificate's message/preimage consistency isn't validated until prepare, i.e. after the certificate proof has already been generated. That's the one case where we still pay for a proof we could have rejected. The message-hash check is state-independent, so calling assert_message_hash_matches_preimage on this branch too (against protocol_message_preimage) would close it the transition-type and parameter checks genuinely need the post-genesis state, so those can stay on the Some arm only. WDYT?

mithril-stm/src/proof_system/ivc_halo2_snark/prover_input.rs
```diff
@@ -66,16 +67,45 @@ impl IvcProverInput {
...
+            certificate_message_hash,
+            certificate_merkle_tree_commitment,
+        })
+    }
+
+    /// Completes `prepare` once the certificate proof is available: verifies it, advances the
+    /// chain state, and folds the accumulator. `checks` must come from [`Self::off_circuit_checks`]
+    /// run against the same `protocol_message_preimage` and `rolling_state`.
+    #[allow(clippy::too_many_arguments)]
+    pub(crate) fn finish_preparation<D: MembershipDigest>(
```
hjeljeli32, line 58: finish_preparation takes certificate_message_hash and certificate_merkle_tree_commitment from checks, but also re-takes message and aggregate_verification_key_for_snark and passes those to verify_certificate_proof. If a caller ever pairs a checks with a different message, the next state commits to one message while the proof verifies another, and we only find out in-circuit . The doc comment says they must match; moving message and the AVK into IvcProverInputChecks would make the mismatch unrepresentable instead of documented. And maybe that way we would also let the #[allow(clippy::too_many_arguments)] go.

hjeljeli32 (approved):
LGTM. I just left 2 non-blocking suggestions.

mithril-stm/src/proof_system/ivc_halo2_snark/prover_input.rs
damrobi: Done

mithril-stm/src/protocol/aggregate_signature/clerk.rs
damrobi: Done

mithril-stm/src/proof_system/ivc_halo2_snark/rolling_state.rs
```diff
@@ -138,6 +138,16 @@ impl IvcRollingState {
             == self.state.current_epoch.as_field() + EpochNumber::new(1).as_field()
     }
 
+    /// Asserts that protocol_parameters and next_protocol_parameters agree.
+    pub(crate) fn assert_protocol_parameters_unchanged(&self) -> StmResult<()> {
+        if self.state.step_counter.is_not_first_step()
```
kitounliu: the unchanged parameter checking should also cover the first step? (it doesn't need checking for genesis as there is no threshold signatures)

mithril-stm/src/proof_system/ivc_halo2_snark/prover_input_helpers.rs
kitounliu, line 174: perhaps verifying `next_accumulator` is valid here, otherwise a corrupted previous rolling_state could pass all the pre-proof checks

mithril-stm/src/protocol/aggregate_signature/clerk.rs
```diff
@@ -290,37 +259,120 @@ fn ivc_prover_input_preparation_and_prove<D: MembershipDigest>(
...
+        &ivc_prover_setup.ivc_verifying_key,
+    )?;
+
+    let avk = snark_clerk.compute_aggregate_verification_key_for_snark::<MithrilMembershipDigest>();
+
+    match current_rolling_state {
+        Some(rolling_state) => {
+            rolling_state.assert_protocol_parameters_unchanged()?;
+
+            off_circuit_checks(msg, &avk, &protocol_message_preimage, rolling_state)?;
```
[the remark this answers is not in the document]
damrobi: Added a call to `ensure_advanceable_rolling_state` before the match statement

mithril-stm/src/protocol/aggregate_signature/clerk.rs
```diff
@@ -245,41 +219,36 @@ impl<D: MembershipDigest> Clerk<D> {
...
-/// Prepares the IVC prover inputs from `ancillary_input` and the cached prover setup to
-/// build the [`Global`] chain constants, and run [`IvcProver::prove`].
-///
-/// Returns `(proof, ancillary_prover_data, ancillary_verifier_data)`. `ancillary_prover_data` is `Some` when
-/// the step advances the epoch and `None` for same-epoch steps.
-///
-/// # Errors
-/// Fails if the genesis Schnorr verifying key is absent, the message preimage is not PREIMAGE_SIZE bytes,
-/// or the proof itself fails.
+/// Bundles the outputs of [`prepare_ivc_snark_request`]: everything [`prove_ivc_snark`] needs to
```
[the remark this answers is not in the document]
damrobi: Fixed

mithril-stm/src/protocol/aggregate_signature/clerk.rs
```diff
@@ -290,37 +259,120 @@ fn ivc_prover_input_preparation_and_prove<D: MembershipDigest>(
...
+        certificate_verifying_key,
+        global,
+        protocol_message_preimage,
+        current_rolling_state,
+        genesis_bootstrap,
+    })
+}
+
+/// Generates the certificate proof and completes IVC proving for a request already validated by
+/// [`prepare_ivc_snark_request`].
```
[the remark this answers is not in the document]
damrobi: Fixed

mithril-stm/src/proof_system/ivc_halo2_snark/rolling_state.rs
damrobi: Fixed, now the function checks `!self.is_genesis()` instead of `self.state.step_counter.is_not_first_step()`

mithril-stm/src/proof_system/ivc_halo2_snark/prover_input_helpers.rs
damrobi: I added a check of the accumulator before returning it in the function

mithril-stm/src/circuits/halo2_ivc/state.rs
```diff
@@ -115,14 +116,15 @@ pub(crate) struct Global {
 }
 
 impl Global {
-    #[allow(dead_code)]
     pub(crate) fn new(
```
jpraynaud:
```suggestion
    pub(crate) fn try_new(
```

Issue #3433, opened 2026-07-27
Reuse the cached unsafe SRS in the SNARK test setups


jpraynaud:
## Why

`TrustedSetupProvider::with_unsafe_srs` regenerates the unsafe SRS on every call even though it already writes it into the content-keyed test cache, so every `build_for_test` pays minutes of SRS generation that the cache should absorb.
This is the highest leverage optimization of #3405 (SRS reuse alone measured at 75 to 61 minutes nightly in PR #3345).

## What

Reuse the SRS file already persisted under `target/mithril-circuit-test-cache` so SRS generation is paid once per machine and per fingerprint, for both `SnarkProverSetup::build_for_test` and `IvcSnarkProverSetup::build_for_test`.

## How

- [x] Add a guard in `with_unsafe_srs`: when the SRS file exists use it
- [x] Replace its direct `File::create` write with the temp-plus-rename pattern already used by `store_srs_bytes_to_file`
- [x] Persist the downsized SRS per target degree as well: profiling shows `ParamsKZG::downsize` (Lagrange basis recomputation) costs about 3x the k=20 SRS generation and is paid on every setup load, warm or cold
- [x] Verify both `build_for_test` constructors hit the cache on a warm second run
  - [x] Non-recursive: works fine, a test was added to verify that (can be removed)
  - [x] Recursive: works but the warm run is not faster due to keeping the downsizing operation and having a slow ProvingKey deserialization
    - [x] Fix the downize operation always applying
    - [x] Check if the ProvingKey can be generated/loaded faster
- [x] Assess the gain (expected in the action plan: nightly from 75 to ~61 minutes, measured in PR #3345) and report it in the issue
  - For now gain looks to be roughly 35 minutes (1h11m => 35min)

Issue #3467, opened 2026-08-04
Persist the SNARK test caches across CI runs


jpraynaud:
## Why

Neither the keygen cache nor the SRS is persisted across CI runs (the `toolchain-and-cache` action only saves dependency artifacts through `Swatinem/rust-cache`), so every nightly and every slow PR leg starts cold.

## What

Add an `actions/cache` layer for `target/mithril-circuit-test-cache` with a three-layer invalidation design: content-keyed directories for semantic staleness, a versions-scoped cache key for serialization format drift, and self-healing readers as backstop.

## How

Paths are relative to `mithril-stm/src/` unless they start with `.github/`.

- [ ] Add an `actions/cache/restore` step for `target/mithril-circuit-test-cache` in the test job of `.github/workflows/test-rust.yml` (next to the `toolchain-and-cache` composite step), with an exact key folding a manual generation token and a hash of the `midnight-*` entries of `Cargo.lock`
- [ ] Scope every `restore-keys` prefix to that versions hash so a `midnight-proofs` upgrade (serialization format drift) can never restore stale bytes
- [ ] Save with `actions/cache/save` only from the runs that execute the full slow suite (the `include-slow-tests: true` invocation from `nightly-dispatcher.yml`); PR runs stay restore-only
- [ ] Make the readers self-healing: on deserialization failure, `KeyProvider::cache_state` and `read_optional` (`circuits/key_provider.rs`) and `TrustedSetupProvider::get_trusted_setup_parameters` (`circuits/trusted_setup.rs`) should delete the offending file and regenerate instead of failing the test
- [ ] Prune fingerprint directories not touched by the current run before saving, so retired fingerprints do not accumulate in the archive
- [ ] Make one recursive verifying-key cache canonical for both the circuit tests and `IvcSnarkProverSetup::build_for_test_degree`, with the prover path storing only the additional proving key: today `proof_system/ivc_halo2_snark/prover_setup.rs` and `circuits/halo2_ivc/tests/common/generators/setup.rs` hold the same key under two protocols, because `KeyProvider::verification_key` treats a missing proving key as a miss and regenerates the pair
- [ ] Assess the gain (expected: removes the cold start from warm CI runs) and report it in the issue

Depends on #3433.

cderinbogaz:
[...]

PR #3481, opened 2026-08-17
refactor(stm): cache the circuit test fixtures on disk

### LINKED ISSUE #3462: Extend the on-disk test cache to the recursive circuit test fixtures
## Why

The `circuits/halo2_ivc` test tree has no caching: every slow file rebuilds the 3000-signer fixture, and the two full-prover tests each regenerate the k=19 SRS and a real certificate proof independently.

## What

Extend the content-keyed `FileMutex` cache to the circuit test fixtures so the SRS, the recursive context, and the heavy signer fixture are computed once and shared across tests and runs.

## How

Paths are relative to `mithril-stm/src/`.

- [x] Replace the `build_deterministic_params` call inside `build_shared_recursive_context` (`circuits/halo2_ivc/tests/common/generators/setup.rs`) with the SRS persisted under the shared content-keyed cache (`FileMutex::for_shared_cache`, `circuits/test_utils/file_mutex.rs`), reusing the file #3433 writes
- [x] Serialize the derived `SharedRecursiveContext` (certificate and recursive verifying keys, commitment parameters) into the same cache directory, fingerprinted like `IvcSnarkProverSetup::build_for_test` (production VK bytes, parameters, depth, degree, seed), so recursive keygen is paid once per fingerprint
- [x] Make the full-prover slow tests load through that one fingerprint:
  - [x] `golden/positive.rs`: `genesis_base_case_circuit_is_accepted`
  - [x] `golden/positive.rs`: `replay_integrity_matches_stored_next_step_output`
  - [x] `in_circuit/accumulator.rs`: `circuit_rejects_with_wrong_next_accumulator_in_same_epoch_step`
- [x] Cache the 3000-signer fixture built by `build_asset_generation_setup` (`generators/setup.rs`, `SIGNER_COUNT = 3000`): serialize the signing keys and Merkle tree into the cache directory, or consolidate its consumers if serialization does not pay
- [x] Remove the `STM_CIRCUIT_KEYS_CACHE` `LazyLock` inside `get_or_build_circuit_keys` (`circuits/halo2/tests/golden/helpers.rs`), inert under the nextest process-per-test model, and route its consumers through the on-disk `KeyProvider` cache (`circuits/key_provider.rs`)
- [x] Assess the gain (expected: minutes off every slow circuit file) and report it in the issue

Depends on #3433.

hjeljeli32:
## Content

This PR extends the content-keyed `FileMutex` cache introduced by #3433 to both circuit test trees, which had none. Under the nextest process-per-test model an in-process cache amortizes nothing, so the SRS, the derived recursive verifying key, the certificate circuit keys and the deterministic 3000-signer fixture are now computed once per fingerprint and shared across test processes and runs.

Measured locally: `circuits::halo2_ivc` slow tier 435 s → 381 s, `circuits::halo2` slow tier 50 s → 44 s (−12 % each, ±8 s run to run). Numbers per ticket item are in #3462.

### Changes

- **Shared SRS** (`halo2_ivc/tests/common/generators/setup.rs`): `build_shared_recursive_context` reads the  degree-19 SRS from the same `"unsafe-srs"` entry `IvcSnarkProverSetup::build_for_test` writes, both deriving from the same seed. A compile-time assertion ties `ASSET_SEED` to `UNSAFE_SRS_SEED`, so a change to either cannot silently alter the committed assets.
- **Recursive verifying key**: cached as a single blob under a fingerprint that folds in the freshly derived certificate verifying key, the committed production recursive key, both circuit degrees and the SRS seed. The certificate key is derived on every call (~93 ms) because its bytes are part of the cache address.
- **Certificate circuit keys** (`halo2/tests/golden/helpers.rs`): the process-local `LazyLock` map is replaced by the on-disk `KeyProvider`, behind the cache lock so parallel processes racing a cold miss derive the pair once. The SRS asset filename now carries the seed that produced it, so a seed change cannot pair an old SRS with keys fingerprinted for a new one.
- **Signer fixture**: the generator-derived values are cached as CBOR; everything else in `AssetGenerationSetup` is a pure function of them and is recomputed on both paths, so the two cannot disagree. Validated on load — vector lengths, plus the genesis signature verified against the message derived from the cached tree, which transitively covers tampering with the tree, the key or the signature.
- **Cache contract**: entries are rebuilt on an absent, unreadable, undecodable, non-canonical or validator-rejected file, and never surface as an error. It makes no wider claim: canonical encoding shows the bytes are self-consistent, not who wrote them.
- **Tests**: 11 new, all fast and hermetic — cold miss, warm hit, truncation, trailing bytes, validator rejection, the real signer-fixture validator (fresh / wrong lengths / changed tree / changed key or signature), and input-by-input coverage of all three production cache fingerprints.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation
  - [x] No new TODOs introduced

## Issue(s)
Closes #3462

### DIFF
--- a/mithril-stm/src/circuits/halo2/tests/golden/helpers.rs
+++ b/mithril-stm/src/circuits/halo2/tests/golden/helpers.rs
@@ -1,8 +1,6 @@
-use std::collections::HashMap;
 use std::fs::create_dir_all;
 use std::panic::{self, AssertUnwindSafe};
 use std::path::PathBuf;
-use std::sync::{Arc, LazyLock, RwLock};
 use std::time::Instant;
 
 use anyhow::{Context, anyhow};
@@ -15,6 +13,7 @@ use midnight_zk_stdlib::MidnightCircuit;
 use rand_chacha::ChaCha20Rng;
 use rand_core::SeedableRng;
 
+use crate::circuits::halo2::NON_RECURSIVE_CIRCUIT_VERIFICATION_KEY_FOR_PRODUCTION;
 use crate::circuits::halo2::circuit::StmCertificateCircuit;
 use crate::circuits::halo2::errors::StmCircuitError;
 use crate::circuits::halo2::keys::{
@@ -25,7 +24,10 @@ use crate::circuits::halo2::witness::{
     CircuitMerkleTreeLeaf, CircuitWitnessEntry, LotteryTargetValue as CircuitLotteryTargetValue,
     MerklePath, MerkleRoot, SignedMessageWithoutPrefix,
 };
+use crate::circuits::key_provider::KeyProvider;
+use crate::circuits::test_utils::file_mutex::FileMutex;
 use crate::circuits::test_utils::setup::{generate_params, load_params};
+use crate::circuits::trusted_setup::UNSAFE_SRS_SEED;
 use crate::hash::poseidon::MidnightPoseidonDigest;
 use crate::membership_commitment::{
     MerkleTree as StmMerkleTree, MerkleTreeSnarkLeaf as StmMerkleTreeSnarkLeaf,
@@ -42,13 +44,11 @@ pub(crate) const LOTTERIES_PER_K: u32 = 10;
 /// Default message value used by golden test cases.
 const DEFAULT_TEST_MSG: u64 = 42;
 
-/// Verification/proving key pair cached per STM circuit configuration.
+/// Verification/proving key pair derived for an STM circuit configuration.
 type CircuitVerificationAndProvingKeyPair = (
     NonRecursiveCircuitVerifyingKey,
     NonRecursiveCircuitProvingKey,
 );
-/// Cache map for verification/proving keys keyed by STM circuit configuration.
-type CircuitKeysCache = HashMap<StmCircuitConfig, Arc<CircuitVerificationAndProvingKeyPair>>;
 
 fn checked_len_u32(actual: usize) -> u32 {
     u32::try_from(actual).unwrap_or(u32::MAX)
@@ -91,15 +91,6 @@ fn validate_relation_for_setup(relation: &StmCertificateCircuit) -> StmResult<()
     relation
         .validate_parameters()
         .with_context(|| "Circuit parameter validation failed before setup")
-}
-
-/// Cache key derived from the STM circuit configuration.
-#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
-struct StmCircuitConfig {
-    circuit_degree: u32,
-    k: u32,
-    m: u32,
-    merkle_tree_depth: u32,
 }
 
 /// Shared environment for STM circuit golden cases (SRS, relation, keys, sizing).
@@ -553,20 +544,13 @@ pub(crate) fn setup_stm_circuit_env(
         println!("{:?}", zk::cost_model(&relation, None));
     }
 
-    let config = StmCircuitConfig {
-        circuit_degree,
-        k,
-        m,
-        merkle_tree_depth: depth,
-    };
-    let key_pair = get_or_build_circuit_keys(config, &relation, &srs)?;
-    let (vk, pk) = (&key_pair.0, &key_pair.1);
+    let (vk, pk) = get_or_build_circuit_keys(&stm_params, depth, circuit_degree, &relation, &srs)?;
 
     Ok(StmCircuitEnv {
         srs,
         relation,
-        vk: vk.clone(),
-        pk: pk.clone(),
+        vk,
+        pk,
         num_signers,
         m,
     })
@@ -650,7 +634,11 @@ pub(crate) fn run_stm_circuit_case(
 fn load_or_generate_params(circuit_degree: u32) -> StmResult<ParamsKZG<Bls12>> {
     let manifest_dir = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
     let assets_dir = manifest_dir.join("src").join("circuits").join("halo2").join("assets");
-    let path = assets_dir.join(format!("params_kzg_unsafe_{}", circuit_degree));
+    // The seed belongs in the name: a file identified only by degree would be reused after a seed
+    // change, pairing an old-seed SRS with keys cached under the new seed.
+    let path = assets_dir.join(format!(
+        "params_kzg_unsafe_degree_{circuit_degree}_seed_{UNSAFE_SRS_SEED}"
+    ));
 
     if path.exists() {
         return Ok(load_params(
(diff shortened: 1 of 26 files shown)


mithril-stm/src/circuits/halo2_ivc/tests/common/generators/setup.rs
damrobi: Should we use the `load_shared_unsafe_srs` function here too?

mithril-stm/src/circuits/halo2_ivc/tests/common/generators/setup.rs
```diff
@@ -192,9 +198,111 @@ fn load_shared_unsafe_srs(circuit_degree: u32) -> ParamsKZG<Bls12> {
         .expect("the shared unsafe SRS should load from the test cache")
 }
 
-/// Builds the shared verifier-side recursive setup from the deterministic SRS.
+/// File holding the cached recursive verifying key inside its fingerprinted cache directory.
+const RECURSIVE_VERIFYING_KEY_CACHE_FILE: &str = "recursive-verifying-key";
+
+/// Where the expensive recursive verifying key comes from.
```
damrobi, line 203: Maybe add to the comment that `Derived` is only for asset generation (at least for now) and `Cached` is for general use in the test functions? I think it would make this easier to understand

mithril-stm/src/circuits/halo2_ivc/tests/common/generators/setup.rs
```diff
@@ -173,12 +182,193 @@ pub(crate) fn build_deterministic_params(circuit_degree: u32) -> ParamsKZG<Bls12
...
+enum RecursiveVerifyingKeySource {
+    /// Always derived. Required wherever the result is written to a committed asset.
+    Derived,
+    /// Loaded from the content-keyed test cache, derived only on a miss.
+    Cached,
+}
+
+/// Derives the recursive verifying key for the default IVC circuit shape (about 8.9 s).
```
damrobi: Can you use an order of magnitude (seconds or tens of seconds) instead fixed timings? Same below with the ms

mithril-stm/src/circuits/halo2_ivc/tests/common/generators/setup.rs
```diff
@@ -173,12 +182,193 @@ pub(crate) fn build_deterministic_params(circuit_degree: u32) -> ParamsKZG<Bls12
...
+/// pinning the unsafe SRS.
+fn recursive_verifying_key_cache(
+    certificate_verifying_key_bytes: &[u8],
+    production_recursive_verifying_key: &[u8],
+    recursive_circuit_degree: u32,
+    certificate_circuit_degree: u32,
+    unsafe_srs_seed: u64,
+) -> FileMutex {
+    FileMutex::for_shared_cache(
+        "ivc-recursive-verifying-key-v1",
```
damrobi: Is there a reason to use the `-v1` suffix here and in the function below?

mithril-stm/src/circuits/halo2_ivc/tests/common/generators/setup.rs
```diff
@@ -173,12 +182,193 @@ pub(crate) fn build_deterministic_params(circuit_degree: u32) -> ParamsKZG<Bls12
...
+/// who wrote them, so an entry that is well-formed but semantically wrong is accepted — acceptable
+/// only because this cache is disposable, is not exposed to a hostile writer, and is never read by
+/// anything that produces committed assets.
+///
+/// Rebuilding rather than reporting is deliberate, and unlike
+/// [`KeyProvider`](crate::circuits::key_provider::KeyProvider), which propagates deserialization
+/// errors: a spoiled test cache must not fail a test run.
+///
+/// `is_valid` carries any invariant the bytes alone cannot express. Values whose encoding is
+/// self-contained pass `|_| true`.
```
damrobi: This comment is a bit long and hard to read, can you make it shorter? maybe with less details

mithril-stm/src/circuits/halo2_ivc/tests/common/generators/setup.rs
```diff
@@ -173,12 +182,193 @@ pub(crate) fn build_deterministic_params(circuit_degree: u32) -> ParamsKZG<Bls12
...
+    // opened this is a no-op, and losing a test-cache entry only costs a rebuild.
+    let _ = std::fs::File::open(directory).and_then(|directory_file| directory_file.sync_all());
+}
+
+/// Builds the shared verifier-side recursive setup, **always deriving** the recursive verifying key.
+///
+/// Asset generators must use this: they write committed assets, and a stale cached key would
+/// silently produce assets derived from it. Behavior tests that only read should call
+/// [`build_shared_recursive_context_from_cache`].
 pub(crate) fn build_shared_recursive_context(
```
damrobi: Can you rename that function to make it explicit it always derives? something like `build_shared_recursive_context_from_scratch`

mithril-stm/src/circuits/halo2_ivc/tests/common/generators/setup.rs
```diff
@@ -173,12 +182,193 @@ pub(crate) fn build_deterministic_params(circuit_degree: u32) -> ParamsKZG<Bls12
...
+    build_shared_recursive_context_with(setup, RecursiveVerifyingKeySource::Derived)
+}
+
+/// Builds the shared verifier-side recursive setup, taking the recursive verifying key from the
+/// content-keyed test cache when one is present.
+///
+/// The cache address folds in the freshly derived certificate verifying key, the committed
+/// production recursive key, both circuit degrees, and the SRS seed, so a change to the certificate
+/// circuit or a regenerated production key resolves to a different entry. **Never call this from an
+/// asset generator** — see [`build_shared_recursive_context`].
```
damrobi: Maybe this can be an explanation of `build_shared_recursive_context_with` instead

mithril-stm/src/circuits/halo2/tests/golden/helpers.rs
```diff
@@ -674,48 +662,76 @@ fn load_or_generate_params(circuit_degree: u32) -> StmResult<ParamsKZG<Bls12>> {
...
 }
 
-/// Get cached verification/proving keys or build and insert them if missing.
+/// Content-keyed cache entry holding the certificate keys derived from these inputs.
+///
+/// Every input that changes the derived keys is a parameter, so the address is a pure function of
+/// them and a test can vary each one: the committed production verifying key as a circuit-version
+/// salt, the protocol parameters, the Merkle-tree depth, the circuit degree, and the seed pinning
+/// the unsafe SRS. Distinct configurations therefore never share a directory, which is what lets
+/// [`KeyProvider`] be built with no expected verifying key.
```
damrobi: I'm not sure this comment is necessary, the function reads well even without it

mithril-stm/src/circuits/halo2/tests/golden/helpers.rs
```diff
@@ -674,48 +662,76 @@ fn load_or_generate_params(circuit_degree: u32) -> StmResult<ParamsKZG<Bls12>> {
...
+}
+
+/// Loads the verification/proving key pair for this configuration from the on-disk cache, deriving
+/// and storing it on a miss.
+///
+/// The cache is shared across processes, unlike the in-process map this replaced, which amortized
+/// nothing under the nextest process-per-test model. The lock is taken before the lookup so that
+/// parallel processes racing a cold miss derive the pair once rather than once each — `KeyProvider`
+/// does not serialize its writers.
 fn get_or_build_circuit_keys(
```
damrobi: Not sure this comment is needed

mithril-stm/src/circuits/halo2_ivc/tests/common/generators/setup.rs
```diff
@@ -173,12 +182,193 @@ pub(crate) fn build_deterministic_params(circuit_degree: u32) -> ParamsKZG<Bls12
...
+    )
+}
+
+/// Content-keyed cache entry holding the recursive verifying key derived from these inputs.
+///
+/// Every input is a parameter, so the address is a pure function of them and a test can vary each
+/// one: the freshly derived certificate verifying key (which tracks the certificate circuit), the
+/// committed production recursive key as a circuit-version salt, both circuit degrees, and the seed
+/// pinning the unsafe SRS.
+fn recursive_verifying_key_cache(
```
damrobi, line 235: It looks like we have two caches for the recursive VK, is it possible to re-use the cache of `IvcSnarkProverSetup::build_for_test_degree` to not regenerate the VK if it exists?

damrobi (commented):
Some none blocking comments but otherwise it looks good 👍

mithril-stm/src/circuits/halo2_ivc/tests/common/generators/setup.rs
hjeljeli32: Good catch! for this configuration they really do hold the same recursive verifying key, so the duplication is real. The reason for the split is the cache contract rather than the key: `IvcSnarkProverSetup::build_for_test_degree` owns a complete VK+PK prover cache, and `KeyProvider::verification_key` deliberately treats a verifying key without its proving key as a miss and regenerates the pair so calling it from the circuit tests would generate the recursive proving key, which is exactly the cost these tests avoid. 
A hit-only probe would dodge that but would be order-dependent under nextest's process-per-test model and would still leave two cache protocols. Real unification means making one VK cache canonical for both paths and having the prover cache store only the extra proving key, which is a KeyProvider refactor. I've added it to #3467. Worth noting we did unify the SRS: both paths now share the same seed-keyed entry, which is safe because it has no pair invariant to violate.

damrobi (approved):
LGTM 👍

PR #3483, opened 2026-08-19
test(stm): consolidate the certificate proof tests around the golden proof

### LINKED ISSUE #3463: Consolidate the `halo2_snark` slow proof tests around shared and golden proofs
## Why

Each of the 16 `halo2_snark::proof::slow` tests generates its own full proof (100s to 620s each, 3458s aggregated) even when the assertion only needs proof bytes to tamper with.

## What

Prove once per configuration and run each negative verify assertion as its own unit test against a committed golden proof or a cached proof, keeping 4 or 5 genuine proving tests.

## How

All tests live in `mithril-stm/src/proof_system/halo2_snark/proof.rs`.

- [x] Route the verify-only tests to the committed golden proof (`GOLDEN_PROOF_JSON` and `golden_proof()`, parameters m=200, k=3, phi_f=0.8, depth 4):
  - [x] `valid_proof_verifies`, then merge it into `golden_conversion` if routing makes them identical
  - [x] `verify_fails_with_random_bytes_or_wrong_number_bytes`
  - [x] `verify_fails_with_wrong_message`
  - [x] `verify_fails_with_wrong_circuit_verification_key`
  - [x] `valid_proof_prepares_and_checks`
  - [x] `prepare_and_check_fails_with_wrong_message`
  - [x] `prepare_and_check_fails_with_random_bytes`
  - [x] `snark_proof_to_from_bytes`
- [x] ~Prove once into the content-keyed cache directory for `different_parameters_prove_and_verify_fails`, which needs two parameter sets the single golden proof cannot serve (a proof is a few KB; fingerprint by parameters, depth, message, and prover seed)~
- [x] Keep as genuine proving tests:
  - [x] `produces_valid_snark_proof`
  - [x] `valid_snark_proof_with_different_path_length`
  - [x] `non_deterministic_proofs_verify`
- [x] Leave unchanged, they fail during aggregation before any proving and their setup cost is absorbed by the cache of #3433:
  - [x] `fails_with_empty_signatures`
  - [x] `fails_with_insufficient_signatures`
- [x] Leave unchanged, they already run against the golden proof:
  - [x] `golden_conversion`
  - [x] `golden_aggregate_signature_verifies_with_ancillary`
- [x] Assess the gain (expected: about 35 minutes aggregated) and report it in the issue

Depends on #3433.

hjeljeli32:
## Content

This PR routes the verify-only tests in `proof_system::halo2_snark::proof` to the committed golden proof. Each of them previously built a full prover setup and generated its own proof, even though the assertion only ever needed proof *bytes* — a valid proof to tamper with, or any valid proof at all. Verification needs neither an SRS nor a proving key: `SnarkProof::verify` reads only the proof bytes plus the aggregate verification key, message and verifying key it is handed, and the verifier parameters are a 192-byte embedded constant.

Measured locally, `--test-threads 1`: the slow tier goes from 33.9 s to 15.3 s warm (-55 %) and from 65.8 s to 47.1 s cold (-28 %), 16 tests becoming 15. Fast tier unchanged at 511 passing. Per-item numbers are in #3463.

### Changes

- **Verify negatives** (`proof.rs`): `verify_fails_with_random_bytes_or_wrong_number_bytes`, `verify_fails_with_wrong_message` and `verify_fails_with_wrong_circuit_verification_key` move into `mod golden` and run against `golden_proof()`. Four of the routed tests previously ran at m=100, k=5; `verify` and `prepare_and_check` never read the proof's own `params` or `merkle_tree_depth`, so the golden proof's configuration serves them without changing what they assert.
- **`valid_proof_verifies` deleted**: once routed its body is identical to `golden_conversion`, which is itself unmodified.
- **Wrong verifying key**: obtained by decoding `NON_RECURSIVE_CIRCUIT_VERIFICATION_KEY_FOR_PRODUCTION` instead of building a second prover. `MidnightVK` is self-describing, so this needs no SRS and no keygen, and the assertion strengthens to "a test proof must not verify under the key that ships".
- **Prepare-and-check and codec tests**: the three `prepare_and_check` tests and `snark_proof_to_from_bytes` routed the same way, every assertion kept.
- **Fresh-proof coverage**: `produces_valid_snark_proof` gains a `prepare_and_check` assertion on the proof it already generates, so that method still runs against a freshly produced proof rather than only committed bytes. Costs 0.03 s.
- **Untouched**: `valid_snark_proof_with_different_path_length` and `non_deterministic_proofs_verify`, the two tests that fail during aggregation before proving, `different_parameters_prove_and_verify_fails`, and the two pre-existing golden tests are byte-identical to `main`. `produces_valid_snark_proof` is the only surviving test that changed, by the assertion above.
- **Not done, the cached forged proof**: implemented and measured, then reverted. Locally it measures +0.65 s on a cold cache against -0.47 s on a warm one, and the content-keyed cache directory is empty on every CI run, so the cold path is the one CI takes. Left for @jpraynaud's decision, with the numbers in a comment on #3463.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation
  - [x] No new TODOs introduced

## Issue(s)
Closes #3463


damrobi (approved):
LGTM 👍 
I just think that you can probably move some of those test function outside of the slow module, maybe in a golden module outside of slow?

PR #3486, opened 2026-08-24
test(stm): consolidate the recursive circuit MockProver tests

### LINKED ISSUE #3464: Consolidate the recursive circuit MockProver tests
## Why

13 of the 20 `MockProver` invocations in `circuits::halo2_ivc` tamper a single public-input element and leave the circuit data correct, but they cannot share a synthesis: `MockProver::run` re-synthesizes on every call, so merging their test functions saves process setup and not synthesis time. They are also weaker than their names suggest — each asserts only `verify().is_err()`, never which constraint failed, so `circuit_rejects_merkle_tree_commitment_carry_violation_in_same_epoch_step` still passes if `assert_merkle_tree_commitment_link` is deleted outright.

## What

Group the public-input-tamper invocations by transition context and assert each case's exact failure signature. Public inputs are bound by copy constraints, so a tampered row surfaces as a `VerifyFailure::Permutation` at that row; asserting the exact row set is what keeps one multi-field synthesis as informative as separate ones. This takes 20 `MockProver` invocations to 11.

## How

Paths relative to `mithril-stm/src/circuits/halo2_ivc/tests/`.

- [x] Add the failure-signature helper, generic over `Circuit<NativeField>` with an `IvcCircuitData` wrapper: require rejection, require every returned failure to be `VerifyFailure::Permutation`, assert the exact set of failing rows on the plaintext instance column, report field names rather than integers
  - [x] Public-input layout module with named row accessors; no row literals in tests
  - [x] Fast guards on the section lengths, plus a distinct-sentinel test pinning `State::as_public_input` order
  - [x] Fast unit test on a minimal circuit modelling both instance columns, with one deliberately unconstrained ordinary input
- [x] Consolidate the public-input-tamper cases into one multi-field case per transition context, renamed as output-binding tests
  - [x] `transitions/negative/genesis.rs`: its own case plus the three global-field cases from `in_circuit/public_inputs.rs`
  - [x] `transitions/negative/same_epoch.rs`: its three cases plus the three from `in_circuit/state_transition.rs`, which cover different fields and are folded in rather than deleted
  - [x] `transitions/negative/next_epoch.rs`: its three cases
  - [x] Delete the emptied `mod slow` of `in_circuit/public_inputs.rs` and `in_circuit/state_transition.rs`
- [x] `encoding/negative.rs`: keep the three range cases as separate invocations, each asserting the row of the field derived from its tampered byte range
  - [x] Add a message-gate case: XOR one snapshot-digest byte, assert the offset lies outside all three extracted ranges, require rejection with an empty instance-row set
- [x] `in_circuit/accumulator.rs`: apply the same exact-row assertion to the accumulator tamper
- [x] Keep both `in_circuit/genesis_gating.rs` cases as separate invocations — the two contributions are gated and then accumulated together, so a joint stimulus would admit cancellation
- [x] Correct the message-gate naming: the gate is SHA-256 via `sha2_256_chip`, not Blake2b, affecting a test name and doc comments in `in_circuit/state_transition.rs` and `transitions/negative/genesis.rs`. The Blake2b in `verify_prepare_blake2b_recursive_proof` is the proof transcript and is correct
- [x] Assess the gain and report it in the issue: measured local aggregate before and after, plus the observed CI delta 

Depends on #3462.

hjeljeli32:
## Content

This PR consolidates the MockProver tests in `circuits::halo2_ivc` and replaces every negative case's `verify().is_err()` with an exact failure signature. Consolidating test *functions* saves nothing, since `MockProver::run` re-synthesizes on every call, so only removing invocations helps. The tests were also weaker than their names: six were found to reject *even with their tampering removed*, because the cheap stimulus pairs empty proof slots with a trivial accumulator, which balances only at genesis where the gating scales both accumulator contributions to the group identity.

Measured locally at `-j 4` on the `ci-tests-snark` profile: the slow tier goes from 403.2 s to 277.9 s aggregated and 107.0 s to 76.5 s wall, 22 tests becoming 15 and 20 MockProver invocations becoming 14. Net of −6.5 s drift on untouched tests, −118.7 s is attributable; about 65 s of the saving went on four deliberately added invocations. Fast tier unchanged at 541 passing. Per-item numbers are in #3464.

### Changes

- **Failure-signature helper** (`tests/common/failure_signature.rs`): a runner generic over `Circuit<NativeField>` asserting that the failing rows on the public-statement instance column are exactly the expected set and that every failure is a permutation failure. Unit-tested on a minimal two-instance-column circuit.
- **Public-input layout** (`tests/common/public_input_layout.rs`): named row accessors for the global, state and accumulator sections, so no test carries a row literal, guarded by section-length and serialization-order tests.
- **Satisfiable non-genesis fixtures** (`tests/common/helpers.rs`, `tests/transitions/positive.rs`): same-epoch and next-epoch stimuli built from the committed step assets, so nothing is proved at test time, plus two untampered acceptance cases; the first in-circuit positive coverage of non-genesis steps.
- **Consolidated binding cases** (`tests/transitions/negative/`): thirteen single-field cases become three, one per transition context. Coverage widens at no cost, the step counter and the two genesis verification-key rows having been untested.
- **Preimage encoding cases** (`tests/encoding/negative.rs`): the three byte-range cases now assert the message row plus the decoded field their window feeds, flipping one byte rather than filling a range. A new case isolates the message equality.
- **Accumulator case** (`tests/in_circuit/accumulator.rs`): an untampered canary on a clone, then the whole accumulator section asserted rather than one element. Keeps its freshly proved certificate, the only unseen proof the circuit verifies.
- **Row-dependent mutation delta**: mutations add `row + 1`, since two tampered cells in one permutation class stay equal under a shared delta and left one accumulator row unreported.
- **Deletions**: `in_circuit/state_transition.rs` in full and `public_inputs.rs`'s slow module: six cases, three vacuous and three duplicating the genesis case. Both weak rejection helpers are gone.
- **Genesis-only stimulus made explicit**: the trivial-accumulator builders are renamed to say genesis and assert the step counter is zero, their generic names having invited the non-genesis use.
- **Not covered, unchanged from `main`**: the chain-link constraints have no negative coverage, which the deleted cases did not provide either, and an exact row set does not prove each row carries the field named for it. Both are recorded for follow-up.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation
  - [x] No new TODOs introduced

## Issue(s)
Closes https://github.com/IntersectMBO/mithril/issues/3464


mithril-stm/src/circuits/halo2_ivc/tests/common/failure_signature.rs
```diff
@@ -0,0 +1,308 @@
...
+    fn helper_accepts_rejection_with_an_empty_expected_signature() {
+        // A break confined to the committed column implicates no public-statement row.
+        let mut instances = honest_minimal_instances();
+        mutate_public_input(&mut instances[0], 0);
+
+        assert_circuit_rejects_public_input_rows(&MinimalCircuit, instances, &BTreeMap::new());
+    }
+
+    #[test]
+    #[should_panic(expected = "public-input failure signature mismatch")]
```
damrobi: Nitpick but can we try to avoid those `should_panic`?

damrobi (approved):
LGTM 👍 
Just one small nitpick and the Copilot comments that seem relevant

jpraynaud (approved):
LGTM 👍

Issue #3466, opened 2026-08-04
Recover the ignored IVC prover input preparation tests


jpraynaud:
## Why

The 7 `prepare_*` tests in `ivc_halo2_snark/prover_input.rs` (genesis, same-epoch, and next-epoch happy paths and rejections) are `#[ignore]` dead coverage that no CI tier ever runs, and they rely on `OnceLock` fixture cells that never amortize anything under the nextest process-per-test model.

Two of them are also broken against the current API: `prepare_at_genesis_produces_advanced_state_and_witness` and `prepare_rejects_invalid_genesis_signature` fail with `invalid epoch transition at last committed epoch 0`, because `IvcProverInput::prepare` no longer dispatches to `prepare_genesis` and `IvcProver::prove` now rejects a genesis rolling state outright. Being ignored, they never reported it.

Each of the remaining tests costs about 8 seconds, measured locally, because it builds a full `IvcSnarkProverSetup`: a 1.0 GB recursive proving key and a 96 MB SRS. `prepare` never reads the proving key: it reads the certificate verifying key, the KZG verifier parameters and the certificate and IVC fixed-base maps, all of which are already committed as test assets.

## What

Make the dependency `prepare` actually has explicit, then recover the tests against it: they need no keygen, no proving key, no SRS and no cache, so they become sub-second and run on every PR instead of nightly only. Keep the two-level fast/slow convention of DEV-ADR section 8 (no new tier).

## How

Paths are relative to `mithril-stm/src/` unless they start with `.github/` or `.config/`.

- [x] Add `IvcProverInputVerificationContext` in `proof_system/ivc_halo2_snark/prover_setup.rs`, holding the certificate verifying key, the IVC verifying key, the certificate and IVC fixed-base maps and the KZG verifier parameters, with private fields
  - [x] Production constructor deriving it from `IvcSnarkProverSetup`, taking the verifier parameters from `self.srs.verifier_params()` and not from the embedded `KZG_VERIFIER_PARAMS` constant, which describes the trusted production SRS and would be wrong for the unsafe SRS used by tests and benchmarks
  - [x] Test-only constructor taking the two verifying keys and the verifier parameters, deriving both fixed-base maps via `fixed_bases_and_names_from_verifying_key` so inconsistent maps are not representable
[...]

PR #3505, opened 2026-08-30
test(stm): split the IVC prover monolith into separated tests

### LINKED ISSUE #3468: Replace `prove_all_scenarios` monolith with separated tests
## Why

Grouping many scenarios into one test to amortize the setup in-process is an anti-pattern: it hides per-case results, defeats per-test retries, and couples unrelated assertions.

`prove_all_scenarios` measures 1100 to 1394 seconds across the last five nightlies. With the two keygen-only tests of the single-threaded `ivc-snark-sequential` group it forms an 1830 second serial tail that runs after all 2261 other tests of the `future_snark` job have finished, and that tail is 59% of the job's 3116 second test step.

Splitting cannot buy wall clock back. The test reports `user / real = 7.39`, so it already saturates the runner's four vCPUs, and its peak RSS is 11.02 GiB against 16 GB, so the sequential group has to stay and a second concurrent prover test would not fit. Every extra test process is therefore pure added nightly time, roughly 70 to 125 seconds each: promoting all eight helpers with their own setups models at 1882 to 2269 seconds against 1394 today, which is the same overhead that forced the June 2026 consolidation, only smaller now that the disk caches turn a cold keygen into a deserialize.

## What

Separate the tests, and pay for the split by removing proving work that is now redundant rather than by adding setup work. The five rejection paths generate no proofs and belong in the fast tier; the six proving invocations of the positive paths cover only four distinct transition-type and transcript pairs, so they reduce to four across two tests (JP approved).

## How

Paths are relative to `mithril-stm/src/` unless they start with `.github/` or `.config/`.

- [x] Move four rejection scenarios of `proof_system/ivc_halo2_snark/proof.rs` into the fast tier of `proof_system/ivc_halo2_snark/prover_input.rs` as `IvcProverInput::prepare` tests asserting typed errors instead of `is_err()`, since all five fail inside `prepare`, which since #3466 needs no SRS, no keygen and no cache: mismatched certificate message, mismatched aggregate verification key, corrupted previous IVC proof, and mismatched `Global`
- [x] Delete the corrupted-certificate-proof rejection scenario, an exact duplicate of `prepare_rejects_invalid_snark_proof`
- [x] Promote `run_bootstrap_path` (three proofs: genesis Poseidon, first non-genesis next-epoch Poseidon, returned next-epoch Blake2b) and `run_same_epoch_path` (one proof, over the committed chain state's real previous IVC proof and non-degenerate folded accumulator) to standalone `#[test]` functions sharing a plain context builder, with no `OnceLock` or `LazyLock`: nextest gives each test its own process, which is why the June consolidation's cells amortized nothing
- [x] Delete `run_next_epoch_path` and its two proving invocations: it contributes no new transition-type and transcript pair, and `prepare_at_next_epoch_carries_lookahead_protocol_parameters` already pins its next state and next accumulator against the same asset, plus four witness fields it does not check
- [x] Keep `prepare_at_next_epoch_carries_lookahead_protocol_parameters`, `next_epoch_step_circuit_is_accepted`, the positive previous-certificate and previous-IVC-proof circuit checks for next-epoch, and `next_epoch_accumulator_passes_check` with its tampering canary: they carry the mature next-epoch coverage the reduction relies on. The decomposition holds only because no branch couples transition type, incoming-accumulator provenance and outer transcript together; restore a mature next-epoch proving case if production later introduces transition-specific folding, public-input assembly, or transcript selection
- [x] Build the slow-test `Global` and genesis bootstrap input from the committed genesis benchmark fixture, so the prover tests consume only the committed inputs they exercise and no longer depend on asset-generation machinery. This removes real work — neither process reassembles the 3000-signer fixture, Merkle tree and certificate relation — but produced no measurable wall-time change, because the ~1.0 GB recursive proving key deserialize dominates setup
- [x] Keep the verifying-key transcript-representation assertions that detect drift between the committed assets and freshly generated keys
- [x] Delete `prove_all_scenarios`, keeping the two positive tests in `mod slow` so they stay in the `ivc-snark-sequential` group (`.config/nextest.toml`) and watched by the `proof.rs` entry of `.github/workflows/scripts/filter-slow-tests.sh`
- [x] Assess the gain with cold and warm caches, reporting setup and proving phases separately, and report it in the issue

The two keygen-only tests of the serial tail (roughly 413 and 98 seconds) never prove, so the memory ceiling that justifies the sequential group likely does not apply to them; moving only those out needs its own measurement and a separate ticket.

Depends on #3433.

hjeljeli32:
## Content

`prove_all_scenarios` ran eight scenarios in one test to amortize a single setup. It costs 1393.9 s on the nightly and, with the two keygen-only `prover_setup` tests of the single-threaded `ivc-snark-sequential` group, forms an **1830 s serial tail, 59 % of the job's 3116 s test step**.

Splitting it naively costs more than it saves: at `user / real = 7.39` the test already saturates the runner's four vCPUs and peaks at 11.02 GiB against 16 GB, so the group must stay serial and each extra process adds 70-125 s. The split is paid for by removing proving instead — the five rejection scenarios generate no proofs, and the six `prove_with_transcript` calls cover only four distinct (transition type, transcript) pairs. Warm and local: **175.5 s to 126.0 s, −28 %**. On CI, normalised against the two untouched keygen-only tests of the same group, the two replacements cost 951 s against the monolith's 1264 s (**−313 s, −25 %**) taking the serial tail to 1408 s and the job's test step to 2348 s.

### Changes

- **Rejection scenarios to the fast tier** (`prover_input.rs`, `proof.rs`): four `run_rejects_*` helpers become `prepare` tests asserting typed `IvcCircuitError` variants instead of `is_err()`, at ~0.36 s on every PR rather than nightly only, each failed inside `prepare`, which since #3466 needs no SRS, keygen or cache. The fifth duplicated `prepare_rejects_invalid_snark_proof` and is deleted.
- **Monolith split, mature next-epoch proving removed** (`proof.rs`): `prove_bootstrap_produces_first_epoch_proof_and_rolling_state` (3 proofs) and `prove_same_epoch_produces_proof_without_rolling_state` (1 proof) share a plain `build_slow_test_context()`, no `OnceLock`, which amortizes nothing under nextest's process-per-test model. `run_next_epoch_path` adds no new pair, and `prepare_at_next_epoch_carries_lookahead_protocol_parameters` already pins its state and accumulator against the same asset.
- **Context decoupled from asset generation** (`proof.rs`): the slow context builds `Global` and the genesis bootstrap input from the committed genesis fixture rather than `AssetGenerationSetup`. No measurable wall-time change — setup spans 7.1-8.7 s for identical artifacts and the ~1.0 GB proving key deserialize dominates — so it is kept for the decoupling, not for speed.
- **Unchanged**: `.config/nextest.toml` and `filter-slow-tests.sh` match on module path, so the renamed tests stay grouped and watched. The coverage the reduction rests on still passes; restore a mature next-epoch proving case if production later couples transition type with accumulator provenance, public inputs or transcript selection.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation
  - [x] No new TODOs introduced

## Issue(s)
Closes https://github.com/IntersectMBO/mithril/issues/3468

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4834,7 +4834,7 @@ dependencies = [
 
 [[package]]
 name = "mithril-stm"
-version = "0.12.10"
+version = "0.12.11"
 dependencies = [
  "anyhow",
  "blake2 0.10.6",
--- a/mithril-common/Cargo.toml
+++ b/mithril-common/Cargo.toml
@@ -51,7 +51,7 @@ fixed = "1.31.0"
 hex = { workspace = true }
 kes-summed-ed25519 = { version = "0.2.1", features = ["serde_enabled", "sk_clone_enabled"] }
 mithril-merkle-tree = { path = "../internal/mithril-merkle-tree", version = "0.1.4" }
-mithril-stm = { path = "../mithril-stm", version = "0.12.10", default-features = false }
+mithril-stm = { path = "../mithril-stm", version = "0.12.11", default-features = false }
 nom = "8.0.0"
 rand_chacha = { workspace = true }
 rand_core = { workspace = true }
--- a/mithril-stm/CHANGELOG.md
+++ b/mithril-stm/CHANGELOG.md
@@ -5,6 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## 0.12.11 (08-31-2026)
+
+### Changed
+
+- Split the monolithic `prove_all_scenarios` slow test into a bootstrap test and a mature same-epoch test, and dropped the mature next-epoch proving, so that the IVC prover tier generates four real proofs instead of six while still covering every reachable transition type and transcript pair.
+- Moved the prover rejection scenarios to the fast tier as `IvcProverInput::prepare` tests asserting typed errors, and deleted the one duplicating an existing case, so that they run on every pull request rather than the nightly only.
+- Built the slow test context from the committed genesis benchmark fixture instead of the asset generation setup, so that the prover tests consume only the committed inputs they exercise.
+
 ## 0.12.10 (08-28-2026)
 
 ### Changed
--- a/mithril-stm/Cargo.toml
+++ b/mithril-stm/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mithril-stm"
-version = "0.12.10"
+version = "0.12.11"
 edition = { workspace = true }
 authors = { workspace = true }
 homepage = { workspace = true }
--- a/mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
+++ b/mithril-stm/src/proof_system/ivc_halo2_snark/proof.rs
@@ -977,10 +977,7 @@ mod tests {
         use std::sync::Arc;
         use std::time::Instant;
 
-        use midnight_circuits::{
-            hash::poseidon::PoseidonState, verifier::Accumulator, verifier::BlstrsEmulation,
-        };
-        use midnight_proofs::utils::SerdeFormat;
+        use midnight_circuits::hash::poseidon::PoseidonState;
         use rand_core::OsRng;
 
         use crate::{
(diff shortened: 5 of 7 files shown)


jpraynaud (approved):
LGTM 👍

damrobi (approved):
LGTM 👍

Issue #3148, opened 2026-04-01
Implement circuit verification key whitelist/revoke mechanism


jpraynaud:
## Why

Any change in a circuit (logic, dependency update, parameters) changes its verification key, and a circuit or a key can turn out to be insecure. There is no mechanism today to invalidate the certificates produced with it: the trust in the circuit verification key is compile time only, and clients have no signal to reject a range of certificates.

## What

Prepare a revocation mechanism for certificates in defined epoch ranges. The revocation list is published, signed by the genesis key, and must be verified by clients prior to verification. 

## How

- [ ] Specify the format of the whitelist and revocation list (circuit verification keys, epoch ranges) and how the genesis key signs it
- [ ] Publish the list at the root of the repository
- [ ] Implement the creation and the signing of the list with the genesis key (new command in the aggregator)
- [ ] Enforce the whitelist and revocation checks on the circuit verification key in the prover and the verifier setup
- [ ] Update the documentation
- [ ] Create a runbook for the revocation procedure

PR #3513, opened 2026-09-02
feat: introduce the circuit verification key registry

jpraynaud:
## Content

This PR includes the **machinery of the genesis signed circuit verification key registry, a whitelist with epoch ranges and retroactive revocations for the circuit verification keys trusted by SNARK certificates**:

- Add `CircuitVerificationKeyDigest` (Poseidon hash of the transcript representation bytes) and a `requires_certified_circuit_verification_keys` predicate on `AggregateSignatureType` in `mithril-stm`, with golden tests pinning the digests
- Add the registry types with revocation check semantics in `mithril-common`, signed over the exact registry JSON bytes with a domain-separation prefix so the schema stays evolvable after publication
- Add the registry retriever trait with a file implementation and test doubles
- Add the certifier trait with a caching decorator 

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Comments

The enforcement in the certificate verifier, the client and the aggregator, along with the genesis tooling and the runbook, comes in a follow-up PR https://github.com/IntersectMBO/mithril/pull/3514 stacked on this one.

## Issue(s)
Relates to #3148


mithril-common/src/crypto_helper/circuit_key_registry/certifier.rs
```diff
@@ -0,0 +1,537 @@
...
+    #[cfg(test)]
+    fn with_cache_time_to_live_in_seconds(mut self, cache_time_to_live_in_seconds: i64) -> Self {
+        self.cache_time_to_live_in_seconds = cache_time_to_live_in_seconds;
+        self
+    }
+
+    /// Whether the cached registry is still within its time to live.
+    fn is_cache_fresh(&self, cache: &VerifiedRegistryCache) -> bool {
+        (Utc::now() - cache.refreshed_at).num_seconds() < self.cache_time_to_live_in_seconds
+    }
```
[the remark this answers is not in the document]
jpraynaud: Fixed

mithril-common/src/crypto_helper/circuit_key_registry/retriever.rs
```diff
@@ -0,0 +1,139 @@
...
+    /// Build a retriever reading the given signed registry JSON file.
+    pub fn new(registry_file_path: PathBuf) -> Self {
+        Self { registry_file_path }
+    }
+
+    /// Read the signed registry JSON file and parse it.
+    fn read_and_parse_registry_file(
+        &self,
+    ) -> Result<SignedCircuitVerificationKeyRegistry, StdError> {
+        let json = std::fs::read_to_string(&self.registry_file_path).with_context(|| {
```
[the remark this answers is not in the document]
jpraynaud: Fixed

mithril-common/src/crypto_helper/circuit_key_registry/registry.rs
```diff
@@ -0,0 +1,503 @@
...
+}
+
+impl CircuitVerificationKeyRegistry {
+    /// Check that every digest is whitelisted and not revoked for the given epoch.
+    ///
+    /// A digest fails with [Revoked](CircuitVerificationKeyRegistryError::Revoked) when any
+    /// revoked entry covers the epoch, and with
+    /// [NotWhitelisted](CircuitVerificationKeyRegistryError::NotWhitelisted) when no allowed
+    /// entry covers it.
+    pub fn check(
```
damrobi, line 98: We only return the first digest that is not allowed. Could it be interesting to get the list of all rejected digests that fails?

damrobi (approved):
LGTM 👍

mithril-stm/src/circuits/verification_key_digest.rs
```diff
@@ -0,0 +1,297 @@
...
+/// Serialized as a lowercase hex string.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, PartialOrd, Ord)]
+pub struct CircuitVerificationKeyDigest([u8; CIRCUIT_VERIFICATION_KEY_DIGEST_SIZE]);
+
+impl CircuitVerificationKeyDigest {
+    /// Compute the digest of a verifying key from its canonical byte serialization.
+    pub(crate) fn try_from_verification_key<K: TryToBytes>(
+        verification_key: &K,
+    ) -> StmResult<Self> {
+        Ok(Self::from_canonical_key_bytes(
```
hjeljeli32: In this implementation, we hash the serialized key bytes. I propose also including `VerifyingKey::transcript_repr()` to explicitly bind the digest to the gate definitions.

Rationale: the recursive key serialization omits the constraint system, which is reconstructed from circuit code when reading the key.

I checked this with two small circuits on `midnight-proofs 0.8.1`, using the same layout, test SRS, and `k = 4`. Only the gate changed: `a * b = c` versus `a * b = 2c`.

```text
Serialized VK bytes: identical (390 bytes each)
transcript_repr(): different
Witness (2,3,6): accepted by the first, rejected by the second
```
Identical bytes necessarily give the same digest here, despite the different gate equations.

Midnight's `transcript_repr()` already includes the pinned constraint system and domain alongside the commitments. Conceptually:

```text
B = serialized key bytes
R = key.transcript_repr()

registry_digest = Poseidon(
    registry-specific versioned domain tag,
    byte length of B,
    B packed into 31-byte little-endian field elements,
    R
)
```
This preserves the serialized metadata and makes the gate binding explicit. 

WDYT?

mithril-stm/src/circuits/verification_key_digest.rs
```diff
@@ -0,0 +1,297 @@
...
+            &verification_key.to_bytes_vec()?,
+        ))
+    }
+
+    /// Compute the digest of a verifying key already in its canonical byte serialization.
+    fn from_canonical_key_bytes(canonical_key_bytes: &[u8]) -> Self {
+        let mut hasher = MidnightPoseidonDigest::new();
```
hjeljeli32: In this implementation, we feed serialized key bytes directly into `MidnightPoseidonDigest.` I propose that we pack them into 31-byte field elements and include the original byte length before hashing.
Rationale: The adapter splits the input into 32-byte chunks and reduces each chunk modulo the field. Different byte inputs, such as `x` and `x + p`, therefore become identical before Poseidon runs. 
With 31-byte chunks, each value fits in the field without reduction. Including the original length also distinguishes inputs that differ only by trailing zeros. This explains the packing step proposed in my previous comment.
WDYT?

hjeljeli32 (commented):
Looks good but I left a couple of suggestions.

mithril-stm/src/circuits/verification_key_digest.rs
jpraynaud: Fixed by Sha256 the bytes before feeding the Poseidon hasher.

mithril-stm/src/circuits/verification_key_digest.rs
jpraynaud: Nice catch @hjeljeli32! 

Fixed by using the transcript representation (and I also added some DST).

mithril-common/src/crypto_helper/circuit_key_registry/registry.rs
jpraynaud: Fixed, it now returns all the digests that fail.

damrobi (approved):
LGTM 👍

hjeljeli32 (approved):
LGTM 👍

turmelclem (approved):
LGTM 👍

mithril-common/Cargo.toml
```diff
@@ -59,7 +59,7 @@ rayon = { workspace = true }
 semver = { workspace = true }
 serde = { workspace = true }
 serde_bytes = "0.11.19"
-serde_json = { workspace = true }
+serde_json = { workspace = true, features = ["raw_value"] }
```
Alenar: Given how central `mithril-common` is, if a feature is enabled on a workspace level crate, it should probably be enabled at the workspace `Cargo.toml` instead.

Alenar (approved):
LGTM

mithril-common/Cargo.toml
jpraynaud: Fixed

Issue #3465, opened 2026-08-04
Add behavior tests for the IVC wiring


jpraynaud:
## Why

The clerk entry point and the prover-input orchestration are exercised only through slow end-to-end tests, and several error paths are unreachable without a real prover.

## What

Add behavior tests with mockall doubles at the IVC wiring seams (clerk entry point and prover input preparation) to verify orchestration, error mapping and state threading, without re-testing what the proof system module already covers: proof generation and verification stay covered by the existing prover and circuit tests.

## How

Paths are relative to `mithril-stm/src/`.

- [x] Introduce one prover trait per proof system, `CertificateProver` (`proof_system/halo2_snark`, implemented by `SnarkProver`) and `IvcStepProver` (`proof_system/ivc_halo2_snark`, implemented by `IvcProver`), with a single-argument `IvcStepInput` carrying all inputs of an IVC step
- [x] Inject the provers into `Clerk` through a `SnarkProverFactory` built lazily per aggregation branch, with a test-gated constructor taking the mock and no public API change
- [x] Add behavior tests for `Clerk::aggregate_signatures_with_type` (`protocol/aggregate_signature/clerk.rs`) with mocked factory and provers, covering input assembly, error mapping and rolling-state threading for the `Snark` and `IvcSnark` branches; move the two existing precondition tests onto `IvcStepInput::try_new`
- [x] Cover the certificate-rejection path of `IvcProverInput::prepare` (`proof_system/ivc_halo2_snark/prover_input.rs`) in the fast tier with empty proof bytes against the dummy IVC setup
- [x] Retire the `prover_input.rs` slow tests made redundant by the fast tier, promote the two that need no setup, and un-ignore the ones that stay so the nightly slow run executes them
  -  If I understand correctly this was done in this [PR](https://github.com/IntersectMBO/mithril/pull/3502)
- [x] Align the trait, type and function names introduced here with the recursive / non-recursive harmonization of #3421 (prefix vs suffix, module layout), so this issue does not introduce a third convention
- [ ] Assess the gain (expected: closes the wiring gaps at zero runtime cost) and report it in the issue

#### Seams: one trait per proof system

Each trait mirrors an existing module and has a single production implementor that already exists. Both are pinned to `MithrilMembershipDigest` because the clerk already pins it, and both are object safe.

[...]

PR #3515, opened 2026-09-03
refactor(stm): harmonize the recursive and non-recursive SNARK stacks

### LINKED ISSUE #3421: Harmonize the recursive and non-recursive SNARK implementations in STM
## Why

The recursive (IVC) and non-recursive SNARK stacks were built incrementally and drifted apart. Naming and module layout diverged for equivalent concepts, and related preimage types follow inconsistent API conventions, which makes the library harder to read, to review and to extend.

## What

Align the two stacks on a single set of names and a single module structure, without changing the proving or verifying behavior. Where a difference turns out to be intrinsic rather than drift, keep it and document it rather than forcing a symmetry that does not hold.

Note that the crate names the same two stacks differently per layer: certificate and IVC for the circuits, non-recursive and recursive for the keys, SNARK and IVC for the proof systems. This ticket makes each layer internally consistent rather than forcing one word per stack, which would reopen the `Snark` names `mithril-common` depends on and the per-circuit key newtypes.

## How

Paths are relative to `mithril-stm/src/` unless they start with `.github/` or `.config/`.

- [x] Delete the three `#[allow(dead_code)]` markers whose stated condition no longer holds: two on `IvcCircuitData` (`circuits/halo2_ivc/circuit.rs`), now that `IvcProver::prove` constructs it, and one on `IvcProofError` (`proof_system/ivc_halo2_snark/errors.rs`), now that `IvcProof::verify` is called from `protocol/aggregate_signature/signature.rs`
- [x] Rename `StmCertificateCircuit` to `CertificateCircuit`, dropping the crate-redundant `Stm` prefix, and keep `IvcCircuitData`: `CertificateCircuit` is a parameter-only Midnight `Relation` whose instance and witness are supplied separately, while the raw Halo2 `IvcCircuitData` bundles the step values and the verifier metadata synthesis needs. Rename its abbreviated `acc` field to `accumulator`, and update the three documentation mentions in `mithril-stm/README.md` and `circuits/halo2_ivc/tests/assets/README.md`, leaving historical changelog entries unchanged
- [x] Name the error types by layer rather than by producing type: `StmCircuitError` becomes `CertificateCircuitError`, and `SnarkError` moves out of `proof_system/halo2_snark/mod.rs` into its own `errors.rs` as `SnarkProofError`
- [x] Replace the error messages that embed Rust symbol names (`circuits/halo2_ivc/errors.rs`, eight of them) with domain-level text: a symbol name tells a reader of the log nothing they can act on, `thiserror` and the anyhow context chain already carry the provenance a developer needs, and the messages then cannot go stale when a symbol is renamed
- [x] Rename `IvcSnarkProverSetup` to `IvcProverSetup`, the only recursive type still carrying `Snark`, and document which differences between the two setup families are intrinsic — the recursive one holds two verifying keys and three fixed-base maps because the recursive circuit verifies certificate proofs
- [x] Finish the `merkle_root` to `merkle_tree_commitment` rename on the non-recursive side (`circuits/halo2/`, `proof_system/halo2_snark/`), which an earlier workstream applied to the recursive stack only. `circuits/halo2/bench/helpers.rs` currently reads `merkle_tree_commitment: MerkleRoot`. The two error variants become `InvalidMerkleTreeCommitmentBytesLength` and `NonCanonicalMerkleTreeCommitmentBytes`, and the byte form is spelled `merkle_tree_commitment_bytes` to match how the rest of the crate already names it
- [x] Keep the two `MerkleTreeCommitment` wrappers distinct — an alias over the arithmetic wrapper `CircuitBaseField` on one side, an opaque domain newtype on the other — since unifying them settles an alias-versus-newtype question that belongs in its own ticket
- [x] Give `GenesisMessagePreimage` (`protocol/aggregate_signature/preimage.rs`) a private field with `From<Vec<u8>>` and `as_bytes()`, and `TryFrom<&GenesisMessagePreimage> for MessageHash` in place of `TryInto`. Keep `PartialEq<[u8]>`, which `mithril-common` consumes through the public `AncillaryGenesisData::genesis_message_preimage()` getter, and retain `Deref` so no existing caller has to change
- [x] Move `IvcTransitionType` into `rolling_state.rs` and replace `try_compute_transition_type` and `assert_correct_parameters` with one `validate_transition` that classifies, validates the linkage rules and returns the transition type. Keep `is_genesis`, which guards `IvcProver::prove`, and delete the test-only `is_next_epoch`
- [x] Delete the `IvcTransitionType::Genesis` validation arm and the `genesis_always_passes` test that pins it: the classifier never returns `Genesis`, and the genesis path builds its prover input without validating a transition. The variant itself stays
- [x] Rename the `Write` and `Read` traits of `circuits/halo2_ivc/io.rs` to `WriteWithFormat` and `ReadWithFormat`, which say what distinguishes them from `std::io` and from `TryToBytes`/`TryFromBytes`: they take a `SerdeFormat`
- [x] Rename `proof_system/ivc_halo2_snark` to `proof_system/halo2_ivc_snark`, so the IVC marker is a suffix as in `circuits/halo2_ivc`, and give it the facade shape `proof_system/halo2_snark` already has
- [x] Update `.config/nextest.toml` and `.github/workflows/scripts/filter-slow-tests.sh` in the same commit as that rename: both key on the module path and both fail silently, the first by dropping the memory-heavy prover tests out of their single-threaded group, the second by no longer enabling the slow tests of the paths it watches
- [x] Remove the duplicated SNARK-message computation in `IvcProverInput::prepare`: `validate_transition` builds it to compare the commitment and discards the message hash, then `prepare` builds it again from the same arguments. Have `validate_transition` return the pair it already computed
- [x] Take #3465's code as it stands when rebasing onto it, renaming nothing while resolving conflicts, then align its names in a pass of its own, since that ticket defers the convention to this one: `SnarkAggregateSignatureProver::verification_key` becomes `verifying_key`, matching `IvcChainProver` and the `*VerifyingKey` both return; and `IvcChainInput` becomes `IvcChainStepBundle`, which keeps `IvcProverInput` symmetric with `SnarkProverInput` — each stack's `prover_input.rs` describing what its circuit consumes

hjeljeli32:
## Content

The recursive (IVC) and non-recursive SNARK stacks drifted apart: equivalent concepts had different names, the module layout differed, and related preimage types followed inconsistent conventions. This aligns both on one set of names and one module structure **without changing proving or verifying behaviour**. Where a difference turned out to be intrinsic rather than drift, it is kept and documented rather than forced into a false symmetry.

### Changes

- **Naming.** `StmCertificateCircuit` → `CertificateCircuit`, deliberately reversing the earlier `StmCircuit` → `StmCertificateCircuit` rename since the prefix says nothing inside `mithril-stm`. Error types named by layer, not by producing type: `StmCircuitError` → `CertificateCircuitError`, and `SnarkError` → `SnarkProofError` in its own `errors.rs`. Eight IVC error messages that embedded Rust symbol names become domain-level text. `merkle_tree_commitment` naming finished on the non-recursive side. `Write`/`Read` in `circuits/halo2_ivc/io.rs` → `WriteWithFormat`/`ReadWithFormat`, which is what distinguishes them from `std::io`.
- **`IvcCircuitData` kept, not renamed**, and documented instead: `CertificateCircuit` is a parameter-only `Relation` whose instance and witness arrive separately, while `IvcCircuitData` bundles a step's values with the verifier metadata synthesis needs.
- **`GenesisMessagePreimage`** gains a private field with `From<Vec<u8>>`/`as_bytes()` and `TryFrom<&GenesisMessagePreimage> for MessageHash`. `Deref` and `PartialEq<[u8]>` are retained — `mithril-common` uses the latter — so no caller changes.
- **Transition validation moves onto `IvcRollingState`** *(control flow)*. `try_compute_transition_type` and `assert_correct_parameters` collapse into `validate_transition`, which classifies once and returns the type it validated, so no caller can pass one that disagrees with the state. Error precedence unchanged. The unreachable `IvcTransitionType::Genesis` validation arm and the test-only `is_next_epoch` are removed; the variant and `is_genesis` stay.
- **The certificate SNARK message is built once per step** *(control flow)*. `validate_transition` already built it and discarded half; it now returns the pair, so the validated values are the ones the next state and witness carry. It stays inside `validate_transition` because classification must run first, so a bad epoch still reports before a malformed message.
- **`proof_system/ivc_halo2_snark` → `halo2_ivc_snark`**, suffixed as in `circuits/halo2_ivc`, with every submodule made private and the surface re-exported explicitly like `halo2_snark`. External visibility is unchanged. `.config/nextest.toml` and `filter-slow-tests.sh` move in the same commit: both key on the module path and **both fail silently**. Verified the nextest selector still finds the same four slow prover tests.
- **Names from #3465 aligned**, which that issue deferred to this one. `SnarkAggregateSignatureProver::verification_key` → `verifying_key`, matching its counterpart and the `*VerifyingKey` it returns. `IvcChainInput` → `IvcChainStepBundle`, since it sat one word from `IvcProverInput` while meaning something else. Renaming this side keeps `SnarkProverInput`/`IvcProverInput` symmetric — each in its stack's `prover_input.rs`, each describing what its circuit consumes.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation
  - [x] No new TODOs introduced

## Comments
Closes https://github.com/IntersectMBO/mithril/issues/3421


mithril-stm/src/proof_system/halo2_snark/message.rs
```diff
@@ -2,24 +2,24 @@ use anyhow::Context;
 
 use crate::{StmResult, signature_scheme::BaseFieldElement};
 
-/// Build the SNARK message from a Merkle tree root and a raw message.
+/// Build the SNARK message from a Merkle tree commitment digest and a raw message.
```
damrobi, line 5: I'm not sure it's necessary to add "digest" to "Merkle tree commitment" here. In the rest of the file I would use "commitment" instead of "digest"

mithril-stm/src/proof_system/halo2_snark/mod.rs
```diff
@@ -154,15 +143,14 @@ mod tests {
         proptest! {
             #![proptest_config(ProptestConfig::with_cases(20))]
 
-            /// Verifies that `build_snark_message` encodes the Merkle root
+            /// Verifies that `build_snark_message` encodes the Merkle tree commitment digest
```
damrobi, line 146:
```suggestion
            /// Verifies that `build_snark_message` encodes the Merkle tree commitment
```
Same as before I think commitment is enough

mithril-stm/src/proof_system/halo2_snark/proof.rs
```diff
@@ -89,11 +89,12 @@ impl<D: MembershipDigest> SnarkProof<D> {
         snark_verifier_data: &SnarkVerifierData,
         verifier_params: &ParamsVerifierKZG<Bls12>,
     ) -> StmResult<()> {
-        let merkle_root = &aggregate_verification_key_for_snark.get_merkle_tree_commitment().root;
-        let proof_message = build_snark_message(merkle_root, message)?;
+        let merkle_tree_commitment_digest =
+            &aggregate_verification_key_for_snark.get_merkle_tree_commitment().root;
+        let proof_message = build_snark_message(merkle_tree_commitment_digest, message)?;
```
damrobi, line 94:
```suggestion
        let merkle_tree_commitment =
            &aggregate_verification_key_for_snark.get_merkle_tree_commitment().root;
        let proof_message = build_snark_message(merkle_tree_commitment, message)?;
```

mithril-stm/src/proof_system/halo2_snark/setup.rs
```diff
@@ -33,12 +33,14 @@ use crate::{
 /// Bundles the one-time setup artifacts needed to prove and verify SNARK proofs.
 ///
 /// This includes the Structured Reference String (SRS), the compiled circuit, and the
-/// proving and verification keys derived from them.
+/// proving and verification keys derived from them. `IvcProverSetup` on the recursive side carries
```
damrobi, line 36: Not sure this addition to the comment is really needed here.

mithril-stm/src/proof_system/halo2_ivc_snark/rolling_state.rs
```diff
@@ -130,56 +137,81 @@ impl IvcRollingState {
...
+                    && self.state.step_counter.is_not_first_step()
             }
             IvcTransitionType::NextEpoch => {
-                self.state().next_merkle_tree_commitment == merkle_tree_commitment
+                self.state.next_merkle_tree_commitment == certificate_merkle_tree_commitment
             }
-            IvcTransitionType::Genesis => true,
+            // The classification above never yields `Genesis`: the genesis step is built by
+            // `IvcProverInput::prepare_genesis`, which never validates a transition.
+            IvcTransitionType::Genesis => unreachable!("genesis bypasses transition validation"),
```
damrobi, line 199: If this is unreachable then maybe we don't need the `IvcTransitionType::Genesis` variant of the enum?

damrobi (commented):
Looks good overall! Can you see if you can remove the Genesis transition type?

damrobi (approved):
LGTM 👍

jpraynaud (approved):
LGTM 👍

PR #3517, opened 2026-09-06
docs(stm): add runnable examples for the three paths

### LINKED ISSUE #3385: Update STM README with recursive/non-recursive examples
## Why

The README still states that Halo2 is not supported, and its only example covers the concatenation proof system, cluttered with inline `#[cfg(feature = "future_snark")]` attributes. A reader has no example of the non-recursive SNARK aggregation, nor of the recursive (IVC) proof.

## What

Update the README so that it describes the supported proof systems and gives a readable example for each of them.

## How

- [x] Update the introduction with the supported proof systems: concatenation, non-recursive SNARK and recursive SNARK, the last two experimental and gated behind `future_snark`
- [x] Add an example for the non-recursive SNARK aggregate signature in `/examples`, covering aggregation and verification
- [x] Add an example for the recursive SNARK aggregate signature in `/examples`, covering aggregation and verification across two chain steps. It runs from committed fixtures: a documentation-only signer seed and three rigid protocol-message preimages at epochs 0, 1 and 2
- [x] Use example-scale rather than production parameters in the recursive example so it can actually be run, at the cost of regenerating circuit keys on each aggregation call; state both in its header, with the measured cost
- [x] Commit the preimage generator in `mithril-common`, which owns the rigid protocol message builder, as an ignored test that proves and verifies the chain before writing anything, plus a fast test comparing rebuilt messages byte for byte with the committed files
- [x] Move the concatenation example into `/examples`, removing both copies: the one in the README and the duplicate in `src/lib.rs`
- [x] Link the examples from the README and include their sources in the crate-level documentation, so the rendered docs and the example folder cannot drift
- [x] Gate the two SNARK examples with `required-features = ["future_snark"]` so the set builds with and without the feature, and mark them `no_run`: both reach the trusted setup
- [x] Configure docs.rs to build with `future_snark`, and add the production verifying keys and the example fixtures to the package `include` list; the keys are embedded by non-test code but unpublished, so the released crate cannot build with the feature

hjeljeli32:
## Content

The README stated that Halo2 was not supported, and its only example covered the concatenation proof system. That example existed twice — in the README and copied into `src/lib.rs` — and the copies had drifted. There is now one runnable example per proof system in `examples/`, each covering aggregation and verification, linked from the README and the crate documentation.

Adding the examples surfaced a packaging bug. `package.include` listed no `.bin` files, while `include_bytes!` pulls two production verifying keys into non-test code, so the released crate could not compile with `future_snark` at all; `cargo publish --no-verify` in the release workflow is why it went unnoticed. `cargo package --features future_snark,rustls` now verify-builds, and all four examples build from the unpacked tarball.

### Changes

- **Three examples** (`concatenation_aggregate_signature.rs`, `non_recursive_snark_aggregate_signature.rs`, `recursive_snark_aggregate_signature.rs`): one per proof system, each aggregating and verifying, with `?` throughout rather than `unwrap`. The recursive one runs two chain steps and produces five proofs, the genesis bootstrap and both transcripts per step. The SNARK pair uses example-scale parameters so it can be run at all, which means circuit keys are regenerated on every aggregation call; each header states that alongside its measured cost.
- **Documentation links, not copies** (`lib.rs`, `README.md`): the README's example section and the `src/lib.rs` duplicate are both removed; the README links each proof system to its example and to its page on the website. CI compiles all three examples through `cargo clippy --all-features --all-targets`.
- **Committed protocol-message fixtures** (`recursive_snark_example_fixtures.rs`, `protocol/mod.rs`, `examples/assets/*.bin`): the recursive example cannot build its own input, because the rigid protocol message is a function of the signer set and its builder lives in `mithril-common`, which depends on `mithril-stm`. An ignored generator test proves and verifies the chain before writing three preimages at epochs 0, 1 and 2; a fast test rebuilds them and compares byte for byte, so a format change fails rather than silently invalidating the example.
- **Feature gating** (`Cargo.toml`): `required-features = ["future_snark"]` on the two SNARK examples, so the example set builds with and without the feature.
- **Packaging and docs.rs** (`Cargo.toml`): the two production verifying keys and the three fixtures added to `package.include`; docs.rs configured to build with `future_snark` and `rustls` rather than all features, which would enable `benchmark-internals` and its unpublished test assets.
- **Introduction rewritten** (`README.md`): the three proof systems described and linked to their examples, with the two SNARK ones marked experimental and their feature requirements stated.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation
  - [x] No new TODOs introduced

## Comments

The website links drop the `next` path segment, as requested in review. Those three pages currently exist only in the Next documentation source, so the links resolve once the documentation rotates — worth keeping in mind for the release order.

## Issue(s)
Closes https://github.com/IntersectMBO/mithril/issues/3385


mithril-stm/README.md
```diff
@@ -4,7 +4,11 @@
 
 - `mithril-stm` is a Rust implementation of the scheme described in the paper [Mithril: Stake-based Threshold Multisignatures](https://eprint.iacr.org/2021/916.pdf) by Pyrros Chaidos and Aggelos Kiayias.
 - The BLS12-381 signature library [blst](https://github.com/supranational/blst) is used as the backend for the implementation of STM.
-- This implementation supports the _trivial concatenation proof system_ (Section 4.3). Other proof systems such as _Bulletproofs_ or _Halo2_ are not supported in this version.
+- Three proof systems are available:
```
damrobi, line 7: Can you add links to the website sections of each proof system?

mithril-stm/README.md
damrobi: I think the `CoreVerifier` is outdate, maybe we can remove that part.

damrobi, line 15: We could update that list with what we added in stm like the schnorr signature types, the digest, the circuits, etc..

mithril-stm/examples/recursive_snark_aggregate_signature.rs
```diff
@@ -0,0 +1,169 @@
...
+//! Run it with:
+//!
+//! ```text
+//! cargo run --release -p mithril-stm --example recursive_snark_aggregate_signature \
+//!     --features future_snark,rustls
+//! ```
+//!
+//! Each step produces two proofs over the same circuit, under different transcripts. The Poseidon
+//! one seeds the rolling state, so the following step can verify it inside the circuit; the Blake2b
+//! one travels on the certificate for verifiers outside it. A step that stays within its epoch
```
damrobi: This phrasing is a bit weird.

mithril-common/src/protocol/recursive_snark_example_fixtures.rs
```diff
@@ -0,0 +1,240 @@
...
+const SIGNER_SEED: [u8; 32] = [0u8; 32];
+
+/// Seed of the example's genesis Schnorr key. Compromised for the same reason.
+const GENESIS_SEED: [u8; 32] = [1u8; 32];
+
+/// Stakes of the example's four signers, fixed so the aggregate verification key is reproducible.
+const SIGNER_STAKES: [Stake; 4] = [1_000, 2_000, 3_000, 4_000];
+
+const QUORUM_PARAMETER: u64 = 2;
+const SECURITY_PARAMETER: u64 = 100;
```
damrobi: Maybe we can use a more explicit name for this value, examples: `SECURITY_PARAMETER_FOR_LOTTERY`, `STM_SECURITY_PARAMETER`

damrobi (commented):
Looks good! I left a few comments on outdated parts of the README.

mithril-common/src/entities/protocol_message/recursive_snark_example_fixtures.rs
jpraynaud, line 1: It'd be better to have this in a test sub-module of `mithril-common/src/entities/protocol_message.rs`. A runbook is also missing to explain how to create the fixtures.

mithril-stm/examples/recursive_snark_aggregate_signature.rs
```diff
@@ -0,0 +1,170 @@
...
+
+type D = MithrilMembershipDigest;
+
+/// The protocol messages the chain advances through, one per epoch. They are committed rather than
+/// built here because assembling one needs the rigid protocol message format, which belongs to the
+/// node rather than to this library. They were generated together with the key material below, and
+/// the two must stay in step: each message announces the aggregate verification key that the next
+/// certificate is checked against.
+const GENESIS_PROTOCOL_MESSAGE_PREIMAGE: &[u8] = include_bytes!(concat!(
+    env!("CARGO_MANIFEST_DIR"),
```
jpraynaud: This should be needed as relative paths also work with `include_bytes!`. (Same for occurrences below).

mithril-stm/examples/recursive_snark_aggregate_signature.rs
```diff
@@ -0,0 +1,170 @@
...
+        )?;
+
+        rolling_state = Some(
+            ancillary_output
+                .prover_data()
+                .cloned()
+                .ok_or("a next-epoch step produces the rolling state the next step consumes")?,
+        );
+
+        println!("certificate aggregated and verified");
```
jpraynaud: This print could be enhanced or removed.

mithril-stm/src/lib.rs
```diff
@@ -1,4 +1,15 @@
 #![doc = include_str!("../README.md")]
+#![cfg_attr(
+    feature = "future_snark",
+    doc = concat!(
+        "\n## Recursive SNARK aggregate signature\n\n```no_run\n",
+        include_str!(concat!(
+            env!("CARGO_MANIFEST_DIR"),
```
jpraynaud: This should not be necessary:
```suggestion
 
```

mithril-stm/examples/recursive_snark_aggregate_signature.rs
```diff
@@ -0,0 +1,170 @@
+//! Aggregating and verifying certificates with the recursive SNARK proof system.
+//!
+//! Each certificate carries one recursive proof attesting to the whole chain behind it, so a
+//! verifier checks a single proof rather than every certificate since genesis. This example anchors
+//! a chain at genesis and advances it by two epochs, verifying each certificate as it goes.
```
jpraynaud: This paragraph is contradicted by the following one which states that 2 proofs are created for different transcript hashs.

mithril-stm/examples/non_recursive_snark_aggregate_signature.rs
```diff
@@ -0,0 +1,109 @@
+//! Aggregating and verifying a certificate with the non-recursive SNARK proof system.
+//!
+//! The aggregate signature is replaced by a single succinct proof that the quorum was met, so a
```
jpraynaud: The formulation is not very clear: the aggregate signature `consists in` is more clear than `replaced by`.

Also, the term `certificate` should be replaced with `aggregate signature`.

mithril-stm/examples/non_recursive_snark_aggregate_signature.rs
```diff
@@ -0,0 +1,109 @@
...
+//! cargo run --release -p mithril-stm --example non_recursive_snark_aggregate_signature \
+//!     --features future_snark,rustls
+//! ```
+//!
+//! Expect about three and a half seconds and roughly 1.3 GB of peak memory once the trusted setup is
+//! cached, measured on an Apple Mac Studio; the first run additionally downloads that setup. The
+//! circuit keys are generated on each run, because the example's parameters are sized so it can be
+//! run at all and its keys are therefore not the production ones the key cache recognises.
+//! Verification is cheap by comparison and needs no trusted setup at all: the verifier parameters
+//! are embedded in the crate.
```
jpraynaud: Can you give the description of your computer (number of cores, memory and hard drive size)?

mithril-stm/examples/non_recursive_snark_aggregate_signature.rs
```diff
@@ -0,0 +1,109 @@
...
+//!
+//! It is far cheaper than the recursive example, which proves the whole chain behind each
+//! certificate rather than a single certificate on its own.
```
jpraynaud: I'm not convinced that it should be compared to other aggregate signature types. It'd be maybe better to link to the upcoming pages on the website (you can safely remove the `next` in the URLs).

mithril-stm/examples/non_recursive_snark_aggregate_signature.rs
```diff
@@ -0,0 +1,109 @@
...
+    AggregateSignatureType, AncillaryGenesisData, AncillaryProofInput, Clerk, Initializer,
+    KeyRegistration, MithrilMembershipDigest, Parameters, Signer, SingleSignature, Stake,
+};
+
+type D = MithrilMembershipDigest;
+
+const SIGNER_STAKES: [Stake; 4] = [1_000, 2_000, 3_000, 4_000];
+
+fn main() -> Result<(), Box<dyn Error>> {
+    // XXX: not production parameters. They are small so the example is runnable, and `phi_f` of 1.0
```
jpraynaud: The `XXX` is not clear.

mithril-stm/examples/non_recursive_snark_aggregate_signature.rs
```diff
@@ -0,0 +1,109 @@
...
+    let message: [u8; 32] = Sha256::digest(b"the message this certificate attests to").into();
+
+    let signatures = signers
+        .iter()
+        .filter_map(|signer| signer.create_single_signature(&message).ok())
+        .collect::<Vec<SingleSignature>>();
+
+    // The ancillary input carries state between certificates for proof systems that need it. The
+    // non-recursive one does not, so an empty input is correct here; the recursive example shows
+    // what it is for.
```
jpraynaud: You should stick to explanations about the proof system in this example.

mithril-stm/src/lib.rs
```diff
@@ -1,4 +1,15 @@
 #![doc = include_str!("../README.md")]
+#![cfg_attr(
+    feature = "future_snark",
+    doc = concat!(
+        "\n## Non-recursive SNARK aggregate signature\n\n```no_run\n",
+        include_str!(concat!(
+            env!("CARGO_MANIFEST_DIR"),
+            "/examples/non_recursive_snark_aggregate_signature.rs"
+        )),
```
jpraynaud: I'm not sure that we want to inline all the examples, maybe having links to the examples is enough

mithril-stm/examples/concatenation_aggregate_signature.rs
```diff
@@ -0,0 +1,116 @@
+//! Aggregating and verifying a certificate with the concatenation proof system.
```
jpraynaud: Can you talk about aggregate signature instead of certificate? (here and in the other examples)?

mithril-stm/examples/recursive_snark_aggregate_signature.rs
```diff
@@ -0,0 +1,170 @@
+//! Aggregating and verifying certificates with the recursive SNARK proof system.
```
jpraynaud: Can you mention that this is an unstable feature?

mithril-stm/examples/non_recursive_snark_aggregate_signature.rs
```diff
@@ -0,0 +1,109 @@
+//! Aggregating and verifying a certificate with the non-recursive SNARK proof system.
```
jpraynaud: Can you mention that this is an unstable feature?

mithril-stm/examples/concatenation_aggregate_signature.rs
jpraynaud: Can you mention that this is a stable feature?

mithril-stm/examples/concatenation_aggregate_signature.rs
```diff
@@ -0,0 +1,116 @@
...
+//!
+//! This is the proof system Mithril uses today. The certificate carries the individual signatures
+//! that won the lottery, concatenated, together with the batched Merkle paths proving each signer
+//! was registered. Verification is cheap and needs no trusted setup, at the cost of a certificate
+//! that grows with the quorum.
```
jpraynaud: That sounds a bit approximate and should be reformulated.

mithril-stm/examples/concatenation_aggregate_signature.rs
```diff
@@ -0,0 +1,116 @@
...
+    AggregateSignatureType, AncillaryGenesisData, AncillaryProofInput, Clerk, Initializer,
+    KeyRegistration, MithrilMembershipDigest, Parameters, Signer, SingleSignature,
+};
+
+type D = MithrilMembershipDigest;
+
+const SIGNER_COUNT: usize = 32;
+
+fn main() -> Result<(), Box<dyn Error>> {
+    // XXX: seeded so the example produces the same certificate on every run, which keeps it usable
```
jpraynaud: Can you fix this comment?

mithril-stm/README.md
```diff
@@ -4,7 +4,11 @@
...
+  - the _trivial concatenation proof system_ (Section 4.3), currently used by the Mithril network. A certificate carries the winning signatures concatenated, so it grows with the quorum, and verification needs no trusted setup.
+  - a _non-recursive SNARK_ proof system, which replaces those signatures with a single succinct proof of the same statement.
+  - a _recursive SNARK_ proof system, in which each certificate proves the whole chain of certificates behind it, so a verifier checks one proof rather than every certificate since genesis.
+- The two SNARK proof systems are **experimental**. They are gated behind the `future_snark` feature, which also requires one of `rustls` or `native-tls` for the trusted setup download.
```
jpraynaud, line 11: Can you impact the modifications applied to the examples here?
Also the assertion `size grows with the quorum` is not true and should be refined.

jpraynaud (commented):
Overall looks good.

mithril-stm/examples/non_recursive_snark_aggregate_signature.rs
hjeljeli32: @jpraynaud 
The example headers and the README now link to `/doc/mithril/advanced/mithril-protocol/aggregation/…` without next, as you suggested. Those three pages currently exist only in the Next source, so the unversioned links resolve once the docs rotate.

mithril-common/src/entities/protocol_message/recursive_snark_example_fixtures.rs
```diff
@@ -48,15 +64,15 @@ fn fixture_directory() -> PathBuf {
 fn protocol_parameters() -> ProtocolParameters {
     ProtocolParameters {
         k: QUORUM_PARAMETER,
-        m: SECURITY_PARAMETER,
+        m: SECURITY_PARAMETER_NUMBER_OF_LOTTERIES,
         phi_f: LOTTERY_PARAMETER,
```
jpraynaud: This would make more sense:
```suggestion
        k: QUORUM_PARAMETER,
        m: TOTAL_LOTTERIES_PARAMETER,
        phi_f: LOTTERY_PROBABILITY_PARAMETER,
```

mithril-common/src/entities/protocol_message/recursive_snark_example_fixtures.rs
jpraynaud: I my previous comment, I was think about adding a `mod` in the `protocol_message.rs` tests section.

However, this code should really belong to the `mithril-stm` crate where there is already a protocol message implementation.

mithril-stm/README.md
```diff
@@ -5,22 +5,21 @@
...
 - The BLS12-381 signature library [blst](https://github.com/supranational/blst) is used as the backend for the implementation of STM.
 - Three proof systems are available:
-  - the _trivial concatenation proof system_ (Section 4.3), currently used by the Mithril network. A certificate carries the winning signatures concatenated, so it grows with the quorum, and verification needs no trusted setup.
-  - a _non-recursive SNARK_ proof system, which replaces those signatures with a single succinct proof of the same statement.
-  - a _recursive SNARK_ proof system, in which each certificate proves the whole chain of certificates behind it, so a verifier checks one proof rather than every certificate since genesis.
+  - the [_trivial concatenation proof system_](https://mithril.network/doc/mithril/advanced/mithril-protocol/aggregation/concatenation) (Section 4.3), currently used by the Mithril network. The aggregate signature carries one entry per contributing signer, together covering at least the `k` winning lottery indices the quorum requires, so its size follows the number of signers needed rather than `k` alone. Verification needs no trusted setup.
```
jpraynaud:
```suggestion
  - the [_concatenation proof system_](https://mithril.network/doc/mithril/advanced/mithril-protocol/aggregation/concatenation) (Section 4.3), currently used by the Mithril network. The aggregate signature carries one entry per contributing signer, together covering at least the `k` winning lottery indices the quorum requires, so its size follows the number of signers needed rather than `k` alone. Verification needs no trusted setup.
```

mithril-common/src/entities/protocol_message.rs
```diff
@@ -883,6 +883,69 @@ mod tests {
         );
     }
 
+    /// `mithril-stm` reproduces this crate's rigid preimage layout so that it can build a protocol
```
jpraynaud: This can probably be moved to `mithril-stm` after all as there is a protocol message implementation in it. Or even better, the protocol message can be constructed in the example.

damrobi (approved):
LGTM 👍

mithril-common/src/entities/protocol_message/recursive_snark_example_fixtures.rs
hjeljeli32: @jpraynaud 
This file was deleted a couple of commits later the example now builds its own protocol messages, so the fixture module, the committed preimages and the generator are all gone.

hjeljeli32: @jpraynaud 
This file was deleted a couple of commits later the example now builds its own protocol messages, so the fixture module, the committed preimages and the generator are all gone.

mithril-common/src/entities/protocol_message.rs
hjeljeli32: @jpraynaud 
The test compares mithril-common's canonical rigid_preimage() with mithril-stm's builder, so it has to live here mithril-stm can't see this crate, and inside mithril-stm it could only compare the builder against itself.

On the second point: the example does construct its protocol messages now; this test is what replaced the committed fixtures, and it's the only executable guard that the two implementations of the 190-byte layout stay identical.

PR #3532, opened 2026-09-10
feat(stm): cache the circuit keys and reuse the prover setups

### LINKED ISSUE #3424: Cache the circuit keys and reuse the SNARK prover setups
## Why

The circuit key cache never hit outside the production parameters, and even a hit is expensive: deserializing the recursive proving key rebuilds its cosets with FFTs, nearly as costly as deriving it (239 s against 293 s on 4 cores), so every aggregation paid the full setup.

## What

Make the circuit key cache hit for every configuration and keep the loaded prover setups in memory, so that a process pays the setup once instead of at every aggregation.

## How

- [x] Key the on-disk circuit key cache by configuration instead of comparing entries against the production verification key
- [x] Reuse the loaded prover setups across the aggregations of a process, as a policy of the prover factory
- [x] Key the in-memory setups by configuration so that a setup is never served to other parameters

jpraynaud:
## Content

This PR includes the **changes that stop a node from rebuilding its SNARK prover setup for every aggregation**:

- Key the circuit key cache by configuration: it compared entries against the embedded production verification key, so any other parameters were always stale and every aggregation re-derived both key pairs.
- Reuse the loaded setups across a process through the prover factory, as a policy on `NonDeterministicSnarkProverFactory::new`: enabled for the clerk, disabled for the circuit key digest computation.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Comments

- Measured at the end-to-end parameters on 4 cores: 
  - deriving the IVC setup 293 s
  - materializing it from the on-disk cache 239 s
  - the proof itself 165 s.
- Before this PR every signing round paid the derivation, now a process pays it once.
- In a real end-to-end run, the aggregation cycle producing one certificate went from 491 s to 93 s.
- The on-disk cache barely helps alone: deserializing the 1GB recursive proving key rebuilds the cosets with FFTs, nearly as costly as deriving them, so keeping the setup in memory is what removes the cost.

## Issue(s)
Closes #3424

### DIFF
--- a/mithril-stm/src/proof_system/snark_setup_cache.rs
+++ b/mithril-stm/src/proof_system/snark_setup_cache.rs
@@ -5,14 +5,14 @@
 //! loads each setup once for the lifetime of the process, at the cost of keeping the SRS and the
 //! proving keys resident.
 //!
-//! A process only ever needs one setup: the circuit is built from the protocol parameters, and
-//! changing them changes the embedded circuit verification keys, so it takes a new release of the
-//! node rather than a new epoch.
+//! Each slot holds the setup of one configuration at a time. The circuit is built from the protocol
+//! parameters, so a caller asking for other parameters is never served a setup derived for the ones
+//! it did not ask for: the setup is loaded again and the previous one released.
 
 use std::sync::{Arc, Mutex, PoisonError};
 
 use crate::{
-    Parameters, StmResult,
+    MERKLE_TREE_DEPTH_FOR_SNARK, Parameters, StmResult,
     proof_system::{halo2_ivc_snark::IvcProverSetup, halo2_snark::SnarkProverSetup},
 };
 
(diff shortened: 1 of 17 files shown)


mithril-stm/src/proof_system/snark_setup_cache.rs
```diff
@@ -0,0 +1,203 @@
...
+        Self {
+            setup: Mutex::new(None),
+        }
+    }
+
+    /// Returns the setup, loading it with `load` if this is the first caller.
+    ///
+    /// Callers wait for that one load instead of racing through the expensive path. A failed load
+    /// leaves the slot empty and is retried by the next caller.
+    fn get_or_load(&self, load: impl FnOnce() -> StmResult<T>) -> StmResult<Arc<T>> {
```
hjeljeli32: Related to what Copilot already raised: Two Clerks in the same process can use different parameters, but the second currently receives the first setup. For example, changing k makes the witness length disagree with the cached circuit.

hjeljeli32 (approved):
LGTM apart from the issue already raised by Copilot.

mithril-stm/src/proof_system/snark_setup_cache.rs
jpraynaud: This is a not a case that is very likely to happen today as this would mean supporting multiple signer registrations for the different sets of parameters. However, I made sure that in the worst case scenario no stale cache is retrieved.

damrobi (approved):
LGTM 👍

PR #3535, opened 2026-09-11
test(stm): add property tests for the IVC pure helpers

### LINKED ISSUE #3469: Add property-based tests for the IVC pure helpers
## Why

`proptest` is established in `halo2_snark` (eligibility, clerk, message encoding) but absent from `halo2_ivc_snark`, despite several pure helpers with clear invariants.

## What

Add proptest suites in the fast tier for the pure helpers and codecs. Case counts start at 100 for the byte-level and classification properties and stay low where a case decodes a verifying key; each target is measured as it lands, so the counts come from data rather than convention. Where a generated domain reaches an important case only by chance — an epoch pair that happens to be adjacent, a repeated name — that case is constructed rather than sampled. Per-helper scope and rationale are in the comment below.

## How

Paths are relative to `mithril-stm/src/`.

- [x] Accumulator and MSM byte codec (`circuits/halo2_ivc/io.rs`): round-trips preserve every component over arbitrary counts, fixed-base maps and key strings, and consume the whole encoding
- [x] Preimage accessor range isolation (`circuits/halo2_ivc/types.rs`): each accessor decodes its own range, and is unchanged when every byte outside that range is altered
- [x] Rigid preimage assembly (`circuits/halo2_ivc/protocol_message.rs`, `circuits/halo2_ivc/types.rs`): a preimage built from arbitrary inputs decodes back to them, and its whole 190-byte layout matches expectations computed from those inputs rather than from the helpers under test
- [x] `build_next_state` (`proof_system/halo2_ivc_snark/prover_input_helpers.rs`): all seven next-state fields taken from their declared source, for both transition types on the same generated inputs, with step counters spanning the 32-bit boundary
- [x] Epoch transition classification (`proof_system/halo2_ivc_snark/rolling_state.rs`): same-epoch, next-epoch and both gap directions constructed for each generated epoch, with the typed error payloads, plus deterministic `(MAX - 1, MAX)` and `(MAX, MAX)` cases for the overflow precedence
- [x] Protocol parameter promotion guard (`proof_system/halo2_ivc_snark/rolling_state.rs`): a truth table over step counter, transition type, and whether the two parameter fields hold equal field values
- [x] Aggregate verification key codec (`proof_system/halo2_snark/aggregate_key.rs`): legacy to CBOR preservation, a forced CBOR-version-byte root, and one property pairing the rigid slot layout with rejection of any decoded root that is not 32 bytes
- [x] Input message matching (`proof_system/halo2_ivc_snark/proof.rs`): raw, lowercase, uppercase and mixed hex encodings of the same message are accepted, a message reducing to a different field element is rejected in every one of those forms, and malformed encodings are covered by a small example table
- [x] `State::as_public_input` (`circuits/halo2_ivc/state.rs`): every field value at its row over full-width integers, with the four `{0, MAX}` boundary pairs
- [x] `trivial_accumulator` (`circuits/halo2_ivc/accumulator.rs`): each distinct name once with a zero scalar, duplicates collapsed, empty left side, over a name pool small enough that duplicates actually occur
- [x] Fixed-base availability guards (`circuits/halo2_ivc/accumulator.rs`): a name missing from either accumulator side is reported, and a correct label mapped to a different point is rejected
- [x] `IvcRollingState` serde round-trip (`proof_system/halo2_ivc_snark/rolling_state.rs`, through `AncillaryProverData`): state values, proof bytes, accumulator components and signature compared field by field over non-genesis values
- [x] `IvcVerifierData` (`proof_system/halo2_ivc_snark/verifier_setup.rs`): reject input without the CBOR version prefix, and preserve a nonzero genesis message
- [x] `MessageHash::try_from(&ProtocolMessagePreimage)` (`circuits/halo2_ivc/types.rs`): the result is the field reduction of SHA-256 over all 190 input bytes, pinned by independently computed vectors
- [x] Assess the gain and report it in the issue, with the measured per-case cost and the chosen case counts

hjeljeli32:
## Content

`proptest` is established in `halo2_snark` but absent from `halo2_ivc_snark`, whose helpers and two hand-written byte decoders were covered only by single-fixture examples: every rigid-preimage test ran on stake 1 and epoch 42, so seven of the eight bytes in each integer slot were zero. This adds **21 properties across fourteen helpers and codecs** and retires twelve examples a property subsumes.

They found four defects, all in the two hand-written decoders, each fixed here in its own commit directly after the test that found it. A fifth came out of review: certificate accumulators were decoded with an unchecked point reader that panics rather than erroring on a payload that is not a curve point. Nothing is left ignored.

At proptest's default 256 cases a property costs **6 to 8 ms**; the codec round trips run 10 to 15 ms and the prefix-truncation property 38 ms. The fast tier is 636 tests in 7.4 s. Per-helper rationale and measurements are on the issue.

### Changes

- **Decoder fixes** (`io.rs`, `aggregate_key.rs`): mismatched point and scalar counts, an unvalidated fixed-base key length, a truncated point payload, and a legacy root re-detected as CBOR.
- **Checked point reader at the certificate boundary** (`rolling_state.rs`): `midnight_accumulator_serde` reads with `RawBytes` rather than `RawBytesUnchecked`, covering both accumulator carriers. Same wire bytes, and the identity still decodes, but a full-length payload that is not a curve point is now an error rather than a panic, and non-subgroup points are rejected.
- **Byte-level codecs** (`io.rs`, `types.rs`, `aggregate_key.rs`): round trips preserve every component and consume the whole encoding; each preimage accessor decodes only its own range; legacy aggregate keys survive the CBOR round trip on both paths.
- **Rigid preimage and message digest** (`types.rs`, `tests/encoding/positive.rs`): the 190-byte layout matches expectations computed from the inputs, and `MessageHash::try_from` is pinned to digests computed outside the crate.
- **State transition helpers** (`prover_input_helpers.rs`, `rolling_state.rs`): next-state field sources under both transitions, epoch classification with typed errors, and the parameter promotion truth table. Branch boundaries are constructed, not sampled.
- **Certificate carriers** (`rolling_state.rs`, `verifier_setup.rs`): rolling state and verifier data compared component by component after a round trip, not against reserialized bytes, which agree even when a value is lost.
- **Not covered**: the generic codec still panics on a malformed point under `RawBytesUnchecked`, which its trusted callers keep using — only the certificate boundary is protected, by the format it passes. The allocation fix has no test observing allocation.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation
  - [x] No new TODOs introduced

## Issue(s)
Closes https://github.com/IntersectMBO/mithril/issues/3469


mithril-stm/src/circuits/halo2_ivc/types.rs
```diff
@@ -237,6 +239,84 @@ mod protocol_message_preimage_tests {
...
+        for (offset, byte) in mutated.iter_mut().enumerate() {
+            if !range.contains(&offset) {
+                *byte ^= mask;
+            }
+        }
+        mutated
+    }
+
+    proptest! {
+        #![proptest_config(ProptestConfig::with_cases(100))]
```
jpraynaud: We can probbaly go with the default value (`256`) as the tests are really fast (here and everywhere else):
```suggestion
```

mithril-stm/src/proof_system/halo2_ivc_snark/prover_input_helpers.rs
```diff
@@ -233,116 +210,163 @@ pub(crate) mod tests {
...
-                MessageHash::ZERO,
-                MerkleTreeCommitment::ZERO,
-                &preimage,
-            )
-            .unwrap();
-            assert_eq!(next_state.protocol_parameters, pp_current);
+        // --- Field-source property ---
+        // The helper's whole job is choosing, per output field, between the rolling state, the
+        // certificate arguments and the preimage. Each source is generated independently so that a
+        // wrong choice shows up, rather than being hidden by two sources holding the same value.
```
jpraynaud: Are these comments starting with `// ---` necessary?

mithril-stm/src/proof_system/halo2_ivc_snark/prover_input_helpers.rs
```diff
@@ -140,12 +142,17 @@ pub(crate) mod tests {
...
-            .sign_standard(&[BaseFieldElement::from(1u64)], &mut rng)
-            .expect("standard schnorr signing should succeed for a synthetic message")
+        static SIGNATURE: OnceLock<StandardSchnorrSignature> = OnceLock::new();
+        *SIGNATURE.get_or_init(|| {
+            let mut rng = ChaCha20Rng::from_seed([0u8; 32]);
+            let signing_key = SchnorrSigningKey::generate(&mut rng);
+            signing_key
+                .sign_standard(&[BaseFieldElement::from(1u64)], &mut rng)
+                .expect("standard schnorr signing should succeed for a synthetic message")
+        })
```
jpraynaud, line 156: Is this `OnceLock` relevant in the tests (as they are run in different binairies with `nextest`?

mithril-stm/src/proof_system/halo2_ivc_snark/rolling_state.rs
```diff
@@ -583,6 +585,162 @@ mod tests {
...
         }
 
+        // --- Epoch relation property ---
+        // Overflow is checked first; otherwise the relation between the two epochs selects the
+        // branch, and acceptance additionally needs that branch's state guards. Each relation is
+        // constructed from the generated epoch rather than sampled, because a uniform pair of
+        // `u64` values rarely lands on any of the relations that matter.
+
+        /// The rolling state and the aggregate key do not depend on the incoming epoch, so a case
+        /// builds them once and varies only the preimage.
```
jpraynaud: This should probably a simple comment (not a doc comment)?

jpraynaud (approved):
LGTM 👍

mithril-stm/src/circuits/halo2_ivc/tests/common/public_input_layout.rs
```diff
@@ -233,6 +235,102 @@ mod tests {
         );
     }
 
+    // The existing sentinels pin the order at 11 to 77, where the two integer fields never exceed
+    // a byte. A conversion narrowing them through `u32` maps every sentinel to itself, so only a
+    // full-width value distinguishes it.
+
+    fn reduced_field_element(bytes: &[u8; 32]) -> NativeField {
```
damrobi, line 241:
```suggestion
    // The existing sentinels pin the order at 11 to 77, where the two integer fields never exceed
    // a byte. A conversion narrowing them through `u32` maps every sentinel to itself, so only a
    // full-width value distinguishes it.
    fn reduced_field_element(bytes: &[u8; 32]) -> NativeField {
```
Is this extra line intentional?

mithril-stm/src/circuits/halo2_ivc/tests/encoding/positive.rs
```diff
@@ -230,32 +227,13 @@ fn vk_serialization_round_trip() {
...
-}
+// The generated layout property below covers every offset. This example is kept because its
+// 64-character hex value is the shape production sends, guaranteed on every run rather than
+// sampled.
 
 #[test]
 fn rigid_preimage_dynamic_hash_is_at_offset_6() {
-    let (message, _) = build_test_message();
+    let message = build_test_message();
     let preimage = message
```
damrobi, line 236:
```suggestion
// The generated layout property below covers every offset. This example is kept because its
// 64-character hex value is the shape production sends, guaranteed on every run rather than
// sampled.
#[test]
fn rigid_preimage_dynamic_hash_is_at_offset_6() {
    let message = build_test_message();
let preimage = message
```
Same with this extra line

damrobi (approved):
LGTM 👍 
I saw a few extra lines between comments and code and was wondering if they were on purpose (I did not mark them all)

PR #3536, opened 2026-09-11
refactor(aggregator): accelerate build times

### LINKED ISSUE #3538: Reduce the aggregator compile time
## Why

The aggregator compiles much slower than the other nodes: its warp router is a single nested generic type of about 200 combinators and closures, and every crate that instantiates it (the library, the binary, the unit-test binary) pays for it in rustc's monomorphization collector, in dev and in release.

## What

Box the filters at the 14 route module boundaries so that the router composes type-erased filters, and check that the change does not regress serving performance.

## How

- [x] Return a `BoxedFilter` from the `routes()` function of each route module
- [x] Verify the compile time of the binary and unit-test targets in dev and release
- [x] Make sure there are no performance regression

jpraynaud:
This PR includes the **change that stops the aggregator's warp router from being one giant generic type**:

- Return a `BoxedFilter` from the `routes()` function of each of the 14 route modules, so the router composes 14 type-erased filters instead of a single nested tree of about 200 combinators and closures whose symbol names reached 137,000 characters.
- Leave the router and the 30 individual routes untouched: matching order, rejections, status codes and reply types are unchanged, and the 150 route tests pass as they are.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Comments

The cost was in rustc's monomorphization collector, not in the linker: every crate that instantiates the router walked the whole type again, while `rust-lld` relinks the 667 MB debug binary in 2.5 s.

Binary crate after a library edit, dev profile, `-Ztime-passes`:

| Pass | Before | After |
|---|---|---|
| total | 58.7 s | 7.7 s |
| monomorphization collector walk | 48.7 s | 1.2 s |
| link | 1.6 s | 1.3 s |

Measured against `main` on 6 cores, after a library edit:

| Metric | Before | After |
|---|---|---|
| binary step, dev | 58.5 s | 7.4 s |
| full crate build, dev | 1 min 08 s | 29 s |
| debug binary size | 667 MB | 476 MB |

Measured on the earlier baseline branch:

| Metric | Before | After |
|---|---|---|
| binary step, release | 70 s | 17 s |
| unit-test binary compile | 122 s | 77 s |
| unit-test binary size | 880 MB | 562 MB |
| longest symbol name | 137,231 chars | 11,965 chars |
| `.debug_str` section | 361 MB | 180 MB |
| library compile, dev | 10 s | 20 s |
| library compile, release | 24 s | 36 s |

- The library's own compile gets slower because each module now instantiates its boxed wrapper once; the net effect is still a large win for every workflow.
- Serving performance was checked with an in-process A/B/A benchmark (release profile, 100,000 requests per path): the boxed router answers every route 6 to 8 µs faster, since warp's `Or` no longer deep-clones the remaining modules on each request.

| Path | Before, median ns | After, median ns |
|---|---|---|
| `/aggregator/certificates` | 24,176 | 16,180 |
| `/aggregator/status` | 12,883 | 7,083 |
| unknown path (404) | 12,129 | 6,453 |
- Integration test binaries are unchanged, as they never instantiate the router.

## Issue(s)
Closes #3538


Alenar (approved):
LGTM

PR #3539, opened 2026-09-14
feat(common): add proof bound possession for Schnorr signing key

### LINKED ISSUE #3537: Add a SNARK verification key Proof of Bound Possession
## Why

We want to make sure that no registered SNARK verification key can be used by an adversary to pollute the membership registration. A way to avoid this problem is to implement a Proof of Bound Possession for the SNARK verification key.

## What

Implement a Proof of Bound Possession for the SNARK verification keys.

## How

- [x] Create a document explaining the Proof of Bound Possession mechanism
- [x] Add the SNARK Proof of Bound Possession to the SNARK verification keys
- [ ] Adapt the SNARK key registration to not return errors on collision of verification key (with a deterministic resolution in case of collision)
- [ ] Make backward compatibility for codecs of the signer registration messages

damrobi:
## Content

This PR includes the addition of a proof of bound possession for the Schnorr signing key. This proof is added in mithril-common under the `future_snark` feature. It is computed by the Initializers and verified during the registration of the signers. This PR also includes a deduplication mechanism for the Schnorr verification key at the STM level.

## Summary

Adds **Proof of Bound Possession (PoBP)** for the Schnorr/SNARK signer verification key, entirely behind `future_snark`. 

## Changes

- **Crypto core** (`mithril-stm`): Schnorr signing/verification keys gain `create_pobp`/`verify_pobp`; `Initializer`/`StmInitializerWrapper` create and store the proof. `KeyRegistration::register` gains a deduplication of entries based on the Schnorr verification key.
- **Challenge encoding** (`mithril-common`): new `ProofOfBoundPossessionPrefix` (stake || epoch || pool_id) feeding the Schnorr challenge; pool_id uses the raw fixed-length Blake2b-224 hash rather than the bech32 string, removing the need for a length-prefixed encoding.
- **Verification wiring** (`mithril-common`, `mithril-aggregator`): `KeyRegWrapper::register` now verifies the PoBP alongside the existing KES check; `SignerRegistrationVerifier`/leader/follower services thread the registration `epoch` through to that verification.
- **Entities & messages**: `Signer`/`SignerWithStake` gain `proof_of_bound_possession_for_snark`; mirrored into `RegisterSignerMessage`, message adapters, and DB persistence (new nullable column via migration) so it survives aggregator restarts and follower sync.
- **Epoch threading**: the registration epoch is now plumbed end-to-end — signer runtime → aggregator services → `SignerBuilder`/fixture builders — since PoBP verification must use the exact epoch the proof was bound to.
- **Test infrastructure**: `MithrilFixtureBuilder`/`fake_data` gain epoch-aware builders (`build_at_epoch`, `signers_with_stakes_at_epoch`) to replace the implicit "always epoch 0" fixtures that broke once epoch-bound verification was enforced; extensive unit and integration test fixes across `mithril-common`/`mithril-aggregator` (including the aggregator's follower/leader integration suite) to build fixtures at the epoch they're actually registered/re-registered at, including cases where stake is updated mid-test.

## Notes

- All changes are `#[cfg(feature = "future_snark")]`-gated; default builds are unaffected.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [ ] Reviewer requested

## Issue(s)

Closes #3537


mithril-common/src/crypto_helper/cardano/key_certification.rs
```diff
@@ -203,6 +269,15 @@ impl StmInitializerWrapper {
         self.kes_signature_for_snark.map(|k| k.into())
     }
 
+    /// Extract the Proof of Bound Possession of the Schnorr signing key
+    /// for the SNARK proof system.
+    #[cfg(feature = "future_snark")]
+    pub fn proof_of_bound_possession_for_snark(
+        &self,
+    ) -> Option<ProtocolSignerProofOfBoundPossessionForSnark> {
+        self.proof_of_bound_possession_for_snark.map(|pobp| pobp.into())
```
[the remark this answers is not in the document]
damrobi: Fixed it by removing the proof of bound possession in the strip_snark_keys function

mithril-test-lab/mithril-end-to-end/src/stress_test/payload_builder.rs
```diff
@@ -58,6 +58,10 @@ pub fn generate_register_signer_message(
             verification_key_signature_for_snark: signer
                 .verification_key_signature_for_snark
                 .map(|s| s.to_json_hex().unwrap()),
+            #[cfg(feature = "future_snark")]
```
hjeljeli32, line 63: Could we generate the proofs for each registration epoch here? generate_signer_data() currently creates them for epoch 0, but the load-test bootstrap registers these signers for epochs 3 and 4.

hjeljeli32 (approved):
LGTM 👍 Just left one comment.

mithril-common/src/crypto_helper/cardano/key_certification.rs
```diff
@@ -113,6 +124,11 @@ pub struct StmInitializerWrapper {
     #[cfg(feature = "future_snark")]
     #[serde(skip_serializing_if = "Option::is_none", default)]
     kes_signature_for_snark: Option<Sum6KesSig>,
+
+    /// The Proof of Bound Possession of the Schnorr signing key for the SNARK proof system
+    #[cfg(feature = "future_snark")]
+    #[serde(skip_serializing_if = "Option::is_none", default)]
+    proof_of_bound_possession_for_snark: Option<StandardSchnorrSignature>,
```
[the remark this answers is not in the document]
damrobi: Those functions will be removed and we will update the crate minor version to reflect the breaking change.

mithril-stm/src/signature_scheme/schnorr_signature/signing_key.rs
```diff
@@ -160,6 +164,32 @@ impl SchnorrSigningKey {
...
 
+    /// Implementation of the creation of the Proof of Bound Possession (PoBP) of the Schnorr signing key.
+    ///
+    /// This function receives a prefix and uses a standard Schnorr signature to
+    /// create a PoBP of the signing key by signing:
+    /// Sha256("SCHNORR_POBP_DST" || prefix || verification_key_bytes)
+    /// The prefix is used to bound the signing key to a given stake, epoch and pool_id.
+    pub fn create_proof_of_bound_possession<R: RngCore + CryptoRng>(
+        &self,
+        prefix: &[u8],
```
jpraynaud: **question(blocking)**: Shouldn't the prefix length be fixed size?

mithril-stm/src/signature_scheme/schnorr_signature/verification_key.rs
```diff
@@ -39,6 +46,30 @@ impl SchnorrVerificationKey {
...
+        prefix: &[u8],
+        signature: &StandardSchnorrSignature,
+    ) -> StmResult<()> {
+        let digest: [u8; 32] = Sha256::digest(
+            [
+                DOMAIN_SEPARATION_TAG_SCHNORR_PROOF_OF_BOUND_POSSESSION,
+                prefix,
+                &self.to_bytes(),
+            ]
+            .concat(),
```
jpraynaud: **suggestion(non-blocking)**: the digest could be computed in the same function for both creation and verification.

mithril-common/src/crypto_helper/cardano/opcert.rs
```diff
@@ -242,18 +242,22 @@ impl OpCert {
...
-        pool_id.copy_from_slice(hasher.finalize().as_bytes());
-        encode_bech32("pool", &pool_id).map_err(|_| OpCertError::PoolAddressEncoding)
+        encode_bech32("pool", &self.compute_protocol_party_id_as_bytes())
+            .map_err(|_| OpCertError::PoolAddressEncoding)
     }
 
     /// Compute protocol party id as hash
     pub fn compute_protocol_party_id_as_hash(&self) -> String {
+        hex::encode(self.compute_protocol_party_id_as_bytes())
+    }
```
jpraynaud, line 252: **issue(blocking)**: we should have some tests to make sure that we didn't introduce a breaking change here. I'd recommend to add golden tests first with the previous version of the code and then make sure they are still green with the modifications.

mithril-common/src/crypto_helper/cardano/proof_of_bound_possession.rs
```diff
@@ -0,0 +1,92 @@
...
+};
+
+/// Structure representing the information needed to create a prefix
+/// for the Proof of Bound Possession (PoBP).
+/// It is used to compute the hash value signed for the PoBP :
+/// H(DST || prefix || vk)
+pub(crate) struct ProofOfBoundPossessionPrefix {
+    stake: Stake,
+    epoch: Epoch,
+    pool_id: ProtocolPartyIdHash,
```
jpraynaud: **issue(non-blocking)**: the name `ProtocolPartyIdHash` is a bit misleading here, as there are 2 versions of the pool id, one of which is called a hash version. I'd go for `ProtocolPartyIdBytes` instead to avoid confusions.

mithril-common/src/crypto_helper/types/alias.rs
```diff
@@ -17,6 +17,9 @@ pub type ProtocolMembershipDigest = MithrilMembershipDigest;
 /// The id of a mithril party.
 pub type ProtocolPartyId = String;
 
+/// The raw hash of a mithril party id.
+pub type ProtocolPartyIdHash = [u8; 28];
```
jpraynaud: **nitpick**:
```suggestion
/// The id of a Mithril party.
pub type ProtocolPartyId = String;

/// The raw hash of a Mithril party id.
pub type ProtocolPartyIdHash = [u8; 28];
```

mithril-common/src/crypto_helper/cardano/key_certification.rs
```diff
@@ -416,6 +376,26 @@ impl KeyRegWrapper {
...
+    /// key attached to the signing key.
+    #[cfg(feature = "future_snark")]
+    fn verify_proof_of_bound_possession_for_snark(
+        &self,
+        stake: Stake,
+        epoch: Epoch,
+        pool_id: ProtocolPartyIdHash,
+        schnorr_verification_key: VerificationKeyForSnark,
+        proof_of_bound_possession_for_snark: Option<StandardSchnorrSignature>,
+    ) -> StdResult<()> {
```
jpraynaud, line 429: **todo**: the function should take a `StandardSchnorrSignature` instead of an `Option<StandardSchnorrSignature>`, and the caller should handle the error if there is none.

mithril-common/src/crypto_helper/cardano/key_certification.rs
jpraynaud: **nitpick**:  the comment is not accurate anymore (missing the epoch).

mithril-common/src/crypto_helper/cardano/key_certification.rs
```diff
@@ -425,6 +405,7 @@ impl KeyRegWrapper {
     pub fn register(
         &mut self,
         parameters: SignerRegistrationParameters,
+        #[cfg(feature = "future_snark")] epoch: Epoch,
```
jpraynaud: **todo(blocking)**: the epoch should be treated symmetrically to the stake distribution (not gated and passed in the `init` function to be stored as a field of the `KeyRegWrapper`).

mithril-aggregator/src/services/signer_registration/api.rs
```diff
@@ -87,6 +87,8 @@ pub trait SignerRegistrationVerifier: Send + Sync {
         &self,
         signer: &Signer,
         stake_distribution: &StakeDistribution,
+        // no future_snark here to not mess with the mock derivation
```
jpraynaud: **nitpick**: this comment can be removed.

mithril-aggregator/src/services/epoch_service.rs
```diff
@@ -422,6 +422,8 @@ impl EpochService for MithrilEpochService {
                 .network_configuration
                 .configuration_for_aggregation
                 .protocol_parameters,
+            #[cfg(feature = "future_snark")]
+            data.epoch.offset_to_signer_retrieval_epoch_saturating(),
```
jpraynaud, line 426: **question(blocking)**: why did you choose to `offset_to_signer_retrieval_epoch_saturating`?

mithril-aggregator/src/services/epoch_service.rs
```diff
@@ -432,6 +434,8 @@ impl EpochService for MithrilEpochService {
                 .network_configuration
                 .configuration_for_next_aggregation
                 .protocol_parameters,
+            #[cfg(feature = "future_snark")]
+            data.epoch.offset_to_next_signer_retrieval_epoch(),
```
jpraynaud, line 438: **question(blocking)**: why did you choose to `offset_to_next_signer_retrieval_epoch`?

mithril-aggregator/tests/create_certificate_follower.rs
```diff
@@ -53,7 +53,7 @@ impl EpochFixturesMapBuilder {
                                 min_stake: 10,
                             },
                         )
-                        .build(),
+                        .build_at_epoch(Epoch(epoch as u64 + 1)),
```
jpraynaud: **todo(blocking)**: it'd be better to use the offset functions of `Epoch` instead of adding ` 1` which looks like magic number.

mithril-stm/src/protocol/key_registration/register.rs
```diff
@@ -43,21 +43,39 @@ impl KeyRegistration {
...
-        if is_already_registered {
-            return Err(RegisterError::EntryAlreadyRegistered.into());
+        if let Some(vk_snark) = entry.get_verification_key_for_snark()
+            && self.registered_keys_for_snark.contains(&vk_snark)
+        {
+            let existing = *self
+                .registration_entries
+                .iter()
+                .find(|e| e.get_verification_key_for_snark() == Some(vk_snark))
+                .expect("registered_keys_for_snark and registration_entries are kept in sync");
```
jpraynaud: **todo(blocking)**: we don't want any panic in this code, this should be replaced by an error.

mithril-stm/src/protocol/key_registration/register.rs
```diff
@@ -43,21 +43,39 @@ impl KeyRegistration {
...
+
+            // first compares the stake then the BLS key value as bytes
+            if *entry < existing {
+                self.registered_keys_for_concatenation
+                    .remove(&existing.get_verification_key_for_concatenation());
+                self.registration_entries.remove(&existing);
+            } else {
+                return Ok(());
+            }
```
jpraynaud: **issue(blocking)**: if there is a collision with an existing Schnorr verification key, we keep the one with lower stake, but the BLS verification key is also removed. This could create a split in the effective computation of the AVK by the different versions of the signers and aggregators, and thus stop the certification. Only the Schnorr part should be impacted and not the BLS part with an update of the existing entry instead of a replace. Can you add a test that verifies the expected behavior first, make sure that it breaks and then make it green?

mithril-stm/src/signature_scheme/schnorr_signature/signing_key.rs
damrobi: Done

mithril-stm/src/signature_scheme/schnorr_signature/verification_key.rs
damrobi: Moved to an helper function in the crate mod.rs

mithril-common/src/crypto_helper/cardano/proof_of_bound_possession.rs
damrobi: Renamed

mithril-common/src/crypto_helper/types/alias.rs
damrobi: Fixed

mithril-common/src/crypto_helper/cardano/key_certification.rs
damrobi: Done, should I do the same for the KES signature verification function?

mithril-common/src/crypto_helper/cardano/key_certification.rs
damrobi: Added a mention of the epoch in the comment

mithril-common/src/crypto_helper/cardano/key_certification.rs
damrobi: Done

mithril-aggregator/tests/create_certificate_follower.rs
damrobi: I used the `.next()` function of Epoch to represent the next epoch

mithril-aggregator/src/services/epoch_service.rs
damrobi: I chose those values because they match what is done in the `inform_epoch()` function to get the `signer_retrieval_epoch` and `next_signer_retrieval_epoch` that are used to get `current_signers_with_stake` and `next_signers_with_stake`. The same offset function is used in `update_next_signers_with_stake` to get `next_signers_with_stake` before calling `precompute_epoch_data`. It is possible that this is wrong but it looks consistent

mithril-aggregator/src/services/epoch_service.rs
damrobi: See the answer above

mithril-stm/src/protocol/key_registration/register.rs
damrobi: I replaced the panic with an error

mithril-stm/src/protocol/key_registration/register.rs
damrobi: I update the registration to remove the snark key of entry when it is a duplicate with higher stake

mithril-common/src/crypto_helper/cardano/opcert.rs
damrobi: I think that the test already present is acting as a golden test but I will add more specific golden tests for those two functions

damrobi: I added the new golden tests

mithril-aggregator/src/services/signer_registration/api.rs
damrobi: removed

mithril-common/src/protocol/signer_builder.rs
```diff
@@ -51,19 +56,27 @@ impl SignerBuilder {
...
+                        proof_of_bound_possession_for_snark: signer
+                            .proof_of_bound_possession_for_snark,
+                    },
                     #[cfg(feature = "future_snark")]
-                    verification_key_for_snark: signer.verification_key_for_snark,
-                    #[cfg(feature = "future_snark")]
-                    verification_key_signature_for_snark: signer
-                        .verification_key_signature_for_snark,
-                })
+                    epoch,
```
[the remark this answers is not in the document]
damrobi: I'm not sure how to handle that

mithril-test-lab/mithril-end-to-end/src/stress_test/payload_builder.rs
damrobi: I updated the test case to generate the Proof of Bound Possession at every epoch so that it is always valid

mithril-common/src/protocol/signer_builder.rs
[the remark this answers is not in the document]
jpraynaud: This is not as problem as the `future_snark` feature is not activated yet on any network (except dev networks that will need to be reset), which means that no SNARK verification key has been registered yet.

mithril-common/src/test/builder/fixture_builder.rs
```diff
@@ -91,18 +95,41 @@ impl MithrilFixtureBuilder {
...
+            self.epoch,
         );
 
         MithrilFixture::new(
             self.protocol_parameters,
             signers,
             protocol_stake_distribution,
+            #[cfg(feature = "future_snark")]
+            self.epoch,
         )
```
Alenar, line 133: suggestion(blocking): I think you should reverse the method call order as it just simpler, and remove `with_epoch` and the associated `epoch` field since it will be mandatory in the future and is only used when building and prevent easy re-using (see next paragraph).

Another important change is that the `build`/`build_at_epoch` method should not consume self anymore, meaning you will be able to call it multiple times to produce signers for different epoch using the same options.

```suggestion
    /// Transform the specified parameters to a [MithrilFixture], creating each signer's Proof of
    /// Bound Possession (when applicable) for the given epoch.
    pub fn build_at_epoch(&self, epoch: Epoch) -> MithrilFixture {
        let protocol_stake_distribution = self.generate_stake_distribution();
        let signers = crypto_helper::setup_signers_from_stake_distribution(
            &protocol_stake_distribution,
            &self.protocol_parameters.clone().into(),
            #[cfg(feature = "future_snark")]
            epoch,
        );

        MithrilFixture::new(
            self.protocol_parameters.clone(),
            signers,
            protocol_stake_distribution,
            #[cfg(feature = "future_snark")]
            epoch,
        )
    }

/// Transform the specified parameters to a [MithrilFixture], creating each signer's Proof of
    /// Bound Possession (when applicable) for the Epoch 1.
   pub fn build(&self) -> MithrilFixture {
     Self::build_at_epoch(Epoch(1))
   }
```

mithril-aggregator/tests/test_extensions/runtime_tester.rs
```diff
@@ -278,6 +278,29 @@ impl RuntimeTester {
...
+            let current_retrieval_epoch =
+                time_point.epoch.offset_to_signer_retrieval_epoch().unwrap();
+            let current_fixture = MithrilFixtureBuilder::default()
+                .with_signers(signers_count)
+                .with_protocol_parameters(protocol_parameters.clone())
+                .build_at_epoch(current_retrieval_epoch);
+            let next_fixture = MithrilFixtureBuilder::default()
+                .with_signers(signers_count)
+                .with_protocol_parameters(protocol_parameters)
+                .build_at_epoch(time_point.epoch);
```
Alenar, line 294: issue(blocking): you should not rebuild a fixture builder like this, since you may be missing parameters that were used when the fixture was built (e.g. the `StakeDistributionGenerationMethod`). This caller to use any other builder parameters than the three used here.

suggestion: pass the `MithrilFixtureBuilder` by reference and refactor it so the build methods does not consume self (see comment in `mithril-common/src/test/builder/fixture_builder.rs`).

important: this comment apply for all other cases where a builder is rebuilt from a fixture like this in the code base.

mithril-common/src/test/builder/fixture_builder.rs
damrobi: I applied this change

mithril-aggregator/tests/test_extensions/runtime_tester.rs
damrobi: I updated the tests that were creating several similar builders to only create one and call the build function several times

mithril-stm/src/lib.rs
```diff
@@ -102,11 +102,13 @@ pub type StmError = anyhow::Error;
...
 
 #[cfg(feature = "future_snark")]
-// TODO: remove this allow dead_code directive when function is called or future_snark is activated
-#[allow(dead_code)]
 /// Target value type used in the lottery for snark proof system
 pub type LotteryTargetValue = crate::signature_scheme::BaseFieldElement;
 
+#[cfg(feature = "future_snark")]
+/// Prefix of a proof of bound possession for signing key
+pub type ProofOfBoundPossessionPrefix = [u8; 44];
```
jpraynaud: **question(blocking)**: maybe the length of this type should always be 32 bytes (computed as the hash of the 44 bytes)? This would allow to have a braod variety of PoBP prefix.

mithril-common/src/crypto_helper/cardano/key_certification.rs
```diff
@@ -373,16 +373,23 @@ pub struct KeyRegWrapper {
...
+    current_epoch: Epoch,
 }
 
 impl KeyRegWrapper {
     /// New Initialisation function. We temporarily keep the other init function,
     /// but we should eventually transition to only use this one.
-    pub fn init(stake_dist: &ProtocolStakeDistribution) -> Self {
+    pub fn init(
+        stake_dist: &ProtocolStakeDistribution,
+        #[cfg(feature = "future_snark")] epoch: Epoch,
```
jpraynaud: **suggestion**: the `init` function could be simpler as the epoch is available with/without `future_snark`
```suggestion
        epoch: Epoch,
```

mithril-common/src/crypto_helper/cardano/key_certification.rs
```diff
@@ -418,83 +425,85 @@ impl KeyRegWrapper {
...
     ///
     /// When the `future_snark` feature is enabled, an optional Schnorr verification key
-    /// and its KES signature can be provided for SNARK proof system authentication.
+    /// and its KES signature as well as the current epoch can be provided for
+    /// SNARK proof system authentication.
     pub fn register(
         &mut self,
         parameters: SignerRegistrationParameters,
-        #[cfg(feature = "future_snark")] epoch: Epoch,
     ) -> StdResult<ProtocolPartyId> {
```
jpraynaud, line 510: **suggestion**: this function is a bit too long which prevents readability. Could you split it in sub-functions?

mithril-aggregator/tests/create_certificate_follower.rs
```diff
@@ -53,7 +53,7 @@ impl EpochFixturesMapBuilder {
                                 min_stake: 10,
                             },
                         )
-                        .build_at_epoch(Epoch(epoch as u64 + 1)),
+                        .build_at_epoch(Epoch(epoch as u64).next()),
```
jpraynaud: **issue(blocking)**: this is still sort of magic. You should use one of `offset_to_` prefixed methods of Epoch instead. The idea is to explain the choice of the offset we have made through the name of the offset function used.

mithril-common/src/crypto_helper/cardano/opcert.rs
jpraynaud, line 1: **question(blocking)**: did you compute the golden values with the previous methods?

mithril-aggregator/tests/test_extensions/runtime_tester.rs
```diff
@@ -278,6 +278,27 @@ impl RuntimeTester {
...
+                fixture_with_signers_and_parameters.build_at_epoch(current_retrieval_epoch);
+            let next_fixture = fixture_with_signers_and_parameters.build_at_epoch(time_point.epoch);
+            self.dependencies
+                .init_state_from_fixtures_for_genesis(
+                    &current_fixture,
+                    &next_fixture,
+                    time_point.epoch,
+                )
+                .await;
+        }
```
jpraynaud, line 300: **suggestion(blocking)**: could you apply the suggestions about the fixture builder here as well?

mithril-stm/src/signature_scheme/schnorr_signature/signing_key.rs
jpraynaud, line 1: **issue(blocking)**: the `get_snark_registration_entry` function of the non recursive clerk is using the index in the full closed key registration. This could lead to retrieving the path for a wrong signer (this problem may have pre-existed this PR). 

Also a nitpick, the type `LotteryIndex` used for the `signer_index` is incorrect.

mithril-aggregator/tests/certificate_chain.rs
jpraynaud, line 1: **issue(blocking)**: there are a lot of modifications that make this integration test (and all others) more verbose and less readable.

mithril-common/src/crypto_helper/cardano/opcert.rs
damrobi: Yes I computed them using the old methods to make sure they passed before updating to the new methods

mithril-stm/src/lib.rs
damrobi: Done, I used Sha256 for the hash function

mithril-common/src/crypto_helper/cardano/key_certification.rs
damrobi: Fixed

mithril-common/src/crypto_helper/cardano/key_certification.rs
damrobi: I added two sub-functions to improve readability, let me know if it should be improved further

mithril-aggregator/tests/create_certificate_follower.rs
damrobi: clarified the offset using one of the offset functions

mithril-stm/src/proof_system/halo2_snark/clerk.rs
```diff
@@ -29,6 +29,9 @@ pub struct SnarkClerk {
     pub(crate) closed_key_registration: ClosedKeyRegistration,
     /// Protocol parameters
     pub(crate) parameters: Parameters,
+    /// A mapping between the signer index in the list of all signers
+    /// and their index in the reduced list of signers with SNARK keys.
+    global_to_snark_index_mapping: HashMap<u64, usize>,
```
jpraynaud: **issue**: the HashMap entry and values represent the same type `SignerIndex` so the type used should be 
```suggestion
    concatenation_to_snark_indexes: HashMap<SignerIndex, SignerIndex>,
```

mithril-stm/src/proof_system/halo2_snark/clerk.rs
```diff
@@ -69,6 +79,12 @@ impl SnarkClerk {
         Ok(closed_registration_entry.into())
     }
 
+    /// Returns the index mapping between all registered entries and the ones with
+    /// a snark key.
+    pub(crate) fn get_global_to_snark_index_mapping(&self) -> &HashMap<u64, usize> {
+        &self.global_to_snark_index_mapping
+    }
```
jpraynaud: **issue**: instead of returning the HashMap, it'd be better to provide a function which returns a SNARK signer index given a Concatenation signer index.

mithril-stm/src/proof_system/halo2_snark/clerk.rs
damrobi: Done

mithril-stm/src/proof_system/halo2_snark/clerk.rs
damrobi: Done I used global index as an input name for the concatenation signer index

mithril-stm/src/protocol/key_registration/register.rs
```diff
@@ -43,29 +45,55 @@ impl KeyRegistration {
...
+        {
+            let mut existing = *self
+                .registration_entries
+                .iter()
+                .find(|e| e.get_verification_key_for_snark() == Some(vk_snark))
+                .ok_or(RegisterError::SnarkKeyRegistrationInconsistent)?;
+
+            // first compares the stake then the BLS key value as bytes
+            if entry < existing {
+                self.registration_entries.remove(&existing);
```
hjeljeli32: If I understood correctly, a signer whose duplicate SNARK key is removed can no longer create its BLS signer. The new behavior preserves the BLS registration and clears only the duplicate SNARK key. However, `Initializer::try_create_signer() `in `initializer.rs` still searches for an entry containing the initializer’s original SNARK key. Equality includes that field, so the lookup returns `UnregisteredInitializer`.

hjeljeli32 (commented):
I left one comment about duplicated SNARK registration.

PR #3541, opened 2026-09-15
feat: add the circuit verification key registry crate and tools

jpraynaud:
## Content

This PR includes the **circuit verification key registry crate and the tools to manage a registry**, prepared for its enforcement in #3514:

- Move the circuit verification key registry introduced in #3513 to its own `mithril-circuit-key-registry` internal crate, behind its `future_snark` feature (registry format, certifiers and retrievers), `mithril-common` keeping only the `CircuitVerificationKeyCertifier` trait used by the certificate verifier
- Remove the compiled minimum registry version: the genesis signature and the cached certifier cover the registry, a compiled floor would need a release per publication
- Keep a single entry per circuit verification key digest, either `allowed` over an epoch range or `revoked`, a revoked key being rejected for every epoch
- Add an HTTP retriever of the signed registry with a bounded downloader (timeout, retries and size cap)
- Keep the last verified registry in memory when a refresh fails or would lower the registry version, so an outage of the registry source does not stop a running node
- Add the `circuit-key-registry` command to the aggregator: `export` the circuit key digests for the network protocol parameters, `whitelist` a key (creating the registry of a Mithril network when missing), `revoke` an allowed key with its revocation epoch and comment, `sign` a hand-authored registry, and the test-only `bootstrap`, the registry being updated in place and genesis-signed
- Document the command in the aggregator CLI documentation.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested

## Issue(s)
Relates to #3148


internal/mithril-circuit-key-registry/Cargo.toml
```diff
@@ -0,0 +1,47 @@
...
+
+# Enables `rug-backend` features for `mithril-common` dependency
+rug-backend = ["mithril-common/rug-backend"]
+# Enables `num-integer-backend` features for `mithril-common` dependency
+num-integer-backend = ["mithril-common/num-integer-backend"]
+
+# Enables the TLS features for the `mithril-common` dependency
+# Exactly one of these must be enabled
+native-tls = ["mithril-common/native-tls"]
+rustls = ["mithril-common/rustls"]
```
Alenar: **issue(blocking):** internal crates should minimize their features and avoid adding some only to bubbles up one from their dependencies (this applies to `mithril-common` but also third parties like `reqwest`).
Note: We have no other internal crate that depends on `mithril-common` which re-declare those.

I think we should only keep `future_snark`.

internal/mithril-circuit-key-registry/src/lib.rs
```diff
@@ -0,0 +1,28 @@
...
+pub mod test;
+
+#[cfg(feature = "future_snark")]
+pub use certifier::*;
+#[cfg(feature = "future_snark")]
+pub use http_downloader::*;
+#[cfg(feature = "future_snark")]
+pub use registry::*;
+#[cfg(feature = "future_snark")]
+pub use retriever::*;
```
Alenar: **suggestion(non-blocking,if-minor):** adding documentation tags would be nice.

> [!NOTE]
> As the `future_snark` feature won't be long leaved you can dismiss this comment if you think it's not worth the work.

```suggestion
#![cfg_attr(docsrs, feature(doc_cfg))]

macro_rules! cfg_future_snark {
    ($($item:item)*) => {
        $(
            #[cfg(feature = "future_snark")]
            #[cfg_attr(docsrs, doc(cfg(feature = "future_snark")))]
            $item
        )*
    }
}

#[cfg(feature = "future_snark")]
mod certifier;
#[cfg(feature = "future_snark")]
mod http_downloader;
#[cfg(feature = "future_snark")]
mod registry;
#[cfg(feature = "future_snark")]
mod retriever;
#[cfg(feature = "future_snark")]
pub mod test;

cfg_future_snark! {
    pub use certifier::*;
    pub use http_downloader::*;
    pub use registry::*;
    pub use retriever::*;
}

```

Also, the cargo toml should be modified to tell `doc.rs` to add the tags:
```toml
[package.metadata.docs.rs]
all-features = true
# enable unstable features in the documentation
rustdoc-args = ["--cfg", "docsrs"]
```

internal/mithril-circuit-key-registry/src/certifier.rs
```diff
@@ -0,0 +1,812 @@
...
+
+        Self {
+            registry,
+            verified_at: now,
+            refreshed_at: now,
+            next_refresh_at: now + TimeDelta::seconds(time_to_live_in_seconds),
+        }
+    }
+
+    /// Whether the cached registry does not need a refresh yet.
```
damrobi:
```suggestion
    /// Whether the cached registry is fresh or not.
```
suggestion: remove the double negative

mithril-aggregator/src/tools/circuit_key_registry.rs
```diff
@@ -0,0 +1,1440 @@
...
+            })?;
+        let registry_json = Self::edited_registry_json(&current_registry_json, digest, edit)?;
+
+        Self::check_sign_and_write_json(&registry_json, &genesis_signer, registry_path)
+    }
+
+    /// Append the entry to the registry JSON, refusing a digest already listed, and increment
+    /// the version.
+    ///
+    /// The edit is applied to the JSON document rather than to the parsed registry, so the fields
```
damrobi: The term edit can be a bit confusing here since there is not edit parameter as in the other functions.
```suggestion
    /// The extension is applied to the JSON document rather than to the parsed registry, so the fields
```

mithril-aggregator/src/tools/circuit_key_registry.rs
```diff
@@ -0,0 +1,1440 @@
...
+    /// added by a future schema version survive an entry added by an older binary.
+    fn extended_registry_json(
+        current_registry_json: &str,
+        entry: &CircuitVerificationKeyEntry,
+    ) -> StdResult<String> {
+        let mut registry_value: Value = serde_json::from_str(current_registry_json)?;
+        Self::increment_version(&mut registry_value)?;
```
damrobi: nit: move the version incrementation after the entry collision check?

mithril-aggregator/src/tools/circuit_key_registry.rs
```diff
@@ -0,0 +1,1440 @@
...
+
+    /// Apply the edit to the allowed entry of the digest in the registry JSON and increment the
+    /// version.
+    fn edited_registry_json(
+        current_registry_json: &str,
+        digest: &CircuitVerificationKeyDigest,
+        edit: impl FnOnce(&mut Map<String, Value>),
+    ) -> StdResult<String> {
+        let mut registry_value: Value = serde_json::from_str(current_registry_json)?;
+        Self::increment_version(&mut registry_value)?;
```
damrobi: nit: move the version incrementation after the entry collision check?

mithril-aggregator/src/tools/circuit_key_registry.rs
```diff
@@ -0,0 +1,1440 @@
...
+                "The registry version {} must be {expected_version}: the version following the signed registry at '{}', or {INITIAL_REGISTRY_VERSION} when there is none",
+                registry.version,
+                signed_registry_path.display()
+            ));
+        }
+
+        Ok(())
+    }
+
+    /// Create and sign the circuit verification key registry whitelisting from epoch 0 the
```
damrobi: nit: add a comma to help how the comment reads
```suggestion
    /// Create and sign the circuit verification key registry, whitelisting from epoch 0 the
```

mithril-aggregator/src/tools/circuit_key_registry.rs
```diff
@@ -0,0 +1,1440 @@
...
+            if registry.entries[..index]
+                .iter()
+                .any(|listed| listed.digest == entry.digest)
+            {
+                return Err(anyhow!(
+                    "The circuit verification key '{}' has several entries",
+                    entry.digest
+                ));
+            }
+        }
```
damrobi: nit: this could be improved with a HashSet
```suggestion
        let mut seen_digests = std::collections::HashSet::new();
        for entry in &registry.entries {
            if !seen_digests.insert(entry.digest) {
               return Err(anyhow!(
                  "The circuit verification key '{}' has several entries",
                  entry.digest
              ));
          }
        }
```

internal/mithril-circuit-key-registry/src/certifier.rs
```diff
@@ -0,0 +1,812 @@
...
+            Some(cache) => self.refresh_cached_registry(cache, refreshed),
+            None => {
+                let registry = refreshed?;
+                *cache = Some(VerifiedRegistryCache::verified(
+                    registry.clone(),
+                    self.cache_time_to_live_in_seconds,
+                ));
+
+                Ok(registry)
+            }
```
damrobi: question: This branch accepts any valid registry given even if outdated, this could lead to some rollback on a restart. Are we planning to add some guard against that?

internal/mithril-circuit-key-registry/src/registry.rs
```diff
@@ -90,13 +91,14 @@ pub struct CircuitVerificationKeyEntry {
...
+    /// Whether the key is allowed over the entry's range or revoked.
     pub status: CircuitVerificationKeyStatus,
 
     /// First epoch (inclusive) covered by the statement.
     pub start_epoch: Epoch,
 
-    /// Last epoch (inclusive) covered by the statement, open-ended when absent.
+    /// Last epoch (inclusive) covered by an allowed entry, open-ended when absent, or the
+    /// revocation epoch of a revoked entry.
     pub end_epoch: Option<Epoch>,
```
damrobi, line 102: suggestion: this value is written by `Revoke` but never used after, maybe we can dedicate it to `Allowed` and put the revocation epoch in the comment of `Revoked`

internal/mithril-circuit-key-registry/src/http_downloader.rs
```diff
@@ -0,0 +1,208 @@
...
+        assert_eq!(0, document.calls());
+    }
+
+    #[tokio::test]
+    async fn fails_on_a_response_exceeding_the_body_size_limit() {
+        let server = MockServer::start();
+        server.mock(|when, then| {
+            when.method(Method::GET).path("/document");
+            then.status(200)
+                .body(" ".repeat(DOWNLOAD_MAX_BODY_SIZE_IN_BYTES as usize + 1));
```
damrobi, line 234: This test doesn't really exercise the oversize check of the paquets after the first one

internal/mithril-circuit-key-registry/src/http_downloader.rs
```diff
@@ -0,0 +1,208 @@
...
+            .get(url)
+            .send()
+            .await
+            .with_context(|| format!("Failed to download '{url}'"))?;
+        if !response.status().is_success() {
+            return Err(anyhow!(
+                "Failed to download '{url}': status {}",
+                response.status()
+            ));
+        }
```
damrobi, line 109: suggestion: it might be worth adding a check that the response is https to add a guard even for the wasm case

internal/mithril-circuit-key-registry/src/certifier.rs
```diff
@@ -0,0 +1,812 @@
...
+    /// A cache refreshed in the future (the clock jumped backwards) is stale, so it forces a
+    /// refresh instead of staying fresh until the clock catches up.
+    fn is_fresh(&self) -> bool {
+        (self.refreshed_at..self.next_refresh_at).contains(&Utc::now())
+    }
+
+    /// Whether a refreshed registry replaces the cached one: a newer version, or the cached
+    /// registry verified again.
+    fn is_superseded_by(&self, refreshed: &CircuitVerificationKeyRegistry) -> bool {
+        refreshed.version > self.registry.version || *refreshed == self.registry
```
damrobi, line 181: suggestion: according to Claude the second part of the statement has no dedicated test

mithril-aggregator/src/tools/circuit_key_registry.rs
```diff
@@ -0,0 +1,1440 @@
...
+    /// Apply the edit to the allowed entry of the digest in the signed registry at the given
+    /// path, then sign the incremented version with the genesis secret key and write it back in
+    /// place.
+    fn edit_allowed_entry(
+        registry_path: &Path,
+        genesis_secret_key_path: &Path,
+        digest: &CircuitVerificationKeyDigest,
+        edit: impl FnOnce(&mut Map<String, Value>),
+    ) -> StdResult<CircuitVerificationKeyRegistry> {
+        let genesis_signer = GenesisSigner::read_from_file(genesis_secret_key_path)?;
```
damrobi, line 161: nit(not blocking): This loads the secret key but it is only necessary in the `check_sign_and_write_json` so we could first only load the verifier and delay the secret key load

mithril-aggregator/src/tools/circuit_key_registry.rs
```diff
@@ -0,0 +1,1440 @@
...
+            )
+            .unwrap();
+
+            read_signed_registry(&registry_path)
+                .verify(&genesis_signer.create_verifier())
+                .expect("the replaced registry must carry a valid genesis signature");
+        }
+    }
+
+    mod bootstrap {
```
damrobi, line 1351: the calls to `bootstrap()` never exercise what happens with explicit protocol parameters, they only default to the production ones

damrobi (commented):
It looks good to me but there are a few comments worth discussing I believe

internal/mithril-circuit-key-registry/src/certifier.rs
jpraynaud: This is complicated to implement currently as this would require the update of all the registries as soon as a registry is modified. Also the risk that an outdated registry is served is very limited and thus acceptable at the moment. However, we can re-challenge the usage of a monotonic counter in the future.

mithril-aggregator/src/tools/circuit_key_registry.rs
jpraynaud: The cost is marginal, it will not be changed.

internal/mithril-circuit-key-registry/src/certifier.rs
jpraynaud: Fixed

mithril-aggregator/src/tools/circuit_key_registry.rs
jpraynaud: Fixed

mithril-aggregator/src/tools/circuit_key_registry.rs
jpraynaud: Fixed

mithril-aggregator/src/tools/circuit_key_registry.rs
jpraynaud: Fixed

mithril-aggregator/src/tools/circuit_key_registry.rs
jpraynaud: Fixed

mithril-aggregator/src/tools/circuit_key_registry.rs
jpraynaud: Fixed

internal/mithril-circuit-key-registry/src/registry.rs
jpraynaud: We keep this field for audit trail only.

internal/mithril-circuit-key-registry/src/http_downloader.rs
jpraynaud: Fixed

internal/mithril-circuit-key-registry/src/http_downloader.rs
jpraynaud: Fixed

internal/mithril-circuit-key-registry/src/certifier.rs
jpraynaud: Fixed

mithril-aggregator/src/tools/circuit_key_registry.rs
jpraynaud: Fixed

damrobi (approved):
LGTM 👍

PR #3547, opened 2026-09-16
refactor(stm): move the IVC circuit onto the Relation trait

### LINKED ISSUE #3428: Implement `Relation` trait for the IVC circuit
## Why

`StmCertificateCircuit` implements the `Relation` trait of the Midnight standard library, while `IvcCircuitData` implements the lower level `Circuit` trait with a hand written configuration. The two circuits therefore produce verifying keys in different formats and the IVC configuration duplicates what `ZkStdLibArch` already provides.

## What

Move the IVC circuit onto `Relation`, so that both circuits share the same configuration mechanism and the same verifying key format. The standard library configures its chips in a different order, so the recursive key identity changes and every asset derived from it is regenerated as part of the move.

## How

- [x] Separate the circuit's metadata from a single step's values, so the metadata type can carry the relation and the values become its witness (`circuits/halo2_ivc/circuit.rs`)
- [x] Move the IVC circuit body from `synthesize` to the `circuit` method of `Relation` (`circuits/halo2_ivc/circuit.rs`)
- [x] Declare the chips through `ZkStdLibArch` and drop the hand written column pool computation (`circuits/halo2_ivc/config.rs`)
- [x] Use `VerifierGadget` from `midnight-zk-stdlib` for the in-circuit certificate proof verification
- [x] Check that the IVC and the certificate verifying keys share the same format
- [x] Relocate the shared verifying key codec so a single implementation serves both circuits (`circuits/key_serialization.rs`)
- [x] Store the recursive proving key together with the relation metadata its circuit is rebuilt from (`circuits/halo2_ivc/key_serialization.rs`)
- [x] Keep the recursive key decoder restricted to the approved architecture, so a verifying key belonging to another circuit is not accepted in its place (`circuits/halo2_ivc/key_serialization.rs`)
- [x] Invalidate the recursive non-production key caches when the recursive key identity changes, leaving the certificate cache identities untouched (`circuits/key_provider.rs`)
- [x] Confirm the recursive circuit still fits its current degree under the standard library configuration, measured with the production certificate metadata, before any key or asset is generated
- [x] Check the impact on the circuit degree and on the golden test vectors, including whether the recursive range column count can be lowered without raising the degree
- [x] Regenerate the production and golden recursive verifying keys and every asset that depends on them, then update the affected digest goldens and the stored combiner challenge

```rust
fn used_chips(&self) -> ZkStdLibArch {
        ZkStdLibArch {
            jubjub: true,
            poseidon: true,
            sha2_256: true,
            sha2_512: false,
            keccak_256: false,
            sha3_256: false,
            secp256k1: false,
            bls12_381: true,
            base64: false,
            nr_pow2range_cols: 4,
            automaton: false,
            blake2b: false,
        }
    }
```
> Then  move the ivc circuit code in `fn synthesize`  to `fn circuit` from `Relation` .
> If we do this, both ivc_vk and cert_vk will have the same format `MidnightVK`
> `VerifierGadget` is available here https://github.com/midnightntwrk/midnight-zk/blob/main/zk_stdlib/src/lib.rs#L847
> it is also available in `midnight-zk-stdlib = { version = "2.3.0" }`
> Then  move the ivc circuit code in `fn synthesize`  to `fn circuit` from `Relation` .
> If we do this, both ivc_vk and cert_vk will have the same format `MidnightVK`
> `VerifierGadget` is available here https://github.com/midnightntwrk/midnight-zk/blob/main/zk_stdlib/src/lib.rs#L847
> it is also available in `midnight-zk-stdlib = { version = "2.3.0" }`

hjeljeli32:
## Content

The IVC circuit implements Halo2's low level `Circuit` trait with a hand written configuration and its own column pool, while the certificate circuit implements the Midnight standard library's `Relation`. The two configure their chips differently, serialize their verifying keys in different formats, and the IVC configuration duplicates what `ZkStdLibArch` already provides. The IVC circuit now implements `Relation`, declares its chips through `ZkStdLibArch`, takes its verifier gadget from the standard library, and shares the `MidnightVK` and `MidnightPK` key formats. 

**This is a breaking circuit key change.** The standard library registers gates and selectors in a different order, so the recursive verifying key's commitments change although its shape does not — still degree 19, still 45 fixed commitments — and every asset derived from it was regenerated. The degree was measured against production certificate metadata before any key was generated and holds at 19; **lowering `nr_pow2range_cols` below 4 was checked as the runbook requires, and 1 to 3 columns all require degree 20**.

### Changes

- **`Relation` implementation** (`circuit.rs`, `constraint_builder.rs`, `witness_assignments.rs`, `schnorr_signature.rs`): the synthesis body moves to `Relation::circuit`, `IvcCircuit` carries the metadata fixing the constraint system and `IvcCircuitData` becomes its witness. The builder borrows the standard library's chips instead of owning its own.
- **Standard library configuration** (`circuit.rs`, `config.rs` deleted): one `recursive_circuit_architecture()` feeds both the declared chips and the key generation metadata, so they cannot drift. The hand written configuration, its column pool and the errors guarding it are removed.
- **Shared key format** (`key_serialization.rs` added at `circuits/`, `halo2_ivc/key_serialization.rs` deleted, `keys.rs`): one `MidnightVK` codec serves both circuits, the recursive newtypes wrap the Midnight types, and key generation goes through `zk::setup_vk` and `zk::setup_pk`. The inner verifying key is byte identical, so this adds a 23 byte envelope without altering circuit identity.
- **Recursive decoder guards** (`keys.rs`, `errors.rs`): both circuits now encode keys identically, so only the declared architecture distinguishes them. The decoders check the architecture and every declared degree before the bytes reach the dependency's readers, require a standalone encoding to be fully consumed, and check the fixed commitment count against the configured constraint system. Serde uses the same guard, which is the path the verifier data takes.
- **Recursive cache identity** (`key_provider.rs`): non production entries were fingerprinted from the certificate key alone and are trusted without comparison, so a stale recursive entry would have been loaded against the new constraint system. The recursive fingerprint now includes the recursive key; the certificate fingerprint is unchanged and pinned by a test.
- **Regenerated assets** (`recursive_circuit_verification_key_for_production.bin`, eight files under `tests/assets/`, `tests/assets/README.md`): both recursive verifying keys, the verification context, the three step outputs, the chain state and both accumulator anchors, plus the two recursive key digests, the combiner challenge and the ancillary encoding digest. The README now lists those source constants, which no documentation covered.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation
  - [x] No new TODOs introduced

## Issue(s)

Closes https://github.com/IntersectMBO/mithril/issues/3428


jpraynaud (approved):
LGTM 👍

mithril-stm/src/circuits/halo2_ivc/tests/common/failure_signature.rs
```diff
@@ -5,6 +5,11 @@
 //! Advice-side members of a broken class are not enumerated: their columns and regions are named by
 //! the gadget layer and carry no stability guarantee.
 
+use crate::circuits::halo2_ivc::RECURSIVE_CIRCUIT_DEGREE;
+use crate::circuits::halo2_ivc::circuit::IvcCircuit;
+use midnight_proofs::circuit::Value;
+use midnight_zk_stdlib::MidnightCircuit;
```
damrobi: nitpick: import order (here and in other places)

damrobi (approved):
LGTM 👍 
Some imports are out of order or inside functions so maybe you can rearrange them

mithril-stm/src/circuits/halo2_ivc/circuit.rs
```diff
@@ -64,193 +71,194 @@ impl IvcCircuitData {
...
     ) -> Result<(), Error> {
-        let builder = IvcConstraintBuilder::new(&config);
+        let builder = IvcConstraintBuilder::new(std_lib);
+
+        let global_value = witness.clone().map(|data| data.global);
+        let state_value = witness.clone().map(|data| data.state);
+        let witness_value = witness.clone().map(|data| data.witness);
+        let certificate_proof_value = witness.clone().map(|data| data.certificate_proof);
+        let ivc_proof_value = witness.clone().map(|data| data.ivc_proof);
+        let accumulator_value = witness.map(|data| data.accumulator);
```
kitounliu, line 143: witness.clone() copies the whole witness 5 times.  A simpler way could be using the reference,

let data = witness.as_ref();
let global_value = data.map(|d| d.global.clone());
let state_value = data.map(|d| d.state.clone());
let witness_value = data.map(|d| d.witness.clone());
let certificate_proof_value = data.map(|d| d.certificate_proof.clone());
let ivc_proof_value = data.map(|d| d.ivc_proof.clone());
let accumulator_value = data.map(|d| d.accumulator.clone());

kitounliu:
LGTM

PR #3554, opened 2026-09-18
refactor(client): new certificate verifier cache api

### LINKED ISSUE #3520: Add a certificate chain cache committed after chain verification in the client library
## Why

We want to update the unstable `MithrilCertificateVerifier` and `CertificateVerifierCache` implementations of the client to support a commit mechanism when the full certificate chain is validated. A first `FullVerification` mode of verification using the cache will verify the cryptography of the cached certificate, thus providing the same level of security than without cache.

## What

Implement the new commit mechanism for the client certificate cache and the `FullVerification` mechanism.

## How

- [x] Update the `CertificateVerifierCache` trait with:
  - [x] `stage_certificate(&self, certificate_chain_validation_id: &str, certificate: &CertificateMessage) -> MithrilResult<()>`
  - [x] `commit_staged_certificates(&self, certificate_chain_validation_id: &str) -> MithrilResult<`()>
  - [x] `get_certificate_by_hash(&self, certificate_hash: String) -> MithrilResult<Option<CertificateMessage>>`
  - [x] `reset(&self) -> MithrilResult<()>`
- [x] Adapt the `MemoryCertificateVerifierCache` implementation
- [x] Implement the `FullVerification` algorithm in `verify_chain` of the `MithrilCertificateVerifier`
- [x] Adapt the `ClientBuilder` to configure the cache

Alenar:
## Content

This PR completely reworks the `mithril-client` certificate verifier cache api:
- Certificates are entirely stored in cache instead of only their hash/parent hash
- Cached certificates are now cryptographically verified, before they were assumed correct
- Certificates are staged to cache into a batch keyed by `certificate_chain_validation_id`, and committed only after the whole certificate chain is validated, to limit the chance that a tampered certificate is stored then served from cache
- Caching is now handled by a `CachedCertificateRetriever` decorator sitting one layer below `verify_chain` (checks the cache before falling back to the network), instead of `verify_chain` itself special-casing cache hits.
  - Since every certificate is now unconditionally re-verified regardless of source, this also let us remove the previous two-phase "no cache until crossing an epoch boundary" walk (a workaround for the old blind-trust bug) + `verify_chain` is now a single loop
- Genesis certificates are now cached like any other certificate (previously excluded) - safe now that every cache hit is re-verified regardless of source
- `ClientBuilder::with_certificate_verifier_cache` has a new `mode` parameter, an `CertificateVerifierCacheMode` enum with only only one  value (`FullVerification`).
- Any cache operation failure (read, stage, or commit) is logged and swallowed rather than propagated, a broken cache backend must never fail a cryptographically valid chain verification

### Caveats

- `MithrilEvent::CertificateFetchedFromCache` is no longer emitted. Caching now happens inside `CachedCertificateRetriever`, which has no `certificate_chain_validation_id` to attach to the event.
- `mithril-client-wasm` still has no working cache (`build_certifier_cache` returns `None`) has the local storage implementation had to be removed (else, with only 10Mb, it would have hit browsers limits), a real `IndexedDbCertificateVerifierCache` is a separate, larger follow-up involving actual browser storage bindings.
- Only `FullVerification` mode exists. A future `EarlyStop` mode (trusting a cache hit without re-verification) is intentionally deferred to a later ticket.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation
  - [x] No new TODOs introduced

## Issue(s)

Closes #3520


mithril-client/src/certificate_client/verify_cache/memory_cache.rs
```diff
@@ -35,58 +44,100 @@ impl MemoryCertificateVerifierCache {
...
         &self,
         certificate_hash: &CertificateHash,
     ) -> MithrilResult<Option<MithrilCertificate>> {
-        let cache = self.cache.read().await;
+        let cache = self.committed.read().await;
         Ok(cache
             .get(certificate_hash)
             .filter(|cached| cached.expire_at >= Utc::now())
             .map(|cached| cached.certificate.clone()))
     }
```
jpraynaud, line 143: **issue(non-blocking)**: the expired certificates are filtered out but never deleted which does not bound the memory usage.

mithril-client/src/certificate_client/fetch.rs
```diff
@@ -65,6 +71,77 @@ impl CertificateRetriever for InternalCertificateRetriever {
...
+        // LIMITATION:
+        // A cache hit here can't currently emit `MithrilEvent::CertificateFetchedFromCache` as
+        // we do not have access to the `certificate_chain_validation_id` (or any per-verify_chain-call
+        // context), and this type has no `FeedbackSender`.
+        // Properly restoring the feedback needs `mithril-common` `CertificateVerifier`/`verify_certificate`
+        // to thread a per-validation context or callback down to this call site.
+        let certificate = match self.cache.get_certificate_by_hash(certificate_hash).await {
+            Ok(None) => None,
+            Ok(Some(message)) => match Certificate::try_from(message) {
+                Ok(certificate) if certificate.hash == certificate_hash => Some(certificate),
```
jpraynaud, line 107: **suggestion**: re-validate the computation of the hash of the certificate retrieved from the cache.

jpraynaud (approved):
LGTM 👍

mithril-client/src/certificate_client/fetch.rs
Alenar: It would be nice yes, but we have the certificate message here, which does not have the hash computation method, meaning that we would need to convert it back to a certificate entity first adding a lot of complexity for what is designed a simple check.

mithril-client/src/certificate_client/verify_cache/memory_cache.rs
Alenar: Fixed: committed certificates are now sweep at the start of a new batch and when committing a batch.

mithril-client/src/certificate_client/verify_cache/memory_cache.rs
```diff
@@ -39,103 +44,149 @@ impl MemoryCertificateVerifierCache {
...
         certificate_hash: &CertificateHash,
-    ) -> MithrilResult<Option<String>> {
-        let cache = self.cache.read().await;
+    ) -> MithrilResult<Option<MithrilCertificate>> {
+        let cache = self.committed.read().await;
         Ok(cache
             .get(certificate_hash)
             .filter(|cached| cached.expire_at >= Utc::now())
-            .map(|cached| cached.previous_hash.clone()))
+            .map(|cached| cached.certificate.clone()))
```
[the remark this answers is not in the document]
Alenar: Fixed: committed certificates are now sweep at the start of a new batch and when committing a batch.

mithril-client/src/certificate_client/verify_cache/memory_cache.rs
```diff
@@ -39,103 +44,149 @@ impl MemoryCertificateVerifierCache {
...
-        // todo: should we raise an error if an empty string is given for previous_certificate_hash ? (or any other kind of validation)
-        let mut cache = self.cache.write().await;
-        cache.insert(
-            certificate_hash.to_string(),
-            CachedCertificate::new(
-                previous_certificate_hash,
-                Utc::now().add(self.expiration_delay),
-            ),
-        );
+        let mut staged = self.staged.write().await;
```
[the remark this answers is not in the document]
Alenar: Fixed: committed certificates are now sweep at the start of a new batch and when committing a batch.

PR #3559, opened 2026-09-21
refactor: remove the HTTP client and TLS features from STM

### LINKED ISSUE #3558: Remove the HTTP client and TLS features from STM
## Why

The `future_snark` feature of `mithril-stm` depends on `reqwest` with a blocking runtime and forces every consumer to select a `rustls` or `native-tls` feature, only to download the trusted setup SRS in `TrustedSetupProvider`. Only the aggregator proves, so the signer, the client, the WASM client and the relay compile an HTTP client and a TLS backend they never use. The download also runs on the lazy prover path, reached from a tokio worker thread in the aggregator, where it blocks a runtime worker in release builds and panics in builds with debug assertions.

## What

Move the SRS download out of `mithril-stm` behind a `TrustedSetupDownloader` trait implemented in the aggregator, then remove `reqwest` and the TLS features from `mithril-stm`, `mithril-common` and their consumers.

## How

- [x] Add a `TrustedSetupDownloader` trait and a `NoTrustedSetupDownload` implementation to `mithril-stm`, and inject the `TrustedSetupProvider` from `SnarkProverSetupWarmer::warm` down to the prover setup constructors
- [x] Re-export the trusted setup provider, the downloader trait and the SRS URL constant from `mithril-common`
- [x] Implement the downloader in `mithril-aggregator` over its `reqwest` client with a captured runtime handle, guarded against being called from a runtime thread, with bounded retries, and wire it into the aggregate signature prover warmer
- [x] Retry the aggregate signature prover warm-up in `mithril-aggregator` until it succeeds, with a delay between attempts, since the lazy prover path no longer downloads the SRS
- [x] Drop the TLS forwarding features from `mithril-common`, `mithril-client`, `mithril-signer`, `mithril-relay`, `mithril-end-to-end`, `protocol-demo` and the CI workflows
- [x] Remove the `reqwest` download, the `reqwest` dependency, the TLS features and the `httpmock` dev-dependency from `mithril-stm`, and drop the TLS feature from its docs.rs metadata
- [x] Update the `mithril-stm` README and examples

jpraynaud:
## Content

This PR includes the **inversion of the trusted setup SRS download out of `mithril-stm`, so that only the aggregator, the sole proving node, compiles an HTTP client and a TLS backend**:

- Added a `TrustedSetupDownloader` trait and a `NoTrustedSetupDownload` implementation to `mithril-stm`, and injected the `TrustedSetupProvider` from `SnarkProverSetupWarmer::warm` down to the prover setup constructors, the lazy proving path only reading the local SRS cache instead of downloading from a runtime thread.
- Re-exported the trusted setup provider, the downloader trait and the SRS URL constant from `mithril-common`.
- Implemented `ReqwestTrustedSetupDownloader` in `mithril-aggregator` over its `reqwest` client with a captured runtime handle, refusing calls from a runtime thread, with bounded retries and typed errors, and wired it into the aggregate signature prover warmer.
- Made the aggregate signature prover warm-up retry until the setups are materialized, since the lazy proving path no longer downloads the SRS and a failed download would otherwise leave the aggregator unable to prove until a restart.
- Removed the `reqwest` download, the `reqwest` dependency, the `rustls` and `native-tls` features and the `httpmock` dev-dependency from `mithril-stm`, and dropped the TLS feature from its docs.rs metadata.
- Dropped the TLS forwarding features from `mithril-common`, `mithril-client`, `mithril-signer`, `mithril-relay`, `mithril-end-to-end`, `protocol-demo` and the CI workflows.
- Updated the `mithril-stm` README, examples and bench docs, the circuit keys runbook with a manual SRS download step, and the CHANGELOGs.

## Pre-submit checklist

- Branch
- PR
  - [x] Self-reviewed the diff
  - [x] Reviewer requested
- Documentation
  - [x] No new TODOs introduced

## Comments

The SRS is no longer downloaded outside the aggregator: running the SNARK examples, the ignored production key tests or the circuit keys runbook requires caching it first with the commands documented in the `mithril-stm` README and in the examples.

## Issue(s)
Closes #3558


hjeljeli32 (approved):
LGTM 👍

mithril-stm/src/circuits/trusted_setup.rs
```diff
@@ -148,48 +164,73 @@ impl TrustedSetupProvider {
...
+        let srs_file = match File::open(&srs_file_path) {
+            Ok(file) => file,
+            Err(error) if error.kind() == ErrorKind::NotFound => return Ok(false),
+            Err(error) => {
+                return Err(error).with_context(|| {
+                    format!("Failed to open the cached SRS file at {srs_file_path:?}.")
+                });
+            }
+        };
+        if self.srs_expected_hash.is_empty() {
```
damrobi: I think this behavior is mostly there for testing when the SRS is from an untrusted setup. Maybe we should restrict it since the `new` function of  `TrustedSetupProvider` can be used without a hash value and that could trigger this behavior.

mithril-stm/CHANGELOG.md
```diff
@@ -5,6 +5,17 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## 0.12.21 (09-21-2026)
+
+### Changed
+
+- Moved the SRS download out of the library behind the `TrustedSetupDownloader` trait, implemented by the proving node and handed to `SnarkProverSetupWarmer` through a `TrustedSetupProvider`. The provers only read the SRS from the local cache, `NoTrustedSetupDownload` standing in where nothing downloads it, so the lazy proving path never blocks a runtime thread.
+- Verified the cached SRS against the pinned hash before every load: a cached file whose hash does not match is discarded and downloaded again, and a file the filesystem refuses to read or remove fails the load instead of being read unverified.
```
damrobi, line 13: > Verified the cached SRS against the pinned hash before every load
nit: This is no entirely true as pointed in the other comment on the branch skipping the hash verification when empty

mithril-stm/examples/recursive_snark_aggregate_signature.rs
```diff
@@ -22,10 +42,9 @@
 //!
 //! Expect roughly four and a half minutes and about 12 GB of peak memory, measured on an Apple M4
 //! Max with 16 cores and 48 GB of memory. That measurement had memory to spare; a machine with less
-//! than the peak installed will page, and take correspondingly longer. The first run additionally
-//! downloads the trusted setup. Every aggregation generates the circuit keys afresh, because the
-//! example's parameters are sized so it can be run at all and its keys are therefore not the
-//! production ones the key cache recognises.
+//! than the peak installed will page, and take correspondingly longer. Every aggregation generates
+//! the circuit keys afresh, because the example's parameters are sized so it can be run at all and
+//! its keys are therefore not the production ones the key cache recognises.
```
damrobi: nit: maybe this is slightly misleading since the examples are still going through the cache path so they can reuse cached value if several of them are run in a row

damrobi (approved):
LGTM :+1: 
Maybe it's worth having a look at the empty expected hash behavior just in case

mithril-stm/src/circuits/trusted_setup.rs
jpraynaud: Fixed

mithril-stm/CHANGELOG.md
jpraynaud: Fixed

mithril-stm/examples/recursive_snark_aggregate_signature.rs
jpraynaud: Fixed

